PR #31, opened 2014-12-12
3 lights ( + material and shader injection) implemented and tested on ios, android and rasberrypi

patriciogonzalezvivo:
Lot of changes. Please review


core/resources/directional_light.glsl
```diff
@@ -0,0 +1,29 @@
+struct DirectionalLight {
+	vec4 ambient;
+	vec4 diffuse;
+	vec4 specular;
+
+	vec3 direction;
+};
+
+void calculateLight(in DirectionalLight _light, in vec3 _eye, in vec3 _eyeToPoint, in vec3 _normal, inout vec4 _ambient, inout vec4 _diffuse, inout vec4 _specular){
+	#ifdef MATERIAL_AMBIENT
```
bcamper: Minor point but we might want to namespace these to avoid possible collisions with user-defined #defines. Like `TANGRAM_MATERIAL_AMBIENT` or something.

core/resources/lights_frag.glsl
```diff
@@ -0,0 +1,39 @@
...
+
+vec4 calculateLighting(in vec3 _eyeToPoint, in vec3 _normal) {
+	vec3 eye = vec3(0.0, 0.0, 1.0);
+
+	vec4 amb = v_amb;
+	vec4 diff = v_diff;
+	vec4 spec = v_spec;
+
+//  COMPUTE ALL LIGHTS
+#pragma tangram: lights_to_compute
```
bcamper: We should discuss how to simplify any dependencies between various injected fragments, e.g. avoid injections that themselves have injections in them. It's not yet clear that we want to support this level of complexity. To date I've kept the JS-based injection to simple single injection points mapped to key/value string pairs. cc @blair1618

core/resources/lights_frag.glsl
```diff
@@ -0,0 +1,39 @@
...
+varying vec4 v_spec;
+
+vec4 calculateLighting(in vec3 _eyeToPoint, in vec3 _normal) {
+	vec3 eye = vec3(0.0, 0.0, 1.0);
+
+	vec4 amb = v_amb;
+	vec4 diff = v_diff;
+	vec4 spec = v_spec;
+
+//  COMPUTE ALL LIGHTS
```
bcamper: Superfluous comment.

core/resources/lights_frag.glsl
```diff
@@ -0,0 +1,39 @@
...
+#pragma tangram: lights_to_compute
+
+	//  Final light intensity calculation
+	//
+	vec4 color = vec4(0.0);
+  
+	#ifdef MATERIAL_EMISSION
+	color = g_material.emission;
```
bcamper: So material is referenced via a global? Wondering if/how that will scale with custom shaders etc. Maybe not an issues, but something to discuss.

bcamper: Might be better if the function takes a material as input, and can hardcode it to a single global for now (but expand it later if needed?).

core/resources/material.glsl
```diff
@@ -0,0 +1,21 @@
...
+	#endif
+
+	#ifdef MATERIAL_SPECULAR
+	vec4 specular;
+	float shininess;
+	#endif
+};
+
+uniform Material u_material;
+Material g_material = u_material;
```
bcamper: Is `g_material` instantiated as a copy of `u_material` because we want to allow the shader to modify material properties ad hoc? Just clarifying (could add a comment to this effect).

core/resources/polygon.fs
```diff
@@ -1,12 +1,19 @@
-#ifdef GL_ES
-precision mediump float;
-#endif
+uniform mat4 u_modelView;
+uniform mat4 u_modelViewProj;
+uniform float u_time;
+
+#pragma tangram: material
+#pragma tangram: frag_lighting
```
bcamper: I'd prefer to have these be slightly more verbose but clearer: `fragment_lighting`, `vertex_lighting`.

core/resources/polygon.vs
```diff
@@ -1,30 +1,30 @@
...
+  	v_texcoord = a_texcoord;
 
-  	v_color = a_color;
-  	
-  	float lit = dot(normalize(u_lightDirection), normalize(a_normal));
-  	v_color.rgb *= clamp(lit * 1.5, 0.5, 1.5);
-  	
-	v_texcoord = a_texcoord;
+	v_eyeToPoint = vec3(u_modelView * a_position);
+	calculateLighting(v_eyeToPoint,v_normal);
```
bcamper: So, the `calculateLighting` function gets injected into both vertex and fragment shaders, but can do different pieces of work in each, e.g. directional light can be calculated in the vertex shader, then additional point light computations can be added in the fragment shader call of the same function name (but different function body), with the intermediate ? Is that correct? We'll also want to document this more clearly.

bcamper: Not clear what values this function is operating on. It would be less opaque if `calculateLighting` returned a value that was assigned. (I'm also not clear on how intermediate values are passed between vertex and fragment shaders in cases where some lighting happens in each shader, but I'm probably missing something.)

core/resources/polygon.fs
```diff
@@ -2,11 +2,23 @@
...
 varying vec4 v_color;
+varying vec3 v_eyeToPoint;
 varying vec3 v_normal;
 varying vec2 v_texcoord;
 
 void main(void) {
-  	gl_FragColor = v_color;
+	vec4 color = v_color;
+
+	color *= calculateLighting(v_eyeToPoint,v_normal);
```
blair1618: Ideally the style shaders will be valid GL programs on their own, before any injection is performed. Without injecting the lighting code, compilation would fail on the line using the undefined `calculateLighting`.

core/src/scene/directionalLight.cpp
```diff
@@ -0,0 +1,44 @@
...
+DirectionalLight::~DirectionalLight(){
+
+}
+
+void DirectionalLight::setDirection(const glm::vec3 &_dir){
+    m_direction = _dir;
+}
+
+void DirectionalLight::setupProgram( std::shared_ptr<ShaderProgram> _shader ){
+	if(m_dynamic){
```
bcamper: indentation/spacing off in this block

core/src/scene/directionalLight.cpp
```diff
@@ -0,0 +1,44 @@
...
+    if (bFirst){
+        bFirst = false;
+        return stringFromResource("directional_light.glsl")+"\n";
+    } else {
+        return "\n";
+    }
+}
+
+std::string DirectionalLight::getInstanceDefinesBlock(){
+	return "\n";
```
bcamper: What's going on here?

core/src/scene/directionalLight.cpp
```diff
@@ -0,0 +1,44 @@
...
+void DirectionalLight::setupProgram( std::shared_ptr<ShaderProgram> _shader ){
+	if(m_dynamic){
+		Light::setupProgram(_shader);
+    	_shader->setUniformf(getUniformName()+".direction", m_direction);
+	}
+}
+
+std::string DirectionalLight::getClassBlock(){
+    static bool bFirst = true;
+    if (bFirst){
```
bcamper: Maybe a way to clean this pattern up?

core/src/scene/pointLight.cpp
```diff
@@ -0,0 +1,107 @@
...
+        }
+
+        if (m_quadraticAttenuation!=0.0) {
+            _shader->setUniformf(getUniformName()+".quadraticAttenuation", m_quadraticAttenuation);
+        }
+    }
+}
+
+std::string PointLight::getClassBlock() {
+    static bool bFirst = true;
```
blair1618: Using a static variable here doesn't actually achieve what we want. If we want to inject a type of light into more than one shader, all except the first will get nothing for the "class block".

core/resources/material.glsl
patriciogonzalezvivo: Yes that's exactly the case the famous:

if( isWindow)
   g_material.specular = vec4(1.,1.,1.,0.5);
else 
   g_material.specular = vec4(0.,0.,0.,0.);
   g_material.diffuse = wallColor;

core/resources/polygon.vs
patriciogonzalezvivo: is not the same code neither function between vert/frag. Data is passed by

varying vec4 v_amb;
varying vec4 v_diff;
varying vec4 v_spec;

At the begining calculateLighting return a computed color... but that over-ride the original color and what ever get shadowed on vertex will remain shadowed for ever (very poetic right?)

core/resources/polygon.fs
patriciogonzalezvivo: I will resolve with #defines

core/src/scene/light.h
```diff
@@ -0,0 +1,100 @@
...
+ 
+#include <memory>
+#include "glm/glm.hpp"
+#include "util/shaderProgram.h"
+
+typedef enum {
+    DEFAULT,
+    VERTEX,
+    FRAGMENT,
+    BOTH
```
blair1618: I'm in favor of removing the 'BOTH' injection mode. The result of this mode is equivalent to adding two identical lights to the scene, one in 'VERTEX' mode and one in 'FRAGMENT' mode. The resulting 'light' will not behave as expected and I don't think there's a good use case for this.

patriciogonzalezvivo: +1

core/resources/lights_vert.glsl
```diff
@@ -0,0 +1,79 @@
...
+    #endif
+#endif
+
+#ifdef TANGRAM_FRAGMENT_LIGHTS
+void calculateLighting(in vec3 _eyeToPoint, in vec3 _normal ) {
+#else
+vec4 calculateLighting(in vec3 _eyeToPoint, in vec3 _normal ) {
+#endif
+
+    vec3 eye = vec3(0.0, 0.0, 1.0);
```
blair1618: Think we can get rid of the 'eye' vector altogether? I think this value is never changed and all of our positions should be relative to the eye anyway.

patriciogonzalezvivo: +1

blair1618:
I'm calling this branch ready for merge. :)

If anybody has concerns that can't wait for a later PR, speak now!

patriciogonzalezvivo:
I do 

; )

core/src/scene/pointLight.cpp
```diff
@@ -0,0 +1,100 @@
...
+    if (s_classBlock.empty()) {
+        s_classBlock = stringFromResource("pointLight.glsl")+"\n";
+    }
+    return s_classBlock;
+}
+
+std::string PointLight::getInstanceDefinesBlock() {
+    std::string defines = "";
+    
+    if (m_constantAttenuation!=0.0) {
```
tallytalwar: Is it possible for all the attenuation types be defined, shouldn't the function when one attenuation definition is set?

core/src/scene/light.h
```diff
@@ -0,0 +1,107 @@
+#pragma once
+ 
+#include <memory>
+#include "glm/glm.hpp"
+#include "util/shaderProgram.h"
+
+typedef enum {
```
tallytalwar: Can we have comments here!.. Specifically for the default behavior.

core/src/scene/pointLight.cpp
blair1618: Not sure what you're asking. Can you clarify?

tallytalwar: Should all m_constantAttenuation, m_linearAttenuation and m_quadraticAttenuation be allowed to be != 0.0.

Ideally we should have only one type of attenuation right?

patriciogonzalezvivo: you can combine them

tallytalwar: Ok, got it. Something like discussed here: http://ogldev.atspace.co.uk/www/tutorial20/tutorial20.html

core/resources/lightFragment.glsl
```diff
@@ -0,0 +1,55 @@
...
+
+    #ifdef TANGRAM_MATERIAL_DIFFUSE
+        color += g_light_accumulator_diffuse * g_material.diffuse;
+    #endif
+
+    #ifdef TANGRAM_MATERIAL_SPECULAR
+        color += g_light_accumulator_specular * g_material.specular;
+    #endif
+
+    color = clamp(color, 0.0, 1.0);
```
karimnaaji: Little tweak, but I'd suggest a simple `color = pow(color, vec4(u_gamma));` with gamma at 1.0/2.2 for gamma correction here, you'd never really want to add lightning not gamma corrected.

blair1618: Ah, I see! I'm actually not very familiar with gamma correction but from a little reading it seems like a good idea. It would be nice to not need another operation in the fragment shader, especially when we have a lot of overdraw in most scenes. Maybe it could be a second pass?

tallytalwar: Yes, gamma correction can be a second pass. The point to note with color space correction is, all shader calculations must happen in linear space. And the results be color corrected in the post shader operation.

karimnaaji: It's a nice idea to do that in a second pass! But POI icons and textures wouldn't probably need it since they won't be in linear space most of the time, so we'd probably need some separation, but for sure it gives more control on the overhaul pipeline. For now, I guess it's good to apply gamma correction on each shader, I know it's pretty tedious, but we need it only for fonts and lightning and we can move it when we'd have the post-process pipeline :)

blair1618: Yeah actually I realized that I was mistaken about overdraw here (Since hidden fragments are depth-tested and discarded, we're really only doing this on the same number of fragments that a full-screen pass would do). I'll add this to the lighting code!

blair1618: I'm gonna leave out gamma correction for this PR. The most efficient way to go about it seems to be creating an sRGB framebuffer that automatically encodes the correction in its storage format, but that'll be a task on its own.

bcamper:
:confetti_ball: :confetti_ball: :confetti_ball:

patriciogonzalezvivo:
rejoice

PR #35, opened 2014-12-17
Cont zoom

tallytalwar:
Tested rigorously specially after a crazy crash I found yesterday.

Summary of changes:

1. Uses non integral zoom
2. Arch specific code now handles pinch gesture continuously.
3. Managing Proxy Tiles:
  1. TileManager owns all proxy tiles (m_ProxyTiles)
    - proxy tiles are added serially when a previously loaded tile (in m_tileSet) has its vbos ready and asynclly when it needs to tesselate from already loaded map tile data.
    - proxy tiles are deleted once the corresponding visible tile finished loading.
  2. Update of tile sets takes care of:
    - Logical deletion of tiles from m_tileSet
    - Addition of new visible tiles to m_tileSet (and addition of corresponding proxy tiles)
    - Deletion of tiles (if not in proxy and logically deleted)

One small todo:
- Right now one gesture of zoom in and then zoom out without lifting fingers results in a zoom in first and then zoom out. So we have to lift fingers to start another gesture whenever we need to switch zoom-in/zoom-out. Will fix this once I figure out a solution for this.


core/src/tile/tileManager.cpp
```diff
@@ -31,114 +31,173 @@ bool TileManager::updateTileSet() {
...
                 }
-                
             }
             
-            
+            return tile;
+        } else {
+            // tile was deleted (went out off view) before it could load its data
+            // remove its proxies also
+            cleanProxyTiles(_id);
```
blair1618: Calling `cleanProxyTiles` from async threads risks concurrent modification, we should fix that before merging

PR #36, opened 2015-01-05
Tessfix

blair1618:
Fixes the crashes caused by libtess2 failing assertions when fed degenerate geometry:

I've added libtess2 as a submodule since it is 1. very small, and 2. usable without modification to source files. The only change to libtess2 we need is to compile with the preprocessor definition "NDEBUG", which disables all assertions. The effect of this is that tesselation output from degenerate geometry is technically undefined, but in practice the result is that there is no visible output for these geometries.


blair1618:
Just realized that I still need to apply this change to the RPI build, we should wait until I've tested that to merge.

blair1618:
Another thought that might not belong in this PR: We now have two libraries with source files in our "include" folder for the core project (jsoncpp and libtess2). Maybe we should move these to another folder like "3rdparty" or "deps" or something.

tallytalwar:
+1 on moving jsoncpp and libtess2 to an appropriate directory.

bcamper:
Sometimes called 'vendor', though that is also a kinda gross word. 'deps'
is probably better (or even dependencies, clearer and you're not gonna type
it that often).

blair1618:
Ready for merge!

PR #37, opened 2015-01-06
Ci

tallytalwar:
Uses TracisCI: https://travis-ci.org/tangrams/tangram-es

Does the following:
- builds tests, runs tests
- builds osx, ios and android platforms.


.travis.yml
```diff
@@ -0,0 +1,35 @@
...
+    # Set the core file limit to unlimited so a core file is generated upon crash
+    - ulimit -c unlimited -S
+
+# build test project
+script: 
+    - source ./travis/script_build_tests.sh
+    - source ./travis/script_run_tests.sh
+    - source ./travis/script_build.sh
+
+branches:
```
tallytalwar: Once merged with master, will remove this directly in master!.

If I do now all pushes on all branches will trigger builds and fail.

travis/before_install.sh
```diff
@@ -0,0 +1,52 @@
...
+    sudo add-apt-repository ppa:ubuntu-toolchain-r/test -y
+    sudo apt-get update -qq
+    if [ "$CXX" = "g++" ]; then sudo apt-get install -qq g++-4.8; fi
+    if [ "$CXX" = "g++" ]; then export CXX="g++-4.8" CC="gcc-4.8"; fi
+    # install other required packages
+    sudo apt-get -y install git build-essential automake gdb libtool xutils-dev make cmake pkg-config libcurl4-openssl-dev
+    # not installing mesa and opengl right now
+
+    # install jdk
+    sudo apt-get -qq -y install openjdk-7-jdk ant lib32z1-dev lib32stdc++6
```
blair1618: Do we need the build host to have libcurl, libz, or libstdc++? The android libraries we're linking are in the NDK. And what's xutils?

tallytalwar: libcurl: I can get rid of this, I initially thought of using curl to download sdk and ndk, but instead used wget.
libz: Would have been required if tar files for ndk was downloaded.
libstdc++: I added this for future, when we have a linux build.
xutils: I added this for future, when we have linux build (I think might be required with glfw3 ... might!)

Will clean this out though.

tallytalwar: And we also do not need to install gcc 4.8 right now, since it uses the gcc packaged with the ndk, but will need it for linux builds in future. I will comment that out also.

PR #38, opened 2015-01-07
Text rendering

karimnaaji:
Mainly the integration of https://github.com/tangrams/fontstash-es, with some really basic label logic in order to draw something.
- Added the font rendering library and two .ttf fonts for testing
- Created a font style to draw some labels and test the fonts rendering, for now only some roads 
- Added the shaders used for fonts, which is needed by the texture transform feature to have one draw call for a group of labels that don't share the same transforms


blair1618:
You even added tests, awesome! Looking forward to getting this merged. 

Are the fontstash files the same as those in the fonstash-es repo? Would that project benefit from the changes you've added here? I don't know if it makes sense, but it would be cool to keep fontstash-es as updated and self-contained as possible so others could use it in a similar way (especially since that repo had some outside interest).

karimnaaji:
Yeah that's what I was talking about with @tallytalwar , we probably need to make a submodule for it to keep track of the changes!

core/resources/text.vs
```diff
@@ -0,0 +1,81 @@
...
+attribute lowp float a_fsid;
+attribute vec2 a_position;
+attribute vec2 a_texCoord;
+
+uniform sampler2D u_transforms;
+uniform lowp vec2 u_tresolution;
+uniform lowp vec2 u_resolution;
+uniform mat4 u_proj;
+
+varying vec2 f_uv;
```
patriciogonzalezvivo: Irrelevant comment: as convention varyings start with v_ instead of f_

karimnaaji: Yes good point! Will change that.

core/resources/text.vs
```diff
@@ -0,0 +1,81 @@
...
+#define ty      tdata.y
+#define theta   tdata.z
+#define txp     tdataPrecision.x
+#define typ     tdataPrecision.y
+
+/*
+ * Converts (i, j) pixel coordinates to the corresponding (u, v) in
+ * texture space. The (u,v) targets the center of pixel
+ */
+vec2 ij2uv(float i, float j, float w, float h) {
```
patriciogonzalezvivo: Also annoying and unnecessary: I was using _ at the beginning of functions.

karimnaaji: No code styling is necessary :) But yeah, I was wondering about it, does it really makes sense to put _ in shaders? Since putting it is usually to differentiate those variable from class members?

patriciogonzalezvivo: very true!

karimnaaji: But i'm going to add this diff too!

karimnaaji:
Branch ready for reviews after a refactoring and the merge of lights, it just needs some more comments and a safe exit.

blair1618:
Labels don't seem to be deleted properly when tiles are removed (or possibly tiles aren't being removed when they should be)

screen shot 2015-02-10 at 12 30 26 pm

karimnaaji:
@blair1618 , yeah this issue is visible because the tile are not yet deleted, to solve this we could not display labels when the view and the tile labels are too distant in term of zoom levels.

blair1618:
Yeah I'm pretty sure that this is an issue with tile management, not label management. I'm guessing we overlooked some edge case in proxy tile management, I'll see if I can find it.

blair1618:
@karimnaaji Is it unsafe to call `removeLabels()` on a tileID that hasn't had labels added? It gave me a SIGABRT for freeing a free pointer

karimnaaji:
This should be safe. I wasn't able to reproduce, where did the SIGABRT happen exactly?

blair1618:
http://pastebin.com/0bYhqdLt

I'll try to reproduce it from xcode

core/src/tile/labels/label.cpp
```diff
@@ -0,0 +1,41 @@
...
+}
+
+Label::~Label() {}
+
+void Label::rasterize() {
+    m_buffer->rasterize(m_text, m_id);
+}
+
+void Label::updateTransform(const LabelTransform& _transform, const glm::mat4& _mvp, const glm::vec2& _screenSize) {
+    m_transform = _transform;
```
tallytalwar: No use of having a LabelTransform argument here.

If ever transforms of a label needs any changing, another method could be used to set that ("setLabelTransform").

PR #39, opened 2015-01-10
Fix a crash occurring when the application exits

blair1618:
Currently, running Tangram from GLFW and closing the app will produce a segmentation fault during the exit process. The same (bad) behavior occurs on every platform, but errors get suppressed on Android and iOS. The reasons for this error are somewhat nuanced, but have been found and resolved (so feel free to skip the following description). 

In order to recover from GL context loss, classes that contain handles to OpenGL objects (`ShaderProgram` and `VboMesh`) previously tracked all instances of themselves in static `std::set`s so that they can be efficiently invalidated when context loss occurs. Instances of these classes add themselves to these `set`s in their constructor and remove themselves from it in their destructor. In typical program operation this works as expected, but it fails when exiting the program. On exit, the main resource containers (tileManager, scene) are automatically destructed, further destructing all of their members all the way down to `ShaderProgram`s and `VboMesh`es. When these destructors are called, we find that the static `set`s for these classes have already been destructed! As it turns out, our main resource containers are globally scoped in the program, meaning that they have static storage duration. All objects with static storage duration persist for the entire main function, but the order of destruction of these objects is undefined between different compilation units (source files). The result is that these instances of `ShaderProgram` and `VboMesh` weren't getting destructed until after other parts of static memory (namely, the static `set`s in their classes). 

The simplest solution was to modify our context-loss-recovery strategy to not require tracking all instances of those classes. I modified both classes to instead statically track the number of times the GL context has been invalidated and check with each use whether they are 'up to date'.

### DIFF
--- a/core/src/tangram.cpp
+++ b/core/src/tangram.cpp
@@ -16,13 +16,13 @@
 
 namespace Tangram {
 
-std::unique_ptr<TileManager> m_tileManager;    
+std::unique_ptr<TileManager> m_tileManager;
 std::shared_ptr<Scene> m_scene;
 std::shared_ptr<View> m_view;
-
+    
 void initialize() {
     
-    logMsg("%s\n", "initialize");
+    logMsg("initialize\n");
 
     // Create view
     if (!m_view) {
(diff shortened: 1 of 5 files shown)


tallytalwar:
Good to merge.

Will have to add teardown usage on ios for consistency and we will most likely be using teardown in future. That being said, we have already discussed this and have a plan for its implementation.

Merging..

PR #43, opened 2015-01-16
Full-screen display on iOS devices

karimnaaji:
iOS apps need a splashscreen in order to display full screen on iOS 8+ (no black border). Added a splash screen .xib file and modified .plist to target it so that the os knows at which size the gl context should be created at.


blair1618:
@karimnaaji if you have some time today see if you can get the images from 34192d6 to display at launch

karimnaaji:
@blair1618 sure

blair1618:
Build is fine locally; Travis just failed to download GLFW for unknown reasons. Merging!

PR #44, opened 2015-01-19
Fix wrong glm inclusion

karimnaaji:
Including all the subdirectory in the darwin cmake toolchain for homebrew-installed package [here](https://github.com/tangrams/tangram-es/blob/master/toolchains/darwin.cmake#L9), results in wrong inclusion by the preprocessor in the case the user has it installed on its machine, since this happens before the core/include inclusion. To solve it, we should only include the brew dependency we need in the project.

### DIFF
--- a/core/src/gl.h
+++ b/core/src/gl.h
@@ -11,7 +11,7 @@
 #endif
 
 #ifdef PLATFORM_OSX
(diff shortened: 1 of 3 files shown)


blair1618:
Looks good

PR #45, opened 2015-01-19
Submodule glm

tallytalwar:


blair1618:
Seems like everything works here. Just 2 points:

1. My hesitation with making GLM a submodule earlier was that the whole GLM repo is about 15MB of stuff, of which we only need 2.5MB. This only bloats our repo (and has no effect on executables) but I wanted to point out that side effect. 

2. Would it make more sense to keep GLM in the include folder? I'd been separating "dependencies" and "include" as "things that need to be compiled" and "things that just need to be included".

PR #46, opened 2015-01-21
Style setup refactor

blair1618:
Splits style setup into a per-frame setup and a per-tile setup (which are implemented as no-ops in the base style class)


blair1618:
@karimnaaji @tallytalwar Does this make sense?

karimnaaji:
Yes definitely!

tallytalwar:
Yes, looks good.

blair1618:
Travis is failing on ndk extraction but I just tested on Android and all seems well

PR #47, opened 2015-01-22
Checking for geometry for a specific style

tallytalwar:
- change for more readibility. tile->draw should not be called for styles for which there is no geometry/vbo in the
tile.


blair1618:
Is this different from the check that happens here: https://github.com/tangrams/tangram-es/blob/master/core/src/tile/mapTile.cpp#L46 ?

tallytalwar:
Nopes its the same, but it makes more sense to have tile-draw called for something which exists... and yes I can remove the check inside the draw call!

tallytalwar:
AT the same time, another argument can be made against this PR, that "find" is called twice for a draw call then!

PR #48, opened 2015-01-23
Matrix optimizations

blair1618:
By separating the global translations from the model and view matrices, we can store these matrices in single precision values instead of double precision (and prevent looping through matrices and casting like we currently do). This also adds some comments to clarify the usage of the relevant matrices.

### DIFF
--- a/core/src/tile/mapTile.cpp
+++ b/core/src/tile/mapTile.cpp
@@ -16,13 +16,10 @@ MapTile::MapTile(TileID _id, const MapProjection& _projection) : m_id(_id),  m_p
     // negative y coordinate: to change from y down to y up (tile system has y down and gl context we use has y up).
     m_tileOrigin = glm::dvec2(0.5*(bounds.x + bounds.z), -0.5*(bounds.y + bounds.w));
     
-    m_modelMatrix = glm::dmat4(1.0);
-    
-    // Translate model matrix to origin of tile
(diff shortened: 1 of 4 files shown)


core/src/tile/mapTile.cpp
```diff
@@ -47,13 +44,15 @@ void MapTile::draw(const Style& _style, const View& _view) {
 
         std::shared_ptr<ShaderProgram> shader = _style.getShaderProgram();
 
-        glm::dmat4 modelViewProjMatrix = _view.getViewProjectionMatrix() * m_modelMatrix;
+        // Apply tile-view translation to the model matrix
+        const auto& viewOrigin = _view.getPosition();
+        m_modelMatrix[3][0] = m_tileOrigin.x - viewOrigin.x;
```
tallytalwar: double to float, will there be any precision loss here?

blair1618: Technically yes, but this is essentially the same as our current implementation (we converted to floats [here](https://github.com/tangrams/tangram-es/pull/48/files#diff-1563124aa3d1828325c97ec833bb3c88L54)). The idea with this is that these translation values are individually too large to represent well with floats, but their difference is perfectly fine to keep as a float.

tallytalwar: Yup, sounds good. I know we did discuss doing something like this when we found that the tesselator only handles floats. :D

PR #49, opened 2015-01-23
Make styling functions const to forbid potential concurrent modification

blair1618:
This may become irrelevant after future revisions to the tile building process; I expect that when we have a pool of workers, each worker can hold a copy of any data needed for the building process so concurrent access won't be a problem. However, for now this enforces that Style members won't be modified inside functions that are called from multiple threads.

### DIFF
--- a/core/src/style/polygonStyle.cpp
+++ b/core/src/style/polygonStyle.cpp
@@ -35,11 +35,11 @@ void PolygonStyle::setupFrame() {
     m_shaderProgram->setUniformf("u_lightDirection", -1.0, -1.0, 1.0);
 }
 
-void PolygonStyle::buildPoint(Point& _point, std::string& _layer, Properties& _props, VboMesh& _mesh) {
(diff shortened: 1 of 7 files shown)


tallytalwar:
I am good with this :D.

However I do not understand your point here "I expect that when we have a pool of workers, each worker can hold a copy of any data needed for the building process so concurrent access won't be a problem".

I do not expect any of these builder methods to modify any of internal data, so should forever be consts.. correct me if I am wrong. :D

karimnaaji:
Seems good, I think it's the right way to solve the issue I pointed out! And yeah they shouldn't modify internal data, except in the case of the fontStyle since glyph rasterization needs the context to produce vertices for the mesh.. Also the atlas gets updated in cpu on the fly when new glyph get rasterized. Any idea concerning this? Since those two are members of the style, should it be something the style doesn't own, and what should be the flow in that case?

blair1618:
The case of font styles is tricky, there may be no way around that except to mutex lock those resources when they're needed. This means that we might not be able to uphold those `const` declarations..

PR #50, opened 2015-01-28
Linux build

blair1618:
Very similar to the RPi build, but using GLFW for windowing and input. Also shares a lot with the OS X code. At some point I'd like to combine the OS X and Linux platforms such that they share the same source files.


tallytalwar:
Can you also update travis.yml to do a linux build also?

blair1618:
Yeah, adding a Linux build to Travis probably wouldn't hurt. Not sure how best to add it to the yaml file though, currently we treat anything running on "Linux" as if it's doing the Android build.

blair1618:
Let's wait on merging this until after we merge the PR that fixes Travis builds, then I can add the linux build to our Travis config and make sure it works.

PR #51, opened 2015-01-29
Rapidjson

blair1618:
RapidJson replaces JsonCPP, our current Json parser. From brief profiling on an iPhone, I measured rapidjson to be roughly 7.5 times faster at parsing a map scene in Manhattan. 

RapidJson is also a header-only library, which slightly simplifies our build process compared to JsonCPP.


tallytalwar:
This looks good, though failing on travic-android, because of the same android ndk extraction. Merging.

PR #52, opened 2015-01-30
Travis android osx

blair1618:
Building Android on an OS X machine seems to resolve the issue we've been having where Travis fails when extracting the NDK archive (extraction has been successful all 3 times it's run).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,9 @@ matrix:
           env: PLATFORM=osx
           compiler: clang
         - os: osx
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -6,8 +6,10 @@ language: cpp
 matrix:
     include:
         - os: osx
+          env: PLATFORM=osx
```
tallytalwar: Right now osx platform will build for osx and ios, should we add another platform to separate osx and ios?

PR #53, opened 2015-01-31
Metameshes

blair1618:
This allows us to add over 65535 vertices to a VboMesh while still using indexed geometry. 

A single vertex buffer and index buffer pair can't draw more than a fixed number of vertices because the index type in OpenGL ES is an unsigned 16-bit short, whose maximum value is 65535. To support larger collections of vertices, this changes modifies the internal behavior of VboMesh to instantiate additional vertex and index buffers when needed. 

VboMesh now holds a smart pointer to a "backup" VboMesh, which is allocated only when the containing mesh runs out of index space. The original mesh behaves no differently from a mesh with < 65535 vertices and the backup mesh can recursively instantiate further backups, if needed. 

In order to make use of the larger capacity of these meshes, the index type used elsewhere in the geometry pipeline had to be changed from `unsigned short` to `int` (`unsigned int` was a possibility as well, but `int` is the  output index type for our tesselator, so `int` prevents an additional cast).


core/src/util/builders.cpp
```diff
@@ -7,20 +7,20 @@
...
-void Builders::buildPolygon(const Polygon& _polygon, std::vector<glm::vec3>& _pointsOut, std::vector<glm::vec3>& _normalOut, std::vector<ushort>& _indicesOut, std::vector<glm::vec2>& _texcoordOut) {
+void Builders::buildPolygon(const Polygon& _polygon, std::vector<glm::vec3>& _pointsOut, std::vector<glm::vec3>& _normalOut, std::vector<int>& _indicesOut, std::vector<glm::vec2>& _texcoordOut) {
     
     TESStesselator* tesselator = tessNewTess(nullptr);
     
     bool useTexCoords = &_texcoordOut != &NO_TEXCOORDS;
     
     // get the number of vertices already added
-    ushort vertexDataOffset = (ushort)_pointsOut.size();
+    int vertexDataOffset = _pointsOut.size();
```
tallytalwar: This will result in a warning for inconsistent type. size_t :D.

core/src/util/vboMesh.cpp
```diff
@@ -68,15 +68,18 @@ void VboMesh::addVertex(GLbyte* _vertex) {
...
+        logMsg("ERROR: VboMesh cannot add vertices after upload!\n");
         return;
     }
     
     // Only add up to 65535 vertices, any more will overflow our 16-bit indices
     int indexSpace = MAX_INDEX_VALUE - m_nVertices;
-    if (_nVertices > indexSpace) {
-        _nVertices = indexSpace;
-        logMsg("WARNING: Tried to add more vertices than available in index space\n");
+    if (_nVertices > indexSpace || m_backupMesh) {
```
tallytalwar: If ever we got to a dataset with any iteration of addVertices adding vertices more than "indexSpace" then it will be a infinite loop of m_backupMeshes. We should have a catch for that.

blair1618: Yeah good catch. We can just crop any (gigantic) single polygons that won't
fit in a mesh.

core/src/util/vboMesh.cpp
```diff
@@ -68,15 +68,22 @@ void VboMesh::addVertex(GLbyte* _vertex) {
...
+        if (!m_backupMesh) {
+            m_backupMesh.reset(new VboMesh(m_vertexLayout, m_drawMode));
+        }
+        if (_nVertices > MAX_INDEX_VALUE) {
```
tallytalwar: This check should happen before backupMesh is created. @blair1618 is looking into this.

blair1618: Corrected by cf6b242

PR #55, opened 2015-02-03
Configure libtess2 allocator

blair1618:
By providing more sensible values for allocation sizes in libtess2, polygon tesselation is made roughly 2 times faster (in a city scene). 

Further optimization would include:
1. Extended experimentation with these values
2. Changing the allocation functions to use pooled memory

### DIFF
--- a/core/src/util/builders.cpp
+++ b/core/src/util/builders.cpp
@@ -4,8 +4,31 @@
 #include "rectangle.h"
 #include "geom.h"
 
+#include <memory>
+
 static auto& NO_TEXCOORDS = *(new std::vector<glm::vec2>); // denotes that texture coordinates should not be used
 static auto& NO_SCALING_VECS = *(new std::vector<glm::vec2>); // denotes that scaling vectors should not be used
+
+void* alloc(void* _userData, unsigned int _size) {
+    return malloc(_size);
+}
+
+void* realloc(void* _userData, void* _ptr, unsigned int _size) {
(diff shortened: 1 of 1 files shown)


core/src/util/builders.cpp
```diff
@@ -4,9 +4,32 @@
...
+
 static auto& NO_TEXCOORDS = *(new std::vector<glm::vec2>); // denotes that texture coordinates should not be used
 static auto& NO_SCALING_VECS = *(new std::vector<glm::vec2>); // denotes that scaling vectors should not be used
 
+void* alloc(void* _userData, unsigned int _size) {
+    return malloc(_size);
+}
+
+void* realloc(void* _userData, void* _ptr, unsigned int _size) {
+    return realloc(_ptr, _size);
```
karimnaaji: I removed all realloc from fontstash-es because they were really slow, I think it could be nice allocate a memory chunk at the beginning and test in this method if the size goes over it, then just use malloc in that case.

blair1618: Yeah it might be nice to not use the default allocators at all and just create a pool of memory for the tesselator to draw from, but that might add a lot complexity.

blair1618: The tessAlloc struct also allows passing nullptr for the realloc function, which tells libtess2 to not use realloc. I tried that in the configuration but it didn't make a measurable difference in my tests, so I left realloc in.

PR #56, opened 2015-02-04
Protobuf

tallytalwar:
Parsing speedup: 15-20% compared to rapid json
Network access speedup: 50%


blair1618:
There's some spurious geometry being generated here, haven't determined the cause yet.

Protobuf:
protobuff

Master (rapidjson):
master

tallytalwar:
Interesting. Looks like more polys...

core/src/util/pbfParser.cpp
```diff
@@ -0,0 +1,241 @@
...
+    pbfGeomCmd cmd = pbfGeomCmd::moveTo;
+    uint32_t cmdRepeat = 0;
+    protobuf::message geomItr = _in;
+    
+    double invTileExtent = (1.0/(double)_tileExtent);
+    
+    std::vector<Point> line;
+    
+    int x = 0.0f;
+    int y = 0.0f;
```
blair1618: ?

core/src/util/pbfParser.cpp
```diff
@@ -0,0 +1,251 @@
...
+                            break;
+                        case 7: // bool value
+                            numericValues[valueCount] = valueItr.boolean();
+                            break;
+                        default:
+                            valueItr.skip();
+                            break;
+                    }
+                }
+                valueCount++;
```
blair1618: Seems like this should be in the scope immediately above this

PR #57, opened 2015-02-05
Outlines

patriciogonzalezvivo:


blair1618:
Isn't there a polyline branch that this depends on? I think we should merge that first

patriciogonzalezvivo:
I thought we already did that

blair1618:
Nope, there hasn't been a PR made for the polylines branch yet

patriciogonzalezvivo:
Master was merged on Polyline. Polyline is on PR. Polyline was merged on outline.
In Polygon Style I have left commented a water outline example for testing

core/src/util/builders.h
```diff
@@ -5,82 +5,72 @@
...
+    MITER = 0, // No points added at line join
+    BEVEL = 1, // One point added to flatten the corner of a join
+    ROUND = 5 // Five points added in a fan to make a round outer join
+};
+
+struct PolyLineOptions {
+    CapTypes cap;
+    JoinTypes join;
+    float halfWidth;
+};
```
tallytalwar: Need to have a constructor definition!

A PolyLineOptions is initialized with default values in PolygonStyle::buildLine and since halfWidth has no default value, it will result inconsistent behavior.

core/src/util/builders.cpp
```diff
@@ -171,143 +154,237 @@ void Builders::buildPolygonExtrusion(const Polygon& _polygon, const float& _minH
...
+// Get 2D perpendicular of two points
+glm::vec2 perp2d(const glm::vec3& _v1, const glm::vec3& _v2 ){
+    return glm::vec2(_v2.y - _v1.y, _v1.x - _v2.x);
+}
+
+// Get 2D vector rotated 
+glm::vec2 rotate(const glm::vec2& _v, float _radians) {
+    float cosine = cos(_radians);
+    float sine = sin(_radians);
+    return glm::vec2(_v.x * cosine - _v.y * sine, _v.x * sine + _v.y * cosine);
```
tallytalwar: use of glm::rotate would be better.

blair1618: Fixed a0bd3a5

core/src/util/builders.h
blair1618: Looked into this; the default constructor generated by the compiler will initialize all non-static members to the default values for their types (as per the C++ spec), so we actually don't need to do anything

blair1618: Realized that defaulting to 0 width for lines could be potentially super-confusing, so added an explicit default constructor with non-zero width (e30ee1a)

PR #68, opened 2015-03-12
Tilt

blair1618:
Changes:

 - Adds camera 'pitch', allowing views at a range of angles to the z-axis
 - Adds a gesture (two-finger 'shove') to control camera pitch
 - Adapts existing input gestures (pan, rotate) to maintain their physically intuitive behavior. Panning will move the view such that the point on the ground plane that started under your finger will remain under your finger. Rotating will orbit the view around the central point of rotation. 
 - Adds a simplistic tile coverage algorithm that overestimates visible tiles, but fully covers the viewable area with a tilted view. 

Future:

 - The tile coverage algorithm will be improved in a subsequent PR. For the next implementation we will want to cover the viewable area with tiles of varying zoom levels, resulting in lower-detail geometry for more distant parts of the viewable area.

### DIFF
--- a/android/src/com/mapzen/tangram/Tangram.java
+++ b/android/src/com/mapzen/tangram/Tangram.java
@@ -37,7 +37,7 @@ public class Tangram extends GLSurfaceView implements Renderer, OnScaleGestureLi
     private static native void handleDoubleTapGesture(float posX, float posY);
     private static native void handlePanGesture(float startX, float startY, float endX, float endY);
     private static native void handlePinchGesture(float posX, float posY, float scale);
-    private static native void handleRotateGesture(float rotation);
+    private static native void handleRotateGesture(float posX, float posY, float rotation);
     private static native void handleShoveGesture(float distance);
 
     private long time = System.nanoTime();
@@ -191,7 +191,10 @@ public class Tangram extends GLSurfaceView implements Renderer, OnScaleGestureLi
     }
 
     public boolean onRotate(RotateGestureDetector detector) {
-        handleRotateGesture(-detector.getRotationDegreesDelta() * (float)(Math.PI / 180));
+        float x = scaleGestureDetector.getFocusX();
+        float y = scaleGestureDetector.getFocusY();
+        float rotation = -detector.getRotationDegreesDelta() * (float)(Math.PI / 180);
+        handleRotateGesture(x, y, rotation);
         return true;
     }
 
--- a/android/src/com/mapzen/tangram/Tangram.java
+++ b/android/src/com/mapzen/tangram/Tangram.java
@@ -202,7 +202,7 @@ public class Tangram extends GLSurfaceView implements Renderer, OnScaleGestureLi
         return;
     }
 
-    // ShoveGestureDetecrot.OnShoveGestureListener methods
+    // ShoveGestureDetector.OnShoveGestureListener methods
     // ===================================================
 
     public boolean onShoveBegin(ShoveGestureDetector detector) {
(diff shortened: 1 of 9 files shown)


android/src/com/mapzen/tangram/Tangram.java
```diff
@@ -130,11 +136,16 @@ public boolean onDoubleTap(MotionEvent event) {
...
     public boolean onScroll(MotionEvent e1, MotionEvent e2, float distanceX, float distanceY) {
-        
-        // We flip the signs of distanceX and distanceY because onScroll provides the distances
-        // by which the view being scrolled should move, while handlePanGesture expects the 
-        // distances by which the touch point has moved on the screen (these are opposite)
-        handlePanGesture(-distanceX, -distanceY);
+        // Only pan for scrolling events with just one pointer; otherwise vertical scrolling will
+        // cause a simultaneous shove gesture
+        if (e1.getPointerCount() == 1 && e2.getPointerCount() == 1) {
+            // We flip the signs of distanceX and distanceY because onScroll provides the distances
```
tallytalwar: Dont think this comment is applicable now considering both start and end positions are used.

android/src/com/mapzen/tangram/Tangram.java
```diff
@@ -180,12 +191,31 @@ public boolean onRotateBegin(RotateGestureDetector detector) {
...
+        float rotation = -detector.getRotationDegreesDelta() * (float)(Math.PI / 180);
+        handleRotateGesture(x, y, rotation);
         return true;
     }
 
     public void onRotateEnd(RotateGestureDetector detector) {
         return;
     }
+
+    // ShoveGestureDetecrot.OnShoveGestureListener methods
```
tallytalwar: nitpick ... "ShoveGestureDetector"

core/src/view/view.cpp
```diff
@@ -128,23 +152,24 @@ glm::dmat2 View::getBoundsRect() const {
...
+    float t; // Distance along ray to ground plane
+    if (ray_world.z != 0.f) {
+        t = -m_pos.z / ray_world.z;
+    } else {
+        t = 0;
+    }
     
-    _screenX = x;
-    _screenY = y;
+    ray_world *= t;
```
tallytalwar: I know this will never happen considering we are restricting how much we can shove!!... but should we put in a check for t being positive? Negative t values mean ray will never intersect the ground plane (which is when world is upside down :D)

core/src/view/view.cpp
```diff
@@ -103,6 +112,21 @@ void View::roll(float _droll) {
...
+
+    setPitch(m_pitch + _dpitch);
+
+}
+
+void View::orbit(float _x, float _y, float _radians) {
+    
+    glm::vec2 radial = { _x, _y };
+    glm::vec2 displacement = glm::rotate(radial, _radians) - radial;
+    translate(-displacement.x, -displacement.y);
```
tallytalwar: This is good. :D

core/src/view/view.cpp
```diff
@@ -176,37 +201,63 @@ void View::updateMatrices() {
...
     
     float tileSize = 2 * MapProjection::HALF_CIRCUMFERENCE * pow(2, -(int)m_zoom);
     float invTileSize = 1.0 / tileSize;
     
-    // Adjust width and height for view rotation
-    float cos_r = cos(m_roll);
-    float sin_r = sin(m_roll);
-    float width = fabsf(m_height * sin_r) + fabsf(m_width * cos_r);
-    float height = fabsf(m_width * sin_r) + fabsf(m_height * cos_r);
+    // Find bounds of view frustum in world space (i.e. project view frustum onto z = 0 plane)
```
tallytalwar: Rename variables appropriately to give more contextual explanation.

tallytalwar:
Looks good aparts from the comments I made... good to merge once changes are done :D.

PR #71, opened 2015-03-18
Tile coverage

blair1618:
View is now able to accurately determine the tiles that are visible in the current view frustum (instead of approximating with a bounding box)


blair1618:
Don't merge this yet, I'm actually reconsidering the approach I had planned for tile coverage and LOD and I want to try an approach using the scan-line technique before we move ahead with this.

blair1618:
Closing in favor of https://github.com/tangrams/tangram-es/pull/83

PR #77, opened 2015-03-31
Ok http

tallytalwar:
- First version to use okHttp for android platform instead of curl.
- Abstracted out networking from dataSource. 
- Each platform now implements "fetchData" method to do network tasks.
- jni to patch OkHttpInterface class and native code
- OkHttpInterface class which handles network operations using synchronous okHttp library. 
  - Receives data in a bytearray to handle both geojson and protobuf formats.

Todos:
- Async okhttp
- Network caching
- Network call cancellation
- Decoupling network calls from native worker threads.


core/src/platform.h
```diff
@@ -33,3 +38,6 @@ std::string stringFromResource(const char* _path);
  * allocated file
  */ 
 unsigned char* bytesFromResource(const char* _path, unsigned int* _size);
+
+bool fetchData(std::unique_ptr<std::string> _url, std::stringstream& _rawData);
```
blair1618: "fetchData" isn't the most descriptive, maybe something like "stringFromURL" or "streamFromURL"

tallytalwar: Hmm!.

We can do streamFromURL as it is string in case of json source and binary data in case of protobuf.
But I feel fetchData is still fine, because this is making the actual http request to fetch the desired data (string or binary)

blair1618: Using a `unique_ptr` for the url seems unnecessary, in every implementation you simply grab the `c_str()` so you may as well pass one in and make it less verbose

tallytalwar: Sure. I didnt modify how url was being constructed, hence this. WIll change this. Thanks

blair1618: I agree it's definitely "fetching data", I just think it can be more descriptive. "streamFromHttpSync" would be even better because now it's immediately clear that this is going over the network using http, getting a stream, and running synchronously

tallytalwar: Sure! Consider it changed :D.

tallytalwar:
Closing in favor of https://github.com/tangrams/tangram-es/tree/okHTTP-async branch.

PR #79, opened 2015-04-03
Reallocless

hjanetzek:
This one needs a bit of review I guess. The basic idea is to move the vertices and indices as they are to the mesh and only copy into a continuous buffer once before upload.

### DIFF
--- a/core/src/style/debugStyle.cpp
+++ b/core/src/style/debugStyle.cpp
@@ -1,6 +1,7 @@
 #include "debugStyle.h"
 
 #include "tangram.h"
+#include "rawVboMesh.h"
 
 #include <vector>
 #include <memory>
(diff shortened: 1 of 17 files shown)


blair1618:
Nice idea, I'm measuring a good boost to tile construction times using this. Two things I thought about as I was reviewing the changes:

1. Adding the 'compile' step to VBO uploading puts a bit more work on the UI thread when VBOs are first drawn. Since this step doesn't actually require any GL calls it'd be ideal to perform this on the background thread as well.

2. Creating multiple nearly-identical 'Mesh' classes in each calling class is a bit unwieldy. As I'm discovering, there isn't an obvious way to handle the polymorphism of the vertex structs, but I'll poke around a bit more to see if there's a cleaner way to go about this. 

Gonna play around with these things a bit more, but hope to get this merged soon :)

hjanetzek:
Thanks for the comments :)  I'm currently refreshing my c++, so some advice is appreciated 

To me it look quite a bit better now with using templates just for the subclasses.
Not sure about the validity check in VboMesh though

Issue #81, opened 2015-04-06
Rpi build references g++-4.7 directly instead of testing for a C++11 compiler


xorgy:
I am running a Raspberry Pi with GCC 4.9 instead of GCC 4.7, and had to manually switch `g++-4.7` in `toolchains/raspberrypi.toolchain.cmake` to just `g++`.

Ideally you would use some sort of C++11 detection in CMake, or at least tell people that they'll need to do this to build with anything other than g++ 4.7.

karimnaaji:
Hi,
Thanks for reporting,

xorgy:
That hits the spot, thanks.

PR #84, opened 2015-04-06
Update toolchain build for Raspberry Pi #81

karimnaaji:
Don't default the compiler version to g++4.7. This was done for c++11 compiler compatibility on the Raspberry pi platform.
Later, it would be interesting to move to above CMake 3.1, maybe only on the raspberry pi platform, and use [this property](http://www.cmake.org/cmake/help/v3.1/prop_tgt/CXX_STANDARD.html#prop_tgt:CXX_STANDARD) instead.


toolchains/raspberrypi.cmake
```diff
@@ -1,11 +1,26 @@
...
+execute_process(COMMAND ${CMAKE_CXX_COMPILER} -dumpversion OUTPUT_VARIABLE GCC_VERSION)
+
+if("${CMAKE_CXX_COMPILER_ID}" MATCHES "GNU")
+	if(NOT (GCC_VERSION VERSION_GREATER 4.7 OR GCC_VERSION VERSION_EQUAL 4.7))
+		message(FATAL_ERROR "Please install g++ version 4.7 or greater")
+	else()
+		set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
+	endif()
+elseif("${CMAKE_CXX_COMPILER_ID}" MATCHES "Clang")
+		set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -stdlib=libc++")
```
blair1618: Did you try compiling with Clang and libc++? I'd be interested if that worked because I've had trouble getting Android to build with Clang

karimnaaji: I assumed it'd have been working for clang++ but it's actually not for some reasons. I'm going to remove it from possibilities for now.

blair1618:
Bummer that clang doesn't work, I'd prefer using clang and libc++ on all platforms but we'll have to figure that out another time. Otherwise, looks good!

PR #85, opened 2015-04-10
zoom by point

hanchao:
handle double tap and pinch zoom point


blair1618:
Looks good, thanks :)

Double-tap gesture recognition isn't quite handled correctly on the Android and GLFW clients but we intend to update that before too long and this seems like a good behavior to aim for.

PR #88, opened 2015-04-15
Debug text tile infos

karimnaaji:
Add a way to display some tile debug infos with text, defaulting to tile ids but could be any tile informations (loading time, number of labels...)

screen shot 2015-04-15 at 4 55 09 pm


blair1618:
Looks good! Small question: on my machine the tile info labels aren't centered, the left edge of the labels are in the center of the tile. Are labels always anchored by one of their left corners?

tallytalwar:
I believe thats true (at least what Karim told me last). He can confirm.

karimnaaji:
Yes! it's anchored by lower left corner, it's misleading because the image comes from some test on the upcoming logic on label-occlusion branch where text are centered.

karimnaaji:
I'm planning on making a configurable anchor point anyway

Issue #80, opened 2015-04-03
Label collision detection


hanchao:
![collision](https://cloud.githubusercontent.com/assets/1942091/6977327/91eb4c88-d9ed-11e4-8ea4-d0c9e8606d22.png)

tallytalwar:
Hi Han.

Thanks for trying our cross platform map renderer. 

Tangram-es is currently in development and in no way feature complete.
This specific feature is being worked on at present by @karimnaaji, a developer on our team.

Thanks

PR #89, opened 2015-04-16
Label occlusion-culling

karimnaaji:
Naive label occlusion-culling using two phases from [isec2d](https://github.com/tangrams/isect2d/tree/bvh) repo:
- broad phase collision test using labels AABBs (bruteforced)
- narrow phase collision test using labels OBBs and SAT

There is a lot of improvements to do on the broad phase using an appropriate acceleration structure or sweep and prune but this first implementation already gives correct interactive results.

Known issue : culled labels don't go from culled state to non-culled state
Fixes https://github.com/tangrams/tangram-es/issues/80


blair1618:
I was trying to test this out and I got an error cloning the isect2d submodule

```
Permission denied (publickey).
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

Have you seen this before?

blair1618:
Alright I fixed this by running a `git submodule deinit` and then a `git submodule init`. Git must've cached something behind the scenes.

karimnaaji:
Nice!

core/resources/sdf.fs
```diff
@@ -11,25 +11,42 @@ uniform LOWP vec3 u_color;
...
-    a1 = pow(a1, 1.0 / gamma);
-    a2 = pow(a2, 1.0 / gamma);
+float sample(in vec2 uv, float w, in float off) {
+    return contour(texture2D(u_tex, uv).a, w, off);
+}
 
-    float d = clamp(distance / (1.0 - distance), 0.0, 1.0);
+float sampleAlpha(in vec2 uv, float distance, in float off) {
+    float dscale = 0.354; // 1 / sqrt(2)
+    vec2 duv = dscale * (dFdx(uv) + dFdy(uv));
```
blair1618: These derivatives require the "GL_OES_standard_derivatives" extension to be enabled. It's a common extension (available on all iOS GPUs, notably) but it's not enabled by default on Android so we'd need to check/enable that.

core/resources/sdf.fs
```diff
@@ -11,25 +11,42 @@ uniform LOWP vec3 u_color;
...
+    vec4 box = vec4(uv - duv, uv + duv);
+    float alpha = contour(distance, distance, off);
+    
+    float asum = sample(box.xy, distance, off)
+               + sample(box.zw, distance, off)
+               + sample(box.xw, distance, off)
+               + sample(box.zy, distance, off);
```
blair1618: The issue I brought up yesterday regarding texture coordinates modified in the shader is evidently called "dependent texture reads" and it's explained [here](https://developer.apple.com/library/ios/documentation/3DDrawing/Conceptual/OpenGLES_ProgrammingGuide/BestPracticesforShaders/BestPracticesforShaders.html) in the section labeled "Be Aware of Dynamic Texture Lookups"

karimnaaji: Right, I remember reading this; but those can't be moved to the vertex shader because of the derivatives function used.. I think the best solution is to make it configurable and use the injection system to activate it if needed (since it has no cost on hardware that could use opengl es 3.0).

core/resources/sdf.fs
```diff
@@ -5,31 +5,55 @@ precision mediump float;
...
 
-float minOutlineD = 0.25;
-float maxOutlineD = 0.35;
-float minInsideD = 0.38;
-float maxInsideD = 0.55;
-float mixFactor = 0.7;
-const vec3 outlineColor = vec3(0.7);
-
 const float gamma = 2.2;
+const float tint = 1.8;
```
tallytalwar: @karimnaaji can you explain what tint is used for?

PR #90, opened 2015-04-16
No raw vbo mesh

tallytalwar:
Removed usage of RawVboMesh from fontStyle.

Currently because of how fonstash returns the vertex data, vertex data is copied from std::vector<float> to std::vector<PosTexID>... 

These change does not result in any extra copy (just moves the vertex data copy happening rawVboMesh compile to FontStyle).


karimnaaji:
Will follow up on this and update the signatures of the fontstash api to avoid extra copies.

PR #91, opened 2015-04-18
Tile zoom reduction for oblique views

blair1618:
Views with large 'pitch' angles cover a very large area of tiles relative to a top-down view. Loading and rendering tiles at a uniform zoom level for the entire view is wasteful because tiles that are distant from the view point only cover a very small area of the screen. 

This PR changes the visible tile set such that the relative zoom level of tiles is a configurable function of distance from the viewpoint. The visible tile calculation ensures that tiles visible in a top-down view are unchanged by this distance function and that no overlapping tiles will be added to the set. 

TODO:

 - ~~Tile visibility behaves unexpectedly in a small range of pitch angles just before the lower edge of the view frustum goes above the horizon~~
 - ~~Panning input is discontinuous when crossing the horizon~~
 - Tapping input behaves unexpectedly when tapping above the horizon


core/src/view/view.cpp
```diff
@@ -317,11 +343,54 @@ void View::updateTiles() {
...
+
+    // Location of the view center in tile space
+    glm::dvec2 e = (glm::dvec2(m_pos.x, m_pos.y) - tileSpaceOrigin) * tileSpaceAxes;
+    
+    // Determine zoom reduction for tiles far from the center of view
+    double tilesAtFullZoom = std::max(m_width, m_height) * invTileSize * 0.5;
+    double viewCenterX = (m_pos.x + hc) * invTileSize;
+    double viewCenterY = (m_pos.y - hc) * -invTileSize;
+    
+    int x_l_pos[MAX_LOD] = { 0 };
```
tallytalwar: Logic is great!!

Just a small concern with the naming of these lod threshold maps. Can you name these appropriately?

PR #95, opened 2015-04-23
Lights

patriciogonzalezvivo:


core/src/scene/light.h
```diff
@@ -4,20 +4,21 @@
...
     DIRECTIONAL,
     POINT,
     SPOT,
     CUSTOM
 };
 
+enum class LightOrigin {
+    CAMERA,
+    GROUND,
+    WORLD
```
blair1618: What would the difference between "ground" and "world" origins be?

bcamper: `ground` is like a point dropped from the current camera location on the ground:

https://github.com/tangrams/tangram/wiki/Lights-Overview#origin

Perhaps it should be a cast ray intersecting the ground plane in this case? On web, it is (currently) always looking straight down...

patriciogonzalezvivo: GROUND is the camera position (in XY) projected on the floor with Z is on the ground
WORLD is absolute world position

More info at https://github.com/tangrams/tangram/wiki/Lights-Overview#origin

blair1618: I see, thanks. For the first pass we can take the easy route and drop straight down from the camera. We should experiment with projecting from the center from the screen onto the ground too, it's not obvious to me a priori that one is always better.

patriciogonzalezvivo: https://github.com/tangrams/tangram-es/blob/lights_origin/core/src/scene/pointLight.cpp#L54

blair1618:
Looks good to me. @tallytalwar you had a chance to look over this?

PR #103, opened 2015-05-04
Lights origin

patriciogonzalezvivo:
Fixing lights behavior to a origin (camera/ground/world)


core/src/scene/directionalLight.cpp
```diff
@@ -19,10 +19,19 @@ void DirectionalLight::setDirection(const glm::vec3 &_dir) {
...
 
-void DirectionalLight::setupProgram( std::shared_ptr<ShaderProgram> _shader ) {
+void DirectionalLight::setupProgram(const std::shared_ptr<View>& _view, std::shared_ptr<ShaderProgram> _shader ) {
+
+    m_direction_eye = m_direction;
+    if (m_origin != LightOrigin::CAMERA) {
+        glm::mat3 normalMatrix = glm::mat3(_view->getViewMatrix()); // Transforms surface normals into camera space
+        normalMatrix = glm::transpose(glm::inverse(normalMatrix));
+
+        m_direction_eye = normalMatrix * m_direction;
```
blair1618: This looks like it gives the right results for world-origin lights, but it's kind of a waste to generate the same normal matrix so many times! The tile object itself calculates the same normal matrix and makes it available as 'u_normalMatrix' in our shaders. I think it makes sense to just use this matrix in the shader if the light is world-space-fixed.

patriciogonzalezvivo: you mean, not rotating the normals of the geometry on the shader?

blair1618: I mean using the matrix that we already use to rotate normals (u_normalMatrix) to rotate light directions, since it ends up being exactly the same matrix.

core/src/scene/directionalLight.cpp
```diff
@@ -19,10 +19,16 @@ void DirectionalLight::setDirection(const glm::vec3 &_dir) {
     m_direction = glm::normalize(_dir);
 }
 
-void DirectionalLight::setupProgram( std::shared_ptr<ShaderProgram> _shader ) {
+void DirectionalLight::setupProgram(const std::shared_ptr<View>& _view, std::shared_ptr<ShaderProgram> _shader ) {
+
+    glm::vec3 direction = m_direction;
+    if (m_origin == LightOrigin::WORLD) {
+        direction = _view->getNormalMatrix() * direction;
```
tallytalwar: direction should be normalized here?

blair1618: It kind of seems like it, doesn't it? The results look correct though

core/src/scene/pointLight.cpp
```diff
@@ -40,10 +40,31 @@ void PointLight::setRadius(float _inner, float _outer){
...
+            position.y = camSpace.y - _view->getPosition().y;
+            position.z = position.z - _view->getPosition().z;
+
+        } else if (m_origin == LightOrigin::GROUND) {
+            // Leave light's xy in camera space, but z needs to be moved relative to ground plane
+            position.z = position.z - _view->getPosition().z;
+        }
+        
+        if (m_origin == LightOrigin::WORLD || m_origin == LightOrigin::GROUND) {
+            position = glm::vec4(_view->getNormalMatrix() * glm::vec3(position), 1.f);
```
tallytalwar: I did not get why we need to multiply normal matrix to the light position here? 

It makes sense to me in the directional light case but not here ... If someone can explain :D.

PR #105, opened 2015-05-04
Loading images usign stb_image : ) 

patriciogonzalezvivo:


core/src/util/textureImage.h
```diff
@@ -0,0 +1,13 @@
+#pragma once
+
+#include "texture.h"
+#include <string>
+
+class TextureImage : public Texture {
```
blair1618: Instead of creating a new class that differs from texture only in the way it's constructed, what if we just create a static function in Texture that creates a new texture object from an image file; something like `Texture::fromImage(const std::string& _file)`. That seems a little simpler to me.

karimnaaji: Yes it makes sense!

blair1618:
Looks good to me!

PR #107, opened 2015-05-05
Refactor Android classes to separate UI from rendering

blair1618:
GLSurfaceView creates its own managed thread for rendering and it is generally good practice to separate this rendering thread from the UI thread on Android. 

Subjectively this seems to produce a smoother framerate as well, but it's hard to measure.

### DIFF
--- a/android/src/com/mapzen/tangram/Tangram.java
+++ b/android/src/com/mapzen/tangram/Tangram.java
@@ -8,18 +8,20 @@ import android.content.res.AssetManager;
 import android.opengl.GLSurfaceView;
 import android.opengl.GLSurfaceView.Renderer;
 import android.util.DisplayMetrics;
+import android.view.View;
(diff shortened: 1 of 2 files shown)


android/src/com/mapzen/tangram/Tangram.java
```diff
@@ -48,15 +50,26 @@
...
-        setEGLContextClientVersion(2);
-        setPreserveEGLContextOnPause(true);
-        setEGLConfigChooser(8, 8, 8, 8, 24, 0);
-        setRenderer(this);
-        setRenderMode(RENDERMODE_WHEN_DIRTY);
+        view = new GLSurfaceView(mainApp) {
+
+            @Override
+            public void surfaceDestroyed(SurfaceHolder holder) {
+                contextDestroyed = true;
```
tallytalwar: Is this an internal/member variable for GLSurfaceView?

PR #108, opened 2015-05-06
Consistency with the way memory is allocated

karimnaaji:
Deallocating with delete after a c-style allocation can be an undefined behavior depending on the platform. Also use of the stb_image deallocation function.

### DIFF
--- a/core/src/util/texture.cpp
+++ b/core/src/util/texture.cpp
@@ -2,17 +2,17 @@
 #define STB_IMAGE_IMPLEMENTATION
 #include "stb_image.h"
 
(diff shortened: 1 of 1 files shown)


core/src/util/texture.cpp
```diff
@@ -26,9 +26,8 @@ Texture::Texture(const std::string& _file, GLuint _slot, TextureOptions _options
     setData(reinterpret_cast<GLuint*>(pixels), width * height);
     update();
 
-    delete [] pixels;
-    delete [] data;
-
+    free(data);
```
tallytalwar: good catch.

PR #110, opened 2015-05-08
Materials

patriciogonzalezvivo:


patriciogonzalezvivo:
I didn't add the images but you can use any from here https://github.com/tangrams/tangram-sandbox/tree/gh-pages/styles/imgs

core/src/style/material.cpp
```diff
@@ -0,0 +1,286 @@
...
+}
+
+void Material::setAmbient(const glm::vec4 _ambient){
+	m_ambient = _ambient;
+    m_ambient_texture.reset();
+    setAmbientEnabled(true);
+}
+
+void Material::setAmbient(const std::string &_file, MappingType _type, glm::vec3 _scale, glm::vec4 _amount){
+    std::shared_ptr<Texture> texture(new Texture(_file));
```
karimnaaji: There's something about the texture deallocation that needs to be fixed here. Texture has a `destroy()` method that needs to be explicitly called by user since in the case of transform lookup for label (concurrent thread). In this particular case the texture is created from threads, pushed to gpu whenever possible from main thread, but the destructor can be  called from another thread whenever the automatic deallocation from shared_ptr happens. I don't know exactly who owns the texture in the material case, but a `destroy()` needs to be explicitly called somewhere (probably in the material destruction?).

patriciogonzalezvivo: True, also I don't exactly where is the best place. Material destructor seams to be a right place. Also I don't where to do the unbind() of textures

karimnaaji: Yeah material destructor is ok since materials will always be deallocated from main thread. For unbind material could have a kind `teardown`?

blair1618: Texture binding and unbinding should probably follow a model similar to what we've done for shaders. ShaderProgram tracks the currently bound program in a static variable. When a program needs to be used, ShaderProgram can then determine whether the current program needs to be unbound. This allows us to make the fewest possible state changes, which is even more important for textures because they are expensive to bind. In the Texture class we could use a static array (sized according to the number of texture units we allow) to track the currently bound textures according to their slot.

blair1618: ^ Updating the Texture class to work this way might actually be a good thing to do as a small, separate PR that we can then pull into this branch.

karimnaaji: Yes it makes sense to follow this, I'll do this change and make a PR in this branch for this.

blair1618: Yeah I think it makes sense for textures in general, unless there's some subtlety of the label textures that breaks this model. I'd say this change to textures should go into master first and then we can update this PR to use it.

karimnaaji: https://github.com/tangrams/tangram-es/pull/111 should fill the needs of this PR. Concerning deletion from threads there's now a in the texture to know whenever it is created from a thread and will wait for the manual deletion, from main thread textures would be deleted when auto pointers would be destroyed.

blair1618:
I'm cool with merging this now. It's missing the ability to use global positions in shaders for now, I intend to address that a bit later.

PR #111, opened 2015-05-08
Reduce change of GL state when binding textures

karimnaaji:
- reduce the change of GL state when binding textures by using a static variable that tracks the currently bound texture and its slot. 
- basic mechanism in shader in order to track the currently available slot when sending texture uniform to it.
- use a flag for manual deletion when texture is created from a thread, automatic when created from main thread and nothing specified


blair1618:
Interesting, this isn't exactly what I imagined but it's better in some ways! 

This ends up doing some extra work in `Texture::bind()`; It looks like even if a texture is already bound to the desired texture unit the method will make it the active texture unit and call bind again.

core/src/util/texture.cpp
```diff
@@ -24,32 +26,35 @@ Texture::Texture(const std::string& _file, GLuint _slot, TextureOptions _options
...
+    update(0);
 
     free(data);
     stbi_image_free(pixels);
 }
 
 Texture::~Texture() {
-
+    if (m_autoDelete) {
+        destroy();
```
tallytalwar: Should ShaderProgram::m_textureSlots be updated when a texture gets destroyed?

core/src/util/shaderProgram.cpp
```diff
@@ -278,6 +279,24 @@ void ShaderProgram::invalidateAllPrograms() {
...
+    } else {
+        unit = m_textureSlots[_name].second;
+    }
+    
+    _texture.update(unit);
+    _texture.bind(unit);
+    
+    glUniform1i(location, unit);
+}
+
```
blair1618: I have mixed feelings about this functionality. On one hand, it's a cool convenience to essentially never worry about which texture slot you're binding to when you use a texture in a shader. On the other hand, this could lead to unintentional performance problems that are hidden behind this abstraction. As a slightly contrived example, consider a situation where two shaders are both using two textures. Shader 1 sets texture A as a uniform and then sets texture B as another uniform. Shader 2 sets texture B as a uniform and then sets texture A as another uniform. Now whenever we alternate between using shader 1 and shader 2, textures A and B will both have to be re-bound to alternate texture units (even though we could easily keep the same bindings if we were a little more careful). If we were explicitly binding the textures this would be easy to see. 

PS: I didn't intend to think that much about this PR but it actually brings up some really interesting issues!

karimnaaji: I see, this would probably need some more thoughts to be solved. What I would do to have this fixed is solving the dependencies shader/texture by having a "uniform request" instead of an actual uniform attachment, and before each draw we solve the best unit/textureId combination for each texture for the shaders that needed it.

blair1618: I was actually thinking we could just take out this higher-level uniform logic and manually apply uniforms (probably more efficient than any algorithm, since we can use context-specific information)

PR #112, opened 2015-05-08
Configuration with Yaml file

blair1618:
At long last, we can reconfigure Tangram ES without recompiling!

Scene file parsing is obviously not done, but I'd like to get the basics merged so that the more involved aspects (e.g. filtering) can be worked on independently using this as a foundation. 

The changes here enable configuring data sources, lights, and cameras (though we don't support non-perspective cameras yet). This adds config.yaml file that I stole from one of the Tangram demos. We don't support everything in that file yet, it's an aspirational configuration :)

### DIFF
--- a/core/resources/config.yaml
+++ b/core/resources/config.yaml
@@ -14,7 +14,7 @@ lights:
 
 sources:
     osm:
-        type: MVTSource
+        type: MVT
         url:  http://vector.mapzen.com/osm/all/{z}/{x}/{y}.mapbox
 
 layers:
--- a/core/resources/config.yaml
+++ b/core/resources/config.yaml
@@ -7,6 +7,7 @@ cameras:
 lights:
     light1:
         type: directional
+        origin: world
         direction: [1, 1, -1]
         diffuse: [.7, .7, .7, 1.]
         ambient: [.3, .3, .3, 1.]
--- a/core/resources/config.yaml
+++ b/core/resources/config.yaml
@@ -1,0 +1,51 @@
+cameras:
+    camera1:
+        type: perspective
+        focal_length: [[16, 2], [20, 6]] # pairs of [zoom, focal len]
+        vanishing_point: [-250, -250] # relative to center of screen, in pixels
+
+lights:
+    light1:
+        type: directional
+        direction: [1, 1, -1]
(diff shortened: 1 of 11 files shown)


core/resources/config.yaml
```diff
@@ -0,0 +1,52 @@
...
+        type: directional
+        origin: world
+        direction: [1, 1, -1]
+        diffuse: [.7, .7, .7, 1.]
+        ambient: [.3, .3, .3, 1.]
+
+sources:
+    osm:
+        type: MVT
+        url:  http://vector.mapzen.com/osm/all/{z}/{x}/{y}.mapbox
```
bcamper: Minor note, we changed to `.mvt`

core/src/scene/sceneLoader.cpp
```diff
@@ -0,0 +1,263 @@
...
+        std::string url = source["url"].as<std::string>();
+        
+        std::unique_ptr<DataSource> sourcePtr;
+        
+        if (type == "GeoJSONTiles") {
+            sourcePtr = std::unique_ptr<DataSource>(new GeoJsonSource());
+        } else if (type == "TopoJSONTiles") {
+            // TODO
+        } else if (type == "MVT") {
+            sourcePtr = std::unique_ptr<DataSource>(new ProtobufSource());
```
bcamper: We should probably rename `ProtobufSource` to `MVTSource` or similar, protobuf is generic and there are other protobuf formats like VTM.

core/src/scene/sceneLoader.cpp
```diff
@@ -0,0 +1,268 @@
...
+    for (auto it = sources.begin(); it != sources.end(); ++it) {
+        
+        const Node source = it->second;
+        std::string type = source["type"].as<std::string>();
+        std::string url = source["url"].as<std::string>();
+        
+        std::unique_ptr<DataSource> sourcePtr;
+        
+        if (type == "GeoJSONTiles") {
+            sourcePtr = std::unique_ptr<DataSource>(new GeoJsonSource());
```
tallytalwar: Will "sourcePtr.reset(new GeoJsonSource())" be a better choice over the assignment operator!. 
However that being said, based on the documentation it seems they both perform the same and are more or less analogous.

PR #114, opened 2015-05-13
Make labels as unique_ptr owned by LabelContainer

tallytalwar:
- Previous labels were shared between mapTile and LabelContainer.. but maptile can use a label reference from
labelContainer.
- This will also help us further refine our label and label occlusion work!


karimnaaji:
:+1:

tallytalwar:
Keep this PR on hold. We might need Labels to be a shared_ptr!

tallytalwar:
Closing this!. Because implementation I am working for label placement will require labels to be shared_ptrs ... However we can still have mapTile use labels as reference and not shared_ptr!

PR #116, opened 2015-05-13
Rpi ncr

patriciogonzalezvivo:
Non-continuos rendering (single thread)


blair1618:
The re-organization here seems like a good thing, we should keep that. However the "non-continuous rendering" implemented here is unfortunately not super useful :\ 

The CPU won't be doing any less work, it'll just be running through a while loop for all the cycles that it would have otherwise been working on rendering. To get any actual efficiency improvements we'll need to actually put the rendering thread to sleep and wake it again when input or network events require a re-render. Various event-loop libraries can do this, as well as the very low-level pthread interface.

patriciogonzalezvivo:
I see. I will review and go back to the research

blair1618:
If you're happy with the re-organization changes we can go ahead and merge those. Adding truly event-driven rendering can be another PR since it'll be another substantial change.

patriciogonzalezvivo:
that sound great

blair1618:
Will do! I may have mentioned it before, but SDL may be useful for when you revisit this. I think it's cool how far the RPi implementation has gone without using external dependencies, but using SDL for our implementation here would be a good demonstration to users of how to use Tangram with another very popular cross-platform framework.

PR #117, opened 2015-05-13
Typed Mesh vbo data

karimnaaji:
In the compilation step of the typed mesh, make sure that the data is cleared from the CPU after compilation. Also reduce copy by using references.


karimnaaji:
Didn't know about this trick :)
Thanks @hjanetzek!

tallytalwar:
Yup, I just read about it here: ... http://www.gotw.ca/gotw/054.htm. Thanks

PR #118, opened 2015-05-13
World to clip-space optimization

karimnaaji:
Small optimization, instead of performing a full matrix multiplication based on our view transform matrix and model matrix that only uses translation (~15% speedup).


blair1618:
Something here is making labels move from their world anchor when the view is both tilted and rotated, maybe we're omitting a matrix term that's actually nonzero in this case?

karimnaaji:
Ah yes, I forgot about the rotation feature, which breaks this logic..

PR #119, opened 2015-05-13
Vbo cpu data

hjanetzek:
did you just rebase? =)


hjanetzek:
ah nvm, now it should apply

PR #124, opened 2015-05-19
Datasource cleanup

blair1618:
- Remove NetworkDataSource; it was an unnecessary abstraction layer. When we want to actually implement non-network data sources we should do it with something more like a component model (i.e. sources have a transfer component, a parser component, etc.)

 - Update the comments in DataSource to reflect the current implementation

 - Reorganize WorkerData (now TileTask) to contain an optional shared pointer to previously-parsed tile data. This is to prevent a weird trick we had to do in tileManager where we created a task with an empty 'raw data' vector and then ignored that vector in the tile worker. The data flow is a bit clearer now.

### DIFF
--- a/core/src/data/dataSource.cpp
+++ b/core/src/data/dataSource.cpp
@@ -8,6 +8,11 @@
 
 //---- DataSource Implementation----
 
+DataSource::DataSource(const std::string& _name, const std::string& _urlTemplate) :
+    m_name(_name), m_urlTemplate(_urlTemplate) {
+
+}
+
 bool DataSource::hasTileData(const TileID& _tileID) const {
     
     return m_tileStore.find(_tileID) != m_tileStore.end();
@@ -38,19 +43,7 @@ void DataSource::setTileData(const TileID& _tileID, const std::shared_ptr<TileDa
         m_tileStore[_tileID] = _tileData;
 }
 
-void DataSource::setUrlTemplate(const std::string& _urlTemplate){
-    m_urlTemplate = _urlTemplate;
-}
-
-//---- NetworkDataSource Implementation----
-
-NetworkDataSource::NetworkDataSource() {
-}
-
-NetworkDataSource::~NetworkDataSource() {
-}
-
-void NetworkDataSource::constructURL(const TileID& _tileCoord, std::string& _url) {
+void DataSource::constructURL(const TileID& _tileCoord, std::string& _url) const {
 
     _url.assign(m_urlTemplate);
 
@@ -68,7 +61,7 @@ void NetworkDataSource::constructURL(const TileID& _tileCoord, std::string& _url
     }
 }
 
-bool NetworkDataSource::loadTileData(const TileID& _tileID, TileManager& _tileManager) {
+bool DataSource::loadTileData(const TileID& _tileID, TileManager& _tileManager) {
     
     bool success = true; // Begin optimistically
     
@@ -94,7 +87,7 @@ bool NetworkDataSource::loadTileData(const TileID& _tileID, TileManager& _tileMa
     return success;
 }
 
-void NetworkDataSource::cancelLoadingTile(const TileID& _tileID) {
+void DataSource::cancelLoadingTile(const TileID& _tileID) {
     std::string url;
     constructURL(_tileID, url);
     cancelUrlRequest(url);
--- a/core/src/data/dataSource.cpp
+++ b/core/src/data/dataSource.cpp
@@ -66,7 +66,7 @@ bool DataSource::loadTileData(const TileID& _tileID, TileManager& _tileManager) 
     bool success = true; // Begin optimistically
     
     if (hasTileData(_tileID)) {
-        // Tile has been fetched already!
+        _tileManager.addToWorkerQueue(m_tileStore[_tileID], _tileID, this);
         return success;
     }
 
--- a/core/src/tile/tileWorker.cpp
+++ b/core/src/tile/tileWorker.cpp
@@ -15,27 +15,34 @@ void TileWorker::abort() {
     m_aborted = true;
 }
 
-void TileWorker::processTileData(std::unique_ptr<WorkerData> _workerData,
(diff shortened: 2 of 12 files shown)


core/src/data/dataSource.cpp
```diff
@@ -38,19 +43,7 @@ void DataSource::setTileData(const TileID& _tileID, const std::shared_ptr<TileDa
         m_tileStore[_tileID] = _tileData;
 }
 
-void DataSource::setUrlTemplate(const std::string& _urlTemplate){
```
tallytalwar: Should we still keep a setUrlTemplate around? In case we want to modify url template of an existing dataSource!!!!

But that being said, there is nothing stopping to create a new dataSource with that new url template.

core/src/tile/tileWorker.h
```diff
@@ -7,36 +7,38 @@
...
+        source(_source) {       
     }
 
-    WorkerData& operator=(WorkerData&& _other) {
-        rawTileData = std::move(_other.rawTileData);
-        tileID.reset(new TileID(*(_other.tileID)));
-        source = std::move(_other.source);
-        return *this;
+    TileTask(TileTask&& _other) :
+        tileID(_other.tileID),
```
tallytalwar: Its better to define a move constructor for TileID and use std::move here.

core/src/util/tileID.h
```diff
@@ -16,6 +16,7 @@ struct TileID {
     
     TileID(int _x, int _y, int _z) : x(_x), y(_y), z(_z) {};
     TileID(const TileID& _rhs): x(_rhs.x), y(_rhs.y), z(_rhs.z) {};
+    TileID(const TileID&& _rhs): x(std::move(_rhs.x)), y(std::move(_rhs.y)), z(std::move(_rhs.z)) {};
```
tallytalwar: We should be using move for TileID wherever possible (in a different PR).

core/src/tile/tileWorker.cpp
```diff
@@ -15,27 +15,34 @@ void TileWorker::abort() {
     m_aborted = true;
 }
 
-void TileWorker::processTileData(std::unique_ptr<WorkerData> _workerData,
+void TileWorker::processTileData(std::unique_ptr<TileTask> _task,
                                  const std::vector<std::unique_ptr<Style>>& _styles,
                                  const View& _view) {
 
-    m_workerData = std::move(_workerData);
+    m_task = std::move(_task);
```
tallytalwar: Ohh, I thought a move assignment operator was needed for this!! and hence I had added one.

blair1618: Yeah it's moving the unique_ptr, not the object itself, I often overlook this.

core/src/data/dataSource.cpp
blair1618: I couldn't think of an instance in which we would be modifying a data source that way, especially since they'll be defined statically in the style sheet. And as you say, it's not costly to create a new instance should we need it.

tallytalwar: yup agreed.

core/src/tile/tileWorker.cpp
tallytalwar: Yup!! :D.

core/src/util/tileID.h
blair1618: I would hesitate to use move on TileIDs everywhere; a move is only appropriate if we don't ever need the original object after that point (kind of defeats the point of trying to make them immutable). Objects this small are trivially simple to copy anyway.

hjanetzek: Hi, I'm also wondering how this move could work at all :) I would think move can only be used to say that references to heap blocks should be moved from one object to another and invalidated on the former(but I may be missing something)
Whenever a TileID is allocated, be it on the heap or the stack it is a fixed size contiguous region in memory to hold all members, how could parts from one region be said to be *moved* and not copied to the other?

tallytalwar: I agree on we should not be using move on TileID "everywhere" but in cases where TileID is not needed in the original callee (example: https://github.com/tangrams/tangram-es/blob/datasource-cleanup/core/src/tile/tileManager.cpp#L38 and https://github.com/tangrams/tangram-es/blob/datasource-cleanup/core/src/tile/tileManager.cpp#L45), you can get the rvalue of the temporary object and move it to be used inside TileTask. 
To me it is more readable and yes nothing to do with performance.

Also one thing wrong about the move constructor of TileID is the use of "const T&&". Based on my going through multiple stackoverflow posts (one of these being: http://stackoverflow.com/questions/6307526/c0x-const-rvalue-reference-as-function-parameter), it is making sure a "const rvalue" parameter makes sure it is only bindable to an rvalue and nothing else!! So changing that move constructor to a non const rvalue is better. I could be interpreting this wrong though, so forgive me if I am wrong here :D. Based on a mutual agreement here, we can change the signature of TileID's move constructor.

@hjanetzek http://www.cprogramming.com/c++11/rvalue-references-and-move-semantics-in-c++11.html this describes why this kind of move is legal! (Read the section on "Detecting temporary objects with rvalue references).

Also on a similar but different topic, I also recommend reading the following on rvalue reference and universal reference (when to use std::move and when to use std::forward).. https://isocpp.org/blog/2012/11/universal-references-in-c11-scott-meyers and also http://scottmeyers.blogspot.com/2012/10/parameter-types-in-constructors.html (debate on whether to pass parameters to a constructor by value or const reference)

hjanetzek: @tallytalwar thanks, I'll check it out. The best read discussing c++11 argument passing, that I found so far, was this series
http://www.codesynthesis.com/~boris/blog/2012/06/19/efficient-argument-passing-cxx11-part1/

tallytalwar: With all this discussion, I understood that I was wrong about std::move and stack values :(.
But because of this I read about std::forward and other stuff not related to this discussion (but related) :D. Thanks :D

PR #125, opened 2015-05-20
Lod label discard

karimnaaji:
Add a function that decreases the possible distance to camera for a label at high zoom when level-of-detail is happening. This happens to discard labels at horizon when you're tilting the camera close to the ground plane. We could consolidate by applying this only for line labels. 

Edit: results of this function: http://cpp.sh/3cvb

### DIFF
--- a/core/src/tile/labels/labelContainer.cpp
+++ b/core/src/tile/labels/labelContainer.cpp
@@ -9,7 +9,7 @@ LabelContainer::~LabelContainer() {
 }
 
 int LabelContainer::LODDiscardFunc(float _maxZoom, float _zoom) {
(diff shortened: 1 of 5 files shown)


core/src/tile/labels/labelContainer.cpp
```diff
@@ -34,6 +41,8 @@ bool LabelContainer::addLabel(MapTile& _tile, const std::string& _styleName, Lab
 }
 
 void LabelContainer::updateOcclusions() {
+    m_currentZoom = m_view->getZoom();
```
tallytalwar: We can pass m_view as a reference here instead of storing a shared_ptr for m_view in labelContainer.

karimnaaji: Just figured out it was also needed by the first label update to get the view-projection matrix but was actually not used, fixed by last commit.

PR #126, opened 2015-05-20
Cubemap

karimnaaji:
A way of loading horizontal cross-cubemaps.


blair1618:
Works well! Probably best to abstract a class (Skybox or Sky, maybe). Even though we'd only have one instance we should try to keep the main tangram class as un-cluttered as possible.

karimnaaji:
Yes, makes sense!

PR #127, opened 2015-05-21
Label states

karimnaaji:
Managing label-states for fade effects at occlusion, the unit tests for this feature gives a better understanding of how transitions happen.


core/src/tile/labels/label.cpp
```diff
@@ -76,61 +66,151 @@ void Label::updateScreenTransform(const glm::mat4& _mvp, const glm::vec2& _scree
...
+    if (m_currentState == State::SLEEP) {
+        // no-op state for now, when label-collision has less complexity, this state
+        // would lead to FADE_IN state if no collision occured
+        return;
+    }
+
+    bool occludedLastFrame = m_occludedLastFrame;
+    m_occludedLastFrame = false;
+
+    bool ruleSatisfied = updateScreenTransform(_mvp, _screenSize);
```
tallytalwar: I would rename "ruleSatisfied" this for clarity. maybe labelClipped?

karimnaaji: The thing is there is not only the clipping test that is happening, also some rules that we could add later (now only one rule testing whether a line label fits inside the projected screen space anchor points).

tallytalwar: cool makes sense.

PR #128, opened 2015-05-21
Hud

patriciogonzalezvivo:
A HUD for raspberryPi


blair1618:
I'm gonna have to take your word that this compiles and runs on the Pi, since we don't have build automation for it yet :)

This is probably not too big of a change to add, but in general we shouldn't be adding more to the Pi-specific files. What we want in the tangram-es repo is simply a minimal demonstration of using the core library on each platform. Handling specific hardware configurations is not the focus of this project, and other projects have already done that task far better than we could (e.g. GLFW, SDL) so we should use existing code for those aspects wherever we can. 

In particular, the next time we need to make changes on the Pi I think that we should strongly consider using SDL for the input and display surface management. It would make our own code smaller (more readable) and add functionality (like event-driven rendering). 

I'll make a few cosmetic changes and then merge this.

patriciogonzalezvivo:
I'm compiling right now

patriciogonzalezvivo:
I'm pushing one small correction and compiling your changes to check. I'm sorry for the inconvenience

patriciogonzalezvivo:
Everything compiling upto here

core/include/fontstash-es
```diff
@@ -1 +1 @@
-Subproject commit de034e4f4b5328f03b002ca4365353f6d08a83fe
+Subproject commit 5fecc158a64cd8bc7393f392ddc02191244e5e3d
```
karimnaaji: Please do `git submodule update` every time you checkout to another branch

blair1618: Good catch, I'll fix this real quick

PR #133, opened 2015-05-28
Filters

tallytalwar:
- Yaml Filter Parsing
- Filter evaluation
- Unit tests
- Not integrated with config file parsing at present (soon.. (read next))


core/src/scene/sceneLoader.cpp
```diff
@@ -271,6 +273,156 @@ void SceneLoader::loadCameras(Node cameras, View& view) {
...
+        float minVal = -std::numeric_limits<float>::infinity();
+        float maxVal = std::numeric_limits<float>::infinity();
+
+        for(YAML::const_iterator valItr = _node.begin(); valItr != _node.end(); ++valItr) {
+            if(valItr->first.as<std::string>() == "min") {
+                try {
+                    minVal = valItr->second.as<float>();
+                } catch(const BadConversion& e) {
+                    logMsg("Error: Badly formed filter.\tExpect a float value type, string found.\n");
+                    return nullptr;
```
blair1618: Not sure if `nullptr` is the best thing to return here. This means that if someone writes a malformed filter, the whole app is going to segfault when it tries to filter incoming data. It might be better to keep that error message, but return a filter that always fails (for example) so that the rest of the app can still function. We could return range(-inf, -inf), or we could create a distinct "NullFilter" that simply always returns false, or we could make `Filter` a concrete type and have it always return false.

tallytalwar: Good catch. Should be fixed in the following commits!!

tests/unit/yamlFilterTests.cpp
```diff
@@ -0,0 +1,233 @@
...
+    init();
+    YAML::Node node = YAML::Load("filter: { series: 3}");
+    Filter* filter = sceneLoader.generateFilter(node["filter"]);
+    int count = 0;
+    for(auto& vehicle : vehicles) {
+        if(filter->eval(vehicle, ctx)) {
+            count++;
+        }
+    }
+    REQUIRE(count == 1);
```
blair1618: I'm not sure counting is the best way to evaluate these tests. In this test the filter could pass the wrong 'vehicle' and the test wouldn't complain.

tallytalwar: Sure. We can check for what features matched!!.. should be a easy quick
fix. On it.

blair1618: Ah, I already wrote the change, should be up in a minute

tallytalwar: Cool. Will check it when it's pushed.

core/src/data/filters.h
```diff
@@ -0,0 +1,172 @@
...
+
+    };
+
+    using ValueList = std::vector<Value*>;
+
+    using Context = std::unordered_map<std::string, Value*>;
+
+    struct Filter {
+
+        virtual bool eval(const Feature& f, const Context& c) const { return false; };
```
blair1618: I think I was the one who suggested making an empty filter return false, but now I'm reconsidering what would be most natural. If having no filter on a data layer passes every feature, then would it make more sense for a "no-op" filter to behave similarly and always return `true`?

tallytalwar: True.

So something like this should pass!! But it will not in the present implementation.

'''
YAML::Node node = YAML::Load("filter: ");
Filter* filter = sceneLoader.generateFilter(node["filter"]);

REQUIRE(filter->eval(civic, ctx));
'''

core/src/scene/sceneLoader.cpp
```diff
@@ -271,6 +273,154 @@ void SceneLoader::loadCameras(Node cameras, View& view) {
...
+            filter = generatePredicate(_filter[key], key);
+
+        }
+
+        filters.push_back(filter);
+
+    }
+
+    if(filters.size() > 0) {
+        return (new All(filters));
```
blair1618: In the where `filters.size() == 1` we can just return `filters.front()` and save a level of indirection

blair1618: I'm just nitpicking now so clearly this PR is basically good to go

blair1618:
I forgot about a case we haven't handled yet! Boolean values on properties actually have special behavior, they check for the existence of a property (and ignore its value) as seen here: https://github.com/tangrams/tangram/blob/master/demos/scene.yaml#L154

I added a test case for this that currently fails.

PR #137, opened 2015-06-05
Style param

tallytalwar:


core/src/scene/sceneLoader.cpp
```diff
@@ -31,6 +32,38 @@ void SceneLoader::loadScene(const std::string& _file, Scene& _scene, TileManager
 
 }
 
+/*
+ *MaMaMaMaMaMaMa special coloarSequence parser to store [r, g, b]/[r, g, b, a] color sequence in
```
bcamper: more Ma

core/src/scene/sceneLoader.cpp
```diff
@@ -415,6 +448,33 @@ Filter* SceneLoader::generateNoneFilter(YAML::Node _filter) {
...
+        std::string paramKey;
+        if(propPrefix.length() > 0) {
+            paramKey = propPrefix + ":" + prop.as<std::string>();
+        } else {
+            paramKey = prop.as<std::string>();
+        }
+        if(propItr->second.IsScalar()) {
+            paramMap.emplace(paramKey, propItr->second.as<std::string>());
+        } else if(propItr->second.IsSequence()) {
+            if(prop.as<std::string>() == "color") {
```
bcamper: There are some color properties that don't have the property name `color` (if that's what this is doing... can't tell for sure).

core/src/scene/sceneLoader.cpp
tallytalwar: Ahh... :|. Will remove Ma.....

core/src/scene/sceneLoader.cpp
tallytalwar: Yes, its parsing for color.. 

parseColorSequence ... brings [R, G, B]/[R, G, B, A] in cssColorParser recognizable format, which is rgba(r, g, b, a) (all values other than alpha in 0-255 range and alpha in 0.0-1.0 range).

tallytalwar: Fixed here: https://github.com/tangrams/tangram-es/commit/b66276914e840bb313b60aebab4dd45ce1e26b8a

core/src/style/style.cpp
```diff
@@ -1,12 +1,43 @@
...
+            shift += 8;
+        }
+        if(shift == 24) {
+            color += (255 << 24);
+        }
+    } else { // css color or #hex-num
+        color = parse(_colorPropStr).getInt();
+    }
+    return color;
+}
```
blair1618: It seemed natural at first to simply turn all properties into strings, but now I'm a bit worried by the idea of turning all sequence properties into a string and then hand-writing code to parse it (potentially in many places). In particular this is going to become pretty ridiculous when we integrate _stops_ for properties. 

We may have to bite the bullet and include some of yaml-cpp in other classes. To keep ourselves from getting completely entangled with yaml, we could even create some separate class(es) for the parameter structures and then give them the code to de-serialize themselves from yaml-cpp node objects.

tallytalwar: I agree, I am not in favor of writing parsing code on strings !!!

I am not sure we should bite yaml-cpp bullet, but I was thinking of going back what we were discussing before coming to the conclusion of using string:string map for style parameters.

We can define our own structure to either have, "scalar"/"string" value, "sequence"/"vector" values or "map" values. I think thats a better solution than putting yaml nodes in the core code.

tallytalwar: I will get a version of this out and we can go from there.

core/src/style/polygonStyle.cpp
```diff
@@ -18,33 +21,45 @@ void PolygonStyle::constructVertexLayout() {
...
     std::string fragShaderSrcStr = stringFromResource("polygon.fs");
-    
+
     m_shaderProgram = std::make_shared<ShaderProgram>();
     m_shaderProgram->setSourceStrings(fragShaderSrcStr, vertShaderSrcStr);
 }
 
-void PolygonStyle::buildPoint(Point& _point, StyleParams& _params, Properties& _props, VboMesh& _mesh) const {
+void* PolygonStyle::parseStyleParams(StyleParamMap& _styleParamMap) const {
+    StyleParams* params = new StyleParams();
```
blair1618: This is a memory leak (no matching `delete`)

core/src/style/style.h
```diff
@@ -81,31 +63,42 @@ class Style {
...
-    std::vector< std::pair<std::string, StyleParams> > m_layers;
-    
+    /* Set of strings defining which data layers this style applies to,
+     * along with the style paramter map corresponding to these data layers,
+     * to be parsed explicitly by styles for their style parameters*/
+    std::vector< std::pair<std::string, StyleParamMap> > m_layers;
+
+    // cache the styleParameters for a layer (later for a ruleID(uniquely generated))
+    // NOTE: no use at present!!!
+    std::unordered_map<std::string, void*> m_styleParamCache;
```
blair1618: Might make sense to make one of these for each specific style class and store the actual type of the parsed style parameters (instead of `void*`), then this container would take care of destroying all of those cached parameters when it destructs.

Thinking further down the road, we should be aware that there is also a drawback to caching these in the style classes: we'll need to ensure that this shared cache is threadsafe :\

core/src/style/polygonStyle.cpp
tallytalwar: Ohh :|. I had this in, when I commented out the usage of m_styleParamCache in uber style class.
Will add this back in.

core/src/style/style.cpp
tallytalwar: Fixed: 1c257cd881b5db93552343f279d27bbc1829bef2

core/src/style/style.h
tallytalwar: Fixed: 1c257cd881b5db93552343f279d27bbc1829bef2

tallytalwar: While working on the subLayers feature, I though of another approach of caching!!
Instead of caching the "parsed" style properties in individual base styles, I cache the StyleParamMap in the uber style class (static member).

I could do this change here, but it will not be much different to what the present implementation is doing.

PR #138, opened 2015-06-05
Style configuration from YAML

blair1618:
Configuring styles, materials, and textures in the yaml file is now mostly implemented except for things explicitly noted in the code as "TODO". The biggest TODO is that mixing of styles still needs to be implemented, but that's a relatively separable task. 

PS Sorry for all the whitespace removal! My editors are all set to strip trailing whitespace to this is a one-time thing per file.


core/src/scene/sceneLoader.cpp
```diff
@@ -66,6 +73,233 @@ glm::vec3 parseVec3(const Node& node) {
...
+
+        Style* style = nullptr;
+
+        std::string styleName = styleIt.first.as<std::string>();
+        Node styleNode = styleIt.second;
+
+        Node baseNode = styleNode["base"];
+        if (baseNode) {
+            std::string baseString = baseNode.as<std::string>();
+            if (baseString == "lines") { style = new PolylineStyle(styleName); }
```
tallytalwar: Before adding a "named" style make sure same named style does not exists.

example: if a stylesheet defines a style named "lines" using "lines" base style..

core/src/scene/sceneLoader.cpp
```diff
@@ -66,6 +73,261 @@ glm::vec3 parseVec3(const Node& node) {
...
+    if (!shaders) {
+        return;
+    }
+
+    Node definesNode = shaders["defines"];
+    if (definesNode) {
+        for (const auto& define : definesNode) {
+            std::string name = define.first.as<std::string>();
+            std::string value = define.second.as<std::string>();
+            shader.addSourceBlock("defines", "#define " + name + " " + value);
```
blair1618: Gotta add special treatment for booleans here, the JS code uses booleans to define existence: https://github.com/tangrams/tangram/blob/73657bd3c85eb68605398b2735eaa69b3c79126e/src/gl/shader_program.js#L410-L415

@bcamper if one wanted to #define an actual boolean, do 1 and 0 evaluate correctly?

blair1618: Update: they don't. 

I guess for now we have to deal with not being able to #define booleans.

core/src/scene/sceneLoader.cpp
```diff
@@ -66,6 +73,266 @@ glm::vec3 parseVec3(const Node& node) {
...
+    }
+
+}
+
+MaterialTexture SceneLoader::loadMaterialTexture(YAML::Node matCompNode, Scene& scene) {
+
+    MaterialTexture matTex;
+
+    Node textureNode = matCompNode["texture"];
+    Node mappingNode = matCompNode["mapping"];
```
tallytalwar: texture [scale](https://github.com/tangrams/tangram-docs/blob/gh-pages/pages/textures.md#scale) and [amount](https://github.com/tangrams/tangram-docs/blob/gh-pages/pages/textures.md#amount) specification not handled here.

core/src/style/material.h
```diff
@@ -15,42 +15,47 @@ class Texture;
...
+    planar,
+    triplanar,
+    spheremap
+};
+
+struct MaterialTexture {
+    std::shared_ptr<Texture> tex = nullptr;
+    MappingType mapping = MappingType::uv;
+    glm::vec3 scale = glm::vec3(1.f);
+    glm::vec3 amount = glm::vec3(1.f);
```
tallytalwar: :+1:

PR #142, opened 2015-06-15
Text api

karimnaaji:
Update the data layout for text-rendering, now using more attributes, all packed in vbo.
Possible use of DYNAMIC_DRAW/STREAM hint for vertex buffer objects.The code base gets simpler and some known issues are automatically fixed (out of screen translations not anymore packed to a byte so the text origin/anchor point can be drawn out of the screen).
One thing we could consider later is using double-buffering vbo (draw text with vbo1 at frame t and update vbo2, draw with vbo2 at frame t+1 and update vbo1), but we can stick with this until we find device bottlenecks in using dynamic vbos.


blair1618:
I'm encountering segfaults when I zoom out to roughly zoom 1 or 0. Can't tell precisely why, but it seems like maybe the vertex buffer is being modified after it's freed? At first I suspected this was something bad about the merge I did, but the same thing happens at https://github.com/tangrams/tangram-es/commit/b3283f8b0119aca5d6fe30b0b339c69d6992a6c0

karimnaaji:
Yes I'm experiencing the same, investigating on that.

core/resources/text.vs
```diff
@@ -5,88 +5,34 @@ precision mediump float;
...
+            a_position.x * ct - a_position.y * st + a_screenPosition.x,
+            a_position.x * st + a_position.y * ct + a_screenPosition.y,
+            0.0, 1.0
         );
-
+        
         gl_Position = u_proj * p;
     } else {
-        gl_Position = vec4(0.0, 0.0, 0.0, 1.0);
+        gl_Position = vec4(0.0);
```
blair1618: Can we put the transparent vertices somewhere that's totally outside the view frustum? Then the rasterizer will automatically discard all the fragments and we won't have to branch on the alpha value in the fragment shader. Actually.. they might be discarded anyway because a quad with all of its vertices at (0, 0) is probably considered degenerate.

karimnaaji: yes I changed w from 1.0 to 0.0 because of that, so it gets actually clipped (not even trying to divide by 0 in homogeneous coordinates I guess).

blair1618: Cool! We can get rid of that `if` in the fragment shader in that case

karimnaaji: yep good catch

blair1618:
screen shot 2015-06-19 at 5 17 51 pm I guess these glyphs just aren't in our bundled font?

karimnaaji:
Right, none of those glyphs are bundled

core/src/text/fontContext.cpp
```diff
@@ -84,73 +88,23 @@ void FontContext::setFont(std::string _name, int size) {
...
-            break;
-        }
-
-        default:
-            logMsg("[FontContext] FontError : undefined error\n");
-            break;
+    auto buffer = fontContext->getCurrentBuffer();
+    
+    if (buffer->hasData()) {
+        buffer->getWeakMesh()->update(_offset, _size, reinterpret_cast<unsigned char*>(_newData));
```
tallytalwar: Discussed with @karimnaaji. And it will make sense for TextBuffer to be inherited from VboMesh. 
Which also clarifies the ownership of VboMesh to MapTile.

PR #144, opened 2015-06-16
Split Android gradle project into library and demo

blair1618:
Or "Towards an alternative MapBurrito"

This allows us to build an Android ARchive (AAR) for integration with Eraser Maps

TODO:
 - Think more about package/folder naming
 - Decide which assets (if any) should be part of the library


blair1618:
Found an issue with the build system, we're missing the c++_shared standard library that belongs next to our .so; don't merge yet!

tallytalwar:
Have not started reviewing this yet. Will be doing this tomorrow morning.

PR #145, opened 2015-06-17
some tinkerings

hjanetzek:
Hey Matt, I found the cause for unexpected disappearing proxies: the logic was specifically tailored for switching proxies depending on the zoom direction. So when the view is tilted, panning would only consider children as proxies. The logic could still be improved. I'll look into the sorting of the load queue next.

### DIFF
--- a/core/src/data/dataSource.cpp
+++ b/core/src/data/dataSource.cpp
@@ -5,6 +5,7 @@
 #include "tileData.h"
 #include "mapTile.h"
 #include "tileManager.h"
+#include "labels/labelContainer.h"
 
 //---- DataSource Implementation----
 
--- a/core/src/data/geoJsonSource.cpp
+++ b/core/src/data/geoJsonSource.cpp
@@ -1,6 +1,7 @@
 #include "geoJson.h"
 #include "platform.h"
 #include "tileID.h"
+#include "labels/labelContainer.h"
 
 #include "geoJsonSource.h"
 #include "rapidjson/error/en.h"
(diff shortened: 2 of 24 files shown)


hjanetzek:
ok.. need to check what clang is up to

hjanetzek:
amazing errors for one missing <string> include :)

tallytalwar:
This is good. Thanks Hannes.

.gitignore
```diff
@@ -111,3 +111,5 @@ install_manifest.txt
 # vim 
 *.swp
 *.swo
+
+docs
```
tallytalwar: We don't plan to use doxygen for source code documentation. @blair1618 had mentioned his inclination towards http://jessevdk.github.io/cldoc/.

hjanetzek: ok - The doxy output could also go to some other directory. I wouldn't mind to keep the config local either. Though it might be handy for others as well to generate include- and call-graphs.

blair1618:
Ahh nice. This makes a substantial difference to visual coherence! Reducing #includes is always welcome too. I'll try to look through these changes today.

And the doxyfile: Is this just so that other projects that include tangram-es can generate some basic doxygen about the project? I don't really like Doxygen notation or output, but if it's something that other projects expect to find then it might make sense to keep a doxyfile.

hjanetzek:
Doxygen documentation is not supposed to be imported by other projects. To me the only useful thing about it is to have it create dot graphs of the program structure. I'll remove that commit for now

PR #146, opened 2015-06-17
Screen-ray to ground plane

karimnaaji:
Previously tiles behind the camera were loaded when camera was looking above the horizon, this is a small change which automatically stop updating the tile-set when the screen-ray to ground has intersections solutions behind the camera.

### DIFF
--- a/core/src/view/view.cpp
+++ b/core/src/view/view.cpp
@@ -162,7 +162,7 @@ glm::dmat2 View::getBoundsRect() const {
 
 }
 
-void View::screenToGroundPlane(float& _screenX, float& _screenY) const {
+glm::vec2 View::screenToGroundPlane(float& _screenX, float& _screenY) const {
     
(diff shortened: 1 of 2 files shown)


core/src/view/view.cpp
```diff
@@ -188,6 +188,8 @@ void View::screenToGroundPlane(float& _screenX, float& _screenY) const {
     
     _screenX = ray_world.x;
     _screenY = ray_world.y;
+    
+    return glm::vec2(t, ray_world.z);
```
blair1618: Do we need both `t` and `z`? Isn't `t < 0` sufficient to exclude backwards intersections?

karimnaaji: Conceptually we don't (negative `t` for ray to ground intersections are always behind), but updating the tile set should still occur even if one of the 4 viewport casted rays have intersections behind the camera, the only way to know is by looking up is by looking at the sign of `z` ray component.

PR #147, opened 2015-06-19
Add a clang-format style file

blair1618:
`clang-format` is a command line tool for automatic code formatting, available from homebrew and apt. The style file is just YAML, but clang-format expects a specific file name (with no extension). The style here is based on the LLVM style guide, modified to reflect our current practices.

To see how this applies to our existing code, run
```
clang-format -style=file [file]
```
and you'll get the formatted code in standard output. To edit the files in-place, add the `-i` option. 

I don't think we need to run a mass-reformat on our existing code, but we can use this for new files and enforce this guide (within reason) for all future additions.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,64 @@
+---
+Language: Cpp
+AccessModifierOffset: -4
+AlignAfterOpenBracket: true
+AlignEscapedNewlinesLeft: false
+AlignOperands: true
+AlignTrailingComments: true
+AllowAllParametersOfDeclarationOnNextLine: true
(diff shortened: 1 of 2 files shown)


tallytalwar:
@hjanetzek Yes, I/we should be rebasing more :D.

blair1618:
@hjanetzek Yeah it's definitely possible to clutter your history if you constantly merge in other branches. I don't find it necessary to merge in other branches more than about once per PR. Personally I kind of appreciate having a verbose history and I feel like it's less error-prone to simply merge. Of course you can rebase if you prefer, maybe you'll change my mind :)

PR #148, opened 2015-06-22
Debug draw

karimnaaji:
Some ways to draw debug primitive informations in screen space. This doesn't have to be necessarily in master, I just needed this feature so now that it's done it can be used later for any debugging purposes.


blair1618:
Debug drawing is a good thing to have available. If we find ourselves needing more functionality later we can refactor this for better resource management, but I like that it's simple for now.

PR #153, opened 2015-06-23
Add clang-format auto-routine in Makefile

zarov:
Few lines @karimnaaji asks me to help with. There is an auto selection of last modified .cpp and .h files using git diff, on which clang-format is applied.


tallytalwar:
:+1:

PR #155, opened 2015-06-24
update with grid based collisions

karimnaaji:
10x faster on isect2d examples


core/src/tile/labels/labelContainer.cpp
```diff
@@ -79,7 +79,7 @@ void LabelContainer::updateOcclusions() {
     }
 
     // broad phase
-    auto pairs = intersect(aabbs);
+    auto pairs = intersect(aabbs, { 4, 4 }, { m_view->getWidth(), m_view->getHeight() });
```
karimnaaji: This second parameter `{4, 4}` could be computed based on the screen size, we need some profiling to see whats the best approximation though.

blair1618:
Nice! Seems to be almost all formatting changes so I'll just go ahead and merge.

PR #156, opened 2015-06-24
Header cleaning

blair1618:
Aggressive cleaning of #includes to reduce incremental compile times (and for general code hygiene).


core/src/style/style.h
```diff
@@ -1,22 +1,22 @@
 #pragma once
 
-#include <string>
-#include <vector>
-
 #include "data/tileData.h"
```
tallytalwar: We can also take this out and forward declare all the structs!!! Thoughts?

hjanetzek: It depends - when including a struct, like e.g. TileID, does not pull further includes I would not mind to just include it with the header - especially when TileID would need to be included by any user of the header

tallytalwar: :+1:

blair1618:
I think this branch represents a useful lesson. Managing #includes effectively is a ton of pretty mindless work. I saw that @hjanetzek had experimented with include-what-you-use, which (theoretically) provides all the benefits of clean headers without the manual drudgery. How has that been going Hannes?

hjanetzek:
@blair1618, yea I thought setting up iwyu would at least be more interesting :) so far it works quite well, once set up -- the clang-3.6 tag (not branch) works with the llvm.org packages. For GLM one needs to make a mapping file to get the public includes instead of the internal ones.
I guess the last two quirks when applying the created diff automatically can also be fixed. Though for small changes looking at the diff should suffice to make the appropriate changes manually. 

Overall compilation time went down from 155 to 143 seconds (147 for the header-cleaning branch), though the real gain could be to have less files to recompile on some header changes.

core/src/util/vboMesh.cpp
```diff
@@ -73,10 +71,10 @@ void VboMesh::subDataUpload() {
     if (m_dirtySize != 0) {
         glBindBuffer(GL_ARRAY_BUFFER, m_glVertexBuffer);
 
-        long vertexBytes = m_nVertices * m_vertexLayout->getStride();
+        size_t vertexBytes = m_nVertices * m_vertexLayout->getStride();
 
         // updating the entire buffer
-        if (std::abs(m_dirtySize - vertexBytes) < m_vertexLayout->getStride()) {
+        if (vertexBytes - m_dirtySize < m_vertexLayout->getStride()) {
```
hjanetzek: Anyone knows where this change comes from? -  this is not necessarily part of header-cleaning.

hjanetzek:
@blair1618, @tallytalwar  would you mind to squash this branch or apply the diff against master? The commit messages are not that enlightening for git historians :)

core/src/util/vboMesh.cpp
blair1618: Ah, I made this edit along with some of the #include changes, could have been better documented. I only changed it to prevent a warning about signed/unsigned comparison. I believe this comparison could also reasonably be "<= 0" but I was only concerned with the warning.

blair1618:
Yes I can take a shot at squashing this, this branch in particular had several less-than-useful commits.

tallytalwar:
@hjanetzek I never knew about squashing in git. Thanks :D.
Now I do.

blair1618:
Oh err, I wasn't quite thinking clearly last night, I'm not going to squash commits that are already in master (it's not worth the annoyance for everybody using the repo) but your point about using the git history is well-taken and I will squash silly commits in the future.

hjanetzek:
@blair1618 ok. yes, someone might already have merged current master locally.. or might get confused by a detached HEAD otherwise :)

PR #159, opened 2015-06-25
Use of a single data structure for aabbs

karimnaaji:
Reduce the number of structures representing the same thing


blair1618:
yesssss

PR #161, opened 2015-06-25
tile management

hjanetzek:
This is far from ready - but works much better already :)

Next step would be to let the workers pull their jobs from the TileManager. Suggestions are welcome! I'll probably be off until Monday.

### DIFF
--- a/core/src/tangram.cpp
+++ b/core/src/tangram.cpp
@@ -113,14 +113,18 @@ namespace Tangram {
                 Label::s_needUpdate = false;
 
                 for (const auto& mapIDandTile : m_tileManager->getVisibleTiles()) {
-                    const std::shared_ptr<MapTile>& tile = mapIDandTile.second;
+                    const auto& tile = mapIDandTile.second;
+                    if (tile->state() != MapTile::Ready)
+                        continue;
                     tile->update(_dt, *m_view);
                 }
 
                 // update labels for specific style
                 for (const auto& style : m_scene->getStyles()) {
                     for (const auto& mapIDandTile : m_tileManager->getVisibleTiles()) {
-                        const std::shared_ptr<MapTile>& tile = mapIDandTile.second;
+                        const auto& tile = mapIDandTile.second;
+                        if (tile->state() != MapTile::Ready)
+                            continue;
                         tile->updateLabels(_dt, *style, *m_view);
                     }
                 }
@@ -130,7 +134,10 @@ namespace Tangram {
 
                 for (const auto& style : m_scene->getStyles()) {
                     for (const auto& mapIDandTile : m_tileManager->getVisibleTiles()) {
-                        const std::shared_ptr<MapTile>& tile = mapIDandTile.second;
+                        const auto& tile = mapIDandTile.second;
+                        if (tile->state() != MapTile::Ready)
+                            continue;
+
                         tile->pushLabelTransforms(*style, m_labels);
                     }
                 }
@@ -158,6 +165,9 @@ namespace Tangram {
             // Loop over all tiles in m_tileSet
             for (const auto& mapIDandTile : m_tileManager->getVisibleTiles()) {
                 const std::shared_ptr<MapTile>& tile = mapIDandTile.second;
+                if (tile->state() != MapTile::Ready)
+                    continue;
+
                 if (tile->hasGeometry()) {
                     // Draw tile!
                     tile->draw(*style, *m_view);
--- a/core/src/tangram.cpp
+++ b/core/src/tangram.cpp
@@ -114,16 +114,18 @@ namespace Tangram {
 
                 for (const auto& mapIDandTile : m_tileManager->getVisibleTiles()) {
                     const auto& tile = mapIDandTile.second;
-                    if (tile->isReady())
+                    if (tile->isReady()) {
                         tile->update(_dt, *m_view);
+                    }
                 }
 
                 // update labels for specific style
                 for (const auto& style : m_scene->getStyles()) {
                     for (const auto& mapIDandTile : m_tileManager->getVisibleTiles()) {
                         const auto& tile = mapIDandTile.second;
-                        if (tile->isReady())
+                        if (tile->isReady()) {
                             tile->updateLabels(_dt, *style, *m_view);
+                        }
                     }
                 }
 
@@ -133,8 +135,9 @@ namespace Tangram {
                 for (const auto& style : m_scene->getStyles()) {
                     for (const auto& mapIDandTile : m_tileManager->getVisibleTiles()) {
                         const auto& tile = mapIDandTile.second;
-                        if (tile->isReady())
+                        if (tile->isReady()) {
                             tile->pushLabelTransforms(*style, m_labels);
+                        }
                     }
                 }
             }
--- a/core/src/tangram.cpp
+++ b/core/src/tangram.cpp
@@ -164,7 +164,7 @@ namespace Tangram {
             // Loop over all tiles in m_tileSet
             for (const auto& mapIDandTile : m_tileManager->getVisibleTiles()) {
                 const std::shared_ptr<MapTile>& tile = mapIDandTile.second;
-                if (tile->isReady() && tile->hasGeometry()) {
+                if (tile->isReady()) {
                     // Draw tile!
                     tile->draw(*style, *m_view);
                 }
--- a/core/src/tangram.cpp
+++ b/core/src/tangram.cpp
@@ -114,18 +114,16 @@ namespace Tangram {
 
                 for (const auto& mapIDandTile : m_tileManager->getVisibleTiles()) {
                     const auto& tile = mapIDandTile.second;
-                    if (tile->state() != MapTile::Ready)
-                        continue;
-                    tile->update(_dt, *m_view);
+                    if (tile->isReady())
+                        tile->update(_dt, *m_view);
                 }
 
                 // update labels for specific style
                 for (const auto& style : m_scene->getStyles()) {
                     for (const auto& mapIDandTile : m_tileManager->getVisibleTiles()) {
                         const auto& tile = mapIDandTile.second;
-                        if (tile->state() != MapTile::Ready)
-                            continue;
-                        tile->updateLabels(_dt, *style, *m_view);
+                        if (tile->isReady())
+                            tile->updateLabels(_dt, *style, *m_view);
                     }
                 }
 
@@ -135,10 +133,8 @@ namespace Tangram {
                 for (const auto& style : m_scene->getStyles()) {
                     for (const auto& mapIDandTile : m_tileManager->getVisibleTiles()) {
                         const auto& tile = mapIDandTile.second;
-                        if (tile->state() != MapTile::Ready)
-                            continue;
-
-                        tile->pushLabelTransforms(*style, m_labels);
+                        if (tile->isReady())
+                            tile->pushLabelTransforms(*style, m_labels);
                     }
                 }
             }
@@ -165,10 +161,7 @@ namespace Tangram {
             // Loop over all tiles in m_tileSet
             for (const auto& mapIDandTile : m_tileManager->getVisibleTiles()) {
                 const std::shared_ptr<MapTile>& tile = mapIDandTile.second;
-                if (tile->state() != MapTile::Ready)
-                    continue;
-
-                if (tile->hasGeometry()) {
+                if (tile->isReady() && tile->hasGeometry()) {
                     // Draw tile!
                     tile->draw(*style, *m_view);
                 }
--- a/core/src/tile/labels/labels.cpp
+++ b/core/src/tile/labels/labels.cpp
@@ -5,6 +5,8 @@
 #include "util/primitives.h"
 #include "view/view.h"
 
+#include "glm/gtc/matrix_transform.hpp"
+
 Labels::Labels() {}
 
 Labels::~Labels() {
@@ -19,6 +21,8 @@ int Labels::LODDiscardFunc(float _maxZoom, float _zoom) {
 bool Labels::addLabel(MapTile& _tile, const std::string& _styleName, Label::Transform _transform, std::string _text, Label::Type _type) {
     auto currentBuffer = m_ftContext->getCurrentBuffer();
 
+    // FIXME: the current view should not be used to determine whether a label is shown at all
+    // otherwise results will be random
     if ( (m_currentZoom - _tile.getID().z) > LODDiscardFunc(View::s_maxZoom, m_currentZoom)) {
         return false;
     }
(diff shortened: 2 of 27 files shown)


core/src/tangram.cpp
```diff
@@ -121,15 +121,17 @@ namespace Tangram {
                 Label::s_needUpdate = false;
 
                 for (const auto& mapIDandTile : m_tileManager->getVisibleTiles()) {
-                    const std::shared_ptr<MapTile>& tile = mapIDandTile.second;
-                    tile->update(_dt, *m_view);
+                    const auto& tile = mapIDandTile.second;
+                    if (tile->isReady())
```
tallytalwar: Styling nitpick.
@hjanetzek we have been using explicit {} for single line if blocks.

hjanetzek: ok

core/src/tile/tileManager.cpp
```diff
@@ -3,237 +3,350 @@
...
+                    ++visTilesIter;
+                }
+            }
+        }
+    }
+
+    if (!m_loadQueue.empty() && m_loadPending < MAX_PARALLEL_DOWNLOADS) {
+        glm::dvec2 center(m_view->getPosition().x, -m_view->getPosition().y);
+
+        m_loadQueue.sort([&](const TileID& a, const TileID& b) {
```
tallytalwar: Did you benchmark this? How much time does it add?

core/src/tile/tileManager.cpp
```diff
@@ -3,237 +3,350 @@
...
-            worker->abort();
-            // Proxy tiles will be cleaned in update loop
+        // 1. Remove from Datasource
+        // Make sure to cancel the network request associated with this tile.
+        for(auto& dataSource : m_dataSources) {
+            dataSource->cancelLoadingTile(id);
         }
+
+        //// Canceled tiles will be ignored ////
+        // 2. Remove from tiles queued for processing
```
tallytalwar: Is this a Todo? Can the comments be cleaned here?

Some old code is sitting here (commented out)

core/src/tile/tileManager.h
```diff
@@ -7,13 +7,15 @@
 #include <future>
 #include <set>
 #include <mutex>
+#include <deque>
+#include <atomic>
```
tallytalwar: Atomic not being used here!

core/src/tile/tileManager.cpp
hjanetzek: No, this part is not optimized yet and wip. Instead of pushing MAX_PARALLEL_DOWNLOADS to the backend  I would like to change it so that the loadQueue is polled by the platform download workers. I'm not sure if this should be done at the moment because it will require changes in all platform implementations.

A quick check on my laptop gave a maximum of 0.000322 ms for sorting 80 tiles (usually there are less tiles when the view is not fully tilted or tiles are partially loaded ~ 0.000020 ms). It is likely faster to just pick the tiles with the lowest distance though.

core/src/tile/tileManager.h
hjanetzek: yes, it's gone :)

core/src/tile/tileManager.cpp
hjanetzek: this was part of the old code - just kept for reference and about to be removed

core/src/tile/tileManager.cpp
```diff
@@ -3,237 +3,350 @@
...
+
+            while (visTilesIter != visibleTiles.end()) {
+
+                auto& visTileId = *visTilesIter;
+                auto& curTileId = setTilesIter == m_tileSet.end() ? NOT_A_TILE : setTilesIter->first;
+
+                if (visTileId == curTileId) {
+                    if (setTilesIter->second->state() == MapTile::None) {
+                        logMsg("[%d, %d, %d] - Enqueue\n", curTileId.z, curTileId.x, curTileId.y);
+                        m_loadQueue.push_back(curTileId);
```
tallytalwar: I get whats the relationship here with MapTile::None, m_tilesetChanged and enqueueing tiles.

But don't you feel its wasteful to clear the loadQueue every iteration. I could be missing something though.

core/src/tile/tileManager.cpp
tallytalwar: Numbers are good for sorting.

We should be careful with polling implementation, if a bunch of tiles (tilted view) are being fetched and with std::async we can see bad thread performance because of trashing and context switches. Also platforms like raspberry pi might not perform well with this.

core/src/tile/tileManager.cpp
hjanetzek: I assume that clearing a deque is like clearing a vector an O(1) operation for types with trivial destructor (when doing optimized builds). Otherwise one would have to search and remove each item.

hjanetzek: Actually loadQueue using a std::list since vector (or deque) would not allow to sort TileIDs. Which reminds me to ask why TileID needs to have const members?

tallytalwar: That's because we want tileid to be immutable. It's used as a key in
std::map.

hjanetzek: If I recall correctly the key of a map entry is const which also makes the embedded TileId members const. This actually seems to be the case :) http://goo.gl/8OZ44G

core/src/tile/tileManager.cpp
hjanetzek: Ouch  the numbers were seconds.. :/  Needs to be changed before merging.


In general some #defines for collecting benchmarks within the application would be nice to have. As well as some benchmark for individual components. (I was about to start this for the Style processing)

karimnaaji: :+1: on collecting benchmarks! There's also good libraries for that https://github.com/google/benchmark

tallytalwar: :+1:

blair1618:
@hjanetzek are you going to add more to this branch? If not I'll review it and get it merged.

hjanetzek:
@blair1618, it's a good time to merge this before proceeding further. 

One blocker for me to continue with this is to figure out how multiple datasources should be handled. Especially how to handle the case when datasources provide tiles for different zoomlevel ranges, when tiles from different zoomlevels are shown at the same time. For example one could have overzoomed  arial image from z=12 while showing street tiles from z=15.

hjanetzek:
Ah one thing I want to change before the merge all occurences of TileTask should be written out to shared_ptr<TileTask> again

blair1618:
Reviewing! Seems to perform very well on phones.

tallytalwar:
In some cases, some boundry tiles are not fetched/drawn. See images below:

screen shot 2015-07-08 at 9 23 06 am

screen shot 2015-07-08 at 9 28 02 am

screen shot 2015-07-08 at 9 28 55 am

blair1618:
I think this has actually been an existing issue, check master

core/src/tile/tileWorker.cpp
```diff
@@ -2,76 +2,129 @@
...
+                continue;
+            }
+
+            // Pop highest priority tile from queue
+            auto it = std::min_element(m_queue.begin(), m_queue.end(),
+                [](const std::shared_ptr<TileTask>& a, const std::shared_ptr<TileTask>& b) {
+                    if (a->tile->isVisible() != b->tile->isVisible())
+                        return a->tile->isVisible();
+
+                    return a->tile->getPriority() < b->tile->getPriority();
```
tallytalwar: What happens when both a and b are not visible! (Is it possible to have a tile which is not visible but still not cancelled)?

core/src/tangram.cpp
```diff
@@ -160,7 +166,7 @@ namespace Tangram {
             // Loop over all tiles in m_tileSet
             for (const auto& mapIDandTile : m_tileManager->getVisibleTiles()) {
                 const std::shared_ptr<MapTile>& tile = mapIDandTile.second;
-                if (tile->hasGeometry()) {
+                if (tile->isReady() && tile->hasGeometry()) {
```
tallytalwar: Shouldn't a ready tile always have geometry?

core/src/tile/labels/labels.cpp
```diff
@@ -32,7 +36,13 @@ bool Labels::addLabel(MapTile& _tile, const std::string& _styleName, Label::Tran
...
         }
 
-        l->update(m_view->getViewProjectionMatrix() * _tile.getModelMatrix(), m_screenSize, 0);
+        // NB: viewOrigin.z is only determined by screen width and height.
+        const auto& viewOrigin = m_view->getPosition();
+
+        auto modelMatrix = glm::scale(glm::mat4(1.0), glm::vec3(_tile.getScale()));
+        modelMatrix[3][2] = -viewOrigin.z;
+
+        l->update(m_view->getViewProjectionMatrix() * modelMatrix, m_screenSize, 0);
```
tallytalwar: I am not sure I get this! Can you explain this?

core/src/tile/mapTile.h
```diff
@@ -89,30 +92,102 @@ class MapTile {
...
+
+    bool setProxy(ProxyID id) {
+      if ((m_proxies & id) == 0) {
+        m_proxies |= id;
+        return true;
+      }
+      return false;
+    }
+
+    bool unsetProxy(ProxyID id) {
```
tallytalwar: Please conform to code style. (4 space tab instead of 2)

core/src/tile/mapTile.h
```diff
@@ -89,30 +92,102 @@ class MapTile {
...
+    TileState state() {
+        return m_state;
+    }
+
+    void setState(TileState _state) {
+      m_state = _state;
+    }
+
+    void setPriority(double _priority) {
+      m_priority = _priority;
```
tallytalwar: Should a m_priority.store(_priority) be used here?

osx/src/platform_osx.mm
```diff
@@ -142,4 +142,6 @@ void cancelUrlRequest(const std::string& _url) {
     }];
 }
 
+void setCurrentThreadPriority(int priority) {}
```
blair1618: `setpriority` is available on OS X too. Unless there was a specific reason for leaving this blank I'll add the same functionality that linux has.

core/src/tile/tileManager.cpp
```diff
@@ -6,32 +6,30 @@
 #include "tile/mapTile.h"
 #include "view/view.h"
 
+#include "glm/gtx/norm.hpp"
+
+#include <chrono>
 #include <algorithm>
 
-TileManager::TileManager() {
+#define DBG(...)
```
tallytalwar: Maybe define this in the platform header.

We also have a trello card to expose debugging toggles.

osx/src/platform_osx.mm
hjanetzek: There should have been a TODO - I just couldn't test it. Yes, this likely works the same for OSX.

core/src/tile/tileWorker.cpp
hjanetzek: isVisible says whether a tile is in the visible tiles set determined by view. The tile still be used as proxy. Maybe isCurrent would be a appropriate name? (+ documentation in any case)

core/src/tangram.cpp
hjanetzek: Yes, hasGeometry was probably the old test for isReady :)

core/src/tile/labels/labels.cpp
hjanetzek: This is how the modelmatrix of a tile is calculated by the call tile->update(view) in tileworker, minus the use of the current view position. 
I wanted to remove the access the current view state from a worker thread. So this (should) set up the labels as if the tile was at the map center.

core/src/tile/tileWorker.cpp
tallytalwar: yup, makes sense now. I forgot about the proxies :D.

core/src/tile/mapTile.h
hjanetzek: fixed

core/src/tile/mapTile.h
hjanetzek: Just checked: the operator is overloaded to use store()

core/src/tile/tileManager.cpp
hjanetzek: Until we have decided on a more elaborate logging mechanism, local DBG macros are handy to switch on while developing a particular feature. Can be removed now or later

blair1618: Agreed that it's nice to switch debug messages on/off per file.

core/src/tile/labels/labels.cpp
karimnaaji: Yes, we need the projection matrix to project labels to screen ensuring the label is 'valid' position/state speaking after being created. We could probably store the current projection matrix and position in a member though.

core/src/tile/labels/labels.cpp
```diff
@@ -19,6 +21,8 @@ int Labels::LODDiscardFunc(float _maxZoom, float _zoom) {
 bool Labels::addLabel(MapTile& _tile, const std::string& _styleName, Label::Transform _transform, std::string _text, Label::Type _type) {
     auto currentBuffer = m_ftContext->getCurrentBuffer();
 
+    // FIXME: the current view should not be used to determine whether a label is shown at all
+    // otherwise results will be random
     if ( (m_currentZoom - _tile.getID().z) > LODDiscardFunc(View::s_maxZoom, m_currentZoom)) {
```
karimnaaji: Right! This should be moved to the update, to discard based on lod interactively.

core/src/tangram.cpp
blair1618: Yep, `hasGeometry()` is redundant now.

core/src/tile/labels/labels.cpp
blair1618: What's the reason for doing this update here and not, say, in the main update loop after the tile is added (https://github.com/tangrams/tangram-es/blob/master/core/src/tangram.cpp#L124)?

hjanetzek: @karimnaaji, couldn't one use a default matrix for the initial update to project the tile to a fixed size at which it is intended to be displayed? (basically ignoring tilt and rotation from the Views viewProjection matrix)

karimnaaji: Well, some reason is having the projected position to build geometry out of it. I guess we could remove it, if update is always called before render on tangram api it should be fine. Eventually, this could result in an extra geometry buffer update, i'll check this. Also, I think that this was needed before we fixed concurrency issues and introduce the merge of pending labels from threads to main thread.
@hjanetzek I guess this could work, but feel free to remove the view and this update if it makes things easier!

hjanetzek:
Any objections to merge this branch that I missed?

blair1618:
@hjanetzek Is that discussion in labels.cpp resolved? I was waiting on that, if it's resolved then feel free to rebase onto master!

hjanetzek:
@blair1618 I think it is - as far as the TileWorker itself was involved. The more general issue of accessing view position on the worker thread (in labels.cpp) is out of scope for this branch :)

blair1618:
@hjanetzek Sounds good. I'll let you do the honors!

hjanetzek:
Thanks!

Nice, github is smart enough to close pull requests when pushing the locally merged branch. I guess their merge button does not have the --no-ff switch, or does it?

blair1618:
Yeah github always seems to create a merge commit so I would expect that it's using --no-ff, not totally sure though.

PR #164, opened 2015-06-26
Generation of mipmaps

karimnaaji:
Adding the possible generation of mipmaps since this option is available on the web version (and tilting the view can generally requests a lot of mipmaping usage).
This options probably needs to have more control in the stylesheet (cf [docs](https://github.com/tangrams/tangram-docs/blob/a1b65cfc7fa0dcbc17e8cf52a3eb870f4a2677fa/pages/textures.md)), since mipmap generation, how you want the mipmap generation to be filtered at minification (happening once) and filtering are different concepts. 
I'd actually would expect to separate those concept in the stylesheet, having a control on the mipmapping generation/filtering, and actual filtering at sampling. cc @bcamper


karimnaaji:
Actually, the control over the mipmap filtering is not possible on ES except if you generate them yourself. I guess it's fine to let the options this way, the only thing is using mipmap generation doesn't reflect which filtering / mipmap fetching is used, so we could probably make it a boolean.

blair1618:
Yeah it's a bit odd that we have syntax for mipmaps but can't specify how they're interpolated. Covering the whole range of texture filtering options could get pretty involved, but it can all be optional and default to sensible values. We could cover all of the filtering possibilities pretty well with syntax like:
```
filtering: nearest | linear | { min: ... , mag: ... }
mipmap: nearest | linear | none
```
Defaulting to NEAREST_MIPMAP_LINEAR for minification and LINEAR for magnification (for instance).

@karimnaaji the documentation I've read makes it seem like we can specify mipmap filtering in ES 2.0 (http://docs.gl/es2/glGenerateMipmap, http://docs.gl/es2/glTexParameter) I'm not sure what you mean about this.

karimnaaji:
Yes having this syntax definitely makes sense to me, since it reflects the difference. Also, we should suggest not to use it for sprites which are most of the time minified or magnified at relatively small ratio, and would use more memory. 

Concerning the es 2.0 docs, I was thinking of something else (controlling how the actual mipmaps are filtered at generation, but what you can control is actually how to interpolate between two levels, saying that when a level is selected, giving nearest mipmap means a single mipmap level will be used, whereas linear may interpolate between levels).

blair1618:
Will review and merge, though we should keep a trello card for the task of improving stylesheet configuration of filtering and mipmapping.

bcamper:
Some additional stylesheet syntax to provide more control over mipmapping/filtering as you propose makes sense. We can keep the current `mipmap` setting to have the current defaults for min/max, and allow those to be overridden with new settings (or object syntax like `filtering: { min: ..., max: ... }`).

PR #165, opened 2015-06-26
Adds gradle snapshot deployment

ecgreb:
Deploys a snapshot AAR to Sonatype OSS snapshot repository on each successful build of master.

### DIFF
--- a/android/tangram/build.gradle
+++ b/android/tangram/build.gradle
@@ -1,12 +1,7 @@
 buildscript {
-  buildscript {
-    repositories {
-      mavenCentral()
-    }
-    dependencies {
-      classpath 'com.android.tools.build:gradle:1.2.3'
-      classpath 'com.github.dcendents:android-maven-plugin:1.2'
-    }
+  dependencies {
+    classpath 'com.android.tools.build:gradle:1.2.3'
+    classpath 'com.github.dcendents:android-maven-plugin:1.2'
   }
 }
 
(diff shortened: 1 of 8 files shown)


blair1618:
Is there a convention of indenting with 2 spaces in gradle files? We use 4 spaces everywhere but it's not a huge deal

ecgreb:
Yeah 2 spaces is the standard indent for gradle/groovy.

android/tangram/build.gradle
```diff
@@ -1,20 +1,31 @@
...
+  }
+}
 
-    compileSdkVersion 15
-    buildToolsVersion "21.1.2"
+apply plugin: 'com.android.library'
+apply plugin: 'com.github.dcendents.android-maven'
 
-    defaultConfig {
-        minSdkVersion 15
```
tallytalwar: @ecgreb Don't we have to mention the minSdkVersion and targetVersion?

blair1618:
@ecgreb you may need to update the android platform version in the travis setup script to "22": https://github.com/tangrams/tangram-es/blob/6c17f78970e097ef4fd5511fa58489760d22d6b1/travis/before_install.sh#L48

Travis only installs the parts of the sdk that we need for building, to save time. It's a bit of a pain to configure this in two places, but we don't change it very often.

android/tangram/build.gradle
ecgreb: Nice catch. Re-added.

PR #166, opened 2015-06-26
Sublayer

tallytalwar:
We have subLayers

Yaml parsing
- Recursive addition of sublayers to layers

Filtering layers, style paramater mixing, style paramater caching
- Does a BFS layer graph traversal to determine appropriate style parameter for a feature.

NOTES:
Because of how unique set of layer match IDs are determined, this puts a cap on the number of layers in the scene file (64 aka 64 bit key for the cache).


blair1618:
Reviewing now

core/src/scene/sceneLayer.h
```diff
@@ -0,0 +1,36 @@
...
+
+#include "filters.h"
+#include "style/styleParamMap.h"
+
+namespace Tangram {
+
+    class SceneLayer {
+        std::vector<SceneLayer*> m_subLayers;
+        /*
+         * void* could be any value corresponding to the the named style property key
```
blair1618: can you clarify or remove this comment?

tallytalwar: I can not recollect now why I had that comment. Will take it out for now.

core/src/scene/sceneLayer.h
```diff
@@ -0,0 +1,36 @@
...
+#include <vector>
+#include <unordered_map>
+
+#include "filters.h"
+#include "style/styleParamMap.h"
+
+namespace Tangram {
+
+    class SceneLayer {
+        std::vector<SceneLayer*> m_subLayers;
```
blair1618: Why not a vector of smart pointers or just values?

core/src/style/style.cpp
```diff
@@ -69,10 +79,59 @@ void Style::setLightingType(LightingType _type){
...
+    sLayers.push_back(_uberLayer);
+
+    auto sLayerItr = sLayers.begin();
+
+    //A BFS traversal of the SceneLayer graph
+    while (sLayerItr != sLayers.end()) {
+
+        if ( (*sLayerItr)->getFilter()->eval(_feature, _ctx)) { // filter matches
+
+            _uniqueID += (1 << (*sLayerItr)->getID());
```
blair1618: Using bitmasks would be a really efficient way of indexing style combinations, but as you mention below it restricts the number of possible sublayers and that's not gonna be acceptable in the long run. Also, I'm not sure this will behave correctly in every case; is style parameter mixing order-dependent?

tallytalwar: I agree, hence the comments for that.

Do you have any suggestions? I can look more into it.

bcamper: There's not really a concept of order in which styles match. The order that matters is to ensure that properties from "deeper" styles take precedence over shallower ones. (There is ambiguity when multiple styles at the same depth level modify the same properties.) 

@blair1618 is right that we can't limit the # of styles prematurely though. Is it expensive to use a string key? Probably. Or maybe some kind of hashing function.

tallytalwar: @bcamper nopes not too expensive. I am going to push a commit with that change.

blair1618: @bcamper right, the resulting "mixed" style will depend on the order in which the layers are traversed, but there is only one sensible order of traversal so I think we don't actually need to worry about that. We might just use a string key.

core/src/scene/sceneLayer.h
tallytalwar: I can do vector of shared_ptr. 

Why not vector of values: in applyLayerFiltering I am using a local vector to store all the layers for BFS traversal. If I do by value I will be make a copy for every layer, which I can avoid.

tallytalwar: shared_ptr here: a2d8ade21d46d337252a0c099e41d1c3ed8f7512

core/src/style/style.cpp
hjanetzek: I guess there is some reasonable limit for how many sublayers to support - In this case one could use std::bitset<256> which also has the advantage (compared to string) to not use additional heap allocations.

tallytalwar: :+1:

core/src/scene/sceneLoader.cpp
```diff
@@ -744,12 +744,40 @@ void SceneLoader::parseStyleProps(Node styleProps, StyleParamMap& paramMap, cons
...
+            logMsg("Error: Badly formed Style property, need to be a scalar, sequence or map."
+                    "%s will not be added to stype properties.\n", paramKey.c_str());
         }
     }
 
 }
 
+void SceneLoader::loadSublayers(YAML::Node layer, std::vector<std::shared_ptr<SceneLayer>>& subLayers) {
+
+    for (auto subLayerItr = layer.begin(); subLayerItr != layer.end(); ++subLayerItr ) {
```
blair1618: Use range-based for loops where possible. e.g.
```
for (auto sublayer : layer)
```

core/src/scene/sceneLoader.cpp
```diff
@@ -744,12 +744,40 @@ void SceneLoader::parseStyleProps(Node styleProps, StyleParamMap& paramMap, cons
...
+
+        std::string subLayerName = subLayerItr->first.as<std::string>();
+        if( subLayerName != "data" && subLayerName != "filter" && subLayerName != "draw" &&
```
blair1618: This can be nicely written as an early return
```
for (auto reserved : { "data", "filter", "draw", "properties" }) { if (subLayerName == reserved) { return; } }
```

tallytalwar: How about something like this: https://github.com/tangrams/tangram-es/blob/0d8a138dfaade55b61aedaa625d2b5665e7c9af2/core/src/scene/sceneLoader.cpp#L764-L767

core/src/style/style.cpp
```diff
@@ -68,57 +75,110 @@ void Style::setLightingType(LightingType _type){
...
+                /* update StyleParam with subLayer parameters */
+                auto& sLayerStyleParamMap = sceneLyr->getStyleParamMap();
+                for(auto& styleParam : sLayerStyleParamMap) {
+                    auto it = _styleParamMapMix.find(styleParam.first);
+                    // C++17 has an insert_or_assign which does the same thing.
+                    if(it != _styleParamMapMix.end()) {
+                        it->second = styleParam.second;
+                    } else {
+                        _styleParamMapMix.insert(styleParam);
+                    }
```
hjanetzek: How is this different from

    _styleParamMapMix[styleParam.first] = styleParam.second; 

The separate insert would only save a copy if styleParam could be moved.

tallytalwar: You are suggesting to replace 
```
                    if(it != _styleParamMapMix.end()) {
                        it->second = styleParam.second;
                    } else {
                        _styleParamMapMix.insert(styleParam);
                    }
```
with
```
_styleParamMapMix[styleParam.first] = styleParam.second; 
```
?

tallytalwar: Ok got it ... Yes its the same.. will change it.

core/src/scene/sceneLoader.cpp
hjanetzek: Just got an idea for a syntax to separate keywords from sublayers: Keys don't have to be scalars so using a one item list like [sublayer]. This would allow to differentiate keyword from layers (and would also avoid some problems when introducing new keywords later on :)

hjanetzek: Or maybe simpler, one nice looking prefix character for sublayers should do it.

blair1618: @hjanetzek I agree, that would make things easier. However we're following the syntax that's been established by tangram-js: https://github.com/tangrams/tangram-docs/blob/gh-pages/pages/layers.md#sublayer-name
To keep from needing a list of reserved words we could check if the element has a child called 'draw'

tallytalwar: @blair1618 Yes thats what I did, when I fixed a change in the config yaml to add draw groups for all sublayer .. I realized we can do this.

core/src/style/style.cpp
```diff
@@ -68,57 +75,104 @@ void Style::setLightingType(LightingType _type){
...
+                        // Build lines
+                        for (auto& line : feature.lines) {
+                            buildLine(line, parseStyleParams(styleParamMapMix), feature.props, *mesh);
+                        }
+                        break;
+                    case GeometryType::polygons:
+                        // Build polygons
+                        for (auto& polygon : feature.polygons) {
+                            buildPolygon(polygon, parseStyleParams(styleParamMapMix), feature.props, *mesh);
+                        }
```
hjanetzek: Where does the result of parseStyleParams go? And why is it computed for every feature?

tallytalwar: Good you pointed this out. There is also a memory leak wrt to this. Fixing it.

tallytalwar: @hjanetzek Refer 97e087d.

I made some improvement, but on profiling noticed cssColorParser library is causing some performance hiccups. In some random profiling tries, CSSColorParser::parse is using about 12-16% of the total time. 
But if I do not use any css style color in the style sheet parseStyleParams method blazes through.

core/src/scene/sceneLayer.cpp
```diff
@@ -0,0 +1,19 @@
...
+
+    SceneLayer::SceneLayer(const std::vector<std::shared_ptr<SceneLayer>>&& _subLayers, const StyleParamMap&& _styleParamMap, const std::string _name, Filter* _filter) :
+        m_subLayers(std::move(_subLayers)), m_styleParams(std::move(_styleParamMap)), m_name(_name), m_filter(_filter) {
+
+        m_id = s_layerCount++;
+
+    }
+
+    SceneLayer::~SceneLayer() {
+        m_subLayers.clear();
```
blair1618: No need to clean up STL containers - there may be other instances of this in the repo from back when I was still learning c++

tallytalwar: :+1:

core/src/style/style.cpp
```diff
@@ -68,57 +75,104 @@ void Style::setLightingType(LightingType _type){
...
 
     m_layers.push_back(std::move(_layer));
 
 }
 
+void Style::applyLayerFiltering(const Feature& _feature, const Context& _ctx, std::bitset<MAX_LAYERS>& _uniqueID,
+                                   StyleParamMap& _styleParamMapMix, std::weak_ptr<SceneLayer> _uberLayer) const {
+
+    std::vector<std::weak_ptr<SceneLayer>> sLayers;
+    sLayers.reserve(_uberLayer.lock()->getSublayers().size() + 1);
```
blair1618: What's the benefit of a `weak_ptr` if you're not going to check the result of `lock`?

tallytalwar: yes good catch. My bad. Will fix.

blair1618:
@tallytalwar You overwrote a commit I made, it was small so no big deal this time but do try to be careful when rewriting history

tallytalwar:
@blair1618 oops... so sorry about that :|.

blair1618:
The tile construction process is going to need some radical changes. Currently we have no way to change the base style from within a sublayer or even to use different data sources among top-level layers. I have some ideas on this that I can start implementing ASAP. 

I think it might make the most sense to just merge these changes with the understanding that many things in here will be substantially altered soon.

tallytalwar:
@blair1618 I agree ... We have already discussed that this branch is not addressing such refactors.

I have comments in the code stating the limitations you mentioned above.

Also I am currently working on multiple datasource for the uber layer.

bcamper:
I am generally not a fan of rewriting history on any shared branches for
this reason. I've seen it cause confusion many times. I prefer merge
commits, they are not as pretty, but IMO more likely to "just work".

(I regularly alter history on local changes before pushing, or on a branch
only I am working on.)

tallytalwar:
@bcamper I just learnt altering history so I was doing that a lot here, but I was the only one working on this branch, but unfortunately missed a small commit from Matt. Will take care of this in the future.

bcamper:
@tallytalwar right, the point is that such mistakes are much less likely if the mechanism that enables them isn't used in the first place.

tallytalwar:
Any more suggestions or improvements on this?

Good to be merged?

PR #167, opened 2015-06-29
Specify glfw3>=3.1 in pkg-config check

moskvax:
Versions of glfw3 < 3.1, such as Debian's which is currently stuck at 3.0.4, can't build tangram-es as they lack glfwPostEmptyEvent, so in such cases it's better to fail the build at config time with a comprehensible error.


tallytalwar:
Thanks Stephen, this is good suggestion.

PR #168, opened 2015-06-30
Linux requires libcurl4-openssl-dev package to compile successfully

vespakoen:
Title said it all


vespakoen:
I thought I was writing a pull request while in reality I was writing a commit message, sorry for that.
Also, you are probably able to document installing the of the libcurl dev package better than me, since there are also multiple options to install, I chose libcurl4-openssl-dev, which works.

Thanks for the great work on this package btw, it is super sweet!

PR #173, opened 2015-07-07
add BoundingBox struct

hjanetzek:
- used for MapProjection Tile*Bounds methods


blair1618:
There's this bounding box that we use in other places: https://github.com/tangrams/isect2d/blob/master/include/aabb.h
Though double precision is probably important for global tile boundaries.

PR #176, opened 2015-07-07
Reduce vbo binding calls

karimnaaji:
At upload/sub-data-upload, vertex buffer objects are bounds two times, this reduces the buffer binding.

### DIFF
--- a/core/src/util/vboMesh.cpp
+++ b/core/src/util/vboMesh.cpp
@@ -67,34 +67,42 @@ void VboMesh::update(GLintptr _offset, GLsizei _size, unsigned char* _data) {
     m_dirty = true;
 }
 
-void VboMesh::subDataUpload() {
-    if (m_dirtySize != 0) {
-        glBindBuffer(GL_ARRAY_BUFFER, m_glVertexBuffer);
+bool VboMesh::subDataUpload() {
+    if (m_dirtySize == 0) {
+        return false;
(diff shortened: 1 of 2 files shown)


core/src/util/vboMesh.cpp
```diff
@@ -67,34 +67,42 @@ void VboMesh::update(GLintptr _offset, GLsizei _size, unsigned char* _data) {
...
+    glBindBuffer(GL_ARRAY_BUFFER, m_glVertexBuffer);
 
-        // updating the entire buffer
-        if (vertexBytes - m_dirtySize < m_vertexLayout->getStride()) {
+    long vertexBytes = m_nVertices * m_vertexLayout->getStride();
 
-            // invalidate the data store on the driver
-            glBufferData(GL_ARRAY_BUFFER, vertexBytes, NULL, m_hint);
+    // updating the entire buffer
+    if (m_dirtySize - vertexBytes < m_vertexLayout->getStride()) {
```
blair1618: Maybe I'm confused, but shouldn't this be `vertexBytes - m_dirtySize` ? `vertexBytes` should always be larger than `m_dirtySize` so this comparison doesn't make sense to me.

karimnaaji: Oh yes! This was making sense with  std::abs, not anymore.

PR #177, opened 2015-07-07
Remove autodelete from texture

karimnaaji:
Since textures are no longer created from threads.


core/src/style/spriteStyle.cpp
```diff
@@ -7,21 +7,18 @@
 #include "util/vertexLayout.h"
 
 SpriteStyle::SpriteStyle(std::string _name, GLenum _drawMode) : Style(_name, _drawMode) {
-}
-
-SpriteStyle::~SpriteStyle() {
-
-    m_texture->destroy();
 
+    constructVertexLayout();
```
tallytalwar: Already happening in style::build()

PR #178, opened 2015-07-07
Disabling logs

karimnaaji:
Simple way of disabling log for now, `make platform LOG=0` to disable, we could introduce log levels if needed.


karimnaaji:
Just talked with @tallytalwar about this, and it'd probably makes sense to have control over this through tangram api. Closing

PR #182, opened 2015-07-08
Dynamic meshes

karimnaaji:
Following the work on the transformation of the text buffer to a mesh. 

This PR introduces dynamic mesh updates, it basically adds the possibility to update an attribute/vertex on a specific location inside the mesh, useful for text/sprites. 

Once you have updated an attribute/vertex, the dirtiness is then computed in order to send only the most efficient part to update to OpenGL, some sync could still happen while buffering the data since the vbo could be in use on current frame but I have some thoughts on how to fix that if it becomes an issue in render time.
`updateAttribute` method is not cache-friendly, we'd have to consolidate that.

Once this gets merged in text-buffer-mesh branch, I'd follow up with the sprites PR.

### DIFF
--- a/tests/unit/meshTests.cpp
+++ b/tests/unit/meshTests.cpp
@@ -73,7 +73,7 @@ TEST_CASE( "Right merge on vertices dirtiness", "[Core][TypedMesh]" ) {
 TEST_CASE( "Update on second attribute of the mesh for n vertices", "[Core][TypedMesh]" ) {
     auto mesh = newMesh(10);
     int nVert = 5;
-    size_t stride_b = 1 * sizeof(float); // stride of a in the struct
+    size_t stride_b = sizeof(float); // stride of a in the struct
 
     mesh->updateAttribute(stride_b + sizeof(Vertex), nVert, 0.f);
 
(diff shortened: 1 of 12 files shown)


blair1618:
`SpriteStyle::SetSpriteAtlas` seems to be declared but not implemented anywhere

tallytalwar:
Reviewing this now.

tests/unit/meshTests.cpp
```diff
@@ -0,0 +1,120 @@
...
+    mesh->updateVertices(4 * sizeof(Vertex), 2, Vertex());
+
+    REQUIRE(mesh->getDirtyOffset() == 2 * sizeof(Vertex));
+    REQUIRE(mesh->getDirtySize() == 4 * sizeof(Vertex));
+}
+
+TEST_CASE( "Update on second attribute of the mesh for n vertices", "[Core][TypedMesh]" ) {
+    auto mesh = newMesh(10);
+    int nVert = 5;
+    size_t stride_a = 1 * sizeof(float); // stride of a in the struct
```
tallytalwar: just naming ... stride of a is actually 0. Should be named stride of b.

stride_a : 0
stride_b : sizeof(a)
stride_c : sizeof(a) + sizeof(b)
stride_d : sizeof(a) + sizeof(b) + sizeof(c)

Line: 96, 97 have the correct naming.
(refer https://github.com/tangrams/tangram-es/pull/182/files#diff-5a38a481e64efdae8ee018efbfd7b98aR96)

PR #186, opened 2015-07-10
Render states

karimnaaji:
Keep track of the render states to reduce redundant driver state changes, it lazily updates them. For now it's only concerning the states we are modifying/would need to modify style-wise (blending/depth test/depth write), we could update with stencil if needed in future.


hjanetzek:
:+1:  Just wondering if using the () operator wouldn't be clearer - I recall that I first wondered what the assignment does (if it actually triggers an action)

karimnaaji:
Yes good suggestion! I'll make that change

karimnaaji:
One thing to note in here, blending geometry may have blending with not all of the opaque layers, this is because non-ordering between opaque/non-opaque geometry.

core/src/util/renderState.h
```diff
@@ -0,0 +1,101 @@
...
+        inline static void set(const Type& _type) {
+            if (_type.culling) {
+                glEnable(GL_CULL_FACE);
+                glFrontFace(_type.frontFaceOrder);
+                glCullFace(_type.face);
+            }
+        }
+    };
+
+    inline bool operator!=(const Culling::Type& _lhs, const Culling::Type& _rhs) {
```
tallytalwar: Why not make operator!= a member for Culling?

karimnaaji: No particular reason, I'll change that

PR #189, opened 2015-07-14
Reorg proposal

karimnaaji:
The `util` folder has become too big in my opinion to quickly reflect what's the purpose of the files inside it in general.
This PR is a suggestion to split this folder, adding a `gl`, rename `mapTile` to `tile`, and move some other files where they should be for some more consistency with the webgl side.


tallytalwar:
With this I will also recommend using a util namespace.

blair1618:
@karimnaaji you read my mind, I've been wanting to do this for weeks now.

I don't find it that important to add a sub-namespace. Namespaces exist to solve naming collisions and we're not having any problems with parts of our own code name-colliding.

blair1618:
Somewhat related: I've been kind of undecided on #include convention - our build commands include every source folder, so we don't *need* to provide the full path to a file inclusion. On the one hand I find it informative to #include a full path if the files aren't in the same folder (you immediately know where to find the file) but on the other hand it's wordier and breaks if the file moves. We should pick one or the other though, instead of the mix of conventions that we've used so far.

tallytalwar:
I prefer to have the extra subfolder name in the include folder, even though its redundant for building because our build process takes care of it.

karimnaaji:
I feel like it's easier to see the structure by having extra subfolder structure in includes, and also helps to quickly see usage between different modules.

karimnaaji:
For future update on organization I would suggest moving the platform-dependent folders into a `platform` one where we would have the different platform layers, with a `common` subfolder in it. This could help to easily share common code (osx and ios are sharing some, same for linux and raspberry pi). Except if we want to keep the platform layers example relatively isolated between each other.

blair1618:
While we're at it, you think it makes sense to move the "labels" folder out of "tile"? There's a lot in there now and I sometimes forget where to find it

karimnaaji:
Yes it makes sense, I'll move that up to `src/`

blair1618:
Lookin' good. I like the idea of trying to reorganize some of the platform specific/common files as well, let's make that a separate discussion and PR. @karimnaaji if you're done with edits feel free to merge.

PR #192, opened 2015-07-15
Filters are concrete types

blair1618:
Can't leak them anymore ;)

I also like that this eliminates all the virtual calls that filters had to make


blair1618:
This should be good to go as soon as travis is done.

As a follow-up we should separate the filter tests from the filter parsing tests, they're not super-useful as unit tests as-is.

tallytalwar:
I am in process of reviewing this.

PR #196, opened 2015-07-17
Merged sprites

hjanetzek:
git show a78c5e15 (or git diff-tree --cc a78c5e15) helps to spot which parts were touched during conflict resolution

PR #197, opened 2015-07-17
Blend/Opaque style ordering

karimnaaji:
This fixes styles that uses blending (sprite & text for now) that are drawn in arbitrary order. 
PR for #196


blair1618:
Nice easy fix - I like it.

When we start using the `blend` property from the stylesheet we may want to generalize `isOpaque()` to something that accounts for a configurable blend mode on any style.

karimnaaji:
Yes that's what I was thinking about, we'll have to generalize a bit the process for this.

PR #199, opened 2015-07-17
Sub layer fixes

tallytalwar:
Minor but critical fixes to subLayers.
This branch is based on "concrete-filter" should be merged post that.

### DIFF
--- a/core/src/scene/sceneLayer.cpp
+++ b/core/src/scene/sceneLayer.cpp
@@ -4,7 +4,7 @@ namespace Tangram {
 
     uint8_t SceneLayer::s_layerCount = 0;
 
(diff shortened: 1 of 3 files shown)


tallytalwar:
Yes in the BFS traversal for Style:: applyLayerFiltering.

blair1618:
Raw pointers could be safely used there regardless, no allocation or deletion is happening. 

I also try to avoid type aliases. Naming is one of the two hardest problems in computing, so if we can avoid doing it, why not? :)

tallytalwar:
I reverted the last commit.

and yes shared_ptr are needed for SceneLayer since, multiple raw groups can share these layers.

tallytalwar:
#199 PR.

PR #202, opened 2015-07-17
Sprite and label tweaks

hjanetzek:
@karimnaaji, could you comment on this?

I've change the use of GLvoid* and GLintptr to size_t as these are treated as offsets into the vbo when a vbo is bound during the call to glVertexAttribPointer [https://www.khronos.org/opengles/sdk/docs/man/xhtml/glVertexAttribPointer.xml]. And I find this conceptually more clear than to provide a pointer type just because the c api has no overloads :)

Edit: I'm also ok with GLintptr instead of GLvoid* for the getAttribOffset. Though I would prefer to use the types provided by c++ than to pull in GL headers into everything just for some typedefs.
GLintptr is http://en.cppreference.com/w/cpp/types/ptrdiff_t


karimnaaji:
Concerning GL types, we've had some discussions around it some time ago, and decided to use GL types if used for GL calls, but your point on headers totally makes sense.

core/src/labels/spriteLabel.cpp
```diff
@@ -12,28 +12,16 @@ void SpriteLabel::pushTransform(VboMesh& _mesh) {
...
-
-        GLintptr stride = std::min<GLintptr>(m_attribOffsets.m_position, std::min<GLintptr>(m_attribOffsets.m_alpha, m_attribOffsets.m_rotation));
-
-        VertexAttributes newAttributes {
-            m_transform.m_screenPosition + m_offset,
-            m_transform.m_alpha,
-            m_transform.m_rotation,
-        };
+        BufferVert::State state(m_transform.state);
+        state.screenPos += m_offset;
```
karimnaaji: I think we could skip this extra copy by moving `m_offset` up to the label class, (so that text label would also benefit from it) and apply it in `updateScreenTransform`.

hjanetzek: Either that, offset could also be used for TextLabels or override setScreenPosition in SpriteLabel. Any preference?

karimnaaji: I feel like it'd be better to implement offsets for both. The only difference would be to offset based on the orientation for `line` labels, if that makes sense?

hjanetzek:
the issue with the most :+1: to be resolved is :drumroll: https://github.com/isaacs/github/issues/18

core/src/labels/spriteLabel.cpp
hjanetzek: I've made offset part of Label::Transform - does that fit?

karimnaaji: Yes!

PR #205, opened 2015-07-20
State invalidation

karimnaaji:
When context gets destroyed, invalidate the render states.

### DIFF
--- a/core/src/gl/renderState.cpp
+++ b/core/src/gl/renderState.cpp
@@ -1,32 +1,25 @@
 #include "renderState.h"
 
 namespace RenderState {
-    
+
     State<Blending> blending = State<Blending>();
     State<DepthWrite> depthWrite = State<DepthWrite>();
     State<Culling> culling = State<Culling>();
     State<BlendingFunc> blendingFunc = State<BlendingFunc>();
     State<DepthTest> depthTest = State<DepthTest>();
-    
+
     void configure() {
         blending.init(GL_FALSE);
         culling.init({GL_TRUE, GL_CCW, GL_BACK});
+        blendingFunc.init({GL_SRC_ALPHA, GL_ONE_MINUS_SRC_ALPHA});
         depthTest.init(GL_TRUE);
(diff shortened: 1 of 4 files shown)


core/src/gl/renderState.cpp
```diff
@@ -21,4 +21,12 @@ namespace RenderState {
...
     }
     
+    void invalidateAllStates() {
+        blending.invalidate();
+        culling.invalidate();
+        depthTest.invalidate();
+        depthWrite.invalidate();
+        blendingFunc.invalidate();
+    }
+
```
hjanetzek: Couldn't we use an 'invalidate' flag in render() and call 

    if (invalidate) { RenderState::configure();  invalidate = false; }

instead?

karimnaaji: Yes this works, I didn't want to break the pattern followed so far for invalidation though (the gl object itself is aware of invalidation and regenerate/reinit itself on context destroy).

blair1618: We get a call to `onContextDestroyed()` whenever invalidation is needed, so an extra state variable seems unnecessary

blair1618: But it does seem like we should be able to call `configure()` to set the states again, right?

karimnaaji: Yeah we could reinit whenever context gets loss, but does `onContextDestroyed()` happens before or after the new context gets created?

hjanetzek: I mean one could set the flag 'invalidate' in onContextDestroyed on then check in render whether to re-init the state.

... below code misses one m_valid = true, btw

blair1618:
This makes sense to me now. `onContextDestroyed()` is called after a context loss immediately before rendering begins again.

hjanetzek:
@blair, right - I wasn't sure whether onContextDestroyed will only be called on the android render thread

blair1618:
Ah yes, that's an important qualification! It's true for the way we're using it.

PR #209, opened 2015-07-22
csscolorparser tweaks

hjanetzek:
Using  binary instead of linear search for NamedColor lookup and moved the NameColor part after checking the other format options.

Still, this should be done only once when loading the config

### DIFF
--- a/core/dependencies/cssColorParser/CMakeLists.txt
+++ b/core/dependencies/cssColorParser/CMakeLists.txt
@@ -1,6 +1,5 @@
 set(SOURCE_FILE csscolorparser.cpp)
 set(HEADER_FILE csscolorparser.hpp)
 
-include_directories(${HEADER_FILE})
 add_library(libcsscolorparser ${SOURCE_FILE})
 
--- a/core/dependencies/cssColorParser/csscolorparser.cpp
+++ b/core/dependencies/cssColorParser/csscolorparser.cpp
@@ -203,27 +203,34 @@ float clamp_css_float(T f) {  // Clamp to float 0.0 .. 1.0.
     return f < 0 ? 0 : f > 1 ? 1 : f;
 }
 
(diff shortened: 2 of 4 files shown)


blair1618:
Hey do you think this would make sense to submit to the original css-color-parser? (https://github.com/kkaefer/css-color-parser-cpp)

We could also make our own fork of that and copy the file in from there, just to track changes more neatly.

bcamper:
Right, unclear to me how important this is as a general purpose change. As you note, the overarching speed issue is that we are parsing these colors way too many times and can address that with style caching. 

If we think this is a general improvement, we should benchmark and then can submit back as a PR. Either way I think it's not ideal to be maintaining changes to a local copy of the file in a larger repo - can we fork the repo to tangrams/css-color-parser-cpp and make changes there?

hjanetzek:
@blair1618 I think Konstantin will never have used the port for something serious, considering how unoptimized it is.
@bcamper, sure I can move csscolorparser to a separate repo and add it as submodule. The performance benefit is obvious: First there was always an exhaustive iteration through all named colors even in the case where it is a \#rrggbb or rgb(...) formatted string. When the string is a named color it changed from O(n) to O(log_2 n) for the lookup

hjanetzek:
Any objections that I fork csscolorparser to tangrams organization and apply my changes to it?
I think it may be a bit over the top to include it as submodule given that the chances are low that the code will change in important ways in future. 
@bcamper I found that these files are actually included in mapbox-gl - So, it is indeed used seriously but also only for parsing once in style loading.

blair1618:
@hjanetzek sounds good to me

bcamper:
@hjanetzek sounds good. We also want to be parsing these colors one time, but that's a future optimization. Still good to have the parsing speed improved either way :)

PR #210, opened 2015-07-22
Add Tangram namespace

blair1618:
All headers and source files are now within the Tangram namespace. 

I don't like touching many files at once, but it had to be done eventually and it was becoming problematic for the refactoring work I'm doing. 

There are several formatting conventions to choose from regarding namespaces. I've opted to not add indentation for namespaces, a convention followed by LLVM, Valhalla, and others (furthermore, adding indentation for a namespace would require changing every existing line of code in the repo for this pr :P )

### DIFF
--- a/core/src/data/dataSource.cpp
+++ b/core/src/data/dataSource.cpp
@@ -7,7 +7,7 @@
 #include "tile/tileManager.h"
 #include "labels/labels.h"
 
-//---- DataSource Implementation----
+namespace Tangram {
 
 DataSource::DataSource(const std::string& _name, const std::string& _urlTemplate) :
     m_name(_name), m_urlTemplate(_urlTemplate) {
@@ -98,3 +98,4 @@ void DataSource::cancelLoadingTile(const TileID& _tileID) {
     cancelUrlRequest(constructURL(_tileID));
 }
 
+}
--- a/core/src/gl/texture.cpp
+++ b/core/src/gl/texture.cpp
@@ -8,6 +8,8 @@
 
 #include <cstring> // for memset
 
+namespace Tangram {
+
 GLuint Texture::s_boundTextures[] = { 0 };
 GLuint Texture::s_activeSlot = GL_TEXTURE0;
 int Texture::s_validGeneration = 0;
@@ -207,5 +209,7 @@ void Texture::invalidateAllTextures() {
     s_activeSlot = GL_TEXTURE0;
 
     ++s_validGeneration;
+
+}
 
 }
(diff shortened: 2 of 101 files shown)


core/src/data/dataSource.cpp
```diff
@@ -7,7 +7,7 @@
 #include "tile/tileManager.h"
 #include "labels/labels.h"
 
-//---- DataSource Implementation----
+namespace Tangram {
```
tallytalwar: I would have done "using namespace Tangram" in a cpp file.

core/src/gl/texture.cpp
```diff
@@ -209,3 +211,5 @@ void Texture::invalidateAllTextures() {
     ++s_validGeneration;
 
 }
+
+}
```
tallytalwar: github's :heart: for newlines.

core/src/data/dataSource.cpp
blair1618: That's a possibility. It's functionally equivalent, right?

tallytalwar: Yes its functionally equivalent.

blair1618:
An minor drawback of this formatting is that Xcode seems totally unable to conceive the idea that we might not want to indent within namespaces, I can't even find an option to change it's behavior.

tallytalwar:
I use xcode for debugging purpose :P... 
I can have my VI configured to handle this (and probably my xvim for xcode) :D.

blair1618:
Another possibility here is to indent namespaces in headers (to please Varun and Xcode) and then use `using namespace Tangram;` in implementation files.

tallytalwar:
Ohh please me :D. I said I am in love with namespace indentation but I am already married :P.

core/src/data/dataSource.cpp
blair1618: Actually I've just realized that these are *not* totally equivalent. With 'using', anything not in a class must have its name prefixed by the relevant namespace.

hjanetzek:
maybe a script similar to the uncrustify suggestion helps for xcode http://stackoverflow.com/questions/1035366/is-it-possible-to-customize-the-indent-style-of-xcode

blair1618:
Thanks @hjanetzek I'll check out that script, I'm also going to try the beta for the next xcode, just in case Apple fixed this already (unlikely).

core/src/data/dataSource.cpp
tallytalwar: Also I was reading more on this ... Its recommended not to use "using" unless really needed and that too in limited scopes. So I am good with this.

PR #211, opened 2015-07-22
Camera momentum - fling

karimnaaji:
Let camera move/zoom with exponential momentum while moving with non-zero ending gesture velocity (pan/slide and zoom/pinch, no momentum on rotation). I put some default values to control those gesture momentum with something that I consider giving a good control over the view, and also by comparing the gesture with other maps.
Update tangram API to receive a `velocity` variable on pinch for iOS and Android. For desktop, no velocity is given since GLFW keeps sending events on `scroll` with exponential decrease. For android this velocity on pinch is not available with [`ScaleGestureDetector`](http://developer.android.com/reference/android/view/ScaleGestureDetector.html), so we need to compute it, it's directly available through [`UIPinchGestureRecognizer`](https://developer.apple.com/library/ios/documentation/UIKit/Reference/UIPinchGestureRecognizer_Class/#//apple_ref/occ/instp/UIPinchGestureRecognizer/velocity) for iOS.
Also move input code from `tangram.cpp` to `util/inputHandler` class to keep things clear concerning handling.


blair1618:
Nice, this looks good!

As for scrolling velocity on desktop: I suspect that the decaying scroll events sent to GLFW are from OS X or the macbook trackpad, I can confirm this on my home machine tonight. That said, I don't know if there's a reliable way to detect whether or not we should add momentum on desktop.

hjanetzek:
:+1:  flings smoothly! (with glfw and trackpoint)

karimnaaji:
Cool :)

core/src/util/inputHandler.h
```diff
@@ -0,0 +1,47 @@
...
+
+private:
+
+    void setDeltas(float _zoom, glm::vec2 _translate);
+
+    void onEndGesture();
+
+    std::shared_ptr<View> m_view;
+
+    bool m_gestureOccured;
```
tallytalwar: Constructor should set the default value to true or false. Right now nothing is set.

core/src/util/inputHandler.h
```diff
@@ -0,0 +1,47 @@
...
+
+    // fling deltas on zoom and translation
+    glm::vec2 m_deltaTranslate;
+    float m_deltaZoom;
+
+    /* Momentum config */
+
+    const float m_expDecrease = 0.95f;
```
tallytalwar: You can maybe not have magic numbers and use macros or using declaratives.

core/src/util/inputHandler.cpp
```diff
@@ -0,0 +1,124 @@
...
+#include "platform.h"
+
+InputHandler::InputHandler(std::shared_ptr<View> _view) : m_view(_view) {
+    m_deltaZoom = 0.f;
+    m_deltaTranslate = glm::vec2(0.f);
+}
+
+void InputHandler::update(float _dt) {
+
+    bool hasMomentum = glm::length(m_deltaTranslate) > m_minDeltaLength || std::fabs(m_deltaZoom) > m_minDeltaZoomLength;
```
tallytalwar: I see you are using std::abs and std::fabs both in the code.
c++ has std::abs overloaded to take int/float/double values. It should suffice to use std::abs here (also for consistency).

core/src/util/inputHandler.h
karimnaaji: What would be the difference of using macros?  This number would be magic if used straight away in code without a name on it.

tallytalwar: Fair enough :D.

core/src/util/inputHandler.cpp
```diff
@@ -0,0 +1,124 @@
...
+    m_view->screenToGroundPlane(_endX, _endY);
+
+    float dx = _startX - _endX;
+    float dy = _startY - _endY;
+
+    setDeltas(0.f, glm::vec2(dx, dy));
+
+    m_view->translate(dx, dy);
+
+    onEndGesture();
```
tallytalwar: Correct me if I am wrong here.

Say a Pan gesture occurs with sufficient delta.
onEndGesture marks the m_gestureOccured as "true", which means in the next InputHandler::update() call this delta will not cause an update of view. (However the subsequent update will, since InputHandler::update is setting m_gestureOccured to false).

core/src/util/inputHandler.cpp
karimnaaji: Ok I will fix that, didn't know about that.

core/src/util/inputHandler.cpp
karimnaaji: No you're right this is what's happening, there's a one frame delay, but inputHandler receive the correct information since a gesture really occured. If you ask to use this delta when a gesture happens, then you accumulate a fling on a regular pan.

PR #212, opened 2015-07-22
Cleanup update loop

karimnaaji:
Clean update loop of `tangram.cpp`

### DIFF
--- a/core/src/labels/labels.cpp
+++ b/core/src/labels/labels.cpp
@@ -4,6 +4,7 @@
 #include "text/fontContext.h"
 #include "gl/primitives.h"
 #include "view/view.h"
+#include "style/style.h"
 
 #include "glm/gtc/matrix_transform.hpp"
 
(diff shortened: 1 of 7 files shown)


hjanetzek:
The updateVisibleTiles() doesn't really fit in the area of responsibility of TileManager imo.  (Other than that this commit is rebased and included in my label-refactoring branch)

karimnaaji:
Well.. It doesn't seem like an overwhelming area of responsibility to extend it to.

hjanetzek:
Sure, but I would still argue that it doesn't really fit into it :) 

Consider its current external interface updateTileSet() and getVisibileTiles() - It's sole responsibility is to provide tiles for rendering (and doing a lot of work internally to achieve this). 

Then what the proposed updateVisibleTiles actually does is to updateModelMatricesOfVisibleTilesRelativeToCurrentView() - To me this sounds more like a part of the general update/render loop than anything related to the tile managers work. From the call it is also not clear that the View is used to update the tiles.

karimnaaji:
Merged in #219

PR #213, opened 2015-07-22
Cache screenspace proj matrix

karimnaaji:
Sprites and text now uses the same projection matrix so it's better to move it to the view module and cache it when viewport changes.


hjanetzek:
Looks good to me!

blair1618:
Yeah this makes good sense to me - I'll merge the rebased PR

PR #214, opened 2015-07-23
Benchmarks

karimnaaji:
A simple benchmark integration.

To be used properly, benchmarks should be run on the same machine to have consistency over numbers. We could individually run those benchmarks when working on a specific feature and compare/store the numbers on this machine. It's important to consider that this is desktop only but would give comparison ratio for mobile.

This is the kind of result the execution of  `make benchmark && ./bench/run_bench.sh` is giving on some example benchs:

```shell
Running benchmarks
Running builders.out
Benchmark                        Time(ns)    CPU(ns) Iterations
---------------------------------------------------------------
BM_Tangram_BuildButtMiterLine        1606       1599     417024                                 
BM_Tangram_BuildRoundRoundLine       3239       3230     226536                                 
Running loading.out
Benchmark             Time(ns)    CPU(ns) Iterations
----------------------------------------------------
BM_Tangram_LoadFont     553193     548621       1232     
```


tallytalwar:
This is great. :+1:

blair1618:
Cool, this is a good start. Personally I'm not going to trust a benchmark's numbers unless it's run on an actual phone; I've seen relative performance differ by an order of magnitude between a laptop and a phone (there are very good reasons for this too: memory speed, ARM SSE instructions, et cetera). That said, we can probably adapt this to build a phone-runnable test suite and even run automated benchmarks on remote devices (like [these](https://aws.amazon.com/device-farm/))!

karimnaaji:
Since most of our target are ARM based, we could run those benchmarks on a raspberry Pi. But yes, certainly, the point of this isn't to use those numbers to certify that something runs fast or not on a mobile.

blair1618:
I'm good with merging this after the conflicts are fixed - being able to script benchmarks will almost certainly be useful. 

I think [this line](https://github.com/tangrams/tangram-es/pull/214/files#diff-1c059d175998d31aea1f8bfc1d2efbc7R2) will need to change to say 'c++14' but I don't see any other issues.

blair1618:
Merged with #229

PR #216, opened 2015-07-24
c++14

blair1618:
Use c++14 standard in build process

For android, this required moving to the llvm toolchain and the mindk for travis has been updated appropriately. Everything else "just works".

Possible TODOs:
 - ~~Test on Raspberry Pi~~
 - Consolidate CXX flags in cmake files


tallytalwar:
You can include ndk r10e with this only and I will close the other PR.

blair1618:
It's a bit of an open question to me whether this odd minimal ndk thing will continue to be the best option for our automated builds. Right now it still saves us a solid 5 minutes on each build and I've updated the mindk generator such that we can build for multiple architectures as well, so I'm thinking we'll stick with it for now and see if we find good reasons to try something else in the future.

blair1618:
All platforms tested and working (with the exception of some unrelated issues on raspberry pi)

PR #219, opened 2015-07-26
Labels refactor

hjanetzek:
@karimnaaji, while looking into simplifying the labels logic to work with the new tile-cache (labels of inactive but cached tiles should not have been kept in Labels.m_labels) I stumbled over a few opportunities for optimization and also some issues regarding concurrent use of the font context.

Happy reviewing! :)

### DIFF
--- a/core/src/labels/label.cpp
+++ b/core/src/labels/label.cpp
@@ -131,8 +131,11 @@ void Label::enterState(State _state, float _alpha) {
 }
 
 void Label::setAlpha(float _alpha) {
-    m_transform.state.alpha = CLAMP(_alpha, 0.0, 1.0);
-    m_dirty = true;
+    float alpha = CLAMP(_alpha, 0.0, 1.0);
+    if (m_transform.state.alpha != alpha) {
+        m_transform.state.alpha = alpha;
+        m_dirty = true;
+    }
 }
 
 void Label::setScreenPosition(const glm::vec2& _screenPosition) {
--- a/core/src/labels/label.cpp
+++ b/core/src/labels/label.cpp
@@ -4,8 +4,6 @@
 
 namespace Tangram {
 
-bool Label::s_needUpdate = false;
-
 Label::Label(Label::Transform _transform, Type _type) :
     m_type(_type),
     m_transform(_transform) {
@@ -90,10 +88,11 @@ bool Label::updateScreenTransform(const glm::mat4& _mvp, const glm::vec2& _scree
     return true;
 }
 
-void Label::update(const glm::mat4& _mvp, const glm::vec2& _screenSize, float _dt) {
-    updateState(_mvp, _screenSize, _dt);
-
+bool Label::update(const glm::mat4& _mvp, const glm::vec2& _screenSize, float _dt) {
+    bool animate =  updateState(_mvp, _screenSize, _dt);
     m_occlusionSolved = false;
+
+    return animate;
 }
 
 bool Label::offViewport(const glm::vec2& _screenSize) {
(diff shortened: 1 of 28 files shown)


core/src/labels/labelMesh.h
```diff
@@ -0,0 +1,27 @@
...
+    virtual ~LabelMesh();
+
+    void addLabel(std::unique_ptr<Label> _label);
+
+    std::vector<std::unique_ptr<Label>>& getLabels() {
+        return m_labels;
+    }
+
+protected:
+    std::vector<std::unique_ptr<Label>> m_labels;
```
karimnaaji: What do you think of using a vector of objects instead of pointers here; mostly to force memory contiguity? Also reserving the containers at the beginning of tile building process with worst size to reduce copies all around while adding labels. We would need either different containers (for `spriteLabel` & `textLabel`) or a template (not for the best).

core/src/labels/label.cpp
```diff
@@ -132,8 +131,11 @@ void Label::enterState(State _state, float _alpha) {
 }
 
 void Label::setAlpha(float _alpha) {
-    m_transform.state.alpha = CLAMP(_alpha, 0.0, 1.0);
-    m_dirty = true;
+    float alpha = CLAMP(_alpha, 0.0, 1.0);
+    if (m_transform.state.alpha != alpha) {
```
karimnaaji: We would need a `compare_float` utility here, with a default epsilon value.

core/src/labels/textLabel.cpp
```diff
@@ -2,43 +2,40 @@
...
-void TextLabel::pushTransform(VboMesh& _mesh) {
+void TextLabel::pushTransform() {
     if (m_dirty) {
-        TextBuffer& buffer = static_cast<TextBuffer&>(_mesh);
-        buffer.transformID(m_id, m_transform.state);
         m_dirty = false;
+        size_t attribOffset = offsetof(Label::Vertex, state);
+        int numVerts = m_numGlyphs * 6;
+
+        m_mesh.updateAttribute(m_bufferOffset + attribOffset, numVerts, m_transform.state);
```
karimnaaji: I like that change, much more easier to use the same interface.

core/src/labels/label.cpp
hjanetzek: The case that is interesting here is when the value stays 0.0 or 1.0 which I would think works with this test.

This is not a case where numeric instability can blow things up, so bit equality should suffice. Wouldn't mind to change this to a test < then some epsilon (which is still hard to get right though, see e.g. http://floating-point-gui.de/errors/comparison/ ).

core/src/labels/labelMesh.h
hjanetzek: Hmm, yes, storing objects in-place would be good. I think one could use a templated version. 
The question is how to get the items as Label instances. 
Either a combination of getLabelsSize() + getLabel(i) or a forEachLabel(LabelVisitor) comes to my mind.

I'll think about it

core/src/labels/label.cpp
karimnaaji: I'm not sure values between 0.0 and 1.0 changes anything, since mostly the fract part matters. This link is interesting though. The bit comparison failed several times for me with equal values:
screen shot 2015-07-27 at 5 13 55 pm

core/src/labels/labelMesh.h
karimnaaji: Yes this could be interesting!

core/src/labels/label.cpp
hjanetzek: I mean just the test equality to 0.0 or 1.0 - while the alpha is in between it is most likely to change each frame. 
The screenshot is missing the value of m_transform.state.alpha to be meaningful :)

karimnaaji: Oh right, comparing those values makes sense.

karimnaaji:
All good!

hjanetzek:
Thanks! that was a bit too quick for me :)  I would have liked to rebase onto master.. nvm - next time

tallytalwar:
@hannesjanetzek by "rebase onto master" .. you mean to not have a "commit" on master with a message of "pr merged"?... if this then I would prefer pull requests to be merged instead of rebased.

If you mean something else then sorry for my git ignorance :).

hjanetzek:
No, the merge commit should still be there. Rebasing on master ensures that the merge commit will be empty and the branch will be clearly identifiable in the history as grouped 'work package'. 
The procedure is:

    git rebase origin/master && git checkout master && merge --no-ff feature-branch 

I'm not sure if this is the method  the eraser-map team uses, but the result will rather look like this:
https://github.com/mapzen/eraser-map/network

PR #221, opened 2015-07-27
Measuring label buffer updates

hjanetzek:
(Edit: this is not a pull request)

I've added some logging to compare the number of used label vertices with the number of vertices that are uploaded for each frame:
It shows that roughly 4 times more vertices are updated than being using in general. The screenshot shows a case with  a tilted view where the updated buffer are 16 times larger than what is actually used (5.8 vs 0.36 MB), with up to 60 meshes being updated.

http://www.enlightenment.org/ss/e-55b60fd612e225.04695096.jpg

I would propose to use one VboMesh per Text- and SpriteStyle that only hold the currently active labels. 

- this should decrease size and number of buffer uploads
- and reduce the number of vertices processed by the shader (instead of processing all label vertices and discard inactive ones).

### DIFF
--- a/core/src/data/dataSource.cpp
+++ b/core/src/data/dataSource.cpp
@@ -5,7 +5,6 @@
 #include "tile/tileID.h"
 #include "tile/tile.h"
 #include "tile/tileManager.h"
-#include "labels/labels.h"
 
 namespace Tangram {
 
--- a/core/src/labels/label.cpp
+++ b/core/src/labels/label.cpp
@@ -131,8 +131,11 @@ void Label::enterState(State _state, float _alpha) {
 }
 
 void Label::setAlpha(float _alpha) {
-    m_transform.state.alpha = CLAMP(_alpha, 0.0, 1.0);
-    m_dirty = true;
+    float alpha = CLAMP(_alpha, 0.0, 1.0);
+    if (m_transform.state.alpha != alpha) {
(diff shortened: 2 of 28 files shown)


karimnaaji:
Using one mesh per text was the original idea but there's several things too consider that we should probably benchmark. I'm not convinced that it would be faster but we could give it a try. To me, processing more vertex data that gets clipped could be faster (since buffering data, including the unused one, through dma would simply use more bandwidth, but end up triggering one buffer update event per mesh in the gl command queue, even if more vertices would get processed later on), having one mesh per texts would end up going through the pipeline more often and probably queuing up way more gl buffer update commands, and this means more sync.

karimnaaji:
To reduce the memory size I would rather consider reducing the number of labels would be discarded for sure at tile building (roads with several small lines, positions too close in world space...), but for now nothing on this has been investigated.

hjanetzek:
60 buffer uploads (for overall larger buffers) and 60 draw calls vs, say, 2 seems to me an optimization worth to investigate. 
If it is a concern that the upload might be blocked by the previous frame still being processed one could use double buffering, i.e. two meshes that will be updated alternating with each frame.

karimnaaji:
Yeah double-buffering is something we considered and would definitely be an optimization on uploads, for now there's only orphaning, which could work differently depending on devices.

PR #222, opened 2015-07-27
Better gestures

tallytalwar:
Work in progress branch for better gestures.

- Much better handling of scale/rotate and shove gestures.
- Shove gestures are exclusive
- Rotate and Scale gestures are exclusive in themselves.
- Better horizontal gesture control (solves multiple issues of gesture detection when finger pointers were horizontal).
- Shove gesture is flipped now, shoving up will tilt the view forward.
- Rotation about finger/stationary finger vs rotation about focus point.
- Pinch and Rotate is consistent about focus points in all possible views.
- Use of DisplayMetrics for THRESHOLDS.
- Handle DoubleTap and DoubleTap (motion) scaling
- Control sensitivity of 1-finger gesture post 2-finger gestures.

### DIFF
--- a/core/src/util/inputHandler.cpp
+++ b/core/src/util/inputHandler.cpp
@@ -147,7 +147,7 @@ void InputHandler::handleShoveGesture(float _distance) {
     if (!clearMomentums()) {
         m_gestures.set(GestureFlags::shove);
 
-        m_view->pitch(_distance);
+        m_view->pitch(-_distance*2.0f);
 
         onEndGesture();
     }
--- a/core/src/util/inputHandler.cpp
+++ b/core/src/util/inputHandler.cpp
@@ -11,40 +11,20 @@ InputHandler::InputHandler(std::shared_ptr<View> _view) : m_view(_view) {}
 
 void InputHandler::update(float _dt) {
 
-    bool renderRequested = false;
+    bool isFlinging = glm::length(m_deltaTranslate) > m_minDeltaLength || std::abs(m_deltaZoom) > m_minDeltaZoomLength;
 
-    if (!m_gestureOccured) {
+    if (!m_gestureOccured && isFlinging) {
 
-        if (glm::length(m_deltaTranslate) > m_minDeltaLength || std::abs(m_deltaZoom) > m_minDeltaZoomLength) {
+        // exponential decrease, m_expDecrease should be between [0..1]
+        m_deltaTranslate *= m_expDecrease;
+        m_view->translate(m_deltaTranslate.x, m_deltaTranslate.y);
 
-            static std::bitset<8> pan(1 << GestureFlags::pan);
-            static std::bitset<8> pinch(1 << GestureFlags::pinch);
+        // to give more control, make zoom decrease faster than translation
+        m_deltaZoom *= m_expDecrease * m_expDecrease;
+        m_view->zoom(m_deltaZoom);
 
-            // exponential decrease, m_expDecrease should be between [0..1]
-            if (pan == m_gestures || m_momentumHandled) {
-                m_deltaTranslate *= m_expDecrease;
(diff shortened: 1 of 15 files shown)


core/src/view/view.cpp
```diff
@@ -118,29 +118,46 @@ void View::roll(float _droll) {
...
+    if (m_tilted) {
+        screenToGroundPlane(oldViewCenterX, oldViewCenterY);
+    }
+
     updateMatrices();
 
+    // origin after tilt
+    if(m_tilted) {
+        screenToGroundPlane(viewCenterX, viewCenterY);
+        translate(oldViewCenterX - viewCenterX, oldViewCenterY - viewCenterY);
```
blair1618: calling `translate()` after `updateMatrices()` won't have any effect until the next view update, this probably compounds the translation you're adding when the view is tilted

tallytalwar: Thats not true, updateMatrices does not modify anything in m_pos.x and m_pos.y.
The reason it has to be done here is because we need to "focus point" values before and after matrices are updated. And do necessary translations.

And if translations were compounded then you will see the focus point move a lot, which does not happen and remains consistent with the center of the view.

updateTiles is what needs updated m_pos values, and this translate happens before updateTiles.

blair1618:
Does this have the changes for stopping the map mid-fling? I'm having trouble getting that to work when I play with it on my phone.

tallytalwar:
@blair1618 50492accb6a9e93e65e3e8052923f884f37bf9da has these changes. Just tried on my android and it works for me.

blair1618:
It stops if you trigger a gesture like a tap, but if you just put your finger down and leave it there it keeps right on flingin'.

tallytalwar:
Ohh there is no trigger for a "SingleTapDown" in any platform right now.

handleTap event occurs on "SingleTapUp".

blair1618:
Maybe we can treat a single touch down as a pan with zero displacement

tallytalwar:
Sure, or not do anything at all. But clear out the last fling (if it was flinging).

karimnaaji:
Yes I think an event for this would make sense to stop the fling at this moment, as soon as you start putting your finger down to the screen.

tallytalwar:
I do not like how I am handling "making focal point static" on scale and rotate.

Would like to get some inputs there. I have another solution which does not have noise/jumps on scaling or rotating, but it does not work when one of the fingers is stationary :|.

blair1618:
Mmm I've been working on this as well, I'll let you know if I find a good
solution.

blair1618:
Alright I think I've found a pretty good solution. See https://github.com/tangrams/tangram-es/tree/bettererGestures

We don't need to keep track of a 'focus' at all, we can just detect the average translation of two-finger gestures using Android's `onScroll` and treat that as a pan.

tallytalwar:
This now has implementation for handling "TapDown" touch events (android, ios, osx and linux).

These make sure InputHandler clears all deltas for a previous fling gesture.

core/src/util/inputHandler.cpp
```diff
@@ -114,15 +187,21 @@ void InputHandler::onEndGesture() {
     requestRender();
 }
 
+bool InputHandler::clearMomentums() {
+    m_deltaZoom = 0.f;
+    m_deltaTranslate = glm::vec2(0.f);
+    if(m_momentumHandled) {
+        m_momentumHandled = false;
+        return true;
```
blair1618: What's the reason for returning a value here?

blair1618:
Turns out the loss of positional accuracy with rapid zoom and rotation is caused by `screenToGroundPlane` using stale view matrices. I've tried out a fix and found it impossible to lose accuracy when the matrices are actually up-to-date. I'll make a fix for that in another PR.

PR #224, opened 2015-07-28
change shared_ptr parameter to pass by ref

hjanetzek:
- no need to copy shared_ptr unless the callee should be able to store it.
- pass view as *const* reference
- 'const std::shared_ptr\<X\>' is not that useful, it does *not* mean std::share_ptr\<const X\>


tallytalwar:
Ohh I think @karimnaaji's PR was based off this PR.

blair1618:
still merge-able

PR #225, opened 2015-07-29
Use more unique_ptr

karimnaaji:
Following the work of #224 , shared_ptr are incredibly slow, so when possible changing those to unique_ptr, it also makes ownership more clear.


karimnaaji:
yes.. especially when using auto :/

karimnaaji:
Will change the based branch to master

PR #226, opened 2015-07-29
update to latest glm

hjanetzek:
The currentl used revision produces lots of warnings when doing release builds (regarding -Wstrict-aliasing)

Edit: Ah, this warning seems to be gcc optimizer specific.

Which is your favourite warning, btw? :) 
http://fuckingclangwarnings.com/
https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html

More seriously: any preferences which warnings we should turn on by default?


blair1618:
Ah, I don't often compile with GCC or release so it makes sense that I haven't noticed these. GLM upgrade sounds good. 

`-Wvexing-parse` is a fun one haha. I didn't know about `-Wmicrosoft-exists` but I like how it implies that microsoft existing is a mistake. I tend to appreciate most flags except `-Wpedantic`, which I find to be, well, pedantic.

hjanetzek:
For gcc '-Wextra -Wno-unused-parameter' seems sensible. It currently shows a few -Wmissing-field-initializers

hjanetzek:
http://www.cmake.org/cmake/help/v3.0/module/CheckCXXCompilerFlag.html seems to be handy

I'll take my clang with -Weverything

PR #227, opened 2015-07-29
Uniforms

karimnaaji:
This is a small improvements since most of those were sent per-frame, but still useful to send them only when necessary.


core/src/style/style.h
```diff
@@ -72,6 +72,9 @@ class Style {
     /* TODO: decouple layers and styles so that sublayers can apply different styles than the parent */
     std::vector<std::shared_ptr<SceneLayer>> m_layers;
 
+    /* Whether the viewport has changed size */
+    bool m_dirtyViewport;
```
tallytalwar: Setting the default value.

core/src/style/spriteStyle.cpp
```diff
@@ -100,8 +100,17 @@ void SpriteStyle::buildPoint(Point& _point, const StyleParamMap&, Properties& _p
 void SpriteStyle::onBeginDrawFrame(const std::shared_ptr<View>& _view, const std::shared_ptr<Scene>& _scene) {
     m_spriteAtlas->bind();
 
-    m_shaderProgram->setUniformi("u_tex", 0);
-    m_shaderProgram->setUniformMatrix4f("u_proj", glm::value_ptr(_view->getOrthoViewportMatrix()));
+    static bool initUniformSampler = false;
+
+    if (initUniformSampler) {
```
tallytalwar: It will never go inside this if?

karimnaaji: :/ initialized with wrong value

tallytalwar: I also do not get the point of this. This if check is redundant?

karimnaaji: `initUniformSampler` is a local static, so it'll go there once only

tallytalwar: :+1:

PR #228, opened 2015-07-29
Use more unique_ptr

karimnaaji:
#225


core/src/style/style.h
```diff
@@ -54,7 +54,7 @@ class Style {
     std::string m_name;
 
     /* <ShaderProgram> used to draw meshes using this style */
-    std::shared_ptr<ShaderProgram> m_shaderProgram = std::make_shared<ShaderProgram>();
+    std::unique_ptr<ShaderProgram> m_shaderProgram = std::make_unique<ShaderProgram>();
```
karimnaaji: `std::make_unique` which should be part of c++14 seems to be breaking the travis build, any idea why?

tallytalwar: I think we might have to specify "libstdc++" instead of "libc++".

PR #229, opened 2015-07-29
benchmark on linux

hjanetzek:
rebased #214


blair1618:
Oh I just commented on #214 that I'm in favor of merging when the conflicts are fixed - same here, we can just merge this one since it appears to be rebased.

hjanetzek:
I'll look into - glm templates crashed gcc. probably just had a bad day (= 

https://travis-ci.org/tangrams/tangram-es/jobs/73509508#L1321

blair1618:
Breaking compilers are we? :)

It's possible that GCC hit a memory limit and was killed by Travis, that's happened with `-j` before.

hjanetzek:
@blair1618 yes, this is more likely..

hjanetzek:
success!

PR #230, opened 2015-07-29
Labels

karimnaaji:
Move `updateTransform` as a common interface for labels and consolidate it to perform those updates only on visible label states when necessary, this reduces subBufferData sizes.


hjanetzek:
Hmm our branches wont merge cleanly. Is it ok when I rebase this one onto direct-fontstash?

karimnaaji:
Yep, go ahead!

karimnaaji:
in #241

PR #232, opened 2015-07-31
use fontstash directly

hjanetzek:
Most of what is provided by glfontstash is now replicated in tangram. This patch removes the dependency on glfontstash by getting the glyph quads directly from fontstash.

- hide FONScontext in FontContext
- pass everything into TextLabel constructor


hjanetzek:
@karimnaaji, should the quad-indices buffer go into a separate pull request? Other than that this branch is ready for review imo.

karimnaaji:
No I think It's fine within the same PR!

hjanetzek:
moved to pr #241

PR #233, opened 2015-07-31
use indices for glyphs verts

karimnaaji:
use indices for glyph verts, based on #232

### DIFF
--- a/core/src/gl/typedMesh.h
+++ b/core/src/gl/typedMesh.h
@@ -15,15 +15,15 @@ public:
 
     void addVertices(std::vector<T>&& _vertices,
                      std::vector<int>&& _indices) {
-        vertices.push_back(_vertices);
-        indices.push_back(_indices);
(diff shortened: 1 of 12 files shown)


hjanetzek:
Yes, using indices was also my plan - I'm working on a patch to use one shared vbo for quad indices for all LabelMeshes

karimnaaji:
Coo! I wonder how it would work without resizing the meshes?

hjanetzek:
I would make the indices buffer as large as is reasonable to batch quads for one draw call. Then advance the vertex pointer and draw the same indices twice (or more times) when there are vertices left. This requires to override the VboMesh::draw call to not advance the indices

2^16 / 4 * 6 would be the maximum vertices to draw with one call (= 16384 quads) - I think one does not loose much when making it half of the size.

karimnaaji:
Makes sense, I think moving the vertex pointer would most likely not happen so often since it lets a pretty good quad budget.

PR #236, opened 2015-08-03
write out 'layer' and 'scene' variable names

hjanetzek:


tallytalwar:
This is all good. But I think @blair1618 is almost at the very edge of filing a PR which changes this code base.

blair1618:
Yeah this is getting blasted away very soon
*stay tuned!*

hjanetzek:
ok!

PR #237, opened 2015-08-03
Layer refactor

blair1618:
This reorganizes the process of building features in a tile: filtering them by data source, then filtering by layer filters, then finding and combining matching rules for drawing. Some new terminology and classes:

 - A data source can contain several "collections" of data (we used to call these "layers" but that term is quite overloaded now)
 - A top-level layer in the stylesheet is a `DataLayer`, which is the same as a `SceneLayer` but with the additional information of the data source and collection to which the layer should apply
 - A `SceneLayer` contains a `Filter`, a list of `DrawRule`s, and a list of `SceneLayer`s which are sublayers to the original. `SceneLayer` can match a feature and context recursively to determine the `DrawRule`s that apply to that feature.
 - A `DrawRule` contains the name of a `Style` and a list of `StyleParam`s. `DrawRule` can merge with another `DrawRule` and quickly find parameters by name. 
 - A `StyleParam` is just a string key and string value, using ":" to delimit nested objects in the stylesheet. (`StyleParam` will probably need to change in the near future or else we'll be doing lots of manual string parsing in the `Style`s.) 

Most of the code 'addition' is tests for `DrawRule` and `SceneLayer` functionality. The actual matching and filtering code is simpler and smaller, in part because this doesn't attempt any caching of draw rules after they are merged. Even so, the tile building process is about 15% faster with these changes than on the master branch. Rule matching and merging isn't the biggest bottleneck right now, but presumably those steps could be sped up even further if we added caching again.


core/src/scene/drawRule.cpp
```diff
@@ -0,0 +1,68 @@
...
+    str += "}\n";
+
+    return str;
+}
+
+bool DrawRule::findParameter(const std::string& _key, std::string* _out) const {
+
+    auto it = std::lower_bound(parameters.begin(), parameters.end(), _key, [](StyleParam p, std::string k) {
+        return p.key < k;
+    });
```
hjanetzek: Missing an ampersand here :)

It might not look as nice on the calling side, but I would prefer to return the StyleParam by const& instead of the out-copy.

hjanetzek:
Looks all good to me. I guess @tallytalwar is more aware of the details of the style logic. Tweaks and more optimization is always possible later.

I also think that we can gain quite a bit by optimizing StyleParam - a good candidate for choosing a variant implementation for values ;)

blair1618:
@hjanetzek Yes definitely. #238 has some good improvements that I'm profiling right now. I've actually considered just taking the memory representation of the YAML node that defines a DrawRule and storing a pointer to that. It would allow the most general set of parameters, but it may be an inefficient data structure to query frequently.

tallytalwar:
I will start reviewing this later in the afternoon today. And yes I agree with @blair1618, querying YAML node frequently might be inefficient. But we might be able to do better with using the mnmlistic library headers. Open for discussion.

tallytalwar:
@blair1618 Will it make sense to have benchmark tests for drawRule merging?

blair1618:
@tallytalwar that wouldn't be a bad thing to measure - getting representative data might be tricky though

tallytalwar:
Yay! Did you Valgrind on OSX or Linux?

core/src/scene/sceneLoader.cpp
```diff
@@ -200,12 +208,15 @@ MaterialTexture SceneLoader::loadMaterialTexture(YAML::Node matCompNode, Scene&
...
+    auto& tex = scene.textures()[name];
+
+    if (!tex) {
+        tex = std::make_shared<Texture>(name);
     }
 
-    matTex.tex = sharedTexs.at(name);
+    matTex.tex = tex;
+
+    if (!matTex.tex) { matTex.tex.reset(new Texture(name)); }
```
tallytalwar: I am not sure I get this :(.
Isn't this same as [Line: 214](https://github.com/tangrams/tangram-es/pull/237/files#diff-7f29a0cf96bc2f3ea18f48b14ef40906R214)?

214: makes a shared ptr of a texture (if it does not exist already).
Line 219 also does the same. Shouldn't Line:217 suffice?

tallytalwar:
Not wrt this PR, but we should have a trello card created for all the TODOs marked in sceneLoader/yaml parsing.

core/src/scene/sceneLoader.cpp
```diff
@@ -364,7 +375,7 @@ void SceneLoader::loadStyles(YAML::Node styles, Scene& scene) {
         Node urlNode = styleNode["url"];
         if (urlNode) { logMsg("WARNING: loading style from URL not yet implemented\n"); } // TODO
 
-        scene.addStyle(std::unique_ptr<Style>(style));
+        scene.styles().push_back(std::unique_ptr<Style>(style));
```
tallytalwar: Should a check be happening to see if a style with the same name already exists?

core/src/scene/sceneLoader.cpp
```diff
@@ -37,10 +40,15 @@ void SceneLoader::loadScene(const std::string& _file, Scene& _scene, TileManager
...
     loadLights(config["lights"], _scene);
 
-    for (auto& style : _scene.getStyles()) {
-        style->build(_scene.getLights());
+    for (auto& style : _scene.styles()) {
+        style->build(_scene.lights());
     }
 
+    // Styles that are opaque must be ordered first in the scene so that they are rendered 'under' styles that require blending
+    std::sort(_scene.styles().begin(), _scene.styles().end(), [](std::unique_ptr<Style>& a, std::unique_ptr<Style>& b) {
```
tallytalwar: There is also [this](http://en.cppreference.com/w/cpp/experimental/parallelism/existing#sort).. But "experimental" :|

core/src/scene/sceneLoader.cpp
```diff
@@ -410,7 +421,7 @@ void SceneLoader::loadLights(Node lights, Scene& scene) {
         // Add an ambient light if nothing else is specified
         std::unique_ptr<AmbientLight> amb(new AmbientLight("defaultLight"));
         amb->setAmbientColor({ .5f, .5f, .5f, 1.f });
-        scene.addLight(std::move(amb));
+        scene.lights().push_back(std::move(amb));
```
tallytalwar: No check for existing light names? (No use of "findLight" method).

core/src/scene/dataLayer.h
```diff
@@ -0,0 +1,26 @@
+#pragma once
+
+#include "scene/sceneLayer.h"
+#include <string>
+
+namespace Tangram {
+
+// DataLayer represents a top-level layer in the stylesheet, distinct from
+// SceneLayer by its association with a collection within a DataSource
```
tallytalwar: Collection is basically a "feature" collection right?

core/src/scene/dataLayer.h
```diff
@@ -0,0 +1,26 @@
...
+class DataLayer {
```
tallytalwar: Will it be better to make DataLayer a specialized class of SceneLayer by making it the child of SceneLayer?

Currently its implemented using "containership" OOPs concept. 
http://www.differencebetween.com/difference-between-inheritance-and-vs-containership/

core/src/scene/sceneLoader.cpp
blair1618: Interesting! But I'm pretty confident we don't need parallelism to sort a list that's typically under 10 items :)

core/src/scene/sceneLoader.cpp
blair1618: Yeah looks redundant, I'll remove it and see

core/src/scene/sceneLoader.cpp
tallytalwar: :+1:  I agree. But I never knew about "parallel sort in cpp" so wanted to mention :D.

core/src/style/style.h
```diff
@@ -131,8 +108,13 @@ class Style {
     /* Add layers to which this style will apply */
     void addLayer(std::shared_ptr<SceneLayer> _layer);
```
tallytalwar: Style::addLayer should be redundant now. Also Style::m_layers.

Styles basically now have no ownership of sceneLayer/dataLayer.

core/src/scene/sceneLoader.cpp
blair1618: A style with the same name would only be added if: 1. the style used the name of a built-in style, which we check for explicitly above, or 2. if two yaml nodes had the same name, which is invalid yaml and therefore won't ever be parsed.

core/src/scene/sceneLoader.cpp
blair1618: the yaml spec takes care of the uniqueness check for us

core/src/scene/dataLayer.h
blair1618: yup, that's the idea!

core/src/scene/sceneLoader.cpp
tallytalwar: Cool, I did not know its invalid yaml :D.

core/src/scene/dataLayer.h
blair1618: Right, it's a little funny but this made more sense to me. If `DataLayer` inherited from `SceneLayer` then the class hierarchy is saying that it would be valid for a `DataLayer` to contain other `DataLayer`s, and that's not the case.

core/src/scene/sceneLoader.cpp

core/src/style/style.h
blair1618: Oh weird, I thought I removed that already, thanks for the catch.

core/src/scene/dataLayer.h
tallytalwar: Hmm.. I am not sure I get this.

How in case of inheritance "DataLayer" will container other "DataLayer"s?

DataLayer being a specialized SceneLayer will still have vector of SceneLayers (subLayers). 
To me inheritance makes more sense.

blair1618: Forget my last comment, my head was in Java land. Does inheritance add
anything here? (Aside from more virtual dispatch at runtime ;) )

tallytalwar: I think it will simplify many things... like how DataLayer->SceneLayers are constructed when yaml is parsed. It will also make access like [this](https://github.com/tangrams/tangram-es/pull/237/files#diff-25abeab428d409be23fa79dd35450968R60), more direct.

And logic wise it seems to be more of a inheritance model rather than a containership model.

(Also my battery running low now :|, and charging slot is occupied by my co-passenger :().

core/src/scene/sceneLoader.cpp
```diff
@@ -733,58 +744,67 @@ Filter SceneLoader::generateNoneFilter(YAML::Node _filter) {
...
-            Node filter = subLayerItr.second["filter"];
-            Filter subLayerFilter = generateFilter(filter);
+        const std::string key = member.first.as<std::string>();
 
-            loadSublayers(layer[subLayerName], ssubLayers);
+        if (key == "data") {
+            // Ignored for sublayers
+        } else if (key == "draw") {
+            // Member is a mapping of draw rules
+            for (auto& ruleNode : member.second) {
```
tallytalwar: Correct me if I am wrong here.
According to what I get from the drawGroup [documentation](https://github.com/tangrams/tangram-docs/blob/gh-pages/pages/draw.md#draw), there can only be one drawGroup per layer/subLayer and rules defined in one layer are passed down sublayers.

Can you give me an example of how size of "rules" here can ever be more than one?

blair1618: a draw "group" is a collection of multiple draw rules

tallytalwar: Hmm, something like this is legal:

```
draw:
        lines:
                order: blah
                color: blah
                ...
        polygons:
                order: blah
                .... 
```

blair1618: Indeed!

core/src/scene/sceneLayer.cpp
```diff
@@ -1,14 +1,61 @@
...
+    m_depth++;
+
+    // Sublayers must be sorted by the depth of their deepest leaf in order to correctly traverse them in match()
+    std::sort(m_sublayers.begin(), m_sublayers.end(), [&](const SceneLayer& a, const SceneLayer& b) {
+        return a.m_depth > b.m_depth;
+    });
+
+}
+
+void SceneLayer::match(const Feature& _feat, const Context& _ctx, std::vector<DrawRule>& _matches) const {
```
tallytalwar: Why not return the vector of DrawRules here, instead of pass by reference and swapping?
(I am thinking in reference to let compiler do its job, based on the explanation given by the llvm guy :D).

core/src/scene/sceneLayer.cpp
```diff
@@ -1,14 +1,61 @@
...
 
-SceneLayer::SceneLayer(const std::vector<std::shared_ptr<SceneLayer>> _subLayers, const StyleParamMap&& _styleParamMap, const std::string _name, Filter _filter) :
-    m_subLayers(std::move(_subLayers)), m_styleParams(std::move(_styleParamMap)), m_name(_name), m_filter(_filter) {
+    std::vector<DrawRule> merged;
 
-    m_id = s_layerCount++;
+    // For all of m_rules that have the same style as an existing matched rule, merge the rules;
+    // for others, take existing rule unchanged
+    {
+        auto pIt = m_rules.begin(), pEnd = m_rules.end();
```
tallytalwar: Can you change the Iterator names to something more explicit?

PR #239, opened 2015-08-04
LineBuilder: reserve space for indices and vertices

hjanetzek:


blair1618:
Looks good to me!

We can also adjust the number of triangles for caps an joins based on angle and maybe screen coverage. In that case this reserved space will be an over-estimate, but it's still better than re-allocating with the default growth factor.

bcamper:
@blair1618: @patriciogonzalezvivo has a branch in Tangram JS for adapting line caps/joins complexity, but we haven't merged it yet. When he's back from vacation let's get it integrated into both...

PR #241, opened 2015-08-06
labels refactor + direct-fontstash + shared quad-indices

hjanetzek:
combined #230 and #232

### DIFF
--- a/core/src/gl/typedMesh.h
+++ b/core/src/gl/typedMesh.h
@@ -15,15 +15,15 @@ public:
 
     void addVertices(std::vector<T>&& _vertices,
                      std::vector<int>&& _indices) {
-        vertices.push_back(_vertices);
-        indices.push_back(_indices);
+        m_vertices.push_back(_vertices);
+        m_indices.push_back(_indices);
 
         m_nVertices += _vertices.size();
         m_nIndices += _indices.size();
     }
 
     virtual void compileVertexBuffer() override {
-        compile(vertices, indices);
+        compile(m_vertices, m_indices);
     }
 
     /*
@@ -66,8 +66,8 @@ protected:
 
     void setDirty(GLintptr _byteOffset, GLsizei _byteSize);
 
-    std::vector<std::vector<T>> vertices;
-    std::vector<std::vector<int>> indices;
+    std::vector<std::vector<T>> m_vertices;
+    std::vector<std::vector<int>> m_indices;
 
 };
 
--- a/core/src/gl/typedMesh.h
+++ b/core/src/gl/typedMesh.h
@@ -2,6 +2,7 @@
 
 #include "vboMesh.h"
 #include <cstdlib> // std::abs
+#include <array>
 
 namespace Tangram {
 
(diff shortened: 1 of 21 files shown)


core/src/labels/label.cpp
```diff
@@ -1,17 +1,23 @@
 #include "label.h"
 
 #include "util/geom.h"
+#include "labels/labelMesh.h"
+#include <cstdint>
```
karimnaaji: I' don't see any usage of this include.

core/src/gl/typedMesh.h
```diff
@@ -37,37 +38,45 @@ class TypedMesh : public VboMesh {
      * after _byteOffset in the mesh vertex data memory
      */
     template<class A>
-    void updateAttribute(GLintptr _byteOffset, unsigned int _nVerts, const A& _newAttributeValue) {
+    void updateAttribute(std::array<int,2> _vertexRange, const A& _newAttributeValue, size_t _attribOffset = 0) {
```
karimnaaji: I may be missing something but any reason of using `std::array<int,2>` rather than `int _vertexRange[2]`? We could probably save an include.

hjanetzek: It's just c++11 version of the c-array + additional API that we don't need here. So I wouldn't mind to change this to standard c arrays.  I thought to rather add an IntRange type in a utils header to make the intended use more clear - but really a plain 2 element array should be ok.

core/src/labels/label.cpp
hjanetzek: yes, it's probably a remainder from a previous approach

core/src/labels/labelMesh.cpp
```diff
@@ -13,4 +19,102 @@ void LabelMesh::addLabel(std::unique_ptr<Label> _label) {
...
+    if (s_quadGeneration == s_validGeneration) {
+        glBindBuffer(GL_ELEMENT_ARRAY_BUFFER, s_quadIndexBuffer);
+
+    } else {
+        s_quadGeneration = s_validGeneration;
+
+        std::vector<GLushort> indices;
+        indices.reserve(maxVertices / 4 * 6);
+
+        for (size_t i = 0; i < maxVertices; i += 4) {
```
karimnaaji: This seems like stateless work we can do at initialization rather than in the render loop.

karimnaaji:
Looking good to me, and glad this works! The only thing I noticed is that we have redundancy in binding the element array buffer, we can probably consolidate on that.

core/src/labels/labelMesh.cpp
```diff
@@ -1,8 +1,14 @@
 #include "labels/labelMesh.h"
 #include "labels/label.h"
+#include "shaderProgram.h"
 
 namespace Tangram {
 
+GLuint s_quadIndexBuffer = 0;
+int s_quadGeneration = -1;
```
karimnaaji: Little nitpick but can we declare those as static so the use `s_*` would make sense here.

core/src/labels/labelMesh.cpp
```diff
@@ -1,8 +1,14 @@
...
+
+const size_t maxVertices = 1 << 14;
```
karimnaaji: Also it seems more clear to me to directly expose `16384` as the max vertices value here.

core/src/labels/labelMesh.cpp
hjanetzek: I assume that compared to all the other things done on the first frame, e.g. compiling the shaders, the impact of this initialization is negligible.

I think one could add async startup tasks that will be processed in parallel before the first frame. Or do you have a suggestion how this should be done?

blair1618:
Can we use a vertex offset instead of a byte offset for TypedMesh::updateVertices? (https://github.com/tangrams/tangram-es/pull/241/files#diff-c238f2dec5420e8749b0b85cd2c56803R34) I feel like I'm likely to use the wrong count somewhere if functions count vertices differently.

core/src/labels/labelMesh.cpp
karimnaaji: This one happens when the first tile is ready for drawing where shaders are usually compiled when tiles aren't ready for drawing.

karimnaaji: I don't see how this could be nicely done, moving this to the constructor would introduce concurrency, maybe statically initialize it? I just noticed that it can take up to 1ms + the buffer upload on this part that's why I was worried about it, but since it's happening once it shouldn't be noticed.

hjanetzek:
@blair1618, right. I'll make this consistent

core/src/gl/typedMesh.h
hjanetzek: Actually the std::array api turned out to be quite handy while trying to change the  to c-arrays. In particular aggregate initialization and copy construction. 

I've now changed it to use a minimal Range type.

PR #242, opened 2015-08-06
WIP Tile Cache

hjanetzek:
This branch replaces the unlimited cache for parsed tile data with an LRU memory cache for raw tile data. The raw data of the most recent tiles is also kept in memory for fast recreation of tiles on context loss.

The new LRU TileCache stores tiles that are ready for rendering. It is limited by maximum tiles and buffer usage. 
The cached tiles are also used for missing proxies. 

To reduce the memory usage of cached tiles the byte buffers for static VboMesh are discarded now.

TODO: 
- cache size config. Where should this go? I would add Tangram::setCacheLimits(raw MB, gl MB) so that it can be set by the backends

### DIFF
--- a/core/src/data/dataSource.h
+++ b/core/src/data/dataSource.h
@@ -5,6 +5,7 @@
 #include <memory>
 #include <vector>
 #include <mutex>
+#include <list>
 #include "tile/tileTask.h"
 
 namespace Tangram {
@@ -35,29 +36,23 @@ public:
      */
     virtual bool loadTileData(std::shared_ptr<TileTask>&& _task, TileTaskCb _cb);
 
-    virtual bool getTileData(std::shared_ptr<TileTask>&& _task, TileTaskCb _cb);
+    /* Lookup TileData in cache and */
+    virtual bool getTileData(std::shared_ptr<TileTask>& _task);
 
     /* Stops any running I/O tasks pertaining to @_tile */
     virtual void cancelLoadingTile(const TileID& _tile);
 
-    /* Checks if data exists for a specific <TileID> */
-    virtual bool hasTileData(const TileID& _tileID) const;
-
-    /* Returns the data corresponding to a <TileID>, if it has been fetched already */
-    virtual std::shared_ptr<TileData> getTileData(const TileID& _tileID) const;
-    
     /* Parse an I/O response into a <TileData>, returning an empty TileData on failure */
     virtual std::shared_ptr<TileData> parse(const Tile& _tile, std::vector<char>& _rawData) const = 0;
 
-    /* Stores tileData in m_tileStore */
-    virtual void setTileData(const TileID& _tileID, const std::shared_ptr<TileData>& _tileData);
-    
     /* Clears all data associated with this DataSource */
     void clearData();
 
     const std::string& name() const { return m_name; }
 
 protected:
+
+    void onTileLoaded(std::vector<char>&& _rawData, std::shared_ptr<TileTask>& _task, TileTaskCb _cb);
 
     /* Constructs the URL of a tile using <m_urlTemplate> */
     virtual void constructURL(const TileID& _tileCoord, std::string& _url) const;
@@ -68,13 +63,24 @@ protected:
         return url;
     }
 
-    std::map< TileID, std::shared_ptr<TileData> > m_tileStore; // Map of tileIDs to data for that tile
-    
-    std::string m_name; // Name used to identify this source in the style sheet
+    // Name used to identify this source in the style sheet
+    std::string m_name;
 
-    std::mutex m_mutex; // Used to ensure safe access from async loading threads
+    // URL template for requesting tiles from a network or filesystem
+    std::string m_urlTemplate;
 
-    std::string m_urlTemplate; // URL template for requesting tiles from a network or filesystem
+    // Used to ensure safe access from async loading threads
+    std::mutex m_mutex;
+
+    // LRU in-memory cache for raw tile data
+    using CacheEntry = std::pair<TileID, std::shared_ptr<std::vector<char>>>;
+    using CacheList = std::list<CacheEntry>;
+    using CacheMap = std::unordered_map<TileID, typename CacheList::iterator>;
+
+    CacheMap m_cacheMap;
+    CacheList m_cacheList;
+    size_t m_cacheUsage;
+    size_t m_cacheMaxUsage;
 
 };
 
--- a/core/src/data/dataSource.h
+++ b/core/src/data/dataSource.h
@@ -84,8 +84,8 @@ protected:
 
     CacheMap m_cacheMap;
     CacheList m_cacheList;
-    size_t m_cacheUsage;
-    size_t m_cacheMaxUsage;
+    int m_cacheUsage;
+    int m_cacheMaxUsage;
 
 };
 
--- a/core/src/data/dataSource.h
+++ b/core/src/data/dataSource.h
@@ -50,7 +50,7 @@ public:
 
     const std::string& name() const { return m_name; }
 
-    /* @_cacheSize: Set size of in-memory cache for tile data in bytes. */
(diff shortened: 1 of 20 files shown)


hjanetzek:
moved the context-recreation workaround to a separate branch

hjanetzek:
I'll add the size config tomorrow and I'm thinking about moving the TileCache into TileManager as it's not that big and pretty much an internal of the TileManager.

blair1618:
@hjanetzek the question of cache configuration is a good one. I agree that adding a simple configuration function to the c++ api is the place to start. I'm unclear whether configuration should separate the limits for CPU memory and GPU memory. For most applications (i.e. on Android and iOS) the memory is a shared resource anyway, so my view would be that we should just expose a single configurable value, e.g. `Tangram::setCacheLimit(total_mb)`.

hjanetzek:
@blair1618, having one memory limit poses the harder question which fraction should be used by each cache :)

blair1618:
@hjanetzek Yeah, I think if I were a developer who didn't know the internals of tangram I wouldn't know how to choose a limit for each cache. 

Yes, it makes sense to me that raw and gl data would both contribute toward some common memory limit.

hjanetzek:
@blair1618 TileManager and DataSource have now their own setCacheSize functions. I would say we should leave it at that for now. If someone needs to change the defaults they can figure it out I guess.

blair1618:
@hjanetzek I think I figured out that segfault I've been encountering: in `TileCache`, `m_cacheUsage` is never initialized, so sometimes it takes on a very large value and the cache tries to erase more tiles than it contains. 

Seems to be fine if we initialize `m_cacheUsage` to zero here:
https://github.com/tangrams/tangram-es/blob/wip-tile-cache/core/src/tile/tileCache.h#L69

But it would also be more robust to check the size of `m_cacheList` in this loop as well:
https://github.com/tangrams/tangram-es/blob/wip-tile-cache/core/src/tile/tileCache.h#L44

core/src/data/dataSource.h
```diff
@@ -68,13 +68,24 @@ class DataSource {
...
+    std::string m_urlTemplate;
+
+    // Used to ensure safe access from async loading threads
+    std::mutex m_mutex;
 
-    std::mutex m_mutex; // Used to ensure safe access from async loading threads
+    // LRU in-memory cache for raw tile data
+    using CacheEntry = std::pair<TileID, std::shared_ptr<std::vector<char>>>;
+    using CacheList = std::list<CacheEntry>;
+    using CacheMap = std::unordered_map<TileID, typename CacheList::iterator>;
```
tallytalwar: Instead of using 2 data containers, we can do away with using a single std::vector of pairs.

Why I suggest this:
- single container to hold our data, instead of 2 containers (and potential issues with using iterator as a MapValue).
- faster iteration on vectors over list
- for accessing we can do linear search first or binary search depending on how many tiles we have to search for in the 32MB cached memory.

hjanetzek:
@blair1618 ah right, thanks. That uninitialized value went in with the change yesterday evening.. 

@tallytalwar I found this approach in two other implementations for LRU caches. The put and get operations are pretty much O(1) + quite efficient removal of items when limiting the cache. 
http://stackoverflow.com/questions/2057424/lru-implementation-in-production-code

tallytalwar:
@hjanetzek I am just concerned about using iterator as map value. And almost close to double memory footprint!! We are not iterating over the map anywhere right?

hjanetzek:
The additional memory of the structure is rather negligible I would assume - It gets up to 200 items (currently with 32MB)

For the map the usage is probably (or maybe slightly more): 

    256 * (key:[sizeof tileid] + iterator:[3 * sizeof pointer]) = 3kb + 6kb (with 64bit addresses)

and similar for the list. I just mean it's not the range where memory consumption should be a concern.

hjanetzek:
@tallytalwar yea, the iterator usage is not something I would have came up with myself :)

tallytalwar:
yes memory should not be a concern, and we are not iterating on the unordered_map so should be good.

tallytalwar:
@hjanetzek yes the iterator usage haunts me more than anything else.

hjanetzek:
http://stackoverflow.com/questions/3329956/do-stl-iterators-guarantee-validity-after-collection-was-changed <- list iterators are special in regards to being hard to invalidate. Need to check the internal again myself..

hjanetzek:
so the std::list iterator must actually be only a single pointer to the list entry... erm, sounds reasonable of course :)

tallytalwar:
:+1: Yes I am aware of std::list iterators being hard to invalidate :D.

But that being said, I would personally never try to do this :P.

core/src/tile/tile.cpp
```diff
@@ -112,4 +112,14 @@ std::unique_ptr<VboMesh>& Tile::getMesh(const Style& _style) {
     return m_geometry[_style.getName()];
 }
 
+size_t Tile::getMemoryUsage() const {
+    size_t sum = 0;
+
+    for (auto& entry : m_geometry) {
+        if (entry.second)
```
tallytalwar: Please conform to code styling and use {} here.

core/src/tile/tileManager.cpp
```diff
@@ -173,8 +188,8 @@ void TileManager::updateTileSet() {
 
             } else if (curTileId == NOT_A_TILE || visTileId < curTileId) {
                 // tileSet is missing an element present in visibleTiles
-                addTile(visTileId);
-                enqueueLoadTask(visTileId, viewCenter);
+                if (!addTile(visTileId))
+                    enqueueLoadTask(visTileId, viewCenter);
```
tallytalwar: Conform to coding style by using {}

karimnaaji:
I may be missing something since I didn't follow everything, but something I have in mind, we are caching gl data as raw buffer data on ram, but about the driver data, are we doing anything on this? As @blair1618 mentioned most devices are SMA so keeping this data in the gpu and skip draw call on those could probably make sense, what do you think @hjanetzek ?

core/src/tile/tile.cpp
hjanetzek: fixed

core/src/tile/tileManager.h
```diff
@@ -26,6 +27,10 @@ class View;
  */
 class TileManager {
 
+    const static size_t MAX_WORKERS = 2;
+    const static size_t MAX_DOWNLOADS = 4;
+    const static size_t DEFAULT_CACHE_SIZE = 32*1024*1024; // 32 MB
```
tallytalwar: Right now our lower cache is bigger than the higher cache... usually its the opposite.

And if we also have gpu cache (as @karimnaaji) mentioned in another comment, we should have:
gpu_cache < tile_cache < dataSource_cache

core/src/tile/tileManager.cpp
hjanetzek: yea.. we got some more suspects

    grep -E -r --include \*.h --include \*.cpp "if ?\(" | grep -v {

core/src/tile/tileManager.h
hjanetzek: Yes, it's actually like that. In number of tiles :)

hjanetzek: The tiles in tile_cache also keep their gl data though. 
So it's rather: visible_tiles < cached_tiles < raw_tiles

hjanetzek: @karimnaaji I'm not quite sure if I understood - but yes, the tiles in the cache are not updated or drawn as long as they are not visible

karimnaaji: I misunderstood a part of the process, stepping through with debug answered my question!

tallytalwar:
I am good with this PR. @hjanetzek can you remove a bunch a DBG statements in there? Once done will merge it.

PR #245, opened 2015-08-18
Check curl response code

hjanetzek:


tallytalwar:
This is good. Can you also replicate the same on rpi, please?

Thanks

hjanetzek:
@tallytalwar rpi uses the files from linux https://github.com/tangrams/tangram-es/blob/master/toolchains/raspberrypi.cmake#L30

tallytalwar:
Ohh yes I forgot about that.

PR #247, opened 2015-08-19
Use variant types

hjanetzek:
Introducing variant for Property::Value and StyleParam::Value

- StyleParams, especially Color values are now only parsed once 
- less allocations: The pbf parser uses a single vector for values (instead of parallel ones for float and string)
- likewise Properties uses a single map now

### DIFF
--- a/core/src/util/pbfParser.cpp
+++ b/core/src/util/pbfParser.cpp
@@ -236,6 +236,7 @@ void PbfParser::extractLayer(ParserContext& _ctx, protobuf::message& _layerIn, L
         }
     }
 
+    _out.features.reserve(_ctx.featureMsgs.size());
     for(auto& featureMsg : _ctx.featureMsgs) {
         _out.features.emplace_back();
         extractFeature(_ctx, featureMsg, _out.features.back());
--- a/core/src/util/pbfParser.cpp
+++ b/core/src/util/pbfParser.cpp
@@ -7,12 +7,12 @@
 
 namespace Tangram {
 
-void PbfParser::extractGeometry(protobuf::message& _geomIn, int _tileExtent, std::vector<Line>& _out) {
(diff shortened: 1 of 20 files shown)


hjanetzek:
ready for review - the pbfParser commits could go into a separate branch though..

blair1618:
Reviewing! I'm already liking the way this improves `TileData`

core/src/util/pbfParser.cpp
```diff
@@ -3,22 +3,20 @@
 #include "tile/tile.h"
 #include "platform.h"
 
-#include <cmath> // for isnan
-
 namespace Tangram {
 
-void PbfParser::extractGeometry(protobuf::message& _geomIn, int _tileExtent, std::vector<Line>& _out, const Tile& _tile) {
+void PbfParser::extractGeometry(ParserContext& ctx, protobuf::message& _geomIn) {
```
tallytalwar: Sorry, for code styling and consistency using "_" with function arguments.

tallytalwar:
We can also do variants for Values in filters.

hjanetzek:
@tallytalwar yes, I think it should also work for filter Value with some special handling for when a number should be interpreted as string or the other way round

tallytalwar:
Yup, we were discussing the same on this end of the planet. :D

tallytalwar:
So YAML has this failsafe type support: http://www.yaml.org/spec/1.2/spec.html#schema/failsafe/.

So the edge case of distinguishing something like a "007" and 007 yaml value can be resolved at users' end (we will have to document this in our style sheet documentation).

Considering what I mentioned above, we can go with having variants for Filter values, and by default something "007" yaml value will be treated as a numeric value, and if a user wants to use it as a string, can explicitly use "yaml failsafe" way of specifying it, i.e. "!!str 007". :D

nvkelso:
That is weird and awesome. Go YAML!

PR #248, opened 2015-08-19
WIP - Support for more than one DataSource

hjanetzek:
The TileManager handles separate TileSets for each DataSource and distributes TileCache and Download resources.

(This is early work in progress - I know that the code needs some clean up)


blair1618:
This works correctly as far as I've been able to test. I think we can do a few things to simplify TileManager but for the sake of enabling client-provided data sources for the eraser map team to start using, I'm in favor of merging this now and revising later.

hjanetzek:
If there's no hurry I would like to remove the use of DataSource* in TileCache

PR #254, opened 2015-08-21
Primitives

karimnaaji:
Better debug primitives handling by using the new render states tracking, and add of colors.


blair1618:
Super handy. Mergin!

PR #255, opened 2015-08-21
Context loss fixes

karimnaaji:
Some context loss fixes:
- Non-tiled assets should be able to keep the vbo memory data for reupload since it's not cached (skybox or future others)
- Tracking the vbo/index buffer bound should be reinitialized with invalid data after context loss
- Notify styles for context loss for uniform reinitialization.


core/src/gl/renderState.h
```diff
@@ -63,7 +63,7 @@ namespace RenderState {
                 call(typename gens<sizeof...(Args)>::type());
             }
         }
-
+
```
tallytalwar: :D. Whitespace :D.

PR #256, opened 2015-08-24
Text style refactor

tallytalwar:
- Specifying font files in the style sheet to be used in the style sheet
- TextStyle respects styling rules now.
- Loading of font file from android systems path. (Future TODO for iOS, OSX, Linux).
- color attribute for labels (to be parsed from styling rule). (thanks to @karimnaaji)
- colored sprites (thanks to @karimnaaji)
- add `captilized` option
- add `visible` option
- font units specified in `px`, `pt`, `em`, `%`, default to `px`

Todos:

- [x] add `font:style`, defaulting to `normal`, supported is `italic`
- [x] rename `font:name` to `font:family`
- [x] merge [font-units](https://github.com/tangrams/tangram-es/tree/font-units) to have `pt`, `em`, `px` conversions for font sizes
- [x] parse android fonts.xml for font filename

### DIFF
--- a/android/tangram/jni/platform_android.cpp
+++ b/android/tangram/jni/platform_android.cpp
@@ -26,6 +26,7 @@ static jmethodID requestRenderMethodID;
 static jmethodID setRenderModeMethodID;
 static jmethodID startUrlRequestMID;
 static jmethodID cancelUrlRequestMID;
+static jmethodID getFontFilePath;
 static AAssetManager* assetManager;
 
 static bool s_isContinuousRendering = false;
@@ -38,6 +39,7 @@ void setupJniEnv(JNIEnv* _jniEnv, jobject _tangramInstance, jobject _assetManage
     jclass tangramClass = jniEnv->FindClass("com/mapzen/tangram/MapController");
     startUrlRequestMID = jniEnv->GetMethodID(tangramClass, "startUrlRequest", "(Ljava/lang/String;J)Z");
     cancelUrlRequestMID = jniEnv->GetMethodID(tangramClass, "cancelUrlRequest", "(Ljava/lang/String;)V");
+    getFontFilePath = jniEnv->GetMethodID(tangramClass, "getFontFilePath", "(Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;)Ljava/lang/String;");
 	requestRenderMethodID = _jniEnv->GetMethodID(tangramClass, "requestRender", "()V");
     setRenderModeMethodID = _jniEnv->GetMethodID(tangramClass, "setRenderMode", "(I)V");
 
@@ -73,8 +75,17 @@ void requestRender() {
     }
 }
 
-std::string systemFontPath(const std::string& _name, const std::string& _weight, const std::string& _face) {
-    return "/system/fonts/" + _name + "-" + _weight + _face + ".ttf";
+std::string systemFontPath(const std::string& _family, const std::string& _weight, const std::string& _style) {
+
+    jstring jfamily = jniEnv->NewStringUTF(_family.c_str());
+    jstring jweight = jniEnv->NewStringUTF(_weight.c_str());
+    jstring jstyle = jniEnv->NewStringUTF(_style.c_str());
+    jstring returnStr = (jstring) jniEnv->CallObjectMethod(tangramInstance, getFontFilePath, jfamily, jweight, jstyle);
+
+    size_t length = jniEnv->GetStringUTFLength(returnStr);
+    char* cdata = (char*) malloc(sizeof(char) * (length));
+    jniEnv->GetStringUTFRegion(returnStr, 0, length, cdata);
+    return std::string(cdata, length);
 }
 
 void setContinuousRendering(bool _isContinuous) {
--- a/android/tangram/jni/platform_android.cpp
+++ b/android/tangram/jni/platform_android.cpp
@@ -73,57 +73,8 @@ void requestRender() {
     }
 }
 
-std::string deviceFontsPath() {
-    return "/system/fonts/";
-}
-
-std::string constructFontFilename(const std::string& _name, const std::string& _weight, const std::string& _face) {
-    std::string fontFilename;
-    fontFilename = _name + "-" + _weight + _face;
-    return fontFilename;
-}
-
-bool parseTypeFaceFontsInfo(const std::string& _typefaceInput, std::string& _fontName, float& _fontSize) {
-
-    char str[4][40];
-    float size;
-    int num = sscanf(_typefaceInput.c_str(), "%s %s %s %s", str[0], str[1], str[2], str[3]);
-    switch(num) {
-        case 1:
-            _fontName = str[0];
-            break;
-        case 2:
-            try {
-                size = std::stof(std::string(str[0]));
-                _fontSize = size;
-                _fontName = str[1];
-            } catch (const std::invalid_argument& e) {
-                _fontName = std::string(str[1]) + "-" + std::string(str[0]);
-            }
-            break;
-        case 3:
-            try {
-                size = std::stof(std::string(str[1]));
-                _fontSize = size;
-                _fontName = std::string(str[2]) + "-" + std::string(str[0]);
-            } catch (const std::invalid_argument& e) {
-                _fontName = std::string(str[2]) + "-" + std::string(str[1]) + std::string(str[0]);
-            }
-            break;
-        case 4:
-            try {
-                size = std::stof(std::string(str[2]));
-                _fontSize = size;
-                _fontName = std::string(str[3]) + "-" + std::string(str[1]) + std::string(str[0]);
-            } catch (const std::invalid_argument& e) {
-                _fontName = std::string(str[3]) + "-" + std::string(str[2]) + std::string(str[1]) + std::string(str[0]);
-            }
-            break;
-        case 0:
-        default:
-            return false;
-    }
-    return true;
+std::string systemFontPath(const std::string& _name, const std::string& _weight, const std::string& _face) {
+    return "/system/fonts/" + _name + "-" + _weight + _face + ".ttf";
 }
 
 void setContinuousRendering(bool _isContinuous) {
@@ -180,7 +131,7 @@ std::string stringFromResource(const char* _path) {
 
 }
 
-unsigned char* bytesFromExtMemory(const char* _path, unsigned int* _size) {
+unsigned char* bytesFromFileSystem(const char* _path, unsigned int* _size) {
 
     std::ifstream resource(_path, std::ifstream::ate | std::ifstream::binary);
 
--- a/android/tangram/jni/platform_android.cpp
+++ b/android/tangram/jni/platform_android.cpp
@@ -77,6 +77,12 @@ std::string deviceFontsPath() {
     return "/system/fonts/";
 }
 
+std::string constructFontFilename(const std::string& _name, const std::string& _weight, const std::string& _face) {
+    std::string fontFilename;
+    fontFilename = _name + "-" + _weight + _face;
+    return fontFilename;
+}
+
 bool parseTypeFaceFontsInfo(const std::string& _typefaceInput, std::string& _fontName, float& _fontSize) {
 
     char str[4][40];
--- a/android/tangram/jni/platform_android.cpp
+++ b/android/tangram/jni/platform_android.cpp
@@ -77,6 +77,49 @@ std::string deviceFontsPath() {
     return "/system/fonts/";
 }
 
+bool parseTypeFaceFontsInfo(const std::string& _typefaceInput, std::string& _fontName, float& _fontSize) {
+
+    char str[4][40];
+    float size;
+    int num = sscanf(_typefaceInput.c_str(), "%s %s %s %s", str[0], str[1], str[2], str[3]);
+    switch(num) {
+        case 1:
+            _fontName = str[0];
+            break;
+        case 2:
+            try {
+                size = std::stof(std::string(str[0]));
+                _fontSize = size;
+                _fontName = str[1];
+            } catch (const std::invalid_argument& e) {
+                _fontName = std::string(str[1]) + "-" + std::string(str[0]);
+            }
+            break;
+        case 3:
+            try {
+                size = std::stof(std::string(str[1]));
+                _fontSize = size;
+                _fontName = std::string(str[2]) + "-" + std::string(str[0]);
+            } catch (const std::invalid_argument& e) {
+                _fontName = std::string(str[2]) + "-" + std::string(str[1]) + std::string(str[0]);
+            }
+            break;
+        case 4:
+            try {
+                size = std::stof(std::string(str[2]));
+                _fontSize = size;
+                _fontName = std::string(str[3]) + "-" + std::string(str[1]) + std::string(str[0]);
+            } catch (const std::invalid_argument& e) {
+                _fontName = std::string(str[3]) + "-" + std::string(str[2]) + std::string(str[1]) + std::string(str[0]);
+            }
+            break;
+        case 0:
+        default:
+            return false;
+    }
+    return true;
+}
+
 void setContinuousRendering(bool _isContinuous) {
 
     s_isContinuousRendering = _isContinuous;
--- a/android/tangram/jni/platform_android.cpp
+++ b/android/tangram/jni/platform_android.cpp
@@ -193,8 +193,6 @@ unsigned char* bytesFromExtMemory(const char* _path, unsigned int* _size) {
     resource.read(cdata, *_size);
     resource.close();
 
-    logMsg("Tangram Testing: Here\n");
-
     return reinterpret_cast<unsigned char *>(cdata);
 
 }
--- a/android/tangram/jni/platform_android.cpp
+++ b/android/tangram/jni/platform_android.cpp
@@ -83,9 +83,9 @@ std::string systemFontPath(const std::string& _family, const std::string& _weigh
     jstring returnStr = (jstring) jniEnv->CallObjectMethod(tangramInstance, getFontFilePath, jfamily, jweight, jstyle);
 
     size_t length = jniEnv->GetStringUTFLength(returnStr);
-    char* cdata = (char*) malloc(sizeof(char) * (length));
-    jniEnv->GetStringUTFRegion(returnStr, 0, length, cdata);
-    return std::string(cdata, length);
+    std::string fontPath = std::string(length, 0);
+    jniEnv->GetStringUTFRegion(returnStr, 0, length, &fontPath[0]);
+    return fontPath;
 }
 
 void setContinuousRendering(bool _isContinuous) {
--- a/android/tangram/src/com/mapzen/tangram/FontFileParser.java
+++ b/android/tangram/src/com/mapzen/tangram/FontFileParser.java
@@ -4,6 +4,7 @@ import java.io.File;
 import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.FileNotFoundException;
 import java.util.HashMap;
 import java.util.Map;
 
@@ -14,19 +15,16 @@ import org.xmlpull.v1.XmlPullParserFactory;
 import android.util.Xml;
 import android.util.Log;
 
-public class FontFileParser {
-    private static final File fontFile = new File("/system/etc/fonts.xml");
-    private static final File systemFontFile = new File("/system/etc/system_fonts.xml");
-    private static final File fallbackFontFile = new File("/system/etc/fallback_fonts.xml");
+class FontFileParser {
 
-    private static Map<String, String> fontDict = new HashMap<String, String>();
+    private Map<String, String> fontDict = new HashMap<String, String>();
 
     private void processDocument(XmlPullParser parser) throws XmlPullParserException, IOException {
 
         parser.nextTag();
         // Parse Families
         parser.require(XmlPullParser.START_TAG, null, "familyset");
-        while (parser.next() != XmlPullParser.END_TAG) {
+        while (parser.next() != XmlPullParser.END_DOCUMENT) {
             if (parser.getEventType() != XmlPullParser.START_TAG) {
                 continue;
             }
@@ -40,30 +38,35 @@ public class FontFileParser {
                     String tag = parser.getName();
                     if (tag.equals("font")) {
                         String weightStr = parser.getAttributeValue(null, "weight");
+                        weightStr = (weightStr == null) ? "400" : weightStr;
+
                         String styleStr = parser.getAttributeValue(null, "style");
+                        styleStr = (styleStr == null) ? "normal" : styleStr;
+
                         String filename = parser.nextText();
                         String fullFilename = "/system/fonts/" + filename;
+
                         String key = name + "_" + weightStr + "_" + styleStr;
                         fontDict.put(key, fullFilename);
                     } else {
                         skip(parser);
                     }
                 }
-            } /*else if (parser.getName().equals("alias")) {
+            } else if (parser.getName().equals("alias")) {
                 // Parse this alias to font to fileName
                 String aliasName = parser.getAttributeValue(null, "name");
                 String toName = parser.getAttributeValue(null, "to");
                 String weightStr = parser.getAttributeValue(null, "weight");
-                if (weightStr == "") { weightStr = "400"; } // normal
-                String fontFilename = fontDict.get(toName + "_" + weightStr + "_normal"); // alias style is default: normal
+                weightStr = (weightStr == null) ? "400" : weightStr;
+                String fontFilename = fontDict.get(toName + "_" + weightStr + "_normal"); // alias style is default: normal (always)
                 fontDict.put(aliasName + "_" + weightStr, fontFilename);
-            } */else {
+            } else {
                 skip(parser);
             }
         }
     }
 
-    private static void skip(XmlPullParser parser) throws XmlPullParserException, IOException {
+    private void skip(XmlPullParser parser) throws XmlPullParserException, IOException {
         int depth = 1;
         while (depth > 0) {
             switch (parser.next()) {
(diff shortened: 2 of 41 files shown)


android/tangram/jni/platform_android.cpp
```diff
@@ -72,6 +73,53 @@ void requestRender() {
...
 
+std::string deviceFontsPath() {
+    return "/system/fonts/";
+}
+
+bool parseTypeFaceFontsInfo(const std::string& _typefaceInput, std::string& _fontName, float& _fontSize) {
+
+    char str[4][40];
+    float size;
+    int num = sscanf(_typefaceInput.c_str(), "%s %s %s %s", str[0], str[1], str[2], str[3]);
```
tallytalwar: This is the most ugly bit of parsing code. Suggestions welcomed.

core/resources/scene/config.yaml
```diff
@@ -16,6 +16,12 @@ textures:
     water-texture:
         url: normals.jpg
         filtering: nearest
+
+# provide global fontfile names to be used for this stylesheet
+# these could be bundled with the tangram repo or could be loaded from system path
+# priority is given to bundled fonts
+fonts: [ Roboto-Regular.ttf, FiraSans-Medium.ttf]
```
tallytalwar: @bcamper this is what I have to specify font files for the style sheet.

android/tangram/jni/platform_android.cpp
```diff
@@ -126,6 +174,31 @@ std::string stringFromResource(const char* _path) {
...
+    *_size = resource.tellg();
+
+    resource.seekg(std::ifstream::beg);
+
+    char* cdata = (char*) malloc(sizeof(char) * (*_size));
+
+    resource.read(cdata, *_size);
+    resource.close();
+
+    logMsg("Tangram Testing: Here\n");
```
blair1618: leftover log

core/resources/scene/config.yaml
blair1618: These file names are platform-specific, we can't require them to be specified in the stylesheet (it defeats the whole purpose of having a cross-platform implementation of the renderer).

tallytalwar: Good point.
We can have a list of strings something respecting the following format:
```
"FontName, Weight, Style, Extension"
```
Which can then be parsed as (for android): FontName-WeightStyle.ext

karimnaaji: What about something like such: 
```yaml
fonts:
    font:
         name: Roboto
         face: Regular
    font: 
        name: FiraSans
        face: Medium
```
As we spoke last time, since we're able to know the names for the platform, we can have an a specific implementation for each platform to figure those name out.

tallytalwar: :+1: 
This seems good to me.

core/resources/scene/config.yaml
```diff
@@ -177,7 +183,10 @@ layers:
         data: {source: osm, layer: roads}
         filter: { name: true, aeroway: false, tunnel: false, railway: false, not: { kind: rail }, $zoom: { min: 10 } }
         draw:
-            text: {}
+            text: 
+                font:
+                    typeface: Regular 15 Roboto
```
tallytalwar: 1. Based on the documentation [here](https://github.com/tangrams/tangram-docs/blob/ec9491758ad578b5607e7549336fdac07a64475b/pages/draw.md#font-parameters), there are 2 ways of specifying a font size, one is the "size" property and other is the size in the "typeface" property.

Also,
2. Instead of having typeface as a string of different values, will it make sense to have it broken down as yaml key value pairs?
```
text:
      font:
            name:
            face:
            size:
            weight:
```

This could be helpful in style mixing, where a sublayer only overrides one of the specific font properties, instead of specifying the whole typeface string.

tallytalwar: With respect to implementation point of view on cpp:
This gets really hard and ugly to parse typeface string into "face", "weight", "size" and "name" for further processing. And given these parameters could be optional, parsing code can get really ugly and convoluted.

But if we have yaml key value pairs for these parameters, it will be trivial implementation for both cpp and js side.

karimnaaji:
When we figured out what to use for font description, we could merge the branch [font-units](https://github.com/tangrams/tangram-es/tree/font-units) into this one to get different specified font unit (em/pt...).

core/resources/scene/config.yaml
karimnaaji: I agree with @tallytalwar  that breaking the font properties into several would make more sense for native. This would let us use the yaml parsing instead of writing parsing specifically for fonts definition.

core/resources/scene/scene.yaml
```diff
@@ -16,6 +16,22 @@ textures:
...
+# priority is given to bundled fonts
+fonts:
+    font1:
+        name: Roboto
+        weight: Regular
+    font2:
+        name: FireSans
+        weight: Medium
+    font3:
+        name: Roboto
```
blair1618: Is there any reason this information can't be inferred from text draw parameters? It seems like we could just infer the necessary fonts, lazily instantiate them, and remove the need for this top-level block entirely

tallytalwar: Thats a good point and yes it can (will) be done. 

I will however wait for discussion on "typeface" font parameter to come to a conclusion and then implement the code using "name", "weight" and "face" parameters of a font.

So, as proposed earlier for style rule matching and inheritance and (native) implementation sakes, its much better to have the following:

```
draw:
        text:
              font:
                     size: 15px
                     name: Roboto
                     weight: Bold
                     face: Italics
```
instead of
```
draw:
        text:
              font:
                     typeface: Italics Bold 15px Roboto
```

blair1618: @bcamper what do you think of the suggested syntax revision above? I too prefer the map-style syntax because of it's suitability for rule inheritance.

android/tangram/jni/platform_android.cpp
```diff
@@ -126,6 +180,29 @@ std::string stringFromResource(const char* _path) {
 
 }
 
+unsigned char* bytesFromExtMemory(const char* _path, unsigned int* _size) {
```
blair1618: This name is kinda confusing - it's reading from storage (not memory) and "Ext" is kinda ambiguous. Maybe `bytesFromFileSystem` ? `bytesFromStorage` ?

core/resources/scene/scene.yaml
```diff
@@ -177,7 +193,12 @@ layers:
         data: {source: osm, layer: roads}
         filter: { name: true, aeroway: false, tunnel: false, railway: false, not: { kind: rail }, $zoom: { min: 10 } }
         draw:
-            text: {}
+            text: 
+                font:
+                    typeface: Regular 17 Roboto
```
blair1618: We're going to split this into fields, yes?

core/resources/shaders/sdf.fs
```diff
@@ -51,14 +50,14 @@ float sampleAlpha(in vec2 uv, float distance, in float off) {
...
-        alpha = pow(alpha, 1.0 / gamma);
+	} else {
+	    float distance = texture2D(u_tex, v_uv).a;
+    	float alpha = sampleAlpha(v_uv, distance, sdf) * tint;
+    	alpha = pow(alpha, 0.4545);
 
-        gl_FragColor = vec4(u_color, v_alpha * alpha);
-    }
+    	gl_FragColor = vec4(v_color.rgb, v_alpha * alpha * v_color.a);
+	}
```
blair1618: Did some tabs sneak into this file? The spacing looks odd.

core/resources/shaders/sprite.fs
```diff
@@ -9,12 +9,13 @@ uniform sampler2D u_tex;
...
 varying float v_alpha;
+varying vec4 v_color;
 
 void main(void) {
     if (v_alpha < TANGRAM_EPSILON) {
         discard;
     } else {
         vec4 color = texture2D(u_tex, v_uv);
-        gl_FragColor = vec4(color.rgb, v_alpha * color.a);
+		gl_FragColor = vec4(color.rgb * v_color.rgb, v_alpha * color.a * v_color.a);
```
blair1618: tabs again?

core/src/platform.h
```diff
@@ -44,35 +44,55 @@ void logMsg(const char* fmt, ...);
...
 
 bool isContinuousRendering();
 
+/* Get the system path for font files
+ *  for each platform this will be different
+ */
+std::string deviceFontsPath();
+
+/* Construct font file name from font properties specified in yaml */
+std::string constructFontFilename(const std::string& _name, const std::string& _weight, const std::string& _face);
```
blair1618: I feel like `constructFontFileName` and `deviceFontsPath` can be consolidated into a single function like `getFontFilePath`, I can't see how these would be used separately and they may not be sufficient for cases like the multiple font file paths in OS X

core/src/scene/drawRule.h
```diff
@@ -15,10 +15,12 @@ namespace Tangram {
 
 enum class StyleParamKey : uint8_t {
     none, order, color, width, cap, join, outline_color, outline_width, outline_cap, outline_join,
+    font_typeface, font_fill, font_stroke, font_stroke_color, font_stroke_width, font_capitalized,
+    font_visible,
 };
 
 struct StyleParam {
-    using Value = variant<none_type, std::string, Color, CapTypes, JoinTypes, int32_t, float>;
+    using Value = variant<none_type, bool, std::string, Color, CapTypes, JoinTypes, int32_t, float>;
```
blair1618: Note for future revisions: We can safely consolidate `bool`, `CapTypes`, and `JoinTypes` as `char`

core/resources/scene/scene.yaml
tallytalwar: yes, I am working on it right now.

core/src/platform.h
tallytalwar: I agree, I made these 2 functions at 2 different times and did not look into pruning them.

tallytalwar:
This looks mostly good now, except I think I broke debugTextStyle drawing :(. Will debug/fix that and update for reviewing.

core/src/scene/drawRule.cpp
```diff
@@ -21,6 +21,16 @@ const std::map<std::string, StyleParamKey> s_StyleParamMap = {
...
+    {"font:name", StyleParamKey::font_name},
+    {"font:weight", StyleParamKey::font_weight},
+    {"font:face", StyleParamKey::font_face},
+    {"font:size", StyleParamKey::font_size},
+    {"font:fill", StyleParamKey::font_fill},
+    {"font:stroke", StyleParamKey::font_stroke},
+    {"font:stroke_color", StyleParamKey::font_stroke_color},
+    {"font:stroke_width", StyleParamKey::font_stroke_width},
+    {"font:capitalized", StyleParamKey::font_capitalized},
+    {"font:visible", StyleParamKey::font_visible}
```
bcamper: Not sure we need a separate `visible` flag for font blocks specifically. What's the thinking here? (Maybe it's a debug thing?)

karimnaaji: No it's an error, I was just modifying this and moving it up to `text`

nvkelso: Separate toggle for label visibility in general and text visibility as a "sub layer" seems useful to me. I've setup Skin and Bones this way thru some hacks.

bcamper: Yes that is covered in the generic `visible` property available for each draw group. This was a further visible property inside the font block.

bcamper:
We should normalize our property names to match CSS properties as much as possible:
http://www.w3.org/TR/css3-fonts/#property-index

Propose we implement the following subset of CSS for now:

- `family`: Helvetica, Futura, etc.
- `size`: in `pt`, `px`, or `em`
- `weight`: `normal`, `bold`, etc.
  - for now only support `bold`, all other values default to `normal`
- `style`: `normal`, `italic`, `oblique`
  - for now only support `italic`, all other values default to `normal`
- `transform`: replace our `capitalized` behavior with a more generic proper that follows CSS `text-transform` (see https://developer.mozilla.org/en-US/docs/Web/CSS/text-transform):
  - `capitalized`: first letter of each word is capitalized -- this is not our current behavior, which is actually what uppercase is, so we should fix that
  - `uppercase`: all letters are upper cased
  - `lowercase`: all letters are lower cased

There are other properties we could add in the future, such as:

`stretch`: `normal`, `condensed`, etc.

tallytalwar:
@bcamper Since we will be using the font (family, weight, style) mapping to font filename on disk, we can support all weight values.

For example a sample mapping of these font properties to a file on android looks like:

```
<family name="sans-serif">
        <font weight="100" style="normal">Roboto-Thin.ttf</font>
        <font weight="100" style="italic">Roboto-ThinItalic.ttf</font>
        <font weight="300" style="normal">Roboto-Light.ttf</font>
        <font weight="300" style="italic">Roboto-LightItalic.ttf</font>
        <font weight="400" style="normal">Roboto-Regular.ttf</font>
        <font weight="400" style="italic">Roboto-Italic.ttf</font>
        <font weight="500" style="normal">Roboto-Medium.ttf</font>
        <font weight="500" style="italic">Roboto-MediumItalic.ttf</font>
        <font weight="900" style="normal">Roboto-Black.ttf</font>
        <font weight="900" style="italic">Roboto-BlackItalic.ttf</font>
        <font weight="700" style="normal">Roboto-Bold.ttf</font>
        <font weight="700" style="italic">Roboto-BoldItalic.ttf</font>
</family>
```

tallytalwar:
This should be reviewable now. @karimnaaji just pushed a fix for debug text and also making sure to use the latest renderStates update.

tallytalwar:
This branch should be ready for review.
- Updated font styleparams to handle css style properties, including "family", "weight" and "style". Note all other platforms will fallback on default bundled font for the time being. However android will try to fetch apt system font file depending on the css style font properties. It can also handle numeric weights, and font aliases like "ariel", "times" and "times new roman".
- If the font specifications specified in the style sheet does not match either the bundled fonts or the system fonts, it will fallback to the default application bundled font.
- This also includes @karimnaaji's work on a bunch of text rendering improvements which brings us much closer to text parsing/rendering on the tangram js renderer.

android/tangram/src/com/mapzen/tangram/MapController.java
```diff
@@ -33,6 +35,8 @@
 
 import okio.BufferedSource;
 
+import org.xmlpull.v1.XmlPullParserException;
```
ecgreb: Unused import?

tallytalwar: :+1: and :metal:  

Removed.

android/tangram/src/com/mapzen/tangram/FontFileParser.java
```diff
@@ -0,0 +1,138 @@
...
+    private void processDocument(XmlPullParser parser) throws XmlPullParserException, IOException {
+
+        parser.nextTag();
+        // Parse Families
+        parser.require(XmlPullParser.START_TAG, null, "familyset");
+        while (parser.next() != XmlPullParser.END_DOCUMENT) {
+            if (parser.getEventType() != XmlPullParser.START_TAG) {
+                continue;
+            }
+            if (parser.getName().equals("family")) {
```
ecgreb: If `parser.getName()` returns null this will throw a NullPointerException

Better to invert the order and place the string literal first.

```java
if ("family".equals(parser.getName()) { ... }
```

tallytalwar: Nice. Every cpp programmer will make this mistake. Thanks... fixing now.

hjanetzek:
Looks good to me!

Some future work could be:
- std::toupper is likely only work for most Latin characters (would need ICU for proper support, which is quite heavy)
- fontKey could be a struct { int family, int weight, fontstyle } with FontContext holding the mappings for  int->string

hjanetzek:
Got a segv after context lost in FontContext::renderUpdate, Texture::setSubData.

```
/libc    (  598): Fatal signal 7 (SIGBUS), code 2, fault addr 0xa2621000 in tid 627 (Thread-534)
I/DEBUG   (  205): property debug.db.uid not set; NOT waiting for gdb.
I/DEBUG   (  205): HINT: adb shell setprop debug.db.uid 100000
I/DEBUG   (  205): HINT: adb forward tcp:5039 tcp:5039
I/DEBUG   (  205): *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
I/DEBUG   (  205): Build fingerprint: 'htc/m7_google/m7:5.0.1/LRX22C.H3/434277:user/release-keys'
I/DEBUG   (  205): Revision: '128'
I/DEBUG   (  205): ABI: 'arm'
I/DEBUG   (  205): pid: 598, tid: 627, name: Thread-534  >>> com.mapzen.tangram.android <<<
I/DEBUG   (  205): signal 7 (SIGBUS), code 2 (BUS_ADRERR), fault addr 0xa2621000
I/DEBUG   (  205):     r0 9e25affc  r1 00000007  r2 9e25b000  r3 a2621000
I/DEBUG   (  205):     r4 00008e00  r5 00004808  r6 b87cd408  r7 a2602008
I/DEBUG   (  205):     r8 00008e00  r9 00000000  sl b875b068  fp 00000200
I/DEBUG   (  205):     ip a14fc918  sp a14fc978  lr b6e666ab  pc afaa2076  cpsr a0010030
I/DEBUG   (  205): 
I/DEBUG   (  205): backtrace:
I/DEBUG   (  205):     #00 pc 00081076  /data/app/com.mapzen.tangram.android-1/lib/arm/libtangram.so (Tangram::Texture::setSubData(unsigned int const*, unsigned int, unsigned int, unsigned int, unsigned int)+89)
I/DEBUG   (  205):     #01 pc 00072587  /data/app/com.mapzen.tangram.android-1/lib/arm/libtangram.so (Tangram::FontContext::renderUpdate(void*, int*, unsigned char const*)+94)
```

android/tangram/src/com/mapzen/tangram/FontFileParser.java
```diff
@@ -0,0 +1,138 @@
...
+
+        if (fontFile.exists()) {
+            fileXml = fontFile.getAbsolutePath();
+        } else if (systemFontFile.exists()) {
+            fileXml = systemFontFile.getAbsolutePath();
+        } else if (fallbackFontFile.exists()) {
+            fileXml = systemFontFile.getAbsolutePath();
+        }
+
+        if(fileXml == "") {
```
blair1618: Using `==` on Java objects checks whether they have the same reference (pointer equality, basically). To check if the contents of two strings are equal, use `.equals()`. In this case though it's probably simpler just leave it uninitialized and check for `null`.

android/tangram/src/com/mapzen/tangram/MapController.java
```diff
@@ -14,6 +14,8 @@
 import android.view.View;
 import android.view.View.OnTouchListener;
 
+import com.mapzen.tangram.FontFileParser;
```
blair1618: Unused import

android/tangram/src/com/mapzen/tangram/MapController.java
```diff
@@ -290,6 +291,8 @@ public void setTapGestureListener(View.OnGenericMotionListener listener) {
     private RotateGestureDetector rotateGestureDetector;
     private ShoveGestureDetector shoveGestureDetector;
 
+    private static final FontFileParser fontFileParser = new FontFileParser();
```
blair1618: Why static? Let's avoid static initialization where it isn't necessary.

android/tangram/src/com/mapzen/tangram/FontFileParser.java
tallytalwar: You mean something like this:

```
String fileXml;
... 
...
...
if (fileXml == null) { blah;}

File: android/tangram/src/com/mapzen/tangram/MapController.java [hunk 18]
```diff
@@ -552,6 +555,12 @@ public void cancelUrlRequest(String url) {
         okClient.cancel(url);
     }
 
+    public String getFontFilePath(String family, String weight, String style) {
```
blair1618: Let's not put this in the middle of a group of methods explicitly labelled as 'networking'

android/tangram/src/com/mapzen/tangram/FontFileParser.java
blair1618: Just so.

tallytalwar: Ha!

So I can not let a variable uninitialized ... it throws a compilation error "fileXml might not have been initialized".

Will do the equals check.

tallytalwar:
@hjanetzek Was there a specific operation you were doing to get that segv. Can you send me repro instructions?

android/tangram/src/com/mapzen/tangram/FontFileParser.java
hjanetzek: @blair1618 oh, yes I haven't checked the java code.. But this case one would work as string constants are interned()  :D

android/tangram/jni/platform_android.cpp
```diff
@@ -72,6 +75,19 @@ void requestRender() {
...
 
+std::string systemFontPath(const std::string& _family, const std::string& _weight, const std::string& _style) {
+
+    jstring jfamily = jniEnv->NewStringUTF(_family.c_str());
+    jstring jweight = jniEnv->NewStringUTF(_weight.c_str());
+    jstring jstyle = jniEnv->NewStringUTF(_style.c_str());
+    jstring returnStr = (jstring) jniEnv->CallObjectMethod(tangramInstance, getFontFilePath, jfamily, jweight, jstyle);
+
+    size_t length = jniEnv->GetStringUTFLength(returnStr);
+    char* cdata = (char*) malloc(sizeof(char) * (length));
```
blair1618: This memory will be leaked after the function returns (`std::string` only copies the data, it doesn't take ownership). You could instead create a string of suitable length `auto out = std::string(length, 0)` and copy the data directly into it `jniEnv->GetStringUTFRegion(returnStr, 0, length, &out[0])`

core/resources/shaders/sdf.fs
```diff
@@ -1,9 +1,9 @@
...
-    #ifdef GL_OES_standard_derivatives
-        #extension GL_OES_standard_derivatives : enable
-    #else
-        #undef TANGRAM_SDF_MULTISAMPLING
-    #endif
+#ifdef GL_OES_standard_derivatives
+#extension GL_OES_standard_derivatives : enable
+#else
+#undef TANGRAM_SDF_MULTISAMPLING
+#endif
```
blair1618: Any reason for un-indenting this? I find it easier to read when indented.

hjanetzek:
@tallytalwar I just rotated a few times to get the tangram instance recreated. It seems that the subData pointer from fontstash becomes invalid - the fault addr lies within its range

android/tangram/jni/platform_android.cpp
tallytalwar: Good catch and good suggestion.

tallytalwar:
@hjanetzek Thanks.
Will check it.

android/tangram/src/com/mapzen/tangram/FontFileParser.java
```diff
@@ -0,0 +1,138 @@
...
+            }
+        }
+    }
+
+    public FontFileParser() {
+
+        InputStream in = null;
+        final File fontFile = new File("/system/etc/fonts.xml");
+        final File systemFontFile = new File("/system/etc/system_fonts.xml");
+        final File fallbackFontFile = new File("/system/etc/fallback_fonts.xml");
```
blair1618: The three files named here all have different layouts (on my phone, at least). We can't really treat them as substitutes because `parseDocument` will simply fail on a file that doesn't have the same layout as fonts.xml

core/src/gl/texture.cpp
```diff
@@ -64,23 +64,27 @@ void Texture::setData(const GLuint* _data, unsigned int _dataSize) {
...
     size_t bpp = bytesPerPixel();
     size_t divisor = sizeof(GLuint) / bpp;
-    for (size_t j = 0; j < _height; j++) {
-        size_t dpos = ((j + _yoff) * m_width + _xoff) / divisor;
-        size_t spos = (j * _width) / divisor;
-        std::memcpy(&m_data[dpos], &_subData[spos], _width * bpp);
+
+    // Init m_data if update() was not called after resize()
+    if (m_data.size() != (m_width * m_height) / divisor) {
+        m_data.resize((m_width * m_height) / divisor);
```
tallytalwar: Don't we have to update the _xoff and _yoff at this point to reflect the new location to be updated in the resized m_data?

tallytalwar: Also it seems we are not using Texture::resize right now!

android/tangram/src/com/mapzen/tangram/FontFileParser.java
blair1618: After checking on several devices now, it seems that `fonts.xml` is a rather new format that only exists on Android L (5.0) or newer. `system_fonts.xml` was the standard name & format prior to 5.0 and is still present on 5.0 so it might make more sense to read that one by default.

core/src/gl/texture.cpp
hjanetzek: At least in Texture constructor

tallytalwar: yup, but shouldn't the offset's be updated to reflect the updated texture size?

hjanetzek: To the first question: This part just makes sure that m_data is actually allocated if Texture::update() was not called before setSubData..

hjanetzek: the only necessary fix is the 'fix segv' commit or:

````
- std::unique_ptr<std::vector<GLuint>> subData(new std::vector<GLuint>(_subData, _subData + (_width * _height)));
+ std::unique_ptr<std::vector<GLuint>> subData(new std::vector<GLuint>(_subData, _subData + (_width * _height)/divider));

````

hjanetzek: ok, the 'fix: initialization of Texture m_data' is also necessary, but I have never hit this case yet :)

PR #257, opened 2015-08-24
Custom scene file

blair1618:
Custom scene files can now be set at initialization

@ecgreb now `MapController` has an optional 3rd constructor parameter to provide a path to a custom scene file


core/src/tangram.cpp
```diff
@@ -61,8 +61,11 @@ void initialize() {
         m_ftContext->addFont("FiraSans-Medium.ttf", "FiraSans");
         m_labels = std::unique_ptr<Labels>(new Labels());
 
+        logMsg("Loading Tangram scene file: %s\n", _scenePath);
```
tallytalwar: _stylePath instead of _scenePath

blair1618: (jedi git magic) these aren't the droids you're looking for

tallytalwar: Github should do more magic to respect comments and rebases :D.

PR #258, opened 2015-08-25
Use render states for textures

karimnaaji:
Use of the render states tracking we've added for texture unit/id binding.


blair1618:
Works for me - if there are no complaints I'll merge this.

PR #259, opened 2015-08-26
WIP - Style mixing

tallytalwar:
- Works with something like this:
```
styles:
    pillows:
        base: polygon
        lighting: fragment
        material:
            normal:
                texture: water-texture
                mapping: uv
            diffuse: 0.75

    heightglow:
        base: polygon
            lighting: vertex
            shaders:
                blocks:
                    color:
                        "color.rgb += vec3(0.3 * position.z);"

    roadglow:
        mix: heightglow
        base: lines
        lighting: fragment

    test:
        mix: [heightglow, pillows]
        base: polygon
        lighting: vertex
```

### DIFF
--- a/android/demo/AndroidManifest.xml
+++ b/android/demo/AndroidManifest.xml
@@ -6,6 +6,7 @@
     <uses-feature android:glEsVersion="0x00020000" android:required="true" />
     <uses-permission android:name="android.permission.INTERNET" />
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
+    <uses-permission android:name="android.permission.READ_EXTERNAL_STORAGE" />
     <application android:label="@string/app_name" android:icon="@drawable/ic_launcher">
         <activity android:name="MainActivity"
                   android:label="@string/app_name">
--- a/android/demo/AndroidManifest.xml
+++ b/android/demo/AndroidManifest.xml
@@ -6,7 +6,6 @@
     <uses-feature android:glEsVersion="0x00020000" android:required="true" />
     <uses-permission android:name="android.permission.INTERNET" />
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
-    <uses-permission android:name="android.permission.READ_EXTERNAL_STORAGE" />
     <application android:label="@string/app_name" android:icon="@drawable/ic_launcher">
         <activity android:name="MainActivity"
                   android:label="@string/app_name">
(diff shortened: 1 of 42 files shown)


blair1618:
Nice. Just one thing that I can think of: From the docs it seems that a style A can `mix` style B and, if no `base` is specified in A, the `base` of B is used (see https://github.com/tangrams/tangram-docs/blob/gh-pages/pages/Styles-Overview.md#style-composition-with-mix). Would that work correctly for us?

bcamper:
Small note, in the example, it's probably better to write this:

```
                    color:
                        "color.rgb += vec3(0.3 * position.z);"
```

as this:

```
                    color: |
                        color.rgb += vec3(0.3 * position.z);
```

Since you often want a multi-line string for shader blocks, and even if you didn't, it should automatically parse it as a string.

bcamper:
So style mixing requires more field-specific handling in order to properly merge styles, if I understand how you've implemented it (I wasn't able to find a generic way to this on the JS side). You can see the JS version here: https://github.com/tangrams/tangram/blob/master/src/styles/style_manager.js#L180-L257

There are 3 types of properties:

- Flags like `animated` or `texcoords` that are OR'ed, so `true` if any style has it set: https://github.com/tangrams/tangram/blob/master/src/styles/style_manager.js#L199
- Overwriting, single properties like `base` (this answers @blair1618's question too), `lighting`, etc. where the "last style wins", e.g. https://github.com/tangrams/tangram/blob/master/src/styles/style_manager.js#L204
- Structures that are deeply merged, like `material` and shader `blocks`. For example, a style B that mixes styles A can modify just the `specular` component of its material, while preserving the rest of style A's material properties. Or, B can add a `color` shader block that is inserted after any `color` blocks from style A. (Single properties that overwrite can be considered a simplified form of a deep merge, but I wanted to call out the deep merge in case it requires explicit handling.)

Is that clear? We can walk through more example.

bcamper:
Oh, there is also the case to consider for shader `extensions`, where you want to build a *unique* set of extensions from all mixed styles. This could be considered a kind of  "post-processing" step after the structures are merged, however.

Now that I'm looking at this... maybe I can find a way to simplify this on the JS side into one big deep merge. If I recall, I had some issues with default properties being set earlier in the pipeline causing issues, but with JS's distinction between `null` and `undefined`, maybe I can make this cleaner. Will take a look.

tallytalwar:
Thanks @bcamper for all these links. 
This branch is still WIP.

tallytalwar:
Closed in favor of https://github.com/tangrams/tangram-es/pull/283

It was easy to do a cherry pick on master than rebasing this to master... because of much different histories :(.

PR #260, opened 2015-08-26
Extrude draw param for polygon style

tallytalwar:


blair1618:
Nice and simple :)

It's unclear what type `extrude` should have, though. From the tangram source it looks like it can be a boolean, a number, a function, or a pair of numbers? (https://github.com/tangrams/tangram/blob/master/src/styles/polygons/polygons.js#L61)

tallytalwar:
Yes, I saw that.

But wanted to have something out there. And was not sure of few things like the default value.
Also it seems we are not using min_height in our extrusion code, which is used in the "pair" usecase for extrude param.

tallytalwar:
Plus it should not be hard to do using variant for extrude parameter.

core/src/scene/drawRule.cpp
```diff
@@ -170,6 +196,17 @@ bool DrawRule::getValue(StyleParamKey _key, int32_t& _value) const {
     return true;
 }
 
+bool DrawRule::getValue(StyleParamKey _key, std::pair<float, float>& _value) const {
+    auto& param = findParameter(_key);
+    if (!param) { return false; }
+    if (!param.value.is<std::pair<float, float>>()) {
+        logMsg("Error: not a bool\n");
```
blair1618: should say 'not a pair', probably

tallytalwar: fixed.

PR #261, opened 2015-08-26
cmake: use public headers

hjanetzek:
When using target_include_directories the public includes will be inherited by other targets when referenced with target_link_libraries

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.8)
+cmake_minimum_required(VERSION 2.8.11)
 project(tangram)
 
 include(${PROJECT_SOURCE_DIR}/toolchains/utils.cmake)
--- a/.gitmodules
+++ b/.gitmodules
@@ -12,7 +12,7 @@
 	url = https://github.com/tangrams/isect2d.git
 [submodule "core/dependencies/yaml-cpp"]
 	path = core/dependencies/yaml-cpp
-	url = https://github.com/blair1618/yaml-cpp.git
+	url = https://github.com/tangrams/yaml-cpp.git
 	branch = no-boost
 	ignore = untracked
 [submodule "glfw"]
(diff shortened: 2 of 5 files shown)


blair1618:
Cool. Will merge if nobody has comments.

CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.8)
+cmake_minimum_required(VERSION 2.8.11)
```
tallytalwar: In the [Readme](https://github.com/tangrams/tangram-es#build) we have mentioned cmake min is 3.0.

hjanetzek:
yea, but we dont get cmake 3.0 packages with apt. just something between 2.8.9 and 2.8.11 :/

blair1618:
I think the 3.0 minimum may have been for a dependency which we no longer use, but I don't remember precisely. 

One consideration is that CMake 2.8.9 is the latest version available on Debian Wheezy (and therefore Raspberry Pi). 2.8.11 is available from wheezy-backports so maybe that's not so bad.

tallytalwar:
Cool. Should be change it in the Readme then?

Varun Talwar
Mailto: varun@mapzen.com

blair1618:
Yep I'll check on the pi and update the readme

core/dependencies/yaml-cpp
```diff
@@ -1 +1 @@
-Subproject commit 95bb6a18f9523b46499e9b870ef4b8e4d8068fbb
+Subproject commit a49337024219f7b3778cbcbfc3e6d26e84f9408f
```
tallytalwar: Am I missing something? But I am not able to update submodule post merge to master.

I get the following on "git submodule update":
```
fatal: reference is not a tree: a49337024219f7b3778cbcbfc3e6d26e84f9408f
Unable to checkout 'a49337024219f7b3778cbcbfc3e6d26e84f9408f' in submodule path 'core/dependencies/yaml-cpp'
```

tallytalwar: I had to do the following to get it working, since the yaml-cpp submodule's fork was updated:

```
git submodule sync
```

PR #262, opened 2015-08-26
duktape integration

hjanetzek:
TODOs

- [x] should duktape be added as a submodule? The problem is that the amalgamated version needs to be build using python tools. So we would have make a repo of the amalgamated version for our use.
- [x] FilterContext needs a better name as it is also used for StyleParam evaluation. So maybe StyleContext? It actually replaces the 'using Context = map\<string,value\>' typedef - But Context sounds also a bit too general - unless.. we move it to a sub-namespace? :)  
- [x] Remove scene.yaml changes used for testing 

- ...

### DIFF
--- a/tests/unit/dukTests.cpp
+++ b/tests/unit/dukTests.cpp
@@ -163,6 +163,7 @@ TEST_CASE( "Test evalStyleFn - StyleParamKey::extrude", "[Duktape][evalStyleFn]"
     ctx.setFeature(feat);
     ctx.addFunction("fn_t", R"(function () { return true; })");
     ctx.addFunction("fn_f", R"(function () { return false; })");
+    ctx.addFunction("fn_a", R"(function () { return [1.1, 2.2]; })");
 
     StyleParam::Value value;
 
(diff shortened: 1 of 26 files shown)


blair1618:
Reviewing this now! I'll try not to take too long. 

What do you think about adding some labels for PRs? It's not always clear whether a branch is still being worked on or whether it should be merged as-is once reviewed. Maybe a "Do not merge" label for PRs with ongoing changes.

tests/unit/dukTests.cpp
```diff
@@ -0,0 +1,266 @@
...
+    {
+        Feature feature;
+        feature.props.add("a", "A");
+        feature.props.add("b", "B");
+        feature.props.add("n", 42);
+
+        StyleContext ctx;
+        ctx.setFeature(feature);
+
+        ctx.addFunction("fn_a", R"(function() { return feature.a == 'A' })");
```
blair1618: Tests should probably use `===` in the js functions, just in case.

tests/unit/dukTests.cpp
```diff
@@ -0,0 +1,266 @@
...
+        REQUIRE(ctx.evalFilterFn("fn_n2") == false);
+
+        // OK?
+        ctx.addFunction("fn_n3", R"(function() { return feature.n == '42' })");
+        REQUIRE(ctx.evalFilterFn("fn_n3") == true);
+    }
+}
+
+TEST_CASE( "Test evalFilterFn with feature and globals", "[Duktape][evalFilterFn]") {
+    {
```
blair1618: What's up with the extra `{}` scope?

blair1618:
Big addition! Code looks very good to me aside from minor comments on the tests. Just needs to be rebased and cleaned of the JS test functions in the scene file, then it should be good to merge.

PR #263, opened 2015-08-27
use uint16 for indices

hjanetzek:
A single mesh may not have more than max GLushort vertices so this
can already be restricted in mesh building.
(otherwise the vertex index will overflow VboMesh::compile)


tallytalwar:
:+1:

PR #264, opened 2015-08-28
Some issues found by coverity

hjanetzek:


core/src/labels/label.cpp
```diff
@@ -17,6 +17,7 @@ Label::Label(Label::Transform _transform, glm::vec2 _size, Type _type, LabelMesh
     m_occludedLastFrame = false;
```
tallytalwar: I think we can move all these to the header and clean the constructor.

core/src/scene/sceneLoader.cpp
```diff
@@ -32,24 +32,29 @@ constexpr size_t CACHE_SIZE = 16 * (1024 * 1024);
...
 
+        // Styles that are opaque must be ordered first in the scene so that they are rendered 'under' styles that require blending
+        std::sort(_scene.styles().begin(), _scene.styles().end(), [](std::unique_ptr<Style>& a, std::unique_ptr<Style>& b) {
+                return a->isOpaque();
+            });
+    } catch (YAML::ParserException e) {
+        logMsg("Error: Parsing scene config '%s'\n", e.what());
+    } catch (YAML::RepresentationException e) {
+        logMsg("Error: Parsing scene config '%s'\n", e.what());
+    }
```

core/src/style/textStyle.cpp
```diff
@@ -97,6 +97,7 @@ void TextStyle::buildPolygon(const Polygon& _polygon, const DrawRule& _rule, con
             n++;
         }
     }
+    if (n == 0) { return; }
```
tallytalwar: Ideally we should return much before n is calculated... when polygon size is zero.

core/src/util/mapProjection.h
```diff
@@ -137,6 +136,8 @@ class MercatorProjection : public MapProjection {
      * Following define the boundry covered by this mercator projection
      */
     int m_TileSize;
+    double m_Res;
```
tallytalwar: Why public?

tallytalwar:
These are all good fixes.

Can we move the class member defaults to the class definition instead of the constructor?

core/src/util/mapProjection.h
hjanetzek: So far it's only used for internal calculations within MercatorProjection. If it will be certainly shared by all MapProjections, then can be moved it back.

hjanetzek:
@tallytalwar I have no strong opinion on that. It seems to be more common to initialize in constructors

tallytalwar:
I think with cpp11/14 the trend is moving towards setting defaults in the class definition. I think we have been (trying) to follow the same.

core/src/util/mapProjection.h
tallytalwar: :+1:

blair1618:
I'm kinda conflicted about where to initialize members with default values. I had started to prefer initializing in the class declaration because it's more compact and allows the default values to be part of  the header as documentation. However this means that a chance as small as tuning a number can trigger long recompilations due to other files including the header. 

That said, I don't feel a need to be obsessive about always using one syntax. These fixes look good to me!

PR #267, opened 2015-08-31
Dynamic non-tiled map data

blair1618:
*Dots on maps! Lines on maps! Wonders abundant!*

These changes incorporate the very handy [geojson-vt-cpp](https://github.com/mapbox/geojson-vt-cpp) library from the good folks at Mapbox to enable two new ways of putting data on tangram maps:

__1. Add geometry from a file of GeoJSON:__

In the scene file, you can declare a data source from a non-tiled file location.
```yaml
sources:
    my-data-file:
        type: GeoJSON
        url: path/to/my-data-file.json
```
The geometry from `my-data-file.json` will be added to the map and styled according to the scene file. 

The usual `layer` property of the data block is ignored - all data from this source is one layer. 

The `url` is currently limited to files in the bundled resources of the application. 

__2. Add points, lines, and polygons through C++ and Java interfaces__

You can also create a data source at run-time in C++
```c++
int data_source_id = Tangram::addDataSource("my-cool-data");
```
or Java
```java
MapData data = new MapData("my-cool-data");
```
Then add geometry however you fancy, in C++
```c++
double coords[2] = { coord_x, coord_y };
Tangram::addSourcePoint(data_source_id, coords);
```
or Java
```java
LngLat point = new LngLat(coord_x, coord_y);
data.addPoint(point);
```
Again, data will be styled according to the scene file. The `source` property of a layer must match the name provided when creating the data source and the `layer` property is, again, ignored. 

Some notes:
 - `MapData` is unsafe to instantiate before the Tangram MapView is initialized, which is not a well-defined point in the application life-cycle.  You can see from the current demo that I've just lazily instantiated it for now. I have in mind a way to remove this limitation, it'll just take a bit more time. 
 - There are several ways to further optimize the usage of geojson-vt-cpp. I've left these for later in favor of making these features available as quickly as possible. 
 - It could also be possible to add string properties to geometry that is programmatically added to the map (e.g. for filtering within the scene file) but this isn't implemented right now (again, for expediency).

@ecgreb I'd be interested in feedback on the Android interface (i.e. [MapData](https://github.com/tangrams/tangram-es/blob/7df7bb40966190d194a9aba606a304e999322d2b/android/tangram/src/com/mapzen/tangram/MapData.java))


android/tangram/src/com/mapzen/tangram/MapData.java
```diff
@@ -0,0 +1,100 @@
...
+        addSourcePoint(id, point);
+        return this;
+    }
+
+    /**
+     * Add a line geometry to this data source
+     * @param line Array of points, where points are as described in {@link #addPoint(double[])}
+     * @return This object, for chaining
+     */
+    public MapData addLine(double[][] line) {
```
ecgreb: While this is certainly correct I wonder if the parameter type will be slightly confusing for developers.

Perhaps we could add an abstraction for (longitude, latitude). This would be similar to the similar to the [LatLng] (https://developers.google.com/android/reference/com/google/android/gms/maps/model/LatLng) class in the Google Maps API. Maybe we could call it `LngLat`?

Then we could make the parameter type `List<LngLat>` instead of a 2d array to make it a little more Java friendly.

This might make it less likely developers would get the array dimensions flipped or order of coordinates reversed.

blair1618: Great idea, I've been wondering the best way to be clear about coordinate order and this sounds like a nice solution.

hjanetzek: As the list will be converted to a one dimensional array anyway, we could support both LngLat *and* LatLng to make it easier for gmaps API users to port their overlays

blair1618: True, since we're iterating over the coordinates anyway it would be simple to flip them if needed. However I feel conflicted about supporting LatLng ordering. We should be aiming for consistency across our data exchange interfaces and the GeoJSON spec is [pretty clear](http://geojson.org/geojson-spec.html#positions) on the ordering that coordinates should follow.

tallytalwar:
Reviewing this now.

core/src/data/clientGeoJsonSource.h
```diff
@@ -0,0 +1,36 @@
...
+
+    virtual bool loadTileData(std::shared_ptr<TileTask>&& _task, TileTaskCb _cb) override;
+    virtual bool getTileData(std::shared_ptr<TileTask>& _task) override;
+    virtual void cancelLoadingTile(const TileID& _tile) override {};
+    virtual void clearData() override;
+
+protected:
+
+    virtual std::shared_ptr<TileData> parse(const Tile& _tile, std::vector<char>& _rawData) const override;
+    Point transformPoint(mapbox::util::geojsonvt::TilePoint pt) const;
```
tallytalwar: Can we have some documentation for these methods?

core/src/data/clientGeoJsonSource.cpp
```diff
@@ -0,0 +1,186 @@
...
+    if (!_url.empty()) {
+        // Load from file
+        const auto& string = stringFromResource(_url.c_str());
+        addData(string);
+    }
+}
+
+void ClientGeoJsonSource::addData(const std::string& _data) {
+
+    m_features = geojsonvt::GeoJSONVT::convertFeatures(_data);
```
tallytalwar: Nothing on tangram side, but was looking at GeoJSONVT code (specially convert code), I think there are a bunch of cpp improvement we can do there. We can definitely contribute back to the library once we are done with some high priority tasks.

core/src/data/clientGeoJsonSource.cpp
```diff
@@ -0,0 +1,186 @@
...
+bool ClientGeoJsonSource::getTileData(std::shared_ptr<TileTask>& _task) {
+
+    return true;
+}
+
+void ClientGeoJsonSource::clearData() {
+
+    m_features.clear();
+    m_store.reset();
+    DataSource::clearData();
```
tallytalwar: DataSource destructor calls DataSource::clearData(), don't need to do it explicitly.

core/src/data/clientGeoJsonSource.cpp
blair1618: Yes there are a few things I'd like to submit to their repo :)

core/src/data/clientGeoJsonSource.cpp
blair1618: This is for just clearing the data (not at destruction time) - for instance if you want to plot a new route or set of search results.

tallytalwar: Oh yes.
DataSource::clearData() is clearing any cached data, which is nothing in this case, correct?

blair1618: I believe tile data will be cached for any subclass of `DataSource`.  `DataSource::onTileLoaded` stores tile data in a cache so `ClientGeoJsonSource` needs to explicitly clear it.

tallytalwar: Correct `DataSource::onTileLoaded` stored tile data in a cache, but `ClientGeoJsonSource` does not use `DataSource::onTileLoaded` (which is called from `DataSource::loadTileData`).

`ClientGeoJsonSource` has its own implementation of `loadTileData`.

android/tangram/src/com/mapzen/tangram/MapData.java
```diff
@@ -0,0 +1,102 @@
...
+        addSourceGeoJSON(id, data);
+        return this;
+    }
+
+    /**
+     * Add a point geometry to this data source
+     * @param point LngLat with the coordinates of the point
+     * @return This object, for chaining
+     */
+    public MapData addPoint(LngLat point) {
```
blair1618: Thinking ahead to use cases beyond simply drawing a feature on a map, we'll want to expand the data-adding functionality to include metadata about the geometry, e.g. a name, id, or multiple properties. We'll need this when we implement feature selection so that when we return the selected feature we can attach some identifying information beyond just the geometry itself. To that end, one solution would be to allow an optional string as the final argument to `addPoint`, `addLine`, and `addPolygon` which would then be interpreted as the "properties" section of a GeoJSON feature. @ecgreb What do you think about this? Is there an approach that would make things easier for the use cases that you foresee? cc @bcamper @tallytalwar

ecgreb: With regards to search results we will only need the index of the selected feature in the initial version. If the order of points added is preserved this is enough.

If the order is not preserved we could include the index (or a feature id) as an additional argument.

core/src/data/clientGeoJsonSource.cpp
```diff
@@ -0,0 +1,186 @@
...
+                                              container.members);
+
+    m_features.push_back(feature);
+    m_store = std::make_unique<GeoJSONVT>(m_features, maxZoom, indexMaxZoom, indexMaxPoints, tolerance);
+
+}
+
+void ClientGeoJsonSource::addLine(double* _coords, int _lineLength) {
+    
+    std::vector<geojsonvt::LonLat> line(_lineLength, { 0, 0 });
```
tallytalwar: Referring [this](http://en.cppreference.com/w/cpp/container/vector/vector), its better to use default inserted values instead of explicitly, since it saves copies, but...
It would have been good if `geojsonvt::LonLat` provided default constructor. (another todo for the geojsonvt library).

tallytalwar: Or will it make sense to do a reserve, instead of constructing a `_lineLength` sized vector with {0, 0} `LonLat` values?

blair1618: Yeah, can't default initialize this list because of the missing constructor for `geojsonvt::LonLat` (I can understand their intent with that, they probably won't change it). The choice is between allocating once and copying values or allocating twice with reserve - I wouldn't worry about things like this until we measure them to be a problem. There are much greater inefficiencies here that I would address first.

tallytalwar: Makes sense. :+1:

core/src/data/clientGeoJsonSource.cpp
```diff
@@ -0,0 +1,186 @@
...
+
+void ClientGeoJsonSource::addPoint(double* _coords) {
+
+    auto container = geojsonvt::Convert::project({ geojsonvt::LonLat(_coords[0], _coords[1]) }, tolerance);
+
+    auto feature = geojsonvt::Convert::create(geojsonvt::Tags(),
+                                              geojsonvt::ProjectedFeatureType::Point,
+                                              container.members);
+
+    m_features.push_back(feature);
```
tallytalwar: can we move the feature here, just like its done for other `add*` methods.

blair1618: Yep!

core/src/data/clientGeoJsonSource.cpp
```diff
@@ -0,0 +1,186 @@
...
+
+    auto data = std::make_shared<TileData>();
+
+    auto id = _tile.getID();
+
+    auto tile = m_store->getTile(id.z, id.x, id.y); // uses a mutex lock internally for thread-safety
+
+    Layer layer(""); // empty name will skip filtering by 'collection'
+
+    for (auto& it : tile.features) {
```
tallytalwar: For code readability will it make sense have mapbox namespace here to distinguish geojsonvt ProjectedFeature vs our (read Tangram) Features?
Just a thought, not too strong on it.

core/src/data/clientGeoJsonSource.cpp
```diff
@@ -0,0 +1,186 @@
...
+
+        const auto& geom = it.tileGeometry;
+        const auto type = it.type;
+
+        switch (type) {
+            case geojsonvt::TileFeatureType::Point: {
+                feat.geometryType = GeometryType::points;
+                for (const auto& pt : geom) {
+                    const auto& point = pt.get<geojsonvt::TilePoint>();
+                    feat.points.push_back(transformPoint(point));
```
tallytalwar: how about doing a `std::move(transformPoint(point))`? will save a copy of the point.

android/tangram/src/com/mapzen/tangram/MapData.java
blair1618: We want to be pretty generic about the data that can be associated with a piece of geometry, but now that I think about it using something as simple as an integer identifier wouldn't necessarily constrain us in that regard. Instead of requiring that the metadata "live" inside tangram, we let the users store the metadata in whatever way is most convenient and when they select a feature, we simply provide an identifier that enables them to query metadata themselves. This would probably be far more convenient for developers, but it diverges a bit from the way that we've been thinking about the results of a feature query.

core/src/tile/tileManager.cpp
```diff
@@ -39,8 +40,18 @@ TileManager::~TileManager() {
...
-void TileManager::addDataSource(std::shared_ptr<DataSource>&& dataSource) {
+int32_t TileManager::addDataSource(std::shared_ptr<DataSource>&& dataSource) {
     m_tileSets.push_back({++m_tileSetSerial, dataSource});
+    return m_tileSetSerial;
+}
+
+std::shared_ptr<ClientGeoJsonSource> TileManager::getClientSourceById(int32_t _id) {
+    for (auto& set : m_tileSets) {
+        if (set.id == _id) {
+            return std::dynamic_pointer_cast<ClientGeoJsonSource>(set.source);
```
tallytalwar: instead of creating another shared_ptr here, we can return a const reference of the shared DataSource ptr. Something like the following diff:

core/src/tangram.cpp:
```
 void addSourcePoint(int _sourceId, double* _coords) {
-    auto source = m_tileManager->getClientSourceById(_sourceId);
+    const auto& source = std::dynamic_pointer_cast<ClientGeoJsonSource>(m_tileManager->getClientSourceById(_sourceId));
 
 void addSourceLine(int _sourceId, double* _coords, int _lineLength) {
-    auto source = m_tileManager->getClientSourceById(_sourceId);
+    const auto& source = std::dynamic_pointer_cast<ClientGeoJsonSource>(m_tileManager->getClientSourceById(_sourceId));
 
 void addSourcePoly(int _sourceId, double* _coords, int* _ringLengths, int _rings) {
-    auto source = m_tileManager->getClientSourceById(_sourceId);
+    const auto& source = std::dynamic_pointer_cast<ClientGeoJsonSource>(m_tileManager->getClientSourceById(_sourceId));
 
 void addSourceGeoJSON(int _sourceId, const char* _data) {
-    auto source = m_tileManager->getClientSourceById(_sourceId);
+    const auto& source = std::dynamic_pointer_cast<ClientGeoJsonSource>(m_tileManager->getClientSourceById(_sourceId));
```

core/src/tile/tileManager.cpp
```
-std::shared_ptr<ClientGeoJsonSource> TileManager::getClientSourceById(int32_t _id) {
+const std::shared_ptr<DataSource>& TileManager::getClientSourceById(int32_t _id) {
     for (auto& set : m_tileSets) {
         if (set.id == _id) {
-            return std::dynamic_pointer_cast<ClientGeoJsonSource>(set.source);
+            return set.source;
         }
     }
     return nullptr;
```

core/src/tile/tileManager.cpp
``` 
-    std::shared_ptr<ClientGeoJsonSource> getClientSourceById(int32_t _id);
+    const std::shared_ptr<DataSource>& getClientSourceById(int32_t _id);
```

android/tangram/src/com/mapzen/tangram/MapData.java
```diff
@@ -0,0 +1,102 @@
...
+        return name;
+    }
+
+    /**
+     * Remove all data from this source
+     * @return This object, for chaining
+     */
+    public MapData clear() {
+        clearSourceData(id);
+        return this;
```
tallytalwar: Why for chaining?

android/tangram/src/com/mapzen/tangram/MapData.java
tallytalwar: Based on the code so far, we are using the client supplied data in order, so for simple cases like search results it should be not be hard to maintain this "id" metadata, which is basically the index of data.

@ecgreb Can there be other cases where its not trivial enough and we need to maintain an explicit unique id with every data point?

android/tangram/src/com/mapzen/tangram/MapData.java
```diff
@@ -0,0 +1,102 @@
...
+import java.util.List;
+
+public class MapData {
+
+    /**
+     * Construct a new MapData object for adding drawable data to the map
+     * @param name Name of the data source in the scene file for styling this object's data
+     */
+    public MapData(String name) {
+        this.name = name;
```
tallytalwar: Can we remove the use of `this` here, unless its a Java way of doing things. 
To me since id and name are private class members its implicit, and rename the argument from `name` to something else.

blair1618: This is a pretty common Java pattern, see:
https://github.com/mapzen/eraser-map/blob/master/app/src/main/java/com/mapzen/erasermap/AndroidModule.java#L25,
https://github.com/mapzen/eraser-map/blob/master/app/src/main/kotlin/com/mapzen/erasermap/presenter/MainPresenterImpl.kt#L53

core/src/data/clientGeoJsonSource.cpp
blair1618: The result of `transformPoint` only resides on the stack, `std::move` can't prevent a copy

android/tangram/src/com/mapzen/tangram/MapData.java
tallytalwar: :+1:

android/tangram/src/com/mapzen/tangram/MapData.java
blair1618:
```java
mapData.clear()
       .addPoint(p0)
       .addPoint(p1);
```

Some people prefer chain-style calls. It's not like we have anything better to return :P

tallytalwar: :+1:

core/src/data/clientGeoJsonSource.cpp
tallytalwar: ohh yes ... didn't see the implementation for that when I was checking this .. my bad.

tallytalwar:
- Have some unanswered questions/queries
- master rebase 
- possible feature to add id/index to client data points. 

Once these are done, this is good to merge.

core/src/tile/tileManager.cpp
blair1618: Evidently `dynamic_pointer_cast` creates a new `shared_ptr` too, so this is probably fine the way it is.

tallytalwar: :+1:

PR #268, opened 2015-09-01
Earcut tessellator

hjanetzek:


tallytalwar:
screen shot 2015-09-01 at 10 10 09 am
This is what I see on iOS with earcut. Will update with more info, as I debug this further.

hjanetzek:
fixed with last commit

tallytalwar:
@hjanetzek do you plan to do thread local storage implementation here like you did in the https://github.com/tangrams/tangram-es/tree/exp-earcutting branch?

hjanetzek:
@tallytalwar c++'s thread_local seems not to be supported by android yet (tried with clang 3.5 and 3.6). We could enable it for other platforms though or use pthread's thread local storage. But this could go into a separate branch IMO, together with looking for other places to reduce temporary allocations during tile building.

blair1618:
@hjanetzek That makes sense to me. This branch is looking good, will merge!

core/src/util/builders.cpp
```diff
@@ -44,65 +35,54 @@ JoinTypes JoinTypeFromString(const std::string& str) {
...
-            for (auto& p : line) {
-                bbox.include(p.x, p.y);
-            }
+    uint16_t vertexDataOffset = _ctx.numVertices;
+
+    if (vertexDataOffset == 0) {
+        _ctx.indices = std::move(earcut.indices);
+    } else {
+        _ctx.indices.reserve(_ctx.indices.size() +  earcut.indices.size());
+        for (auto i : earcut.indices) {
```
tallytalwar: use reference here. Will push it.

blair1618: A reference would actually be larger than a `uint16_t`

If we cared about optimizing this an `insert` may be faster or even a `memcpy` but it's a trivial amount of computation.

hjanetzek: I thought of using std::transform here. But in any case the else branch
will not be taken atm as I moved the extrusion building to come after
tessellation :)

On Wed, Sep 2, 2015 at 11:46 PM, Matt Blair <notifications@github.com>
wrote:

> In core/src/util/builders.cpp
> <https://github.com/tangrams/tangram-es/pull/268#discussion_r38588850>:
>
> >
> > -    // add polygon contour for every ring
> > -    for (auto& line : _polygon) {
> > -        if (_ctx.useTexCoords) {
> > -            for (auto& p : line) {
> > -                bbox.include(p.x, p.y);
> > -            }
> > +    uint16_t vertexDataOffset = _ctx.numVertices;
> > +
> > +    if (vertexDataOffset == 0) {
> > +        _ctx.indices = std::move(earcut.indices);
> > +    } else {
> > +        _ctx.indices.reserve(_ctx.indices.size() +  earcut.indices.size());
> > +        for (auto i : earcut.indices) {
>
> A reference would actually be larger than a uint16_t
>
> If we cared about optimizing this an insert may be faster or even a memcpy
> but it's a trivial amount of computation.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/tangrams/tangram-es/pull/268/files#r38588850>.
>



-- 
Hannes

hjanetzek: If we integrated extrusion building and tessellation into one function we could preallocate the (likely) whole indices and vertices vectors. This would reduce one copy of the tessellated vertices and indices for extruded polygons

blair1618: Good call, that would be an easy change to make too!

PR #269, opened 2015-09-02
Initialize Tangram before GL context creation

blair1618:
This is important primarily for Android, where the time at which the GL context is created is not necessarily the same as the time of application creation. Previously I had to store some state variables in `MapController` and wait until `MapView` called `onSurfaceCreated` before I could make any native calls. The Android library still has issues around the idea of a tangram "instance" (i.e. there's no such thing) but this helps a bit.

### DIFF
--- a/android/tangram/src/com/mapzen/tangram/MapController.java
+++ b/android/tangram/src/com/mapzen/tangram/MapController.java
@@ -334,6 +334,8 @@ public class MapController implements Renderer, OnTouchListener, OnScaleGestureL
 
     public void onSurfaceCreated(GL10 gl, EGLConfig config) {
         init(this, assetManager, scenePath);
+        // init() is safe to call twice, this invocation ensures that the jni
+        // environment is attached to the rendering thread
         setupGL();
     }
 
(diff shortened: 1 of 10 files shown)


tallytalwar:
:+1:  good to go.

android/tangram/src/com/mapzen/tangram/MapController.java
```diff
@@ -88,6 +88,8 @@ public MapController(Activity mainApp, MapView view, String sceneFilePath) {
         view.setRenderer(this);
         view.setRenderMode(GLSurfaceView.RENDERMODE_WHEN_DIRTY);
 
+        init(this, assetManager, scenePath);
```
tallytalwar: Actually can you put a comment here explaining init call being called twice? We had a discussion on this, but good to have a comment in the code.

core/src/tangram.cpp
```diff
@@ -69,12 +69,8 @@ void initialize(const char* _scenePath) {
 
         loader.loadScene(sceneString, *m_scene, *m_tileManager, *m_view);
 
-        Primitives::setColor(0xffffff);
-        RenderState::configure();
     }
 
-    while (Error::hadGlError("Tangram::initialize()")) {}
```
tallytalwar: @karimnaaji is going to put a PR for GL_CHECK macro he has in place in fontstash repo.

Because something like is helpful to find gl errors (as was the case on the S3 issue).

blair1618: Yeah, as a matter of fact I'll add an error check in `setupGL()`. It's a useful thing to have, we just can't use `glGetError()` in `initialize()` because we may not have a GL context at that point.

Issue #252, opened 2015-08-21
Random crash on FontContext::lock() on app exit.


tallytalwar:
Note: Noticed on deviceFonts-android and text-color branch. Try to repro this on master.

Both the feature branches are modifying any of the code path responsible to call FontContext::lock.

Also, if reproduced update with the stacktrace.

hjanetzek:
On a first look it seems the order of destructor calls for TileManager and FontContext is implicitly given by the order they appear in tangram.cpp - So the global FontContext could be destroyed while TileManager is still waiting for it's workers to shut down.

 http://stackoverflow.com/questions/469597/destruction-order-of-static-objects-in-c

tallytalwar:
I just removed m_ftContext in deviceFonts-android feature branch, because now its not needed there. I have not been able to recreate it, but its random :|.

karimnaaji:
Any related crashes since the merge on this issue?

tallytalwar:
Nothing I am aware of. We can mark it resolved for the time being.

karimnaaji:
Ok, there is other issues but it doesn't seem to be related to this specific one.

PR #270, opened 2015-09-03
Fix dead lock from pending threads

karimnaaji:
Fixes: #252


tallytalwar:
:+1: . Looks good to me.

karimnaaji:
I noticed another issue not related, fontContext can have a single reference from threads and and be deleted from there, which involve atlas texture deletion and a bad access. I'll open an issue for that.

PR #271, opened 2015-09-03
Blend modes

blair1618:
This enables blend modes configured per-style. 

Note that the `inlay` blend mode won't behave correctly for text and sprites because they don't render in the same 3D coordinate space as other geometry; this can be addressed separately.

### DIFF
--- a/core/src/style/style.h
+++ b/core/src/style/style.h
@@ -113,7 +113,7 @@ public:
     void viewportHasChanged() { m_dirtyViewport = true; }
 
     /* Whether or not the style uses blending operation for drawing */
-    virtual bool isOpaque() const { return true; };
+    bool isOpaque() const { return m_blend == Blending::none; };
 
     /* Make this style ready to be used (call after all needed properties are set) */
     virtual void build(const std::vector<std::unique_ptr<Light>>& _lights);
--- a/core/src/style/style.h
+++ b/core/src/style/style.h
@@ -28,6 +28,14 @@ enum class LightingType : char {
     fragment
 };
 
+enum class Blending : char {
+    none,
+    add,
+    multiply,
+    overlay,
+    inlay,
+};
+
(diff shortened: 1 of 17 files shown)


karimnaaji:
Looks good!

core/src/style/style.h
```diff
@@ -28,6 +28,14 @@ enum class LightingType : char {
     fragment
 };
 
+enum class Blending : char {
```
tallytalwar: do you make this char to reduce memory footprint of the enum class?

blair1618: yep, doesn't need to be any larger than `char`

core/src/style/textStyle.cpp
```diff
@@ -175,10 +174,9 @@ void TextStyle::onBeginDrawFrame(const View& _view, const Scene& _scene) {
         m_shaderProgram->setUniformMatrix4f("u_proj", glm::value_ptr(_view.getOrthoViewportMatrix()));
         m_dirtyViewport = false;
     }
+    
+    Style::onBeginDrawFrame(_view, _scene);
```
tallytalwar: TextStyle does not defines a material and text shaders do not define u_zoom uniform, calling Style::onBeginDrawFrame will also call these.

core/src/style/style.h
tallytalwar: :+1: nice.

core/src/style/textStyle.cpp
blair1618: All styles have a default "empty" material, so calling a setup function on it is slightly wasteful, but not harmful. `TextStyle` and its brethren don't *yet* define `u_zoom`, but they will eventually because it must be made available for custom shader blocks.

tallytalwar: :+1:

PR #273, opened 2015-09-04
Priority

karimnaaji:
Numeric label priorities, as implemented in tangram JS

### DIFF
--- a/core/src/labels/labels.cpp
+++ b/core/src/labels/labels.cpp
@@ -82,7 +82,6 @@ void Labels::update(const View& _view, float _dt, const std::vector<std::unique_
         if (intersect(l1->getOBB(), l2->getOBB())) { occlusions.insert({l1, l2}); }
     }
 
-    // no priorities, only occlude one of the two occluded label
     for (auto& pair : occlusions) {
         if (!pair.first->occludedLastFrame() || !pair.second->occludedLastFrame()) {
             // lower numeric priority means higher priority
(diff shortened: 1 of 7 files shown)


core/src/style/textStyle.cpp
```diff
@@ -87,10 +96,12 @@ void TextStyle::buildPoint(const Point& _point, const DrawRule& _rule, const Pro
     const auto& text = _props.getString(key_name);
     if (text.length() == 0) { return; }
 
-    Label::Options options;
-    options.color = params.fill;
 
-    buffer.addLabel(text, { glm::vec2(_point), glm::vec2(_point) }, Label::Type::point, params, options);
+    if (Tangram::getDebugFlag(Tangram::DebugFlags::labels)) {
+        buffer.addLabel(std::to_string(params.priority), { glm::vec2(_point), glm::vec2(_point) }, Label::Type::debug, params, optionsFromTextParams(params));
```
tallytalwar: Is it possible to include this in debugTextStyle?

core/src/labels/labels.cpp
```diff
@@ -84,14 +84,14 @@ void Labels::update(const View& _view, float _dt, const std::vector<std::unique_
 
     // no priorities, only occlude one of the two occluded label
```
bcamper: outdated comment?

core/src/style/textStyle.cpp
karimnaaji: I don't think so, since this needs the feature points/lines etc. Not sure if this would be useful for others to debug though, I was using this only to check if options were good, and too see how many labels with high priority were rejected so this can go away.

tallytalwar: :+1:

tallytalwar:
This is good to go. Merging.

PR #274, opened 2015-09-04
Label offsets

karimnaaji:
Mostly useful for points for now, line offsets will behave weirdly (especially when tilting, rotating), will follow up on more advanced label line offsets once we decided of the right convention.

### DIFF
--- a/core/resources/scene/scene.yaml
+++ b/core/resources/scene/scene.yaml
@@ -169,7 +169,7 @@ layers:
             filter: { $zoom: { min: 15 } }
             draw:
                 heightglow:
-                    extrude: [0m, 1.2m]
+                    extrude: true
         high-line:
             filter: {roof_material: grass}
             draw:
@@ -193,7 +193,7 @@ layers:
             draw:
                 text:
                     visible: true
-                    offset: [0px, 10px]
+                    offset: [0px, 100px]
                     font:
                         family: sans-serif
                         weight: 400
--- a/core/resources/scene/scene.yaml
+++ b/core/resources/scene/scene.yaml
@@ -10,7 +10,7 @@ lights:
         origin: world
         direction: [1, 1, -1]
         diffuse: [.3, .3, .3, 1.]
-        ambient: [0.6, 0.6, 0.6, 1.]
+        ambient: [0.7, 0.7, 0.7, 1.]
 
 textures:
     water-texture:
@@ -104,10 +104,10 @@ layers:
                 width: .02
                 cap: round
                 color: white
-                outline:
-                    width: .01
-                    cap: round
-                    color: '#999'
+#outline:
+#                    width: .01
+#                    cap: round
+#                    color: '#999'
         rounded:
             filter: { $zoom: { min: 18 } }
             draw:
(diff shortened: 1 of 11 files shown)


core/src/scene/drawRule.h
```diff
@@ -13,17 +13,18 @@
...
 enum class StyleParamKey : uint8_t {
     none, order, extrude, color, width, cap, join, outline_color, outline_width, outline_cap, outline_join,
     font_family, font_weight, font_style, font_size, font_fill, font_stroke, font_stroke_color, font_stroke_width, font_uppercase,
-    visible, priority
+    visible, priority, offset
 };
 
 struct StyleParam {
-    using Value = variant<none_type, std::string, CapTypes, JoinTypes, Extrusion, int32_t, uint32_t, float, bool>;
+    using Value = variant<none_type, std::string, CapTypes, JoinTypes, Extrusion, Offset, int32_t, uint32_t, float, bool>;
```
blair1618: Might make sense to combine the `Extrusion` and `Offset` variations into a single `std::pair<float, float>` variation, to reduce the generated code size a bit.

core/resources/scene/scene.yaml
```diff
@@ -104,10 +104,10 @@ layers:
                 width: .02
                 cap: round
                 color: white
-                outline:
-                    width: .01
-                    cap: round
-                    color: '#999'
+#outline:
```
tallytalwar: Was this for debugging? We can put this back!

core/src/scene/sceneLoader.cpp
```diff
@@ -64,7 +64,11 @@ std::string parseSequence(const Node& node) {
         try {
             sstream << val.as<float>() << ",";
         } catch (const BadConversion& e) {
-            logMsg("Error: Float value was expected for styleParam sequence value\n");
+            try {
+                sstream << val.as<std::string>() << ",";
+            } catch (const BadConversion& e) {
+                logMsg("Error: Float or Unit expected for styleParam sequence value\n");
```
tallytalwar: I think any yaml scalar value can be converted to string, so not sure if it makes sense to have a try catch for `val.as<std::string>()`

tallytalwar:
Some trivial suggestions in comments, apart from that looks good.

core/src/scene/drawRule.cpp
```diff
@@ -228,6 +234,41 @@ bool DrawRule::operator<(const DrawRule& _rhs) const {
...
+        return false;
+    }
+
+    std::string value = _value;
+
+    // replace all unit occurences
+    for (auto& unit : _allowedUnit) {
+        auto i = value.find(unit);
+        // TODO: conversion
+        while (i != std::string::npos) {
```
karimnaaji: Yes!

karimnaaji: This step is only replacement to have a clean string though, I added an early exit if this step fails.

core/src/scene/sceneLoader.cpp
karimnaaji: But what if you put another sequence in this sequence?

tallytalwar: Thats true.

tallytalwar:
This one is also good to merge, once rebased with master.

PR #275, opened 2015-09-04
Linked labels [proposal]

karimnaaji:
A potential option for labels of different types representing the same location.

Linking labels together help to make them collide whenever any of the linked label collide with any other label.

merge-labels


karimnaaji:
Closing until we put more thoughts on this feature.

PR #276, opened 2015-09-04
Track shaders

karimnaaji:
Track shader in render states, the same way the other gl components are tracked.


blair1618:
Nice, looks good to me.

PR #277, opened 2015-09-04
Text transform options

karimnaaji:
Add `capitalize`, `lowercase` text-transform options.

### DIFF
--- a/core/src/text/textBuffer.cpp
+++ b/core/src/text/textBuffer.cpp
@@ -16,7 +16,7 @@ TextBuffer::~TextBuffer() {
 }
 
 bool TextBuffer::addLabel(const std::string& _text, Label::Transform _transform, Label::Type _type, const Parameters& _params, Label::Options _options) {
-    if (_params.fontSize <= 0.f) {
+    if (_params.fontSize <= 0.f || _text.size() == 0) {
         return false;
     }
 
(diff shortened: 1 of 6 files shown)


karimnaaji:
Right, much simpler and no conflicts

core/src/text/textBuffer.cpp
```diff
@@ -33,18 +33,45 @@ bool TextBuffer::addLabel(const std::string& _text, Label::Transform _transform,
...
+    if (_params.transform != TextTransform::none) {
+        transformedText = _text;
         std::locale loc;
-        for (std::string::size_type i = 0; i < _text.length(); ++i) {
-            text[i] = std::toupper(_text[i], loc);
+
+        // perfom text transforms
+        switch (_params.transform) {
+            case TextTransform::capitalize:
+                transformedText[0] = toupper(_text[0], loc);
```
tallytalwar: what if a text starts with a whitespace ... possible with osm data?

karimnaaji: It would toupper on a whitespace, which I guess does nothing. Noticed it's missing `std::`.

tallytalwar: :+1:

tallytalwar:
looks good.

tallytalwar:
Looks good, once rebased can be merged.

PR #280, opened 2015-09-09
Drawrule tweaks

blair1618:
Cleanup!


core/src/scene/styleContext.h
```diff
@@ -3,6 +3,7 @@
 #include "duktape.h"
 #include "tileData.h"
 #include "scene/drawRule.h"
+#include "scene/styleParam.h"
```
tallytalwar: We can do with a forward declaration here.

PR #281, opened 2015-09-09
Text source option

karimnaaji:
As descibed in [tangram docs](https://github.com/tangrams/tangram-docs/blob/c834ed2013fc6d5734acb9eb0c7028066533a01b/pages/draw.md#text_source).

- [x] text source as a js function
- [x] text source as a non-js function

### DIFF
--- a/core/src/style/textStyle.cpp
+++ b/core/src/style/textStyle.cpp
@@ -12,6 +12,9 @@
 namespace Tangram {
 
 const static std::string key_name("name");
+const static std::string uppercase("uppercase");
+const static std::string lowercase("lowercase");
+const static std::string capitalize("capitalize");
 
 TextStyle::TextStyle(std::string _name, bool _sdf, bool _sdfMultisampling, Blending _blendMode, GLenum _drawMode) :
     Style(_name, _blendMode, _drawMode), m_sdf(_sdf), m_sdfMultisampling(_sdfMultisampling) {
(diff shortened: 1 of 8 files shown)


core/src/style/textStyle.cpp
```diff
@@ -67,12 +70,15 @@ Parameters TextStyle::parseRule(const DrawRule& _rule) const {
...
-    if (transform == "capitalize") {
+    if (transform == capitalize) {
         p.transform = TextTransform::capitalize;
-    } else if (transform == "lowercase") {
+    } else if (transform == lowercase) {
         p.transform = TextTransform::lowercase;
-    } else if (transform == "uppercase") {
+    } else if (transform == uppercase) {
         p.transform = TextTransform::uppercase;
     }
```
hjanetzek: string== has an overload for char* comparison, so here the heap allocated static strings wouldn't be needed. But I would like to move the transform string-to-enum conversion to StyleParam::parseString.

hjanetzek:
Ready to merge when the scene.yaml test code is reverted

core/src/style/textStyle.cpp
karimnaaji: Yeah it makes sense to do that to reduce those comparisons.

tallytalwar: Yes, it make sense to move this to StyleParam::parseString.

For this PR however, can we get away with the static strings and directly compare with "capitalized", etc.
We can have a separate small PR for moving string-to-enum conversion to StyleParam::parseString.

karimnaaji: :+1:

core/src/style/textStyle.cpp
```diff
@@ -95,6 +96,20 @@ Label::Options TextStyle::optionsFromTextParams(const Parameters& _params) const
     return options;
 }
 
+const std::string& TextStyle::applyTextSource(const Parameters& _parameters, const Properties& _props) const {
+
+    if (_parameters.textSource.text.empty()) {
+        // Default: use 'name' property
+        return _props.getString("name");
+    }
```
hjanetzek: Oh, this test must also check for !_parameters.textSource.isFunction, because the function could have returned an empty string

hjanetzek:
rebasing

tallytalwar:
ohh almost done with that ...

hjanetzek:
first :)

tallytalwar:
:D

Varun Talwar
Mailto: varun@mapzen.com

PR #282, opened 2015-09-10
Point style

karimnaaji:
When no sprite atlas is provided, the sprite/point style should draw a dot/circle at the specified point location.

screen shot 2015-09-10 at 2 22 54 pm

Todos: 
- [x] draw dots when no sprite atlas is provided
- [x] give the ability to use a texture for the points
- [x] rename `spriteStyle` -> `pointStyle`,`sprite.fs` -> `point.fs` for consistency with webgl
- [x] add support for polygon, line


nvkelso:
Is there also an ability to turn off the default drawing of dots when no sprite atlas is defined? Some features (like states and country labels) shouldn't have icons associated with them.

bcamper:
I think that is already covered @nvkelso, but to match JS behavior:

- Points require either 
- Points won't render without `size` and `color` properties defined. (POIs usually don't have 
-

bcamper:
I think that is already covered @nvkelso, but to match JS behavior:

- Point-based styles require either a `color` (at the feature level) or `texture` (at the style level) to be defined in order to render. If there's a `texture` defined, it's a sprite, if there's not, it's a plain point.
- `size` property is used to set both sprite and plain point size.
  - `size` is always specified in pixels (should this be expanded? hasn't been a need).
  - Default `size` is `[32px, 32px]`. (Does this make sense?)
- `color` property determines:
   - The color of a plain point.
   - A "tint" value for a sprite, e.g. by setting `color: [0, 1, 0]` on a sprite, you can pull out just the green channel. (There has not been much need for this as of yet, but it was a natural consequence of the shared shaders between sprites and points.)
- For sprites:
  - If the `texture` does **not** have a sprite atlas defined, then it's assumed to be a single complete image. (This is not the typical case for POIs, but it does happen.)
  - If the `texture` **does** have a sprite atlas defined, then a valid sprite property is required to render.
    - `sprite` property is checked first. It can be an explicit sprite name, or a JS function.
    - If `sprite` property isn't present or doesn't evaluate to a valid sprite, then `sprite_default` property is checked.
    - If neither evaluates to a valid sprite, nothing is drawn.

nvkelso:
Great, sounds good!

karimnaaji:
Thanks @bcamper for the complete description. I'll follow up to match some of the missing behavior here!

core/src/style/textStyle.cpp
```diff
@@ -138,55 +138,17 @@ void TextStyle::buildLine(const Line& _line, const DrawRule& _rule, const Proper
...
-            centroid.y += p.y;
-            n++;
-        }
-    }
-    if (n == 0) { return; }
-
-    centroid /= n;
-
-    buffer.addLabel(text, { centroid, centroid }, Label::Type::point, params, optionsFromTextParams(params));
+    Point p = glm::vec3(centroid(_polygon), 0.0);
```
blair1618: In tangram-js, drawing a polygon geometry with a 'points' style seems to render points at each of the vertices of the polygon. Centroid could make sense too, but we should aim at parity for now.

karimnaaji: Oh right, actually it seems to treat both cases coming with a default on each vertices of the polygon, coming from a style property, but haven't seen anything documented on this. I'll make it match the default.

karimnaaji: https://github.com/tangrams/tangram/blob/master/src/styles/points/points.js#L212

blair1618: Ahh well that is legitimately confusing haha. I'll see about amending tangram-docs. If it's actually intended to be a style property then we can support that as well.

karimnaaji:
After this gets reviewed I'll undo the test case modification on the stylesheet.

blair1618:
Looks pretty good, but some points seem to be flickering on and off every other render (???). I would guess z-fighting but the vertices should be "above" everything else, right?

karimnaaji:
Checking this, but yeah no z-fighting happen since they're all above so it must be something else.

blair1618:
Ah, better now! Feel free to merge once you've cleaned up the testing code.

karimnaaji:
Yeah, I still have a couple of small other things to fix here, and then will merge.

PR #283, opened 2015-09-10
Style mixing

tallytalwar:
- [x] ORing of properties like `animated` or `texcoords`.
- [x] overWriting of properties like `base`, `lighting`, `texture`, `blend`.
- [x] overWriting of `material` subProperties: `diffuse`, `ambient`, `specular`, `shininess`, `normal`.
- [x] merging of shader node `uniforms` and `defines`
- [x] deep merging of `shader blocks`
- [x] mix `extensions` (should be trivial once injection is implemented)
- [x] mix `global` shader blocks (should happen automatically, since the current implementation is generic)
- [x] Remove test styling (to be done before this branch is merged)

Todo:
- Separate PR coming for injection of uniforms.

### DIFF
--- a/core/src/scene/sceneLoader.cpp
+++ b/core/src/scene/sceneLoader.cpp
@@ -23,6 +23,7 @@
 #include "yaml-cpp/yaml.h"
 
 #include <algorithm>
+#include <unordered_set>
 
 using YAML::Node;
 using YAML::NodeType;
@@ -331,7 +332,7 @@ void SceneLoader::loadTextures(Node textures, Scene& scene) {
     }
 }
 
-void SceneLoader::loadStyleProps(Style& style, YAML::Node styleNode, Scene& scene) {
+void SceneLoader::loadStyleProps(Style* style, YAML::Node styleNode, Scene& scene) {
 
     if (!styleNode) {
         logMsg("Error: Can not parse style parameters, bad style YAML Node\n");
@@ -344,7 +345,7 @@ void SceneLoader::loadStyleProps(Style& style, YAML::Node styleNode, Scene& scen
         if (!animatedNode.IsScalar()) { logMsg("WARNING: animated flag should be a scalar\n"); }
         else {
             try {
-                style.setAnimated(animatedNode.as<bool>());
+                style->setAnimated(animatedNode.as<bool>());
             } catch(const BadConversion& e) {
                 logMsg("WARNING: Expected a boolean value in animated property. Using default (false).\n");
             }
@@ -355,11 +356,11 @@ void SceneLoader::loadStyleProps(Style& style, YAML::Node styleNode, Scene& scen
 
         std::string str = blendNode.as<std::string>();
 
-        if      (str == "none")     { style.setBlendMode(Blending::none); }
-        else if (str == "add")      { style.setBlendMode(Blending::add); }
-        else if (str == "multiply") { style.setBlendMode(Blending::multiply); }
-        else if (str == "overlay")  { style.setBlendMode(Blending::overlay); }
-        else if (str == "inlay")    { style.setBlendMode(Blending::inlay); }
+        if      (str == "none")     { style->setBlendMode(Blending::none); }
+        else if (str == "add")      { style->setBlendMode(Blending::add); }
+        else if (str == "multiply") { style->setBlendMode(Blending::multiply); }
+        else if (str == "overlay")  { style->setBlendMode(Blending::overlay); }
+        else if (str == "inlay")    { style->setBlendMode(Blending::inlay); }
         else { logMsg("WARNING: invalid blend mode \"%s\"\n", str.c_str()); }
 
     }
@@ -376,20 +377,20 @@ void SceneLoader::loadStyleProps(Style& style, YAML::Node styleNode, Scene& scen
 
     Node shadersNode = styleNode["shaders"];
     if (shadersNode) {
-        loadShaderConfig(shadersNode, *(style.getShaderProgram()));
+        loadShaderConfig(shadersNode, *(style->getShaderProgram()));
     }
 
     Node materialNode = styleNode["material"];
     if (materialNode) {
-        loadMaterial(materialNode, *(style.getMaterial()), scene);
+        loadMaterial(materialNode, *(style->getMaterial()), scene);
     }
 
     Node lightingNode = styleNode["lighting"];
     if (lightingNode) {
         std::string lighting = lightingNode.as<std::string>();
-        if (lighting == "fragment") { style.setLightingType(LightingType::fragment); }
-        else if (lighting == "vertex") { style.setLightingType(LightingType::vertex); }
-        else if (lighting == "false") { style.setLightingType(LightingType::none); }
+        if (lighting == "fragment") { style->setLightingType(LightingType::fragment); }
+        else if (lighting == "vertex") { style->setLightingType(LightingType::vertex); }
+        else if (lighting == "false") { style->setLightingType(LightingType::none); }
         else if (lighting == "true") { } // use default lighting
         else { logMsg("WARNING: unrecognized lighting type \"%s\"\n", lighting.c_str()); }
     }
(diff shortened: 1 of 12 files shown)


core/src/scene/sceneLoader.cpp
```diff
@@ -330,6 +349,236 @@ void SceneLoader::loadTextures(Node textures, Scene& scene) {
...
+            }
+        }
+	}
+
+    Node urlNode = styleNode["url"];
+    if (urlNode) { logMsg("WARNING: loading style from URL not yet implemented\n"); } // TODO
+
+}
+
+Node SceneLoader::propORing(const std::string& propStr, Node& styles, const MIXES& mixes) {
```
bcamper: Not crazy about the name propORing (keep reading it as O-ring), but also not sure what to call it. Maybe propFlag? Though I suppose there's also nothing inherent in having flags use OR behavior.

core/src/scene/sceneLoader.cpp
```diff
@@ -330,6 +349,236 @@ void SceneLoader::loadTextures(Node textures, Scene& scene) {
...
+
+    Node styleNode;
+
+    // Or`ed style properties
+    for (auto property : {"animated", "texcoords"}) {
+        Node node = propORing(property, styles, mixes);
+        if (!node.IsNull()) { styleNode[property] = node; }
+    }
+
+    // Overwriten style properties (last wins)
```
bcamper: overwritten (typo)

core/src/scene/sceneLoader.cpp
```diff
@@ -330,6 +349,236 @@ void SceneLoader::loadTextures(Node textures, Scene& scene) {
...
+    // material properties merging
+    Node materialNode;
+    for (auto subProp : {"diffuse", "ambient", "specular", "normal", "shininess"}) {
+        Node node = propOverwrite("material", subProp, styles, mixes);
+        if (!node.IsNull()) { materialNode[subProp] = node; }
+    }
+    if (!materialNode.IsNull()) { styleNode["material"] = materialNode; }
+
+
+    // shader prooerties merging
```
bcamper: properties (typo)

core/src/scene/sceneLoader.cpp
```diff
@@ -330,6 +349,236 @@ void SceneLoader::loadTextures(Node textures, Scene& scene) {
...
+    Node shadersNode;
+    // Merge shader uniform and defines (concatinate all)
+    for (auto param : {"defines", "uniform", "extensions"}) {
```
bcamper: `uniforms`, not `uniform`

core/src/scene/sceneLoader.cpp
```diff
@@ -330,6 +349,236 @@ void SceneLoader::loadTextures(Node textures, Scene& scene) {
...
+            style = new PolylineStyle(styleName);
+        }
+        else if (baseString == "text") {
+            style = new TextStyle(styleName, true, true);
+        }
+        else if (baseString == "points") {
+            style = new SpriteStyle(styleName);
+        } else {
+            logMsg("WARNING: base style \"%s\" not recognized, can not instantiate.\n", baseString.c_str());
+            style = new PolygonStyle(styleName);
```
bcamper: Shouldn't instantiate in this case

core/src/scene/sceneLoader.cpp
```diff
@@ -330,6 +349,236 @@ void SceneLoader::loadTextures(Node textures, Scene& scene) {
...
+    if (!blocksNode.IsNull()) { shadersNode["blocks"] = blocksNode; }
+
+    if (!shadersNode.IsNull()) { styleNode["shaders"] = shadersNode; }
+
+
+    // Construct style instance using the merged properties
+    const std::string& styleName = mixes.back();
+    if (Node baseNode = styleNode["base"]) {
+        std::string baseString = baseNode.as<std::string>();
+        if (baseString == "lines") {
```
bcamper: Missing `polygons`?

core/src/scene/sceneLoader.cpp
tallytalwar: Ohh, forgot to take this out, when I updated the logMsg.

core/src/scene/sceneLoader.cpp
```diff
@@ -330,6 +349,236 @@ void SceneLoader::loadTextures(Node textures, Scene& scene) {
...
+    for (auto property : {"base", "lighting", "texture", "blend"}) {
+        Node node = propOverwrite(property, "", styles, mixes);
+        if (!node.IsNull()) { styleNode[property] = node; }
+    }
+
+
+    // material properties merging
+    Node materialNode;
+    for (auto subProp : {"diffuse", "ambient", "specular", "normal", "shininess"}) {
+        Node node = propOverwrite("material", subProp, styles, mixes);
```
bcamper: Some material properties can themselves be maps, to define texture url, mapping, etc. So overwriting properties won't work you need to merge.

However, it also looks like `propMerge` would not handle merging properties more than one level deep. I would think that ideally `propMerge` would be recursive, so it would properly merge structures of any depth. Another case of that would be uniforms with structures, though we also discussed deprecating/not supporting that in ES. Nonetheless we do have multi-level merging needs for `material` today.

core/src/scene/sceneLoader.cpp
```diff
@@ -330,6 +349,236 @@ void SceneLoader::loadTextures(Node textures, Scene& scene) {
...
+Node SceneLoader::propMerge(const std::string& propStr, const std::string& subPropStr, Node& styles,
+        const MIXES& mixes) {
+
+    std::unordered_set<std::string> uniqueList;
+    Node node;
+    for (const auto& mixStyleName : mixes) {
+        auto mixStyleNode = styles[mixStyleName];
+        if (auto propNode = mixStyleNode[propStr]) {
+            if (auto subPropNode = propNode[subPropStr]) {
+                if (subPropNode.IsMap()) { // for something like shaders: defines or unique
```
bcamper: See comments about the need for deep/recursive merging re: `material`.

core/src/scene/sceneLoader.cpp
```diff
@@ -330,6 +349,236 @@ void SceneLoader::loadTextures(Node textures, Scene& scene) {
...
+                    for (const auto& val : subPropNode) {
+                        node[val.first] = val.second;
+                    }
+                } else if (subPropNode.IsScalar()) { // for something like shaders: extension (scalar)
```
bcamper: This is not always the behavior you want for scalars, it's specific to `extensions`. In most cases, scalars overwrite the previous value. The way I would think of `extensions` is to convert a single value into a list, before the merging happens.

I'm weary of having feature-specific code that otherwise reads as generic.

core/src/scene/sceneLoader.cpp
bcamper: meaning of comment "defines or unique"?

core/src/scene/sceneLoader.cpp
tallytalwar: Thats why I have another method for overwritting scalar values.

core/src/scene/sceneLoader.cpp
tallytalwar: I can see if I can generalize propOverwrite and propMerge methods.

core/src/scene/sceneLoader.cpp
tallytalwar: Ohh I guess typo ... meant "uniforms"

core/src/scene/sceneLoader.cpp
tallytalwar: I think I might be able to incorporate this in my "merge" generalization.

tallytalwar:
@bcamper Pushed an update with a more generic property merging method. Its also recursive and handles deep merging, which the previous implementation was lacking.

blair1618:
Looking over this

core/src/scene/sceneLoader.cpp
```diff
@@ -330,8 +350,209 @@ void SceneLoader::loadTextures(Node textures, Scene& scene) {
...
+        	std::string textureName = textureNode.as<std::string>();
+            auto atlases = scene.spriteAtlases();
+            auto it = atlases.find(textureName);
+            if (it != atlases.end()) {
+            	spriteStyle->setSpriteAtlas(it->second);
+            } else {
+            	logMsg("WARNING: undefined texture name %s", textureName.c_str());
+            }
+        }
+	}
```
blair1618: several tabs in the preceding scope - I'd just convert the whole file

core/src/scene/sceneLoader.h
```diff
@@ -17,11 +17,14 @@ class SceneLayer;
 class View;
 class ShaderProgram;
 class Material;
+class Style;
 struct StyleParam;
 struct DrawRule;
 struct MaterialTexture;
 struct Filter;
 
+using MIXES = std::vector<YAML::Node>;
```
blair1618: All-caps says "preprocessor" to me - either that or you're shouting :P

bcamper:
Add tests for more cases :)

tallytalwar:
@blair1618, @bcamper .. having unit tests was a good suggestion. Caught an interesting bug :D. 781540f fixes the bug.

tallytalwar:
Updated with tests for `SceneLoader::propMerge`, `SceneLoader::shaderBlockMerge` and `SceneLoader::shaderExtMerge`.

blair1618:
Seems to work!

I think if we spent some more time looking at this we could make the code a bit more concise, but we can revisit this after the eraser-map release. 

Mergin'

tallytalwar:
Sure, we can make many improvement to sceneLoader.cpp, which I noticed while "staring" at the file :P.

PR #284, opened 2015-09-11
Use a proper dead states for discarded labels

karimnaaji:
When occlusion happens to a label that has been in a visible state at least once, mark this label as a label that could go out of the sleeping state.

label-states

This adds some issues, like flickering or cascading fading when more occlusion happens at the same time. To solve that, we would need to consolidate a little bit on which state would be marked as an "occluding" state.


karimnaaji:
This would also improve determinism when combined with proper priorities.

hjanetzek:
@karimnaaji Any objections to merge this now?

tallytalwar:
We were holding this. He did not want to merge this... I dont remember the reason why!!!

karimnaaji:
I wanted to profile this first :/ I don't think it's safe to merge without adding some more thoughts on it, it'll probably drop down the framerate at some zoom levels where label density is really high..

hjanetzek:
Make sure to test in Release mode. It made a huge impact last time I tried the isect2d demo

hjanetzek:
To avoid per frame allocations for intersection pairs in isect2d::intersect it is probably beneficial to have a reusable instance for its temporary data

tallytalwar:
@karimnaaji  (I can officially tag you now :D). Can you update this? Was your main concern performance?

@hjanetzek made isect2d more fast, than what we had so this should be much better now. So this should be good to merge?

karimnaaji:
I checked the isect2d changes, it looks really good to me now.  It must be good to merge. We can follow up on this to improve some subtile flickering I noticed later.

PR #285, opened 2015-09-13
Feature picking

hjanetzek:
- [ ] Currently using std::string for feature IDs. Should we use (long) integers instead?
- [ ] remove test style

### DIFF
--- a/android/tangram/jni/platform_android.cpp
+++ b/android/tangram/jni/platform_android.cpp
@@ -44,7 +44,7 @@ void setupJniEnv(JNIEnv* _jniEnv, jobject _tangramInstance, jobject _assetManage
     getFontFilePath = jniEnv->GetMethodID(tangramClass, "getFontFilePath", "(Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;)Ljava/lang/String;");
 	requestRenderMethodID = _jniEnv->GetMethodID(tangramClass, "requestRender", "()V");
     setRenderModeMethodID = _jniEnv->GetMethodID(tangramClass, "setRenderMode", "(I)V");
-    featureSelectionCbMID = _jniEnv->GetMethodID(tangramClass, "featureSelectionCb", "(ILjava/lang/String;)V");
+    featureSelectionCbMID = _jniEnv->GetMethodID(tangramClass, "featureSelectionCb", "(Ljava/lang/String;)V");
 
     assetManager = AAssetManager_fromJava(jniEnv, _assetManager);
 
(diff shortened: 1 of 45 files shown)


tallytalwar:
Reviewing now.

android/tangram/jni/platform_android.cpp
```diff
@@ -243,5 +246,27 @@ void setCurrentThreadPriority(int priority) {
     setpriority(PRIO_PROCESS, tid, priority);
 }
 
+void featureSelectionCallback(const std::vector<Tangram::TouchItem>& items) {
+
+    JNIEnv *jniEnv;
```
tallytalwar: Will this be called from a native thread?

If not, we can do with the cached jniEnv. And we do not need to attach current native thread to the jni environment.

hjanetzek: Right, this can now be changed to use the jnienv passed into Java_com_mapzen_tangram_MapController_pickFeature (the only caller of featureSelectionCallback).  As far as I know java threads are native threads as well on Android. 


As a side-note: I noticed that there there are two calls to init / setupJniEnv: once on the main-thread and once on androids render-thread.

blair1618:
Going to check out different ways of returning data from queries...

blair1618:
The latest commit changes the picking interface to return a JSON representation of the picked feature's properties, as a string in C++ or a JSON object in Java. 

It's a pretty convenient interface for Android, but I'm unhappy with parts of it. For one, the C++ interface will now essentially *require* using a JSON parser in your project if you need to use picking. Also, when we add the mechanism for adding properties to client-generated features, the data for those properties will end up going through a truly comical number of conversions:

[some Java data] -> [Java JSON object] -> [Java string] -> [C++ string] -> [C++ JSON object] -> [geojsonvt::tags] -> [Tangram::Properties]

then when a feature is clicked:

[Tangram::Properties] -> [C++ string] -> [Java string] -> [Java JSON object] -> [actual desired data]

... hardly ideal. But in terms of speed and memory this isn't prohibitive right now. 

There isn't an obvious solution to this, but one option might be to allow adding a `void*` (or `Object`, in Java) of "user data" to an added feature, in addition to JSON-like properties. Then users who only care about the structure of their own data don't need to bother with JSON at all.

hjanetzek:
@blair1618, why not just keep the Properties object on the client side? This way is can be used directly by c++ client application. We should rather have a jni interface for Properties than to go through 10 conversion steps imho :)

blair1618:
@hjanetzek That's a possibility, it would definitely simplify parts of the library for sure. It seems equally troublesome to the client though - I'm imagining that I have some (let's say Java) object of exactly the data that I care about and I want to associate that with a feature in the map. It's equally annoying to convert my data into either a JSON object or a Tangram::Properties object and then retrieve my data from that format when I get a callback. 

(fixing the tests right now, btw)

blair1618:
Actually, re: my previous comment, the best would be to do both of these. That is, allow JSON-like data in our own Properties structure and also allow custom data in an opaque object.

hjanetzek:
@blair1618, I think it's not that bad (or annoying): The client application does not have to convert all feature properties to Properties, but only those that are used for styling on the map + a feature_id to match the picking-callback item with its local data.

blair1618:
@hjanetzek I think you have a good point there - a client application probably has some sort of index for their data that they can add to the feature. 

The most "native" type we could use for this might be a Map of strings to strings - which also happens to be the format of "tags" that's ingested by geojson-vt so it would make life easy for us in that respect.

hjanetzek:
Just checked that we could simply change geojson-vt to use our Properties as Tags. All that geojson-vt does is to define Tags to be std::map<string,string> but does not touch them internally :)

bcamper:
I'd rather not get into modifying geojson-vt unless it is in generic ways
that we can submit back upstream (like general performance improvements).
(Unless there's a huge benefit justifying a true fork but it seems more
like convenience in this case.)

hjanetzek:
OK. Let's use map<string,string> for the properties exchange through JNI for now. I can take care of the bindings

blair1618:
Closed in favor of #292

Issue #181, opened 2015-07-08
Setting 2D position doesn't always make target position visible


blair1618:
When the position of the map view is set to a 2D location, the location of the virtual camera is placed directly above that position on the ground. If the camera is tilted off of the z-axis, the point in the center of the screen isn't the position that was set (and often the targeted position won't even be visible). 

The simplest fix for this is to add a function that sets the "view focus" instead of just the position and let the library calculate the appropriate position for the camera. 

@ecgreb Would this be useful?

ecgreb:
Useful yes but not a must have at this time. We typically reset zoom/tilt/rotation when setting the map position anyway.

blair1618:
Resolved: https://github.com/tangrams/tangram-es/pull/289

PR #289, opened 2015-09-16
Map position is the focus of the view

blair1618:
This has always been true in top-down views, but there has been some confusion about what corresponds to the "map position" when the view is tilted: the position of the camera in space, or the position at which the camera is looking. 

With these changes, the "map position" is always the point at which the camera is looking. This resolves an issue in eraser-map where the user's current location is not visible in a (tilted) routing view. 

While making these changes, I found a subtle but important issue related to input handling: 

`InputHandler` sets a fling delta every time that a pan above some threshold occurs. On any frame in which `handlePanGesture` isn't called, that delta is repeatedly applied and decremented to give a smooth fling. The problem is that while a finger/mouse is down and moving, *input events may not occur on every frame*. This results in a panning motion from both the fling delta and (later) the input callback, causing a translational drift. Fixing this with our existing input callbacks seems difficult - the information we really need is touch-up and touch-down events.

Since this issue may require API revision and isn't terribly problematic for most use cases, I've elected to leave it unresolved for now.

### DIFF
--- a/core/src/view/view.cpp
+++ b/core/src/view/view.cpp
@@ -17,7 +17,6 @@ double invLodFunc(double d) {
 }
 
 View::View(int _width, int _height, ProjectionType _projType) :
-    m_zoomAltitude(0.0),
     m_width(0),
     m_height(0),
     m_changed(false) {
@@ -98,14 +97,7 @@ void View::setRoll(float _roll) {
 
 void View::setPitch(float _pitch) {
 
-    _pitch = glm::clamp(_pitch, 0.f, (float)HALF_PI);
-
-    // Orbit vertically around the ground plane at the screen center
-    glm::vec2 radial = glm::vec2(-sin(m_roll), cos(m_roll));
-    radial *= m_zoomAltitude * (sin(m_pitch) - sin(_pitch));
-    translate(radial.x, radial.y);
-
-    m_pitch = _pitch;
+    m_pitch = glm::clamp(_pitch, 0.f, (float)HALF_PI);
     m_dirtyMatrices = true;
     m_dirtyTiles = true;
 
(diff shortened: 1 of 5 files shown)


tallytalwar:
This should also fix https://github.com/tangrams/tangram-es/issues/181

core/src/view/view.cpp
```diff
@@ -226,19 +217,19 @@ void View::updateMatrices() {
     // the "top" face of the view frustum with the ground plane
 
     // NOTE: far here can go to infinity and hence a std::min below!
-    float far = m_pos.z / cos(m_pitch + 0.5 * fovy);
+    float far = 2.f * m_pos.z / cos(m_pitch + 0.5 * fovy);
```
tallytalwar: I did not get the reason for this scaling? Can you explain?

blair1618: Hmm I think I added that as a factor of safety while I was messing with the view matrices, it seems that it's not strictly necessary.

tallytalwar: Also ... we have not noticed before but far can go negative.
We should use fabs here.

core/src/view/view.cpp
```diff
@@ -226,19 +217,19 @@ void View::updateMatrices() {
     // the "top" face of the view frustum with the ground plane
 
     // NOTE: far here can go to infinity and hence a std::min below!
-    float far = m_pos.z / cos(m_pitch + 0.5 * fovy);
+    float far = m_pos.z / std::min(0., cos(m_pitch + 0.5 * fovy));
```
tallytalwar: Can we use FLT_EPSILON here instead of 0.?

blair1618: The relevant constant here would be `DBL_MIN` (http://www.cplusplus.com/reference/cfloat/), but more importantly I got that comparison totally backwards. Fixing.

tallytalwar: yes yes that

PR #290, opened 2015-09-18
Specify colors and distances with Stops

blair1618:
To demonstrate and test the implementation, Stops for the "roads" layer have been ported from the tangram-js default scene file. Roads behave much more sanely now!

For parity with tangram-js, we still need:
 - [ ] resolving distance units to a common scale (right now distances specify a fraction of the tile size)
 - [ ] management of camera properties with Stops
 - [ ] sprite sizes as Stops (not in tangram-docs, but used in examples)
 - [ ] "z" parameter as Stops (we don't handle this parameter at all, but it's quite seldom used)

Some items above make sense to approach in a separate change set, so this isn't a completion list for this PR.

### DIFF
--- a/core/src/scene/stops.cpp
+++ b/core/src/scene/stops.cpp
@@ -1,5 +1,4 @@
 #include "stops.h"
-#include "csscolorparser.hpp"
 #include "util/geom.h"
 #include "yaml-cpp/yaml.h"
 #include <cassert>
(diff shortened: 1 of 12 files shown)


core/src/scene/stops.cpp
```diff
@@ -0,0 +1,84 @@
...
+            if (colorNode.IsScalar()) {
+                color = CSSColorParser::parse(colorNode.as<std::string>()).getInt();
+            } else if (colorNode.IsSequence() && colorNode.size() >= 3) {
+                float alpha = colorNode.size() > 3 ? colorNode[3].as<float>() : 1.f;
+                color = alpha * 255.;
+                color = (color << 8) + colorNode[2].as<float>() * 255.;
+                color = (color << 8) + colorNode[1].as<float>() * 255.;
+                color = (color << 8) + colorNode[0].as<float>() * 255.;
+            }
+            frames.emplace_back(key, color);
```
hjanetzek: We should support other interpolations in future: 
CIE-LCh looks quite good. The 'approximation trick' should be possible to implement by generating more intermediate zoom steps:
http://www.stuartdenman.com/improved-color-blending/

core/src/style/polylineStyle.cpp
```diff
@@ -15,8 +16,7 @@ void PolylineStyle::constructVertexLayout() {
     m_vertexLayout = std::shared_ptr<VertexLayout>(new VertexLayout({
         {"a_position", 3, GL_FLOAT, false, 0},
         {"a_texcoord", 2, GL_FLOAT, false, 0},
-        {"a_extrudeNormal", 2, GL_FLOAT, false, 0},
-        {"a_extrudeWidth", 1, GL_FLOAT, false, 0},
+        {"a_extrude", 4, GL_FLOAT, false, 0},
         {"a_color", 4, GL_UNSIGNED_BYTE, true, 0},
```
hjanetzek: extrudeNormal could be premultiplied with extrudeWidth to save 4 bytes per vertex I guess

core/src/scene/stops.cpp
bcamper: Yes! Would like to support different interpolation schemes in the future,
but want to think more about possible syntax, for example letting user
specify color space and/or interpolation curve (cubic, ease in/out, etc.).
So just sticking with simple RGB linear for now :)

blair1618: Agreed, HSL would be great to support and the technique you linked looks even better! We just need RGBA interpolation to match current tangram-js behavior.

blair1618:
closed in favor of: https://github.com/tangrams/tangram-es/pull/295

PR #291, opened 2015-09-18
Scene reloading - was: Decouple SceneLoader from TileManager and View

hjanetzek:
First step for reloading scenes at runtime (and on a worker thread). Enable loading of scenes without modifying tangrams state. 

Update: Scene reloading at runtime works!

This is a big refactoring of SceneLoader. Please have a look when you have time! Merging can wait until  the uniform-injection branch is done


hjanetzek:
Reorderded commits and split up pull request into separate branches

PR #292, opened 2015-09-19
Feature picking

hjanetzek:
Follow up on #285: Store shared_ptr of Properties instance in Label options and return it as feature picking result.

I've started to use SWIG to generate the JNI boilerplate which can do pretty advanced bindings:
http://www.swig.org/Doc3.0/Java.html#Java
http://www.swig.org/Doc3.0/Android.html#Android for a good example

- [X] converted ClienGeoJsonSource bindings to use SWIG 
   - The Java side now holds a shared_ptr reference to the DataSource and can work on it directly  
- [ ] removed Tangram::source* pass-through methods


hjanetzek:
Added support for pushing Tagged features through JNI

core/src/labels/labels.cpp
```diff
@@ -107,6 +106,65 @@ void Labels::update(const View& _view, float _dt, const std::vector<std::unique_
...
+
+                auto& transform = label->getTransform();
+                glm::vec4 v1 = worldToClipSpace(mvp, glm::vec4(transform.modelPosition1, 0.0, 1.0));
+                glm::vec4 v2 = worldToClipSpace(mvp, glm::vec4(transform.modelPosition2, 0.0, 1.0));
+
+                // check whether the label is behind the camera using the perspective division factor
+                if (v1.w <= 0 || v2.w <= 0) { continue; }
+
+                // project to screen space
+                glm::vec2 p1 = clipToScreenSpace(v1, screenSize);
```
karimnaaji: Here we can skip `worldToClipSpace` -> `clipToScreenSpace` and directly use `worldToScreenSpace` to simplify a bit, since the check on perspective factor `w` is useful only for internal state management of labels. The `visibleState()` check would then be sufficient.

core/src/labels/labels.cpp
```diff
@@ -107,6 +106,65 @@ void Labels::update(const View& _view, float _dt, const std::vector<std::unique_
...
+                glm::vec2 p2 = clipToScreenSpace(v2, screenSize);
+
+                float distance = sqrt(sqSegmentDistance(touchPoint, p1, p2));
+
+                if (distance > threshold) { continue; }
+
+                if (!_visibleOnly || (label->visibleState() && isect2d::intersect(label->getOBB(), obb))) {
```
karimnaaji: This test should rather be:
```cpp
if ((!_visibleOnly || label->visibleState()) && isect2d::intersect(label->getOBB(), obb))) {

}
```
Or the threshold would be active only on the anchor points of the label while using `_visibleOnly = false`.
The thing is labels in non-visible states, don't update their bounding boxes for performance reasons; in that case I think that it could make sense to update the bounding box once if `_visibleOnly = false`.

```cpp
if (!_visibleOnly) {
    label->updateScreenTransform(mvp, {_view.getWidth(), _view.getHeight()});
    label->updateBBoxes(); // need to be public now
}
```
And this would update the bounding boxes of the labels within the threshold before actually testing them against the current position.

tallytalwar:
Going to start reviewing this now. @hjanetzek any more big changes coming here?

core/src/labels/labels.cpp
hjanetzek: Maybe we should postpone handling of non-visible labels: I guess for invisible labels it would be more important to check the distance to the feature instead of whether one hits its invisible label.  

I've updated labels.cpp following your suggestions to always check intersection for the labels bbox

hjanetzek:
@tallytalwar, ready to review. I've removed the swig bindings for ClientGeoJsonSource which were previously included in this PR.

core/src/data/tileData.h
```diff
@@ -66,73 +66,20 @@ typedef std::vector<Point> Line;
...
-            return it.get<float>();
-        }
-        return 0;
-    }
-    bool getString(const std::string& key, std::string& value) const {
-        auto& it = get(key);
-        if (it.is<std::string>()) {
-            value = it.get<std::string>();
-            return true;
+struct Geometry {
```
tallytalwar: Why a struct here? We can have a Geometry namespace here, and make it an inline function.. (maybe ... ).

hjanetzek: yes - I'll just make it inline. Though it's used in only one place. So we could actually inline it there :)

core/src/labels/labels.h
```diff
@@ -18,7 +19,7 @@ class FontContext;
 class Tile;
 class View;
 class Style;
-
+struct TouchItem;
```
tallytalwar: redundant forward declaration.

android/tangram/jni/platform_android.cpp
```diff
@@ -266,5 +276,16 @@ void setCurrentThreadPriority(int priority) {
     setpriority(PRIO_PROCESS, tid, priority);
 }
 
+void featureSelectionCallback(JNIEnv* jniEnv, const std::vector<std::shared_ptr<Tangram::Properties>>& items) {
+
+    jlong jresult = 0;
+    jclass propertiesClass = jniEnv->FindClass("com/mapzen/tangram/Properties");
+
+    jresult = reinterpret_cast<jlong>(new std::shared_ptr<Tangram::Properties >(items[0]));
```
tallytalwar: So, it will always return the properties for the "first" feature being picked?

Should we be sending a list all features picked by our code?

tallytalwar: If so,

then either we can call the featureSelectionCb method multiple times for each picked feature, or somehow return a list of all features to the java interface (don't know how that can be done through jni, specially with the swig bindings :|).

core/src/data/properties.cpp
```diff
@@ -0,0 +1,131 @@
...
+void Properties::add(std::string key, std::string value) {
+    props.emplace_back(std::move(key), Value{std::move(value)});
+    sort();
+}
+void Properties::add(std::string key, float value) {
+    props.emplace_back(std::move(key), Value{value});
+    sort();
+}
+
+std::string Properties::toJson() const {
```
tallytalwar: This should be redundant now!!?

tallytalwar:
This is mostly good, apart from the following:

- [ ] Android only gets the first picked feature
- [ ] ios implementation (Ignoring raspberry pi implementation)

android/tangram/jni/platform_android.cpp
hjanetzek: The other code just changed to also only return one feature for now. Passing a vector of properties to java is easy with swig, basically:

    jclass propertiesClass = jniEnv->FindClass("com/mapzen/tangram/Features");
    jresult = reinterpret_cast<jlong>(new std::shared_ptr<Tangram::Features >(items));

Plus %template(Features) std::vector<std::shared_ptr<Tangram::Feature>>;

I would suggest to concentrate on getting it working for the eraser-map use, and add more advanced things later.

core/src/labels/labels.cpp
```diff
@@ -108,6 +109,65 @@ void Labels::update(const View& _view, float _dt, const std::vector<std::unique_
...
+                    if (distance < minDistance) {
+                        minDistance = distance;
+                        selectedItem = options.properties;
+                    }
+                }
+            }
+        }
+    }
+
+    if (selectedItem) {
```
tallytalwar: Should this not be within the for loop?

Else, `selectedItem` will be overwritten by previous and `m_touchItems` will always contain the last `selectedItem`.

android/tangram/jni/platform_android.cpp
hjanetzek: I could change it to not use a vector at all for now if wanted

tallytalwar: We can get @ecgreb's input here. 

I would assume, the API to return a vector of all selected features and the client app to filter on what is picked.
How you are doing for linux and osx: https://github.com/tangrams/tangram-es/blob/feature-nit-picking/osx/src/main.cpp#L53-L57

Its highly likely, if client expects a search result to be picked, but tangram return a poi right adjacent to the search result.

hjanetzek: Yes, I also think that there are cases where it is good to return a set of items near the touch point and let the client decide if it changes its previous selection (or whatever). But I think we can postpone this unless eraser-map plans to use it now

bcamper: Priority is on getting this merged, and we can extend behavior later. I
think either will work initially here.

core/src/labels/labels.cpp
karimnaaji: Yeah you're right, I'm not sure what should be the best behavior for those invisible labels, since selecting nothing doesn't mean no feature are present (even when you try on invisible labels); but  as a temporary solution both ways would be fine (distance or bbox) depending on how much precision we want. We can keep this as is for now

core/src/labels/label.cpp
```diff
@@ -53,7 +53,7 @@ bool Label::updateScreenTransform(const glm::mat4& _mvp, const glm::vec2& _scree
             glm::vec4 v2 = worldToClipSpace(_mvp, glm::vec4(m_transform.modelPosition2, 0.0, 1.0));
 
             // check whether the label is behind the camera using the perspective division factor
-            if (v1.w <= 0 || v2.w <= 0) {
+            if (_testVisibility && v1.w <= 0 || v2.w <= 0) {
```
blair1618: Boolean grouping is incorrect - group the OR

core/src/labels/labels.cpp
hjanetzek: Returning all intersection labels sorted by distance between touch point and the labels center now

core/src/data/properties.cpp
hjanetzek: Why not leave it as a start to serialize features into json? :)

tallytalwar: Sure.

blair1618:
:tada:

PR #294, opened 2015-09-21
Uniform injection

tallytalwar:
Handles the injection of style uniforms and setting the uniform value on a draw call.

- Style Injection:
-- Handles float, boolean (0: false, 1: true), vec2, vec3, vec4, texture samplers.
-- `Style::setupShaderUniforms` sets the uniform value on draw frame call.

TODO Before this goes for final review:
- [x] Tests cases for parsing uniforms

TODO (Future):
- Arbitrary length numeric uniforms.

### DIFF
--- a/core/src/scene/sceneLoader.cpp
+++ b/core/src/scene/sceneLoader.cpp
@@ -119,17 +119,11 @@ void SceneLoader::loadShaderConfig(Node shaders, Style& style, Scene& scene) {
             if (size == 1) {
                 shader.addSourceBlock("uniforms", "uniform " + type + " " + name + ";");
                 style.styleUniforms().emplace_back(name, uniformValues[0]);
-                if (uniformValues[0].is<std::string>()) {
-                    auto textureName = uniformValues[0].get<std::string>();
-                }
             } else {
                 shader.addSourceBlock("uniforms", "uniform " + type + " " + name +
                                                         "[" + std::to_string(size) + "];");
                 for (int i = 0; i < size; i++) {
                     style.styleUniforms().emplace_back(name + "[" + std::to_string(i) + "]", uniformValues[i]);
-                    if (uniformValues[i].is<std::string>()) {
-                        auto textureName = uniformValues[i].get<std::string>();
-                    }
                 }
             }
         }
--- a/core/src/scene/sceneLoader.cpp
+++ b/core/src/scene/sceneLoader.cpp
@@ -70,9 +70,7 @@ void SceneLoader::loadShaderConfig(Node shaders, Style& style, Scene& scene) {
 
     auto& shader = *(style.getShaderProgram());
 
-    if (!shaders) {
-        return;
-    }
+    if (!shaders) { return; }
 
     if (Node extensionsNode = shaders["extensions"]) {
         std::vector<std::string> extensions;
@@ -92,8 +90,7 @@ void SceneLoader::loadShaderConfig(Node shaders, Style& style, Scene& scene) {
         }
     }
 
-    Node definesNode = shaders["defines"];
-    if (definesNode) {
+    if (Node definesNode = shaders["defines"]) {
         for (const auto& define : definesNode) {
             std::string name = define.first.as<std::string>();
             std::string value = define.second.as<std::string>();
@@ -108,8 +105,7 @@ void SceneLoader::loadShaderConfig(Node shaders, Style& style, Scene& scene) {
         }
     }
 
-    Node uniformsNode = shaders["uniforms"];
-    if (uniformsNode) {
+    if (Node uniformsNode = shaders["uniforms"]) {
         for (const auto& uniform : uniformsNode) {
             auto name = uniform.first.as<std::string>();
             auto uniforms = parseStyleUniforms(uniform.second, scene);
--- a/core/src/scene/sceneLoader.cpp
+++ b/core/src/scene/sceneLoader.cpp
@@ -121,7 +121,6 @@ void SceneLoader::loadShaderConfig(Node shaders, Style& style, Scene& scene) {
                 style.styleUniforms().emplace_back(name, uniformValues[0]);
                 if (uniformValues[0].is<std::string>()) {
                     auto textureName = uniformValues[0].get<std::string>();
-                    style.uniformTextures().emplace(textureName, scene.textures()[textureName]);
                 }
             } else {
                 shader.addSourceBlock("uniforms", "uniform " + type + " " + name +
@@ -130,7 +129,6 @@ void SceneLoader::loadShaderConfig(Node shaders, Style& style, Scene& scene) {
                     style.styleUniforms().emplace_back(name + "[" + std::to_string(i) + "]", uniformValues[i]);
                     if (uniformValues[i].is<std::string>()) {
                         auto textureName = uniformValues[i].get<std::string>();
-                        style.uniformTextures().emplace(textureName, scene.textures()[textureName]);
                     }
                 }
             }
@@ -1053,11 +1051,11 @@ StyleUniforms SceneLoader::parseStyleUniforms(const Node& value, Scene& scene) {
                 uVal = 0;
             } else {
                 type = "sampler2D";
+                uVal = strVal;
                 auto texItr = scene.textures().find(strVal);
                 if (texItr == scene.textures().end()) {
                     loadDefaultTexture(strVal, scene);
                 }
-                uVal = strVal;
             }
         }
         uniformValues.push_back(uVal);
--- a/core/src/scene/sceneLoader.cpp
+++ b/core/src/scene/sceneLoader.cpp
@@ -19,6 +19,7 @@
 #include "sceneLayer.h"
 #include "scene/dataLayer.h"
 #include "text/fontContext.h"
+#include "util/yamlHelper.h"
 
 #include "yaml-cpp/yaml.h"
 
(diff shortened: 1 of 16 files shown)


core/src/style/style.h
```diff
@@ -99,10 +103,14 @@ class Style {
...
     bool glContextLost();
 
+    void setupShaderUniforms(int _lastBoundTex, bool _ctxLost);
+
 private:
 
     /* Whether the context has been lost on last frame */
     bool m_contextLost;
+    std::vector<StyleUniform> m_styleUniforms;
+    std::unordered_map<std::string, std::shared_ptr<Texture>> m_uniformTextures;
```
tallytalwar: Will it make sense to not have this extra mapping and include "shared_ptr<Texture>" as a UniformValue?

tallytalwar: I updated it to use Texture in UniformValue variant.
But variant header now includes texture header ... I don't like that.

hjanetzek: I would prefer to have a 'textures' container in Scene and refer to  them via TextureIDs if possible. similar to FunctionID for js functions. 

In general I thought it would be good to use this approach for all Scene resources, also forStyleParams: This would also enable to have no StyleParam copying during feature style evaluation, just a pair of <StyleParamKey, StyleValueId>.

tallytalwar: So do you think the previous way before I pushed 42ce3fd was better?

hjanetzek: Yes, this way one can also figure out when different shaders  refer to the same texture(more easily) and only load one instance

tallytalwar: :+1:

hjanetzek:
Ready to merge!

tallytalwar:
Changes you made look good. 
I just want @bcamper and @blair1618 to confirm the test cases I have to make sure we are not missing anything this time. Once done will merge.

bcamper:
Cool will look in a bit

hjanetzek:
@tallytalwar, we should add a check whether the shader definition tries to overwrite builtin uniforms

bcamper:
Or should we? This way you can *stop time* :)

hjanetzek:
@bcamper, at least it would not work with our current uniform logic since the uniform name will appear twice in the shader source.

bcamper:
Ah good call. I actually have some code in JS that checks for existing
uniform definitions to not repeat them (simple regex, not a real AST
parser), but I was thinking of removing it anyway because it's a lot of
complexity for something we probably shouldn't support in the first place :)

tallytalwar:
Should we then not allow uniform injection which match builtin uniforms for that style?

Or we can also maintain types for each builtin uniform and do appropriate overwrite handling if types differ.

core/src/scene/sceneLoader.cpp
```diff
@@ -147,12 +105,23 @@ void SceneLoader::loadShaderConfig(Node shaders, ShaderProgram& shader) {
...
-            std::string name = uniform.first.as<std::string>();
-            std::string value = uniform.first.as<std::string>();
-            shader.addSourceBlock("uniforms", "uniform " + name + " = " + value + ";");
+            auto name = uniform.first.as<std::string>();
+            auto uniforms = parseStyleUniforms(uniform.second, scene);
+            auto& type = uniforms.first;
+            auto& uniformValues = uniforms.second;
+            int size = uniformValues.size();
+            if (size == 1) {
+                shader.addSourceBlock("uniforms", "uniform " + type + " " + name + ";");
```
bcamper: I put them all in `globals`, but there is no real reason we can't have a `uniforms` block instead, and doing that in JS may also help to more clearly segment things too.

core/src/scene/sceneLoader.cpp
```diff
@@ -1049,6 +1024,69 @@ std::vector<StyleParam> SceneLoader::parseStyleParams(Node params, Scene& scene,
...
+    std::string type = "";
+    std::vector<UniformValue> uniformValues;
+    if (value.IsScalar()) { //float, int (bool), string (texture)
+        UniformValue uVal;
+        try {
+            uVal = value.as<float>();
+            type = "float";
+        } catch (const BadConversion& e) {
+            auto strVal = value.as<std::string>();
+            type = "int";
```
bcamper: Why the `int` / `bool` conversion? Is this because C++ YAML parser interprets `true`/`false` as 0/1?

bcamper: (FWIW we don't support `int` uniform types in JS anyway so we don't need them here either.)

core/src/scene/sceneLoader.cpp
```diff
@@ -1049,6 +1024,69 @@ std::vector<StyleParam> SceneLoader::parseStyleParams(Node params, Scene& scene,
...
+            if (strVal == "true") {
+                uVal = 1;
+            } else if (strVal == "false") {
+                uVal = 0;
+            } else {
+                type = "sampler2D";
+                uVal = strVal;
+                auto texItr = scene.textures().find(strVal);
+                if (texItr == scene.textures().end()) {
```
bcamper: Can you explain what is happening with this logic? What does `loadDefaultTexture` do?

core/src/scene/sceneLoader.cpp
```diff
@@ -1049,6 +1024,69 @@ std::vector<StyleParam> SceneLoader::parseStyleParams(Node params, Scene& scene,
...
+                    break;
+                case 3:
+                    uVal = parseVec<glm::vec3>(value);
+                    break;
+                case 4:
+                    uVal = parseVec<glm::vec4>(value);
+                    break;
+                default:
+                    break;
+                    // TODO: Handle numeric arrays past 4 elements
```
bcamper: I do think it's worth doing this, and also that we can skip supporting `structs` for ES (and deprecate for JS).

core/src/scene/sceneLoader.cpp
tallytalwar: For your first question .. 

YAML cpp can convert "true"/"false" to boolean but will have to be handled in a try catch and then explicitly handle value setting. So saved 3/4 lines of code.

For second question.
`int` type for uniform is only there to set a boolean uniform. All other numeric uniforms will have a `float` type.

core/src/scene/sceneLoader.cpp
tallytalwar: if a texture has not been loaded before then it will go and load it with default texture properties and add it to the list of textures maintained by the scene.

core/src/scene/sceneLoader.cpp
tallytalwar: Should I take the TODO out? 

Is it still good to have a TODO there?

bcamper: Keep it, just commenting.

core/src/scene/sceneLoader.cpp
bcamper: Got it. I find `loadDefaultTexture` slightly odd name there since it's not a *default* texture, it's a new texture with default *settings*.

core/src/scene/sceneLoader.cpp
blair1618: These lines confused me too. An integer value in yaml will always coerce to a float in the preceding `try` block (even if explicitly tagged as an integer, but that's a whole other rant) so only string values (or Booleans written as strings) will enter the catch block. Variables declared as `int` in GLSL can't be used in conditionals, so the type here should probably be `bool` (unless I'm misunderstanding something).

core/src/scene/sceneLoader.cpp
tallytalwar: Renamed: 0dbb92e

core/src/scene/sceneLoader.cpp
tallytalwar: Fixed: 0dbb92e

hjanetzek: cpp-yaml converts quite a lot of things to boolean:  {"y", "n"}, {"yes", "no"}, {"true", "false"}, {"on", "off"} http://yaml.org/type/bool.html. Though chances are low that someone names textures like this :)

@bcamper don't need to doesn't mean we may not add support for int uniforms while we are at it. Or is it about strict feature parity?

tallytalwar: Not just feature parity.

But it also gets harder to distinguish between an `int` or `float` value in yaml. Its doable though!!!! But!!!

With respect usage of `int` uniforms. Based on a quick discussion we had here in office, it seems int uniforms are mostly used when you want to access array indices. We could not come up with any other use case for `int` uniforms.

tests/unit/styleUniformsTests.cpp
```diff
@@ -0,0 +1,109 @@
...
+#include "scene/scene.h"
+#include "util/variant.h"
+
+using namespace Tangram;
+using YAML::Node;
+
+Scene scene;
+SceneLoader sceneLoader;
+
+TEST_CASE( "Style Mixing Test: Float uniform value", "[mixing][core][yaml]") {
```
bcamper: Don't know that I'd call these style "mixing" tests -- mixing to be is about the actual `mix` functionality. Shader uniforms are an independent feature.

tallytalwar: Ohh copy paste dumbness :|. Will fix.

core/src/scene/sceneLoader.cpp
bcamper: I would prefer to not add support for `int` uniforms since they appear to
not be supported in WebGL, and I'd like to keep parity so we don't have
files that work on one platform but not the other (that's clearly pretty
aspirational right now, but we can at least avoid making the gap wider :)

hjanetzek: On Wed, Sep 23, 2015 at 10:33 PM, Brett Camper <notifications@github.com>
wrote:

> In core/src/scene/sceneLoader.cpp
> <https://github.com/tangrams/tangram-es/pull/294#discussion_r40253896>:
>
> > @@ -1049,6 +1024,69 @@ std::vector<StyleParam> SceneLoader::parseStyleParams(Node params, Scene& scene,
> >      return out;
> >  }
> >
> > +StyleUniforms SceneLoader::parseStyleUniforms(const Node& value, Scene& scene) {
> > +    std::string type = "";
> > +    std::vector<UniformValue> uniformValues;
> > +    if (value.IsScalar()) { //float, int (bool), string (texture)
> > +        UniformValue uVal;
> > +        try {
> > +            uVal = value.as<float>();
> > +            type = "float";
> > +        } catch (const BadConversion& e) {
> > +            auto strVal = value.as<std::string>();
> > +            type = "int";
>
> I would prefer to not add support for `int` uniforms since they appear to
> not be supported in WebGL, and I'd like to keep parity so we don't have
> files that work on one platform but not the other (that's clearly pretty
> aspirational right now, but we can at least avoid making the gap wider :)
>

 That would be very reasonable -  I know that GLES does not support int for
vertex attributes either, though it seems WebGL does have int uniforms as
well:
http://webglfundamentals.org/webgl/lessons/webgl-shaders-and-glsl.html#uniforms

bcamper: Oh sorry I was confusing that with int vertex attributes!

Nonetheless, I think the use case for int uniforms is pretty small :)

core/src/scene/sceneLoader.cpp
```diff
@@ -1049,6 +1024,67 @@ std::vector<StyleParam> SceneLoader::parseStyleParams(Node params, Scene& scene,
...
+            }
+            uniformValues.push_back(uVal);
+        } catch (const BadConversion& e) { // array of strings (textures)
+            uniformValues.reserve(size);
+            type = "sampler2D";
+            for (const auto& strVal : value) {
```
blair1618: I'm unclear on the interpretation of an array of strings here, does it create an array of samplers?

PR #295, opened 2015-09-22
more work on stops

hjanetzek:
I've changed our tilesize dependent width to use pixel as units. This implements one of the supported width units of tangram-js and was also somewhat necessary to visually validate that the width interpolation with stops works correctly :)

### DIFF
--- a/core/src/scene/stops.cpp
+++ b/core/src/scene/stops.cpp
@@ -1,5 +1,4 @@
 #include "stops.h"
-#include "csscolorparser.hpp"
 #include "util/geom.h"
 #include "yaml-cpp/yaml.h"
 #include <cassert>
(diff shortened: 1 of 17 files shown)


tallytalwar:
Reviewing this now.

core/src/scene/stops.cpp
```diff
@@ -0,0 +1,84 @@
+#include "stops.h"
+#include "csscolorparser.hpp"
+#include "yaml-cpp/yaml.h"
+#include <cassert>
+#include <algorithm>
+
+#define RED(col) (col % 256)
```
tallytalwar: Thoughts?

We should move these define to the header?

blair1618: A utility file with higher-level functions on colors would be useful in many places, distance conversions too; just not macros in headers :P

tallytalwar: Yes, I got rid of that :D. (hence ... ~~..~~)

PR #297, opened 2015-09-22
Resource paths are resolved relative to scene file

blair1618:
- The "core/resources" directory structure is preserved when the application is built. 
 - At launch, tangram stores the directory that contains the scene file being loaded. All future resource paths are resolved relative to this directory. 
 - If the scene file is provided using a relative path, it is interpreted to be located within the application's bundled resources (and by extension, all other resources will be resolved within the bundled resources).


tallytalwar:
Reviewing this first before https://github.com/tangrams/tangram-es/pull/295. Should be quick.

ios/src/platform_ios.mm
```diff
@@ -57,12 +58,29 @@ bool isContinuousRendering() {
...
+std::string setResourceRoot(const char* _path) {
+
+    NSString* path = [NSString stringWithUTF8String:_path];
+
+    if (*_path != '/') {
+        NSString* resources = [[NSBundle mainBundle] resourcePath];
+        path = [resources stringByAppendingPathComponent:path];
+    }
+
+    s_resourceRoot = [path stringByDeletingLastPathComponent];
```
tallytalwar: We can rework this to be more direct.

```
s_resourceRoot = [[NSBundle mainBundle] resourcePath];
if (*_path != '/') {
    path = [s_resourceRoot stringByAppendingPathComponent:path];
}
```

blair1618: We only want `s_resourceRoot` to contain the bundle resource path if a relative path is used, i.e. if `_path` does not begin with a `/`.

ios/src/platform_ios.mm
```diff
@@ -57,12 +58,29 @@ bool isContinuousRendering() {
...
+}
+
 NSString* resolveResourcePath(const char* _path) {
-    
-    NSString* resourcePath = [[[NSBundle mainBundle] resourcePath] stringByAppendingString:@"/"];
-    NSString* internalPath = [NSString stringWithUTF8String:_path];
-    return [resourcePath stringByAppendingString:internalPath];
-    
+
+    if (s_resourceRoot == NULL) {
```
tallytalwar: Can we move this check to `setResourceRoot` which makes sure `s_resourceRoot` is always set when `resolveResourcePath` is called.

android/tangram/jni/platform_android.cpp
```diff
@@ -112,32 +115,64 @@ bool isContinuousRendering() {
...
 
-std::string stringFromResource(const char* _path) {
+std::string setResourceRoot(const char* _path) {
+
+    s_resourceRoot = std::string(dirname(_path));
 
-    std::string out;
+    s_useInternalResources = (*_path != '/');
+
+    if (s_resourceRoot.front() == '.') {
```
tallytalwar: According to the documentation for `string::front`: "This function shall not be called on empty strings.". Can we do a check for that first?

tallytalwar: Also document this special asset manager behavior, which we just talked about :D.

tallytalwar:
Good to go once the above mentioned issues are resolved.

ios/src/platform_ios.mm
blair1618: Checking `s_resourceRoot` in `resolveResourcePath` is to handle cases where `setResourceRoot` hasn't been called before `resolveResourcePath` (e.g. in some tests).

tallytalwar: but its definite that `setResourceRoot` is called first always, because of `Tangram::initialize`, [here](https://github.com/tangrams/tangram-es/blob/relative-resources/core/src/tangram.cpp#L38)

blair1618: Yes, but Tangram::initialize is not called in, say, tests.

blair1618: Just trying to make sure nothing insane will happen if we forget something.

PR #298, opened 2015-09-23
Fix a potential crash from concurrent access of client geometry

blair1618:


core/src/tangram.cpp
```diff
@@ -94,6 +96,8 @@ void resize(int _newWidth, int _newHeight) {
 
 void update(float _dt) {
 
+    std::lock_guard<std::mutex> lock(m_tilesMutex);
```
tallytalwar: Can we have the lock around the first access for anything to do with m_tileManager?

No need for `inputHandler` or `m_view` to wait on the lock.

tallytalwar:
Looks good to me, once the above comment is addressed.

PR #299, opened 2015-09-24
Scene reloading

hjanetzek:
First in the series of Scene loading related branches, splitting up #291 

This branch add support for reloading scenes at runtime. Edit your style, press 'R' for reload in the app  and see your changes immediately! :)

### DIFF
--- a/core/src/scene/scene.h
+++ b/core/src/scene/scene.h
@@ -10,10 +10,13 @@
 #include <vector>
 #include <unordered_map>
 
+#include "glm/glm.hpp"
+
 namespace Tangram {
 
 class Style;
 class Texture;
+class DataSource;
 
 /* Singleton container of <Style> information
  *
@@ -25,6 +28,7 @@ public:
     Scene();
     ~Scene();
 
+    auto& dataSources() { return m_dataSources; };
     auto& layers() { return m_layers; };
     auto& styles() { return m_styles; };
     auto& lights() { return m_lights; };
@@ -33,6 +37,7 @@ public:
     auto& spriteAtlases() { return m_spriteAtlases; };
     auto& stops() { return m_stops; }
 
+    const auto& dataSources() const { return m_dataSources; };
     const auto& layers() const { return m_layers; };
     const auto& styles() const { return m_styles; };
     const auto& lights() const { return m_lights; };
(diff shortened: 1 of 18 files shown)


tallytalwar:
Reviewing this now.

core/src/tangram.cpp
```diff
@@ -63,19 +62,34 @@ void initialize(const char* _scenePath) {
...
     logMsg("finish initialize\n");
 
 }
 
+void loadScene(const char* _scenePath) {
+
+    m_scene = std::make_shared<Scene>();
+
+    logMsg("Loading Tangram scene file: %s\n", _scenePath);
+    auto sceneString = stringFromResource(_scenePath);
```
blair1618: Before loading resources for a new scene, the resource root location needs to be set, as here: https://github.com/tangrams/tangram-es/blob/master/core/src/tangram.cpp#L40

blair1618:
For a while I thought this was broken because I was modifying the scene file and seeing no changes, but then I realized that on OS X it's loading the scene file that's copied into the application bundle, which of course was unchanged. Using a scene file at an absolute path shows reloading as expected :)

core/src/tangram.cpp
```diff
@@ -63,19 +62,35 @@ void initialize(const char* _scenePath) {
...
         m_skybox->init();
-
     }
 
     logMsg("finish initialize\n");
 
 }
 
+void loadScene(const char* _scenePath) {
+    logMsg("Loading scene file: %s\n", _scenePath);
```
tallytalwar: Should we be clearing all states/contexts (gl, fexture, fonts) loaded from the previous scene file?

hjanetzek: The fonts might need to be cleared. The other things should be managed by the Scene destructor

hjanetzek: A bit harder is to remove individual fonts from FontContext since fontstash has no method to remove fonts. We could move the FontContext to be a member of Scene though, but this could better go into the add-merged-font-style branch.

core/src/tangram.cpp
```diff
@@ -63,19 +62,35 @@ void initialize(const char* _scenePath) {
...
+
+    auto sceneRelPath = setResourceRoot(_scenePath);
+    auto sceneString = stringFromResource(_scenePath);
```
blair1618: `stringFromResource` needs to be called on `sceneRelPath` because after `setResourceRoot` is called, all resources will be resolved relative to the containing directory of `_scenePath`; this doesn't matter for the default location but for an absolute path like `/blah/scene.yaml` the second call will try to access `/blah//blah/scene.yaml`.

core/src/scene/scene.h
```diff
@@ -10,10 +10,13 @@
 #include <vector>
 #include <unordered_map>
 
+#include "glm/glm.hpp"
```
blair1618: I would usually include just `glm/vec2.hpp` if I'm only using that type - not sure how much of a difference it makes on compile times but it can't hurt.

tallytalwar:
Post Matt's commits, this is good to go.

PR #301, opened 2015-09-24
Load merged font styles

hjanetzek:
Font styles (family+weight+style) can be a mix from different sublayers. This branch loads the font definition from merged styles instead of loading a font style for single (sub)layer nodes.

+ Plus: Move font loading to be done lazily on tile worker threads

### DIFF
--- a/core/src/text/fontContext.cpp
+++ b/core/src/text/fontContext.cpp
@@ -55,6 +55,9 @@ bool FontContext::addFont(const std::string& _family, const std::string& _weight
     if ( !(data = bytesFromResource(bundledFontPath.c_str(), &dataSize))) {
         const std::string sysFontPath = systemFontPath(_family, _weight, _style);
         if ( !(data = bytesFromFileSystem(sysFontPath.c_str(), &dataSize)) ) {
+
+            logMsg("[FontContext] Error loading font file %s\n", fontKey.c_str());
+            m_fonts.emplace(std::move(fontKey), FONS_INVALID);
             return false;
         }
     }
@@ -63,6 +66,7 @@ bool FontContext::addFont(const std::string& _family, const std::string& _weight
 
     if (font == FONS_INVALID) {
         logMsg("[FontContext] Error loading font %s\n", fontKey.c_str());
+        m_fonts.emplace(std::move(fontKey), FONS_INVALID);
         return false;
     }
 
--- a/core/src/text/fontContext.cpp
+++ b/core/src/text/fontContext.cpp
@@ -2,6 +2,9 @@
 #define FONTSTASH_IMPLEMENTATION
 #include "fontstash.h"
 
+#define LOGE(fmt, ...) do { logMsg("FontContext/Error: " fmt "\n", ## __VA_ARGS__); } while(0)
+#define LOGW(fmt, ...) do { logMsg("FontContext/Warn: " fmt "\n", ## __VA_ARGS__); } while(0)
+
 namespace Tangram {
 
 #define INVALID_FONT -2
@@ -32,7 +35,7 @@ bool FontContext::lock() {
     try {
         m_contextMutex.lock();
     } catch (std::system_error& e) {
-        logMsg("Dead lock: aborting\n");
+        LOGE("Dead lock: aborting!");
         return false;
     }
     return true;
@@ -66,7 +69,7 @@ FontID FontContext::addFont(const std::string& _family, const std::string& _weig
             const std::string sysFontPath = systemFontPath(_family, _weight, _style);
             if ( !(data = bytesFromFileSystem(sysFontPath.c_str(), &dataSize)) ) {
 
-                logMsg("[FontContext] Error loading font file %s\n", fontKey.c_str());
+                LOGE("Could not load font file %s", fontKey.c_str());
                 m_fonts.emplace(std::move(fontKey), INVALID_FONT);
                 goto fallback;
             }
@@ -76,7 +79,7 @@ FontID FontContext::addFont(const std::string& _family, const std::string& _weig
     font = fonsAddFont(m_fsContext, fontKey.c_str(), data, dataSize);
 
     if (font == FONS_INVALID) {
-        logMsg("[FontContext] Error loading font %s\n", fontKey.c_str());
+        LOGE("Could not load font %s", fontKey.c_str());
         m_fonts.emplace(std::move(fontKey), INVALID_FONT);
         goto fallback;
     }
@@ -99,7 +102,7 @@ void FontContext::setFont(const std::string& _key, int size) {
         fonsSetSize(m_fsContext, size);
         fonsSetFont(m_fsContext, id);
     } else {
-        logMsg("[FontContext] Could not find font %s\n", _key.c_str());
+        LOGW("Could not find font %s", _key.c_str());
     }
 }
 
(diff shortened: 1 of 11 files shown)


bcamper:
Ah so was this a bug, meaning font properties were not inheriting?

hjanetzek:
The properties were inherited but fonts were only loaded for the 'static' nodes, not for the possible combinations of merged style parameters. E.g.

        draw:
            text:
                font:
                    family: Roboto
                    size: 1.em
        highway:
            filter: { kind: highway }
            draw:
                text:
                    font:
                        style: Regular
        highway_link:
            filter: all: { kind: highway, link: yes }
            draw:
                text:
                    font:
                        weight: 400

Would load the fonts Roboto, Roboto-Regular, and Roboto-400 instead of Roboto-400Regular. So the font lookup for drawing highway_link would fail.

tallytalwar:
In the example you gave ... post mixing of styles, `highway_link` should look like:

```
    highway_link:
        filter: all: { kind: highway, link: yes }
        draw:
            text:
                font:
                    family: Roboto
                    size: 1.em
                    weight: 400
```

instead of:
```
    highway_link:
        filter: all: { kind: highway, link: yes }
        draw:
            text:
                font:
                    family: Roboto
                    size: 1.em
                    weight: 400
                    style: Regular
```

highway_link sublayer should not inherit from highway sublayer here. 

However, If highway_link was a sublayer of highway layer, then it should inherit its font properties.

bcamper:
Are the fonts loaded on demand? Because there are potentially may unique
font combinations (especially because styles can combine "laterally" when
multiple match in addition to cascading inheritance), but you only know
which ones are actually needed when you evaluate features (and a new font
might become needed when a new tile loads and a feature matches, etc.).

hjanetzek:
@bcamper yes, this 'on-demand font loading once the styling of a feature is evaluated' is what this patch is intended to add.

@varun, the example was quickly made-up. it should probably work as you say. Though I'm not sure if this is actually what our implementation does at the moment since it collects style parameter from all matched sublayers. 

Hmm, how does this lateral combination intended to work? The order of evaluation cannot be ensured when highway or highway_link items will take precedence (there is no ordering of items in yaml-maps). iirc it was a mentioned somewhere that one has to use if-else filters for the cases highway_and_not_link and highway_link.

It may be good to have a rule checker that tells when matching rules are non-deterministic :)

tallytalwar:
Lateral merging is basically a BFS traversal of the graph (you traverse the subgraph is a node matches a rule).

bcamper:
Yes it's not ideal to have non-deterministic rule combinations, but really people shouldn't write filters that would collide like that :)  In that regard it is similar to CSS.

For nested filters, there is no conflict. If you write multiple filters at the same level that match, you can get potential collisions, but it's also a very useful way for writing rules that are not strictly hierarchical. For example both of these can match:

```
# put all tunnels at a lower level
tunnels:
   filter: { is_tunnel: yes }
   draw: { lines: { order: 1 }

# give all highways a color
highways:
   filter: { is_tunnel: yes }
   draw: { lines: { color: red }
```

bcamper:
For cases like that, they match in non-deterministic order, so if multiple rules set the same properties, the behavior is undefined. In practice I think both JS and C++ parsers are following the maps in order they appear in the file, but it is not guaranteed and should not be relied upon.

hjanetzek:
In yaml-cpp a map uses raw pointers as keys (std::map<node*,node*>) so order off iteration is quite as arbitrary as it can get. 
Writing a full rule checker might be a bit too much at the moment, but we could at least emit warnings when evaluation of one branch overwrites those parameter from another branch.

core/src/text/fontContext.cpp
```diff
@@ -40,35 +45,54 @@ void FontContext::unlock() {
...
-    if (m_fonts.find(fontKey) != m_fonts.end()) {
-        return true;
-    }
 
-    unsigned int dataSize;
-    unsigned char* data = nullptr;
+    auto it = m_fonts.find(fontKey);
+    if (it != m_fonts.end()) {
+        if (it->second < 0) {
+            goto fallback;
```
blair1618: http://imgs.xkcd.com/comics/goto.png

tallytalwar: OMG goto !!!

blair1618: To be clear, I'm fine with this (goto is no worse than exceptions in my opinion), I just find it amusing

tallytalwar: Not gone through the code yet to give any opinion right now on usage of `goto` :D.

But I agree exceptions are in the same league of goto statements.

hjanetzek: The clang devs seems to care so much about goto that it warns about all impossible cases of uninitialized variables before goto statements that are not being used..

Wouldn't mind to repeat the 'fallback' case three times or write a function to wrap that part if we have any purists :)

You'll find another goto usage in the sceneloader refactoring where it is used to break out of a nested loop:
https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Res-goto

core/src/text/fontContext.cpp
```diff
@@ -2,8 +2,13 @@
 #define FONTSTASH_IMPLEMENTATION
 #include "fontstash.h"
 
+#define LOGE(fmt, ...) do { logMsg("FontContext/Error: " fmt "\n", ## __VA_ARGS__); } while(0)
+#define LOGW(fmt, ...) do { logMsg("FontContext/Warn: " fmt "\n", ## __VA_ARGS__); } while(0)
```
blair1618: Getting a whole lot of warnings here:
```
warning: token pasting of ',' and __VA_ARGS__ is a GNU extension [-Wgnu-zero-variadic-macro-arguments]
```
Also, can you help me understand what's going on in these lines? What's the ## and why is it in a loop?

blair1618:
Getting a crash at runtime on Android in `NewStringUTF` in a call to `systemFontPath`. Since this function is now being called by the tile building threads, each invocation will need to check, attach, and detach the current thread to the JVM.

core/src/text/fontContext.cpp
hjanetzek: The comma should only be inserted when there are arguments to be able to use the macro with LOGE("just a message")

The do while wrapping ensures that the macro behaves also syntactically like regular functions:   
http://stackoverflow.com/questions/1067226/c-multi-line-macro-do-while0-vs-scope-block

blair1618: Ah I see, thanks! Can we prevent or suppress the warnings from this?

hjanetzek: I'll look into cmake for setting compiler specific flags. This one needs 
-Wgnu-zero-variadic-macro-arguments, but gcc will complain about unknown flags..

hjanetzek:
@blair1618 I forgot to add that this branch is stacked on top of the refactoring branch. Unmerged now and silenced the clang warnings

core/src/text/fontContext.cpp
tallytalwar: Based on the discussions I have seen .. and based on Stroustrup's recommendation ... goto should be used when there is no other alternative and mostly in cases when you want to exit a nested loop. (this is also the example he gives in his books).

core/src/text/fontContext.cpp
```diff
@@ -40,35 +45,54 @@ void FontContext::unlock() {
...
+        goto fallback;
     }
 
     m_fonts.emplace(std::move(fontKey), font);
 
-    return true;
+    return font;
+
+fallback:
+    if (_tryFallback && m_fonts.size() > 0) {
```
tallytalwar: Are we not always certain that m_fonts will have at least one font.. (We are bundling one font always for debugging texts).

PR #302, opened 2015-09-24
Nested mixins

tallytalwar:
- [x] fix for unique style mixes (test commented right now)

### DIFF
--- a/core/src/scene/sceneLoader.cpp
+++ b/core/src/scene/sceneLoader.cpp
@@ -414,6 +414,7 @@ Node SceneLoader::propMerge(const std::string& propStr, const Mixes& mixes) {
     std::unordered_map<std::string, Mixes> mapMixes;
 
     for (const auto& mixNode: mixes) {
+        if (!mixNode.IsMap()) { continue; }
         if (Node propNode = mixNode[propStr]) {
             if (propNode.IsScalar() && propNode.as<std::string>() == "true") {      // OR Properties
                 node = propNode;
@@ -452,6 +453,7 @@ Node SceneLoader::shaderBlockMerge(const Mixes& mixes) {
 
     Node node;
     for (const auto& mixNode : mixes) {
+        if (!mixNode.IsMap()) { continue; }
         if (Node shaderNode = mixNode["shaders"]) {
             if (Node blocks = shaderNode["blocks"]) {
                 for (const auto& block : blocks) {
@@ -475,6 +477,7 @@ Node SceneLoader::shaderExtMerge(const Mixes& mixes) {
     std::unordered_set<std::string> uniqueList;
 
     for (const auto& mixNode : mixes) {
+        if (!mixNode.IsMap()) { continue; }
         if (Node shaderNode = mixNode["shaders"]) {
             if (Node extNode = shaderNode["extensions"]) {
                 if (extNode.IsScalar()) {
@@ -522,6 +525,44 @@ Node SceneLoader::mixStyle(const Mixes& mixes) {
     return styleNode;
 }
 
+Mixes SceneLoader::recursiveMixins(Mixes mixesIn, const std::string& styleName, Node styles) {
+    Mixes mixes;
+    mixes.swap(mixesIn);
+
+    static std::unordered_set<std::string> mixedStyles;
+
+    auto styleNode = styles[styleName];
+
+    if (Node mixNode = styleNode["mix"]) {
+        if (mixNode.IsScalar()) {
+            auto mixStyleName = mixNode.as<std::string>();
+            mixes.reserve(mixes.size() + 1);
+            if (mixedStyles.find(mixStyleName) != mixedStyles.end()) {
+                mixes.push_back(mixNode);
+            } else {
+                mixes = recursiveMixins(mixes, mixStyleName, styles);
+            }
+        } else if (mixNode.IsSequence()) {
+            mixes.reserve(mixes.size() + mixNode.size() + 1);
+            for (const auto& mixStyleNode : mixNode) {
+                auto mixStyleName = mixStyleNode.as<std::string>();
+                if (mixedStyles.find(mixStyleName) != mixedStyles.end()) {
+                    mixes.push_back(mixStyleNode);
+                } else {
+                    mixes = recursiveMixins(mixes, mixStyleName, styles);
+                }
+            }
+        } else {
+            logMsg("Error parsing mix param for style: %s. Expected scalar or sequence value\n", styleName.c_str());
+        }
+    }
+
+    mixes.push_back(styleNode);
+    mixedStyles.insert(styleName);
+
+    return mixes;
+}
+
 void SceneLoader::loadStyles(Node styles, Scene& scene) {
 
     Style* style = nullptr;
@@ -555,22 +596,11 @@ void SceneLoader::loadStyles(Node styles, Scene& scene) {
         }
 
         Mixes mixes;
-        if (Node mixNode = styleNode["mix"]) {
-            if (mixNode.IsScalar()) {
-                mixes.reserve(2);
-                mixes.push_back(styles[mixNode.as<std::string>()]);
-            } else if (mixNode.IsSequence()) {
-                mixes.reserve(mixNode.size() + 1);
-                for (const auto& mixStyleNode : mixNode) {
-                    mixes.push_back(styles[mixStyleNode.as<std::string>()]);
-                }
-            } else {
-                logMsg("Error parsing mix param for style: %s. Expected scalar or sequence value\n", styleName.c_str());
-            }
-        }
-        mixes.push_back(styles[styleName]);
+        mixes = recursiveMixins(mixes, styleName, styles);
 
         Node mixedStyleNode = mixStyle(mixes);
+        // Update styleNode with mixedStyleNode (for future uses)
+        styles[styleName] = mixedStyleNode;
 
         // Construct style instance using the merged properties
         if (Node baseNode = mixedStyleNode["base"]) {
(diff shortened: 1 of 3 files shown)


core/src/scene/sceneLoader.cpp
```diff
@@ -73,21 +73,15 @@ void SceneLoader::loadShaderConfig(Node shaders, Style& style, Scene& scene) {
...
     if (Node extensionsNode = shaders["extensions"]) {
-        std::vector<std::string> extensions;
-        if (extensionsNode.IsScalar()) {
-            extensions.push_back(extensionsNode.as<std::string>());
-        } else if(extensionsNode.IsSequence()) {
-            extensions.reserve(extensionsNode.size());
+        if(extensionsNode.IsSequence()) {
             for (const auto& extNode : extensionsNode) {
-                extensions.push_back(extNode.as<std::string>());
+                char buffer[1000]; //sufficient space
```
bcamper: http://quotes.lifehack.org/quote/bill-gates/640k-ought-to-be-enough-for-anybody/

core/src/scene/sceneLoader.cpp
```diff
@@ -73,21 +73,15 @@ void SceneLoader::loadShaderConfig(Node shaders, Style& style, Scene& scene) {
...
+                auto extName = extNode.as<std::string>();
+                sprintf(buffer, "#ifdef %s\n    #extension %s : enable\n    #define TANGRAM_EXTENSION_%s\n#endif",
```
bcamper: This is after extensions have been mixed together, so each one will only be defined once, right?

bcamper: There's no requirement for extensions to be enabled at the top of the shader? I always get confused about that.

core/src/scene/sceneLoader.cpp
```diff
@@ -420,6 +414,7 @@ Node SceneLoader::propMerge(const std::string& propStr, const Mixes& mixes) {
     std::unordered_map<std::string, Mixes> mapMixes;
 
     for (const auto& mixNode: mixes) {
+        if (!mixNode.IsMap()) { continue; }
```
bcamper: When does this happen?

tallytalwar: Say someone wrote an empty style in the style sheet... before you start checking for map values in line 418, need to make sure this node is a Map node.

tests/unit/styleMixTests.cpp
```diff
@@ -13,6 +13,54 @@
 using namespace Tangram;
 using YAML::Node;
 
+TEST_CASE( "Style Mixing Test: Nested Style Mixin Nodes", "[mixing][core][yaml]") {
+    SceneLoader sceneLoader;
+    Mixes mix;
+    Node mixNode;
+    Node node = YAML::Load(R"END(
```
bcamper: I would add tests where the styles are defined out of dependency order? I know order is not supposed to be guaranteed, but I had a case in JS where mixing only worked in linear order, but that order was happening most of the time.

core/src/scene/sceneLoader.cpp
tallytalwar: Yes, injection will happen after all mixing has been done.

wrt requirement for extensions to be at the top of the shader... I think there is no documentation saying that, but we have seen some issues in the past which were fixed when extensions were put at the top.

tests/unit/styleMixTests.cpp
tallytalwar: I have a test where a style is mixed out of order: https://github.com/tangrams/tangram-es/pull/302/files#diff-f021195bbca063598e222f41b7a41950R29

tallytalwar:
Kindly review/merge this before merging any of the sceneLoader refactor.

core/src/scene/sceneLoader.cpp
blair1618: More seriously though, there's no reason to use `sprintf` now that `snprintf` is available (http://www.cplusplus.com/reference/cstdio/snprintf/)

tallytalwar: Sure, can use snprintf :D.

tallytalwar: 3ac8916 updates this.

hjanetzek: We should consider using https://github.com/c42f/tinyformat as typesafe c++11 printf replacement. It is likely to be more efficient than our common string concatenation and much nicer to write formatting specifiers with it.

Just tests: When adding a version using std::string concatenation it is actually slower, even without precision formatting specifiers
https://github.com/c42f/tinyformat/blob/master/tinyformat_speed_test.cpp

tallytalwar: There is also this: https://github.com/cppformat/cppformat

hjanetzek: Looks also good and more optimized. Tinyformat is more rather a nice interface to std iostreams.

core/src/scene/sceneLoader.h
```diff
@@ -27,6 +28,7 @@ struct MaterialTexture;
 struct Filter;
 
 using Mixes = std::vector<YAML::Node>;
+using UniqueStyles = std::unordered_set<std::string>;
```
blair1618: `Mixes` and `UniqueStyles` seem like they should rightly be variable names, not types.

tallytalwar: Sorry, did not get you here!
`Mixes` and `UniqueStyles` are just type aliases.

blair1618: I mean that I don't think these type aliases are needed. I think the code would be clearer if the actual type were used; the intent of the relevant variables is already expressed in their names.

core/src/scene/sceneLoader.cpp
```diff
@@ -528,6 +525,51 @@ Node SceneLoader::mixStyle(const Mixes& mixes) {
     return styleNode;
 }
 
+Mixes SceneLoader::recursiveMixins(Mixes mixIn, const std::string& styleName, Node styles, UniqueStyles& uniqueStyles) {
+    Mixes mixes;
+    mixes.swap(mixIn);
+
+    static std::unordered_set<std::string> mixedStyles;
```
blair1618: This is gonna be a problem if scene loading is ever performed more than once; entries will remain in this set for the entire duration of the program execution.

core/src/scene/sceneLoader.cpp
```diff
@@ -528,6 +525,51 @@ Node SceneLoader::mixStyle(const Mixes& mixes) {
     return styleNode;
 }
 
+Mixes SceneLoader::recursiveMixins(Mixes mixIn, const std::string& styleName, Node styles, UniqueStyles& uniqueStyles) {
+    Mixes mixes;
+    mixes.swap(mixIn);
```
blair1618: Seems like just using the input `mixIn` would be equivalent, am I missing something?

core/src/scene/sceneLoader.cpp
tallytalwar: True, good catch. I was just avoiding a member variable there... Can fix it.

core/src/scene/sceneLoader.h
tallytalwar: Sure... will just make these function signatures huge.
And because of that will use the new way of defining function signatures.

core/src/scene/sceneLoader.cpp
```diff
@@ -528,6 +525,51 @@ Node SceneLoader::mixStyle(const Mixes& mixes) {
...
+
+    static std::unordered_set<std::string> mixedStyles;
+
+    auto styleNode = styles[styleName];
+
+    if (Node mixNode = styleNode["mix"]) {
+        if (mixNode.IsScalar()) {
+            auto mixStyleName = mixNode.as<std::string>();
+            mixes.reserve(mixes.size() + 1);
```
blair1618: `reserve` not needed here - `push_back` will do precisely the same thing /nitpick

tallytalwar:
e999646 updates a bunch of stuff mentioned in the past conversations.

hjanetzek:
@tallytalwar I'm all for the new syntax, but there is a pending SceneLoader refactoring PR where this change would better fit in and avoid major pita when merging :/

tallytalwar:
@hjanetzek sure can revert that.

hjanetzek:
Thanks!

core/src/scene/sceneLoader.h
```diff
@@ -43,15 +43,14 @@ class SceneLoader {
...
     void loadShaderConfig(YAML::Node shaders, Style& style, Scene& scene);
     SceneLayer loadSublayer(YAML::Node layer, const std::string& name, Scene& scene);
-    MaterialTexture loadMaterialTexture(YAML::Node matCompNode, Scene& scene);
     Filter generateAnyFilter(YAML::Node filter, Scene& scene);
     Filter generateNoneFilter(YAML::Node filter, Scene& scene);
     Filter generatePredicate(YAML::Node filter, std::string _key);
 
-    // Style Mixing helper methods
-    YAML::Node mixStyle(const Mixes& mixes);
+    std::unordered_set<std::string> m_mixedStyles;
```
blair1618: Looking ahead to the scene-reloading PRs (in which all of `SceneLoader`'s functions become static) it would make sense to ensure that `m_mixedStyles` will behave correctly in that case (or, better yet, remove the need for it at all)

core/src/scene/sceneLoader.h
```diff
@@ -64,16 +63,25 @@ class SceneLoader {
...
+     * @styleName: styleName to be mixed
+     * @styles: YAML::Node for all styles
+     * @uniqueStyles: to make sure Mixes returned is a uniqueSet
+     */
+    std::vector<YAML::Node> recursiveMixins(const std::string& styleName, YAML::Node styles,
+            std::unordered_set<std::string>& uniqueStyles);
 
     // Generic methods to merge properties
-    YAML::Node propMerge(const std::string& propStr, const Mixes& mixes);
+    YAML::Node propMerge (const std::string& propStr, const std::vector<YAML::Node>& mixes);
```
blair1618: stray space after function name

core/src/scene/sceneLoader.h
tallytalwar: Hmm ... I think it can be marked static once all `SceneLoader` functions are static. It gets cleared once style mixin nodes are determined.

blair1618:
Waiting on Travis, but looks alright to me!

PR #303, opened 2015-09-24
Draw text stroke using SDF

blair1618:
screen shot 2015-09-24 at 7 09 22 pm

Text labels are starting to look quite nice! This is just using our current signed distance field technique and (basically) just drawing a stroke color between the "glyph interior" threshold and another threshold. 

Things to revisit later:

1. Sufficiently large strokes will encroach on neighboring glyphs. This is a little complicated to fix, we'll probably need another draw pass on text with stroke. 
2. Corners of strokes are round, as opposed to squared like a canvas-drawn stroke. This can be fixed with a small change to fontstash, but isn't a major issue for now.

### DIFF
--- a/core/src/scene/styleContext.cpp
+++ b/core/src/scene/styleContext.cpp
@@ -233,7 +233,6 @@ bool StyleContext::parseStyleResult(StyleParamKey _key, StyleParam::Value& _val)
             case StyleParamKey::color:
             case StyleParamKey::outline_color:
             case StyleParamKey::font_fill:
-            case StyleParamKey::font_stroke:
             case StyleParamKey::font_stroke_color: {
                 if (len < 3 || len > 4) {
                     logMsg("Warning: Wrong array size for color: '%d'.\n", len);
@@ -283,7 +282,6 @@ bool StyleContext::parseStyleResult(StyleParamKey _key, StyleParam::Value& _val)
             case StyleParamKey::color:
             case StyleParamKey::outline_color:
             case StyleParamKey::font_fill:
-            case StyleParamKey::font_stroke:
             case StyleParamKey::font_stroke_color: {
                 _val = static_cast<uint32_t>(duk_get_uint(m_ctx, -1));
                 break;
--- a/core/src/text/textBuffer.cpp
+++ b/core/src/text/textBuffer.cpp
@@ -87,7 +87,11 @@ bool TextBuffer::addLabel(const std::string& _text, Label::Transform _transform,
     float inf = std::numeric_limits<float>::infinity();
     float x0 = inf, x1 = -inf, y0 = inf, y1 = -inf;
 
-    uint32_t strokeWidth = _params.strokeWidth * 255.;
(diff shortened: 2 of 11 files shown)


core/src/scene/styleContext.cpp
```diff
@@ -233,7 +233,6 @@ bool StyleContext::parseStyleResult(StyleParamKey _key, StyleParam::Value& _val)
             case StyleParamKey::color:
             case StyleParamKey::outline_color:
             case StyleParamKey::font_fill:
-            case StyleParamKey::font_stroke:
```
tallytalwar: [Documentation](https://github.com/tangrams/tangram-docs/blob/gh-pages/pages/draw.md#stroke), says `stroke` can be a `scalar` value, which will get mapped to the color of the stroke. But I am not sure how useful this is!

bcamper: Hmm I believe that documentation is wrong. Should only accept:

`stroke: { color: white, width: 5px }`

cc @meetar

meetar: Thanks - fixed with https://github.com/tangrams/tangram-docs/commit/92fe9234813ab9c2ff2f16e2917c764346044bbc

tallytalwar: :+1:

core/src/text/textBuffer.cpp
```diff
@@ -87,16 +87,23 @@ bool TextBuffer::addLabel(const std::string& _text, Label::Transform _transform,
     float inf = std::numeric_limits<float>::infinity();
     float x0 = inf, x1 = -inf, y0 = inf, y1 = -inf;
 
+    // Stroke width is normalized by the distance of the SDF spread, then scaled
+    // to a char, then packed into the "alpha" channel of stroke. The .25 scaling
+    // probably has to do with how the SDF is generated, but honestly I'm not sure
```
tallytalwar: Also curious on this 0.25 scaling.

blair1618: Ok buckle up, here's my model of this:

The SDF for a glyph goes from 0 to 255 over a distance of `blurSpread` pixels. `strokeWidth` is in pixels, so to find its SDF threshold we divide by `blurSpread` and multiply by 255. Now recall that one half of the SDF value range is the interior of contours, so to make sure our stroke is measured from the outside of a contour we multiply by 0.5. 

This is the part where I start obsessively measuring stroke pixel widths in both tangram-js and tangram-es, because the strokes I was getting were twice the size that they were in the browser. After an extended discussion we discovered that tangram-js is probably producing half-sized strokes, so to match that here we add another factor of 0.5 and now the whole thing is scaled by 0.25.

tallytalwar: :+1:

meetar: I imagine we'll change that behavior in js so that 5px == 5px…

bcamper: Probably but good to maintain parity for now? :)

PR #305, opened 2015-09-25
Fix lighting mix

tallytalwar:
This PR fixes a couple of things:

- Lighting property is special in the case that it can be boolean or scalar/string value. [`true`, `false`, `vertex`, `fragment`]. In which case it should be overwritten when mixing styles properties. This PR, explicitly calls either OR, or Merge methods to appropriate properties explicitly. (As has been done on the JS side). Hence handles `lighting` property correctly.

- Another issue which I noticed was this special behavior yamlcpp Nodes. Nodes are non-copyable and hence needs to be explicitly cloned. Since this was not being done, propMerge method was modifying the original scene yaml nodes. Refer: https://code.google.com/p/yaml-cpp/wiki/HowToParseADocument#Note_about_copying_YAML::Node.

### DIFF
--- a/core/src/scene/sceneLoader.cpp
+++ b/core/src/scene/sceneLoader.cpp
@@ -410,9 +410,13 @@ Node SceneLoader::propOr(const std::string& propStr, const std::vector<Node>& mi
     for (const auto& mixNode: mixes) {
         if (Node propNode = mixNode[propStr]) {
             if (propNode.IsScalar()) {
-                if (propNode.as<std::string>() == "true") {
-                    node = true;
-                    break;
+                try {
+                    if (propNode.as<bool>()) {
+                        node = true;
+                        break;
+                    }
+                } catch (const BadConversion& e) {
+                    logMsg("Error: Expected a boolean value for %s.\n", propStr.c_str());
                 }
             }
         }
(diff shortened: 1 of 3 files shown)


tallytalwar:
As with https://github.com/tangrams/tangram-es/pull/302, request to review/merge this PR before merging any of sceneLoader refactor PRs.

core/src/scene/sceneLoader.cpp
```diff
@@ -408,6 +408,23 @@ void SceneLoader::loadStyleProps(Style* style, YAML::Node styleNode, Scene& scen
...
 }
 
+Node SceneLoader::propOr(const std::string& propStr, const Mixes& mixes) {
+
+    Node node;
+
+    for (const auto& mixNode: mixes) {
+        if (Node propNode = mixNode[propStr]) {
+            if (propNode.IsScalar()) {
+                if (propNode.as<std::string>() == "true") {
```
blair1618: YAML allows a few variations on Booleans (e.g. "true", "True", "TRUE") so for consistency we may want to use `as<bool>` here.

bcamper: Good call

blair1618:
Looks good to me - will merge

hjanetzek:
@tallytalwar, or I lost it during resolving conflicts in merging..

PR #307, opened 2015-09-25
Builtin Uniforms

tallytalwar:
- [x] u_map_position (priority for eraser map)
- [x] u_time
- [x] u_resolution

### DIFF
--- a/core/src/tangram.cpp
+++ b/core/src/tangram.cpp
@@ -104,11 +104,6 @@ void update(float _dt) {
 
     m_view->update();
 
-    // Set time uniforms for all style's shader programs
-    for (auto& style : m_scene->styles()) {
-        style->getShaderProgram()->setUniformf("u_time", _dt);
-    }
-
     {
         std::lock_guard<std::mutex> lock(m_tilesMutex);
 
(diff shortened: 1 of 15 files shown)


core/src/tangram.cpp
```diff
@@ -104,6 +104,11 @@ void update(float _dt) {
 
     m_view->update();
 
+    // Set time uniforms for all style's shader programs
+    for (auto& style : m_scene->styles()) {
+        style->getShaderProgram()->setUniformf("u_time", _dt);
```
blair1618: Should be `g_time`, not `_dt`

karimnaaji: This also might be best in the render loop, we have the update/render in the main loop, but it might be safe to simply let update not involve any gl call if possible.

blair1618: I agree actually, would be better to separate keep gl calls out of update()

tallytalwar: I also agree :D. WIll fix.

blair1618:
All good, will merge!

blair1618:
Err hang on, why aren't we adding all of the built-ins?

https://github.com/tangrams/tangram-docs/blob/gh-pages/pages/shaders.md#built-in-uniforms

tallytalwar:
We already had a chat about it.

I wanted to first get the priority ones in for eraser map.

PR #308, opened 2015-09-26
assemble shader source incrementally

hjanetzek:
Edit: This branch now contains only a slightly more sophisticated way for injecting source blocks. (and much less big string copies and memmoves)

~~This patch makes multiple point lightsources work. The issue was actually the additional '\n' inserted in addSourceBlock~~

### DIFF
--- a/core/src/gl/shaderProgram.cpp
+++ b/core/src/gl/shaderProgram.cpp
@@ -4,6 +4,10 @@
 #include "scene/light.h"
 #include "gl/renderState.h"
 
+#include <sstream>
+#include <regex>
+#include <set>
+
 namespace Tangram {
 
 int ShaderProgram::s_validGeneration = 0;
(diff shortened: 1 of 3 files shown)


bcamper:
Why was the extra line breaking the shader?

hjanetzek:
@bcamper I've pushed the small fix only: https://github.com/tangrams/tangram-es/commit/5aa31503e5eed4ae6cf65aac48b5902aa5c842e4

bcamper:
Ah, thanks, I see, was just wondering why the \n caused a problem, I see
now.

core/src/gl/shaderProgram.cpp
```diff
@@ -215,43 +223,76 @@ GLuint ShaderProgram::makeCompiledShader(const std::string& _src, GLenum _type)
 
 }
 
-void ShaderProgram::applySourceBlocks(std::string& _vertSrcOut, std::string& _fragSrcOut) {
+static const std::regex pragmaLine("^\\s*#pragma tangram:\\s+(\\w+).*$");
```
tallytalwar: this can be a local static const in `ShaderProgram::applySourceBlocks`. I always prefer not to have global variables wherever possible.

hjanetzek: It's still local for the compilation unit (the static is also redundant in that place). But yes, good to reduce scope where possible

PR #309, opened 2015-09-28
Android HTTP handler

blair1618:
Allows network request behavior to be customized on Android (in our case, for adding API keys)

### DIFF
--- a/android/tangram/src/com/mapzen/tangram/HttpHandler.java
+++ b/android/tangram/src/com/mapzen/tangram/HttpHandler.java
@@ -21,12 +21,22 @@ public class HttpHandler {
         okClient.setReadTimeout(30, TimeUnit.SECONDS);
     }
 
+    /**
+     * Begin an HTTP request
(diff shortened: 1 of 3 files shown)


tallytalwar:
Reviewing this now.

android/tangram/src/com/mapzen/tangram/HttpHandler.java
```diff
@@ -0,0 +1,58 @@
...
+public class HttpHandler {
+
+    private OkHttpClient okClient;
+    private Request.Builder okRequestBuilder;
+
+    public HttpHandler() {
+        okRequestBuilder = new Request.Builder();
+        okClient = new OkHttpClient();
+        okClient.setConnectTimeout(10, TimeUnit.SECONDS);
+        okClient.setReadTimeout(30, TimeUnit.SECONDS);
```
tallytalwar: Should we provide setters for these timeouts also, like we do for cache?
Nothing urgent though. (will merge it like this for now).

PR #311, opened 2015-09-28
Layers can select data from multiple collections in a source

blair1618:


core/src/scene/sceneLoader.cpp
```diff
@@ -1228,15 +1228,30 @@ void SceneLoader::loadLayers(Node layers, Scene& scene, TileManager& tileManager
...
+                source = data_source.Scalar();
+            }
+        }
+
+        std::vector<std::string> collections;
+        if (Node data_layer = data["layer"]) {
+            if (data_layer.IsScalar()) {
+                collections.push_back(data_layer.Scalar());
+            } else if (data_layer.IsSequence()) {
+                collections = data_layer.as<std::vector<std::string>>();
```
blair1618: Just want to point out how cool it is that yaml-cpp supports STL types like this, we should use this more often

tallytalwar: This is nice. Didn't know it provides default implementations for stl container conversions.

tallytalwar:
Looks good to me. Once couple of questions are answered/fixed can be merged.

core/src/scene/dataLayer.cpp
```diff
@@ -2,9 +2,9 @@
 
 namespace Tangram {
 
-DataLayer::DataLayer(SceneLayer _layer, const std::string& _source, const std::string& _collection) :
+DataLayer::DataLayer(SceneLayer _layer, const std::string& _source, const std::vector<std::string>& _collections) :
     SceneLayer(_layer),
     m_source(_source),
-    m_collection(_collection) {}
+    m_collections(_collections) {}
```
tallytalwar: We can do a `std::move` here.

P.S. github app does not respect where you comment :(.

core/src/scene/sceneLoader.cpp
```diff
@@ -1228,15 +1228,30 @@ void SceneLoader::loadLayers(Node layers, Scene& scene, TileManager& tileManager
...
         Node data = layer.second["data"];
-        Node data_layer = data["layer"];
-        Node data_source = data["source"];
 
-        std::string collection = data_layer ? data_layer.as<std::string>() : name;
-        std::string source = data_source ? data_source.as<std::string>() : "";
+        std::string source;
+        if (Node data_source = data["source"]) {
+            if (data_source.IsScalar()) {
+                source = data_source.Scalar();
```
tallytalwar: a log message should be good here, if `data_source` is not scalar.

core/src/scene/dataLayer.cpp
blair1618: Could, but it's a small, one-time copy

core/src/scene/sceneLoader.cpp
blair1618: Warnings would be good in many places while loading the style sheet, I could add 3 to 5 different warnings in this block alone, so before peppering log messages everywhere I'd like to think about a more general validation step before reading configuration at all.

PR #312, opened 2015-09-28
Background color configured from scene file

blair1618:
Also adds a color struct that will save us some headache in other parts of the code

### DIFF
--- a/core/src/scene/sceneLoader.h
+++ b/core/src/scene/sceneLoader.h
@@ -33,7 +33,7 @@ using StyleUniforms = std::pair<std::string, std::vector<UniformValue>>;
 class SceneLoader {
 
     static void loadBackground(YAML::Node background, Scene& scene);
-	static void loadSources(YAML::Node sources, Scene& scene);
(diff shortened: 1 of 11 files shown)


core/src/scene/sceneLoader.h
```diff
@@ -32,7 +32,8 @@ using StyleUniforms = std::pair<std::string, std::vector<UniformValue>>;
 
 class SceneLoader {
 
-    static void loadSources(YAML::Node sources, Scene& scene);
+    static void loadBackground(YAML::Node background, Scene& scene);
+	static void loadSources(YAML::Node sources, Scene& scene);
```
tallytalwar: tabs mis-aligned.

core/src/util/color.h
```diff
@@ -0,0 +1,17 @@
+#pragma once
+
+namespace Tangram {
+    
+struct Color {
```
tallytalwar: Getting a bunch of warning on anonymous struct and anonymous unions.

PR #315, opened 2015-09-29
More uniforms (missing u_model)

hjanetzek:
Add one missing uniform for u_model and set all the view uniforms in Style instead of Tile

### DIFF
--- a/core/resources/shaders/polygon.fs
+++ b/core/resources/shaders/polygon.fs
@@ -6,6 +6,7 @@ precision highp float;
 
 #pragma tangram: defines
 
+uniform mat4 u_model;
 uniform mat4 u_modelView;
 uniform mat4 u_modelViewProj;
 uniform mat3 u_normalMatrix;
--- a/core/resources/shaders/polygon.vs
+++ b/core/resources/shaders/polygon.vs
@@ -6,6 +6,7 @@ precision highp float;
 
 #pragma tangram: defines
 
+uniform mat4 u_model;
 uniform mat4 u_modelView;
 uniform mat4 u_modelViewProj;
 uniform mat3 u_normalMatrix;
--- a/core/resources/shaders/polyline.fs
+++ b/core/resources/shaders/polyline.fs
@@ -6,6 +6,7 @@ precision highp float;
 
 #pragma tangram: defines
 
+uniform mat4 u_model;
 uniform mat4 u_modelView;
 uniform mat4 u_modelViewProj;
 uniform mat3 u_normalMatrix;
--- a/core/resources/shaders/polyline.vs
+++ b/core/resources/shaders/polyline.vs
@@ -8,6 +8,7 @@ precision highp float;
 
 #define TANGRAM_WORLD_POSITION_WRAP vec3(100000.0)
 
+uniform mat4 u_model;
 uniform mat4 u_modelView;
 uniform mat4 u_modelViewProj;
 uniform mat3 u_normalMatrix;
--- a/core/src/style/style.cpp
+++ b/core/src/style/style.cpp
@@ -7,6 +7,8 @@
 #include "gl/vboMesh.h"
 #include "view/view.h"
 
+#include "glm/gtc/type_ptr.hpp"
+
 namespace Tangram {
 
     Style::Style(std::string _name, Blending _blendMode, GLenum _drawMode) :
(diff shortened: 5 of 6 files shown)


blair1618:
The `u_model` and `u_modelView` that we use now aren't quite the same as the ones provided in tangram-js:

 - The js `u_model` is from the absolute world origin to the tile while ours is from the center of the map view to the map tile.
 - The js `u_modelView` doesn't include transforming `z` into camera space, so that in the `position` shader block, `position.z = 0` will place something on the ground
 - In js, when you change the value of `position` in a vertex shader, the operation is in meters.. which means that camera viewport is in.. meters?

While trying to enumerate the differences here I've realized that I'm probably a little confused about this myself. Perhaps @bcamper can illuminate? :)

bcamper:
Yes, `u_model` translates tile coordinates into world space (this is my understanding of the [traditional model matrix](http://www.opengl-tutorial.org/beginners-tutorials/tutorial-3-matrices/#The_Model__View_and_Projection_matrices)).

Yes, `u_modelView` does **not** do any `z` transformation, so that users modifying the vertex `position` in a shader block can treat `z = 0` as the ground plane. While this is unconventional, this seemed a lot easier for users to reason about than to make `z` some negative distance from the camera (ground plane would be camera height * -1). I considered calling this matrix something other than "model view" for that reason, but ultimately decided it was a matter of semantics, and might be more confusing than to just use the MVP terminology. (Side note, it wasn't practical to let users modify vertex position directly in absolute world coordinates before the view matrix was applied, which might have been more more straightforward, because of precision issues with 32-bit floats and spherical mercator space.)

`u_modelView` does keep coordinates into meters, with XY relative to the camera. Meters seemed appropriate to me because again, my understanding of a typical view matrix is that it will leave coordinates in world units, but relative to the camera (rather than absolute). 

The `u_projection` matrix is a standard perspective matrix (for perspective camera), with the camera placed at `[0, 0, height]`, where camera height is in meters: https://github.com/tangrams/tangram/blob/master/src/camera.js#L169 (`height * 2` is the far clipping plane). The `z` translate step [is included](https://github.com/tangrams/tangram/blob/master/src/camera.js#L191) in the projection matrix. 

Additional, currently WebGL-specific functionality is applied to [tweak the apparent vanishing point](https://github.com/tangrams/tangram/blob/master/src/camera.js#L171-L177) of the camera while leaving the visible ground area un-changed -- this is to match the viewable area of traditional web mercator maps.

Let me know if any of that doesn't make sense... there are some compromises, but all in all it seemed to be the most practical balance.

blair1618:
Awesome, that helps a lot! I think we can make this work in a compatible way, but there are some further considerations for the view system we have in tangram-es.

In js, the `position` block comes directly after multiplying `position` by `u_modelView`, so the `position` that a user is modifying is relative to the camera in x & y and relative to the ground in z - makes sense so far. But when your view matrix includes tilting and rotation, it's no longer very intuitive to manipulate `position` after being multiplied by `u_modelView`. This problem is why I've made the model matrix in tangram-es is a little different. 

Our `u_model` is the transformation from a world origin to the local coordinates of a model (tile), and the world origin we use is the current center of the map view (as opposed to the origin of the web-mercator coordinate system). This is so that our vertex pipeline could work something like this: 

```
vec4 position = u_model * a_position;

#pragma tangram: position
// do position transformations in meters, x and y correspond to meters East and North 
// from the current map center and z corresponds to meters above the ground

gl_Position =  position * u_viewProjection;
```

In a way, it's like the tangram-es `u_model` is like the tangram-js `u_modelView` and the tangram-es `u_viewProjection` is like the tangram-js `u_projection`

bcamper:
One proposal is to provide a set of constant global variables that give consistent results to commonly needed values, as an alternative to directly exposing the matrices themselves (at least not exposing in a supported/documented way). So far we've discussed the need for:

- position in tile space (`g_model_position`?)
   - this has different scales in JS (tile bound is [0, 4096]) and ES (tile bound is [-1, 1]), a global var would normalize this to a common scale (and/or we can start storing them in the same space)
   - position is VBO packed in different ways in JS (normalized short) and ES (float), a global var would account for this to hide the unpacking from the user
   - `#define UNPACK_SHORT(x) (x * 32767.)`
   - `#define TANGRAM_TILE_SCALE 4096.`
   - `const vec3 g_model_position = UNPACK_SHORT(a_position.xyz) / TANGRAM_TILE_SCALE;`
- position in world space (`g_world_position`?)
   - for procedural textures, re-projecting, etc.
   - this has precision issues but can still be useful
- position in wrapped world space (`g_wrapped_world_position`?)
   - modified form of world space, wrapped to a scale that allows for procedural textures without precision issues

Any others? The proposed names above aren't that great, but just getting notes down for now. Thoughts on this approach?

blair1618:
I like the idea of providing global variables for these values. It would be nice if we could just initialize them as `const` so to enforce that they aren't accidentally written to, but GLSL only allows `const` for compile-time constants and function arguments :\

We could provide macros that act like read-only variables:
`#define WORLD_POSITION vec3(g_world_position)`
Or provide functions that return the desired value:
`vec3 world_position() { return thing; }`
Or just let the globals fend for themselves and document that they *shall not* be written to. Can't really think of anything nicer at the moment. 

Names are always hard. We could say `local` and `global` instead of `model` and `world`, but I could live with either as long as it's consistent and documented. 

There's a tiny voice nagging me that this might be somewhat specific to web-mercator; do other projections have comparable cartesian mappings?

blair1618:
So @bcamper would it be sane to try to normalize `g_model_position` to [0, 1] for coordinates within a tile? The same will need to be done on JS for eraser-map to be cross-compatible.

bcamper:
I think so, will give it a try later today.

blair1618:
Closing for now, will return to this when we decide how to normalize `u_model` between here and tangram-js.

bcamper:
See https://github.com/tangrams/tangram/pull/197

PR #316, opened 2015-09-29
[WIP] add support for Meter width units

hjanetzek:


core/src/scene/stops.cpp
```diff
@@ -11,33 +11,48 @@
...
-            // parse distance from string
-            float value = frameNode[1].as<float>();
-            frames.emplace_back(key, value);
+
+        StyleParam::Width widthValue;
+        widthValue.unit = Unit::meter;
+        size_t start = 0;
+
+        if (StyleParam::parseValueUnitPair(frameNode[1].Scalar(), start, widthValue)){
+            stops.frames.emplace_back(key, widthValue);
```
blair1618: I think we can normalize widths into "pixels" here - since widths are paired with a zoom, the conversion is unambiguous and we don't need to keep their units around.

hjanetzek: I though so too, but then tile-size and map-projection need to be known in the SceneLoader and one could not modify the View map-projection and pixel-scale after sceneloading (which one should probably not do at the moment anyway though)

blair1618: I see what you mean. I just wouldn't want to go through too much trouble for that because I'm not sure how important it is to change projections without reloading the scene.

core/src/scene/styleContext.cpp
```diff
@@ -274,7 +274,7 @@ bool StyleContext::parseStyleResult(StyleParamKey _key, StyleParam::Value& _val)
             case StyleParamKey::outline_width:
             case StyleParamKey::font_stroke_width: {
                 double v = duk_get_number(m_ctx, -1);
-                _val = static_cast<float>(v);
+                _val = StyleParam::Width{static_cast<float>(v)};
```
blair1618: In `StyleParam`, `font_stroke_width` is set as a `float`, here it's set as a `StyleParam::Width`

core/src/scene/stops.cpp
hjanetzek: ok, I'll look into doing the conversion in the Stops::Width constructor

blair1618:
I can see now how it would be nicer to keep the units along with distances in Stops - thanks for figuring out that edge case with the last Frame in meters. Now I think it makes more sense to store a sequence of values with units in Stops. May also make sense to separate the "value-with-unit" type into its own file. 

The behavior seems to be correct, so for the sake of enabling work on the eraser-map style we can merge this now (with the test styles reverted) and I may return to this soon :)

blair1618:
Ok wait - I am no longer convinced that this behaves 100% correctly. With the test styles provided, I'm seeing a discontinuity in width of highway outlines between zooms 15 and 16

hjanetzek:
@blair1618,  indeed there were two bugs ..

For the stops I think it would make sense to store the change to the next stop, i.e. exp2(upper->key - lower->key) - 1.0, instead of introducing width units. But we can decide that later

core/src/scene/stops.cpp
```diff
@@ -11,33 +14,109 @@
...
         } else {
-            // parse distance from string
-            float value = frameNode[1].as<float>();
-            frames.emplace_back(key, value);
+            logMsg("could not parse node %s\n", Dump(frameNode[1]).c_str());
         }
     }
+    // Append stop at max-zoom to continue scaling after the last stop
+    // TODO: define MAX_ZOOM == 24
+    if (lastIsMeter && lastKey < 24) {
```
tallytalwar: Should max zoom be same as the view max zoom, [here](https://github.com/tangrams/tangram-es/blob/master/core/src/view/view.h#L122)?

blair1618: Only has to be not smaller

PR #300, opened 2015-09-24
Sceneloader refactoring

hjanetzek:
No question the SceneLoader got messy - This is a first attempt for a major cleanup. Better look at the new SceneLoader before trying to follow the whole diff

+ improved error logs with dump of unparseable nodes

Actually the diff is too big for github to show it..
https://github.com/tangrams/tangram-es/blob/sceneloader-refactoring/core/src/scene/sceneLoader.cpp

PR #317, opened 2015-09-30
Sceneloader refactoring

hjanetzek:
Continuing #300

### DIFF
--- a/core/src/data/filters.h
+++ b/core/src/data/filters.h
@@ -47,34 +47,38 @@ namespace Tangram {
         };
 
         FilterType type;
-        variant<none_type, Operator, Equality, Range, Existence, Function> data;
+        using Data = variant<none_type, Operator, Equality, Range, Existence, Function>;
+        Data data;
 
         Filter() : type(FilterType::undefined), data(none_type{}) {}
+        Filter(FilterType _type, Data _data) : type(_type), data(std::move(_data)) {}
 
         // Create an 'any', 'all', or 'none' filter
-        Filter(Operators op, const std::vector<Filter>& filters) :
-            type(static_cast<FilterType>(op)),
-            data(Operator{ filters }) {}
-
+        inline static Filter MatchAny(const std::vector<Filter>& filters) {
+            return { FilterType::any,  Operator{ filters }};
+        }
+        inline static Filter MatchAll(const std::vector<Filter>& filters) {
+            return { FilterType::all,  Operator{ filters }};
+        }
+        inline static Filter MatchNone(const std::vector<Filter>& filters) {
+            return { FilterType::none, Operator{ filters }};
+        }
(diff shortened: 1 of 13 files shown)


blair1618:
This is looking good! I think Varun is reviewing as well.

Much cleaner, thanks :)

hjanetzek:
Thanks! 

Just saw that I introduced a small breakage while removing GL_ prefix from extensions..

hjanetzek:
All changes that went in with the latest rebase https://gist.github.com/hjanetzek/4c62b45f4695ee326124

tallytalwar:
@hjanetzek 
We also need to `Clone` the node when `propValue` is a `sequence` in `propMerge`.
https://github.com/tangrams/tangram-es/blob/sceneloader-refactoring/core/src/scene/sceneLoader.cpp#L527

blair1618:
Looks like SDF multi-sampling is back on by default (https://github.com/tangrams/tangram-es/blob/sceneloader-refactoring/core/src/scene/sceneLoader.cpp#L766)

I had turned this off because it breaks font rendering on the Galaxy S3 (and probably others). We should fix this eventually, I just didn't want to worry about it for this launch.

hjanetzek:
@tallytalwar @blair1618, changes are in. thanks! :)

tallytalwar:
One small change and we are good with this:

Can we use yaml bool conversions here?

https://github.com/tangrams/tangram-es/blob/sceneloader-refactoring/core/src/scene/sceneLoader.cpp#L1019-L1022

hjanetzek:
This case is a bit different as we 'steal' here two possible values that could be used for MatchEquality for the Existence filter, when we use yaml booleans here we could no check for things like building_part:yes/no OSM tags

tallytalwar:
That's what we thought before, but based on the yaml 1.2 spec (
http://www.yaml.org/spec/1.2/spec.html#id2803629), it seems only
"true"/"True"/"TRUE" and respectively "false" values are legit.

"yes" or "no" will not be bool converted.

hjanetzek:
ok, then we should also update yaml-cpp which still supports all 1.1 boolean tokens

tallytalwar:
Ohh I thought yamlcpp is based on 1.2 spec and hence we went with it for
our parser.

If that is the case, then yes let's keep it like this for now and get
yamlcpp in place first.

hjanetzek:
Yea. I'll do a follow up patch for this and yaml-cpp

blair1618:
Oh that's awful. yaml-cpp claims to follow the 1.2 spec but you're totally right, it accepts the other boolean tokens D:

Let's fix that in our fork and submit a PR to the main repo (probably won't be merged for a long time, but I suppose it's the nice thing to do)

bcamper:
ahhhhhhhhhhh

Issue #310, opened 2015-09-28
Offsets for line labels should be in local coordinates of line


blair1618:
`offset: [0, 50px]`

in tangram play:
![screen shot 2015-09-28 at 1 03 44 pm](https://cloud.githubusercontent.com/assets/3371850/10142652/fb121dc4-65e1-11e5-9be1-963349152a6d.png)

in tangram-es:
![screen shot 2015-09-28 at 1 04 20 pm](https://cloud.githubusercontent.com/assets/3371850/10142662/076ebf8c-65e2-11e5-9390-d00d4b5ac5e2.png)

blair1618:
Fixed by #321

PR #321, opened 2015-10-01
Offsets relative to local coordinate of line

karimnaaji:
For #310

### DIFF
--- a/core/src/labels/label.cpp
+++ b/core/src/labels/label.cpp
@@ -2,6 +2,7 @@
 
 #include "util/geom.h"
 #include "labels/labelMesh.h"
+#include "glm/gtx/rotate_vector.hpp"
 
 namespace Tangram {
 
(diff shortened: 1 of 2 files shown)


blair1618:
Looks perfect. @karimnaaji all good to merge?

bcamper:
Is this relative to upright text direction (not strictly line direction)
like on JS? (Sorry on a phone, can't look mired closely.)

karimnaaji:
@bcamper Not sure what upright text direction exactly means? But yes this is exactly the same ways as JS is currently implemented.

blair1618:
Ooh so when the text flips to stay upright, the offset flips too?

karimnaaji:
Yeah it does if you rotate

bcamper:
Right text is always oriented upright. So as discussed previously, we will
need to add specific logic for rendering borders. cc @nvkelso

blair1618:
Cool cool. We'll tackle the border issue in tandem with JS later on. This part seems to be ready to go.

bcamper:
Yep

PR #322, opened 2015-10-01
Allow persistent icons

karimnaaji:
I am not so happy with this solution since it adds a `persistent` keyword in the stylesheet not existing in JS, but that's the easiest workaround now to enable persistent icons. We could detect if the icon comes from a client data source but this would get pretty messy since it introduces a lot of assumptions.


bcamper:
Thanks @karimnaaji, before we merge let's think a little more on syntax, in case we can think of future uses. For example, maybe something like a `transition` keyword? Default `transition: fade`, with this example being `transition: false`? Not sure if there are other obvious transitions though, so who knows :)  Could also be something like `fade: false` to be more explicit (`persistent` seems a little vague here).

karimnaaji:
Yeah this keyword should be extended to different options to clearly differentiate each cases. 

The fact of colliding or not, the transition with an ease function and the time for the transition.

karimnaaji:
From the current implementation a proposed syntax could also be:

```yaml
icons:
    collide: true | false # defaults to true
    hide:
        effect: fade | blink ...
        ease: pow | sine | ...
        time: 0.2s # sec or millis
    show:
        effect: fade
        ease: pow
        time: 0.1s
```

blair1618:
Seems like we'll need at least one or two more style sheet options for eraser-map functionality:

**1. Enable/disable label fading:** Could be a key like `transition` or `fade`. For semantic density, I'd suggest that the value be the desired duration of the transition in seconds. For our "find me" icon, we'd then use `fade: 0` for an immediate transition. We could allow other values later for more sophisticated behaviors. 

**2. Enable/disable label collision:** A separate key like `collision` would be one approach, but since this behavior is entangled with `priority`, it might make sense to encode collide-ability within that value. For instance, what if we used `priority: 0` to denote that a label always "survives" collision? What this doesn't encode is how particular groups of labels interact - e.g. if we wanted to "mask" collision between certain groups.

bcamper:
```
    transition:
        event: [hide, show]
        effect: fade | blink ...
        ease: pow | sine | ...
        time: 0.2s # sec or millis
```

```
    transition:
        - event: hide
          effect: fade | blink ...
          ease: pow | sine | ...
          time: 0.2s # sec or millis

        - event: show
          effect: fade
          ease: pow
          time: 0.1s
```

bcamper:
```
icons:
    collide: true | false # defaults to true
    hide: &transition
        effect: fade | blink ...
        ease: pow | sine | ...
        time: 0.2s # sec or millis
    show: *transition
```

bcamper:
```
icons:
    collide: true | false # defaults to true
    transition:
        hide:
            effect: fade | blink ...
            ease: pow | sine | ...
            time: 0.2s # sec or millis
        show: hide
        selected: hide
```

bcamper:
```
icons:
    collide: true | false # defaults to true
    [hide, show]:
        effect: fade | blink ...
        ease: pow | sine | ...
        time: 0.2s # sec or millis
    selected:
        effect: fade
        ease: pow
        time: 0.1s
```

bcamper:
```
icons:
    collide: true | false # defaults to true
    transition:
        [hide, show]:
            effect: fade | blink ...
            ease: pow | sine | ...
            time: 0.2s # sec or millis
        selected:
            effect: fade
            ease: pow
            time: 0.1s
```

bcamper:
Minimal implementation:

```
    transition:
        [hide, show]:
            time: 0.2s # sec or millis
        selected:
            time: 0 # turn off effect
```

bcamper:
Initial with `effect: fade | false `

```
    transition:
        [hide, show]:
            effect: fade
            time: 0.2s
        selected:
            effect: false
```

karimnaaji:
This is actually invalid `yaml`:
```yaml
transition:
    [hide, show]:
```
I'll go with
```yaml
transition:
    hide, show:
```

blair1618:
Really? I tried this in a yaml validator just yesterday, to make sure I
wasn't crazy, and it worked fine

karimnaaji:
This one doesn't want it: http://yaml-online-parser.appspot.com/

tallytalwar:
http://www.yamllint.com/ 
it works here.

bcamper:
Yes I believe it is valid YAML, from the spec, but I also think it is not commonly used, so it isn't shocking if some validators/parsers don't support it properly. In JS a `[hide, show]` key just gets de-serialized to `hide,show` anyway (I checked, the brackets are dropped).

```
transition:
    hide, show:
```

is not bad either and is also valid YAML (and to make JS decoding work, we can't allow commas in event names anyway, so there's functionally little or no difference as far as I can tell). But using a true sequence is probably most consistent.

blair1618:
@karimnaaji I just tried the sequence-key syntax in yaml-cpp and it seems to work in a simple test case:
```c++
TEST_CASE("Load a complex key into yaml", "[yaml]") {

    auto str = R"END(
    transitions:
        [show, hide]: 
            effect: fade
            time: 1
        selected:
            effect: blink
            time: 0.5
    )END";

    YAML::Node node = YAML::Load(str);

    REQUIRE(node["transitions"].size() == 2);
}
```
Am I overlooking something?

karimnaaji:
Yep! This works fine, something else in `sceneLoader.cpp` must be assuming something that breaks this and throw an exception then.

core/src/scene/sceneLoader.cpp
```diff
@@ -1134,6 +1133,11 @@ void SceneLoader::parseStyleParams(Node params, Scene& scene, const std::string&
         }
         if (key == "style") { continue; }
 
+        if (key == "transition") {
+            parseTransition(prop.second, scene, key, out);
```
tallytalwar: isn't `key` always `transition` here?

core/src/scene/sceneLoader.cpp
```diff
@@ -1249,6 +1253,48 @@ StyleUniforms SceneLoader::parseStyleUniforms(const Node& value, Scene& scene) {
...
+            case YAML::NodeType::Sequence:
+            {
+                auto rawKeys = parseSequence(prop.first);
+                std::stringstream ss(rawKeys);
+                std::vector<std::string> keys;
+                std::string key;
+
+                while (std::getline(ss, key, ',')) {
+                    keys.push_back(key);
+                }
```
blair1618: More briefly:
```c++
auto keys = prop.first.as<std::vector<std::string>>();
```
We could also pull `keys` and the loop over `keys` outside the switch/case and just populate`keys` inside the cases, if that makes sense to you. 

Just trying to keep `SceneLoader` on a diet :)

blair1618:
Looking over this now!

PR #323, opened 2015-10-02
Recursive style mixing and mixed shader block

hjanetzek:
- The first patch modifies loadStyle to first load all mixins to avoid merging styles twice.
- The second patch add support for combining shader blocks while ensuring to filter out duplicates by collecting all blocks in a vector instead of concatenating block strings in shaderBlockMerge


tallytalwar:
Few more fixes and rebase and this is good to go.

tallytalwar:
Looks good to me to be merged.

PR #324, opened 2015-10-02
File paths resolved as 'internal', 'resource', or 'absolute'

blair1618:
This solves an issue with using scene files outside the application resource directory where tangram-internal files (like shaders) were trying to be located relative to the scene file, like all other resources. 

Now, whenever a file is accessed, you can specify whether the path is 'internal' (a bundled tangram file), 'absolute' (at an absolute path in the filesystem), or 'resource' (relative to the scene file).

### DIFF
--- a/osx/src/platform_osx.mm
+++ b/osx/src/platform_osx.mm
@@ -58,57 +58,55 @@ std::string setResourceRoot(const char* _path) {
 
 }
 
-NSString* resolveResourcePath(const char* _path) {
+NSString* resolvePath(const char* _path, PathType _type) {
 
     if (s_resourceRoot == NULL) {
         setResourceRoot(".");
     }
 
-    return [s_resourceRoot stringByAppendingPathComponent:[NSString stringWithUTF8String:_path]];
+    NSString* path = [NSString stringWithUTF8String:_path];
 
+    switch (_type) {
+    case PathType::internal:
+        return [[[NSBundle mainBundle] resourcePath] stringByAppendingPathComponent:path];
+    case PathType::resource:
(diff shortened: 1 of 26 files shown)


blair1618:
@karimnaaji can you check this over?

There's some code repetition which is obviously not ideal, I'd like to refactor all the platform functions into class methods that are overridden only where needed, but fixing this specific problem is more important for now.

karimnaaji:
Yep checking at it!

osx/src/platform_osx.mm
```diff
@@ -58,57 +58,55 @@ bool isContinuousRendering() {
...
+    NSString* path = resolvePath(_path, _type);
     NSString* str = [NSString stringWithContentsOfFile:path
                                               encoding:NSASCIIStringEncoding
                                                  error:NULL];
-    
+
     if (str == nil) {
-        LOGW("Failed to read file at path: %s", _path);
+        LOGW("Failed to read file at path: %s", [path UTF8String]);
         return std::move(std::string());
```
karimnaaji: This doesn't come from this PR, but it might be the right time to remove the `std::move` and let the compiler do return value optimization if needed. This would also match the new `osx` way of returning strings.

karimnaaji:
I created a card concerning the restructuration of platform code a while ago https://trello.com/c/FbRwhr4v/590-common-platform-code. 
Otherwise this is is good to merge!

osx/src/platform_osx.mm
tallytalwar: :+1:

blair1618: Re`move`d!

PR #325, opened 2015-10-02
Animated property effective

karimnaaji:


core/src/tangram.cpp
```diff
@@ -131,10 +131,13 @@ void update(float _dt) {
 
             m_labels->update(*m_view, _dt, m_scene->styles(), tiles);
         }
-    }
 
-    if (m_scene) {
-        // Update lights and styles
+        for (const auto& style : m_scene->styles()) {
+            if (style->isAnimated()) {
+                requestRender();
```
blair1618: Cool, seems to work! I'd suggest one change though: The GL containers we use on Android and iOS have distinct modes of operation for continuous rendering - in fact this is the only mode on iOS and `requestRender()` just un-pauses the loop for one frame and then pauses it again. There are probably performance benefits to simply changing the render mode according to whether animated styles are present, using `setRenderMode()`. 

I actually just tried this out locally and it seems to work, if this makes sense to you I can make that change.

karimnaaji: Yep that makes sense!

Issue #288, opened 2015-09-16
No label rendering on Samsung Galaxy SIII (Specifically when multisampling is set to true)


tallytalwar:
This is something specific to S3, where getting `u_tex` uniform position fails on sdf.fs fragment shader.

LogCat Output:
```
D/Tangram ( 3216): finish initialize
D/Tangram ( 3216): initialize
D/Tangram ( 3216): finish initialize
D/Tangram ( 3216): setup GL
D/Tangram ( 3216): resize: 720 x 1230
D/Tangram ( 3216): OpenGL Error: GL_INVALID_VALUE at Tangram::render()
D/Tangram ( 3216): Warning: Using default font for 'sans-serif_400_normal'.
D/Tangram ( 3216): OpenGL Error: GL_INVALID_VALUE at Tangram::render()
D/Tangram ( 3216): OpenGL Error: GL_INVALID_OPERATION at Tangram::render()
D/Tangram ( 3216): OpenGL Error: GL_INVALID_OPERATION at Tangram::render()
D/Tangram ( 3216): OpenGL Error: GL_INVALID_OPERATION at Tangram::render()
D/Tangram ( 3216): OpenGL Error: GL_INVALID_OPERATION at Tangram::render()
D/Tangram ( 3216): OpenGL Error: GL_INVALID_OPERATION at Tangram::render()
D/Tangram ( 3216): OpenGL Error: GL_INVALID_OPERATION at Tangram::render()
D/Tangram ( 3216): OpenGL Error: GL_INVALID_OPERATION at Tangram::render()
D/Tangram ( 3216): OpenGL Error: GL_INVALID_OPERATION at Tangram::render()
D/Tangram ( 3216): OpenGL Error: GL_INVALID_OPERATION at Tangram::render()
D/Tangram ( 3216): OpenGL Error: GL_INVALID_OPERATION at Tangram::render()
D/Tangram ( 3216): OpenGL Error: GL_INVALID_OPERATION at Tangram::render()
D/Tangram ( 3216): OpenGL Error: GL_INVALID_OPERATION at Tangram::render()
D/Tangram ( 3216): OpenGL Error: GL_INVALID_OPERATION at Tangram::render()
D/Tangram ( 3216): OpenGL Error: GL_INVALID_OPERATION at Tangram::render()
D/Tangram ( 3216): OpenGL Error: GL_INVALID_OPERATION at Tangram::render()
[...]

PR #326, opened 2015-10-05
Vertex array objects support

karimnaaji:
Add support for VAOs with appropriate fallbacks when extension is not available.
VAOs are supported by ~63% of mobile devices according to [this dataset](http://delphigl.de/glcapsviewer/gles_extensions.php). 

The current frame trace before and after VAO support:
screen shot 2015-10-05 at 3 13 23 pm
screen shot 2015-10-05 at 3 13 32 pm
By adding VAO binding to the render states, this trace would get even better.

This reduces GL calls whenever this extension is available, and remove the map access bottleneck when enabling the vertex layout that was pretty important.

Also add support for virtual pinned memory on client side for faster dynamic meshes update using `glMapBuffers`.

- [ ] Update label mesh to use VAOs
- [x] Fix byteOffset usage for VAOs
- [ ] Make VAO usage shader independent

### DIFF
--- a/android/tangram/jni/platform_android.cpp
+++ b/android/tangram/jni/platform_android.cpp
@@ -2,6 +2,7 @@
 
 #include "platform.h"
 #include "tangram.h"
+#include "gl.h"
 
 #include <android/log.h>
 #include <android/asset_manager.h>
@@ -39,6 +40,10 @@ static AAssetManager* assetManager;
 static bool s_isContinuousRendering = false;
 static bool s_useInternalResources = true;
 static std::string s_resourceRoot;
+
+PFNGLBINDVERTEXARRAYOESPROC glBindVertexArrayOESEXT = 0;
+PFNGLDELETEVERTEXARRAYSOESPROC glDeleteVertexArraysOESEXT = 0;
+PFNGLGENVERTEXARRAYSOESPROC glGenVertexArraysOESEXT = 0;
 
 void setupJniEnv(JNIEnv* _jniEnv, jobject _tangramInstance, jobject _assetManager) {
     _jniEnv->GetJavaVM(&jvm);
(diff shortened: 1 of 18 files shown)


blair1618:
@karimnaaji can you explain the last checklist item in a little more detail?

karimnaaji:
In our implementation, we *get* shader location attributes, another way would be to *bind* shader location attribute, which I think would be a consistent model for VAOs since the vertex array object is initialized once. 
In this implementation, I assumed that the mesh would be drawn with a shader that uses all the attributes of a vertex layout, and that a `getAttribLocation(attributeName)` on any given `attributeName` of the layout of the mesh would not fail.

A potential use case that makes this fail is:
```
Shader s0; // uses attribute a, b
Shader s1; // uses attribute a
Mesh m(vertex_layout{ a, b });
m.draw(s1); // vao initialized with attribute a only
```
We don't have this use case so I've let this as is for now, but we could fix it by adding a `location` in the vertex layout that we set ourself to any shaders that the mesh need to be drawn on.

karimnaaji:
It seems like there is something wrong in the usage of VAOs in `labelMesh.cpp`, will investigate more on that, but the last commit seems to fix the issue on the devices.

tallytalwar:
@blair1618 possible the s3 wonkiness in this and https://github.com/tangrams/tangram-es/issues/288 are related!

blair1618:
@karimnaaji can you rebase and merge this when you're happy with it? I'd like to revisit that issue with VAOs and TextBuffers, but I don't want it to hold up other tasks.

karimnaaji:
@hjanetzek , I am in the process of rebasing this against master, and I am a little confused about the forwards now, did you edit part of it yourself based on the mesa provided headers for ES2? Because the default one is much more consequent than what we have. Should we edit/complete the forwards file each time we use a new GL define/functions? Just to be sure before I start completing it.
Should we use a separate forward header file for extensions to be consistent with how mesa provides it?

hjanetzek:
@karimnaaji I would stick with a single file. Already had a rebased version here - one moment. I think it makes sense to group the functions according to their use.

hjanetzek:
Maybe it makes sense to merge this one first https://github.com/tangrams/tangram-es/pull/351

It removes gl_fwd.h again which was possible by building the android shared library just like the other platforms, with core included as a static lib.

karimnaaji:
Agreed, it'd make things easier, will review #351

karimnaaji:
Closing in favor of #352

PR #328, opened 2015-10-06
Detect generic motion events

ecgreb:
Used to determine whenever a user interacts with the map.


tallytalwar:
Simple and easy. Merging.

PR #329, opened 2015-10-06
World bounds

tallytalwar:


core/src/view/view.cpp
```diff
@@ -105,7 +105,13 @@ void View::setPitch(float _pitch) {
...
 void View::translate(double _dx, double _dy) {
 
-    setPosition(m_pos.x + _dx, m_pos.y + _dy);
+    auto mapBounds = m_projection->MapBounds();
+    glm::dvec2 newMapPos(
+                        glm::clamp(m_pos.x + _dx, mapBounds.min.x, mapBounds.max.x),
+                        glm::clamp(m_pos.y + _dy, mapBounds.min.y, mapBounds.max.y)
+                        );
+
+    setPosition(newMapPos.x, newMapPos.y);
```
blair1618: Any logic applied in `translate()` should also be applied in `setPosition()`, so the logic should reside in `setPosition()` and then `translate()` can just blindly pass the incremented coordinates (like `zoom()` does).

tallytalwar: :+1:

tallytalwar:
For some interesting reasons (possible floating point precision) `labelsTest` is failing on `linux` and not on `osx`!!. Need to check it, so hold on reviewing this.

tallytalwar:
I am pretty sure, its a floating/double precision issue.

Converting `View::screenToGroundPlane` to use all double should most likely fix this!.

tallytalwar:
Found another way of fixing this right now... 

But it does make sense to use double for `View::screenToGroundPlane` consider meter values can extend beyond float precisions. 

Here, it was failing because `MercatorProjection::MapBounds` was returning a higher precision value, and `View::screenToGroundPlane` was returning a lower precision `float` value. `View::checkMapBounds` was evaluating to false, because of this precision differences in very similar values. (20037602.0 vs 20037508.342789244) to be precise.

blair1618:
Yeah, I see no reason why `View::screenToGroundPlane` shouldn't use `double` values for calculations.

tallytalwar:
Cool, going to make another PR for `View::screenToGroundPlane`, as it also touches some part of our code base not related to this PR.

core/src/view/view.cpp
```diff
@@ -68,12 +68,48 @@ void View::setSize(int _width, int _height) {
...
+    if ( bottomRight.x < mapBounds.min.x || bottomRight.x > mapBounds.max.x ||
+         bottomRight.y < mapBounds.min.y || bottomRight.y > mapBounds.max.y ) { return false; }
+    screenToGroundPlane(topRight.x, topRight.y);
+    topRight += glm::dvec2(m_pos.x, m_pos.y);
+    if ( topRight.x < mapBounds.min.x || topRight.x > mapBounds.max.x ||
+         topRight.y < mapBounds.min.y || topRight.y > mapBounds.max.y ) { return false; }
+    screenToGroundPlane(topLeft.x, topLeft.y);
+    topLeft += glm::dvec2(m_pos.x, m_pos.y);
+    if ( topLeft.x < mapBounds.min.x || topLeft.x > mapBounds.max.x ||
+         topLeft.y < mapBounds.min.y || topLeft.y > mapBounds.max.y ) { return false; }
```
blair1618: Might be worthwhile to add a method like `bool BoundingBox::contains(glm::dvec2 point)`

blair1618:
Interesting case: I just set the camera to  start with `zoom: 2` in the scene file and the map seems permanently stuck in this configuration (I also made the background gray):
screen shot 2015-10-19 at 2 35 26 pm

tallytalwar:
Nice catch. I might have a clue how to fix it.

tallytalwar:
Any further questions on this?

blair1618:
Haven't checked with the latest changes and heading to the airport now, but
maybe @karimnaaji or @hjanetzek can take a look

Any further questions on this?

—
Reply to this email directly or view it on GitHub
<https://github.com/tangrams/tangram-es/pull/329#issuecomment-149583496>.

karimnaaji:
Giving a look at it

karimnaaji:
@tallytalwar let me know if that last commit makes sense to you.

karimnaaji:
Looks good to me! Except a small issue when you zoom out/rotate really fast that we discussed offline.

tallytalwar:
Yes, not sure of why this "jump" occurs when zooming out/rotating really fast when map is at its max bounds.

Will check it on a device also. For the time being I will merge it and create a card for this issue.

PR #330, opened 2015-10-06
Geometry within tiles uses positions in [0, 1]

blair1618:
If it worked, you won't notice a thing ;)


blair1618:
Before merging this I'm going to make sure there's a version of the eraser-map style with which this will be compatible; it will probably break the visuals of the eraser-map app until the scene file there is updated.

PR #331, opened 2015-10-06
[WIP] Swig MapData

hjanetzek:
Create swig bindings for ClientGeoJsonSource (MapData), Tags, Polygon, Line and LngLat.

- [X] bindings to add tagged geometries   
- [ ] ensure thread safety (pretty sure that it is correct, but should be handled in tangram.cpp instead of synchronizing java native methods)


bcamper:
I think I would prefer to stick with terminology like Properties instead of Tags. There's nothing inherently wrong with tags, but it seems more OSM-centric, properties feels a little more generic (e.g. GeoJSON).

hjanetzek:
The Tags are a map of <string,string> as they are used in geojsonvt and I've just added to avoid one more conversion. Our Properties use a sorted vector of <string, variant> instead. 

I would prefer to use Properties also with geojsonvt, which is possible with changing two lines instead of switching completely to a string maps everywhere or converting things back and forth:
For example now when the client adds a geometry with Properties ->  we convert the Properties to a string Map for geojsonvt, -> geojsonvt copies the whole Map for every tile the feature appears in -> finally we convert the Map back to Properties (for styling) to show them as a tile.  Additionally, when the feature is selectable, we allocate some heap to copy the Properties to a shared_ptr (since a Feature can have multiple selectable labels) to pass them back to the client application on feature picking

With a very minimal change to geojsonvt Tags can wrap a shared_ptr<Properties> which reduces the number of conversions and copies.
This can be further improved when we add the option to Feature to hold a shared_ptr (or only a shared_ptr)  to Properties (instead of having it as a direct member), then we could actually get the same object out from feature selection that we have put into geojsonvt.

blair1618:
Storing Tangram `Properties` in geojsonvt would indeed be much more elegant, I think this would be worth diverging slightly from the original fork. Maybe we could even submit changes for a generic way of doing this with a template parameter.

swig/array_nocpy.i
```diff
@@ -0,0 +1,295 @@
+
```
karimnaaji: Can we move the swig files in the `android` folder?

hjanetzek: I would like to add more bindings for other languages later and it is also handy to develop java bindings on desktop :)

karimnaaji: :+1:

PR #332, opened 2015-10-07
Simplify color interpolation

blair1618:


core/src/util/color.h
```diff
@@ -9,10 +9,22 @@ struct Color {
...
-        uint32_t abgr = 0x00000000;
-        // channel order appears reversed because of little-endianness
+        uint32_t abgr = 0;
     };
 
+    Color() = default;
+    Color(uint32_t _abgr) : abgr(_abgr) {}
+    Color(uint8_t _r, uint8_t _g, uint8_t _b, uint8_t _a) : r(_r), g(_g), b(_b), a(_a) {}
+
+    static Color mix(const Color& _x, const Color& _y, float _a) {
```
tallytalwar: Instead of mix, can we call it interpolate or something else.

karimnaaji: `mix` is the standard built in name for glsl color interpolation, so it quite makes sense

tallytalwar: then its fine. :D

blair1618: Yep that was the idea

blair1618:
I'll take that to mean there are no issues with this change :)

PR #333, opened 2015-10-07
Vertex pipeline is compatible with web renderer

blair1618:
Supersedes #330 

 - Normalizes positions within tiles to the range [0, 1]
 - Provides modelPosition() and worldPosition() functions in vertex shaders
 - Sets u_model, u_view, and u_proj instead of combined matrices (turns out we actually need all of the intermediate values, so pre-multiplying won't save any work)
 - Rename v_eyeToPoint to v_position
 - Ensure v_world_position is set in all shaders
 - Use u_ortho for orthographic projection matrix, to distinguish from perspective u_proj
 - Organize variables in shaders
 - Hoist some uniform assignments out of Tile::draw and into Style::onBeginFrame

### DIFF
--- a/core/resources/shaders/point.vs
+++ b/core/resources/shaders/point.vs
@@ -9,15 +9,7 @@ precision mediump float;
 
 #pragma tangram: defines
 
-attribute LOWP vec2 a_position;
-attribute LOWP vec2 a_screenPosition;
-attribute vec2 a_uv;
-attribute LOWP float a_alpha;
-attribute LOWP float a_rotation;
-attribute LOWP vec4 a_color;
-attribute LOWP vec4 a_stroke;
-
-uniform mat4 u_proj;
+uniform mat4 u_ortho;
 uniform vec3 u_map_position;
 uniform vec3 u_tile_origin;
 uniform vec2 u_resolution;
@@ -26,11 +18,19 @@ uniform float u_meters_per_pixel;
 
 #pragma tangram: uniforms
 
-varying float v_alpha;
-varying vec2 v_texcoords;
+attribute vec2 a_uv;
+attribute LOWP vec2 a_position;
+attribute LOWP vec2 a_screenPosition;
+attribute LOWP float a_alpha;
+attribute LOWP float a_rotation;
+attribute LOWP vec4 a_color;
+attribute LOWP vec4 a_stroke;
+
 varying vec4 v_color;
 varying vec4 v_strokeColor;
+varying vec2 v_texcoords;
 varying float v_strokeWidth;
+varying float v_alpha;
 
 #pragma tangram: global
 
@@ -54,7 +54,7 @@ void main() {
 
         #pragma tangram: position
 
-        gl_Position = u_proj * position;
+        gl_Position = u_ortho * position;
     } else {
         gl_Position = vec4(0.0, 0.0, 0.0, 1.0);
     }
--- a/core/resources/shaders/polyline.vs
+++ b/core/resources/shaders/polyline.vs
@@ -6,8 +6,6 @@ precision highp float;
 
 #pragma tangram: defines
 
-#define TANGRAM_WORLD_POSITION_WRAP 100000.0
-
 uniform mat4 u_model;
 uniform mat4 u_view;
 uniform mat4 u_proj;
@@ -104,7 +102,7 @@ void main() {
 
     gl_Position = u_proj * v_position;
     
-    // Proxy tiles have u_tile_zoom < 0, so this re-scaling will place proxy tiles deeper in
+    // Proxy tiles have u_tile_origin.z < 0, so this re-scaling will place proxy tiles deeper in
     // the depth buffer than non-proxy tiles by a distance that increases with tile zoom
     gl_Position.z /= 1. + .1 * (abs(u_tile_origin.z) - u_tile_origin.z);
     
--- a/core/resources/shaders/polyline.vs
+++ b/core/resources/shaders/polyline.vs
@@ -41,11 +41,7 @@ vec4 modelPosition() {
 }
 
 vec4 worldPosition() {
-    vec4 worldPosition = a_position * u_model + vec4(u_map_position.xy, 0., 1.);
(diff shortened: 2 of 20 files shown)


blair1618:
@bcamper hopefully this jives with what you were imagining. 

As we had talked about, the matrix uniforms won't be the same between web and native, but there's no need to access them now that we've added position functions (though we can't stop a rogue h4x0r from doing so). I'm actually pretty happy with the way this turned out!

core/resources/shaders/polyline.vs
```diff
@@ -35,6 +36,18 @@ varying vec2 v_texcoord;
     varying vec4 v_lighting;
 #endif
 
+vec4 modelPosition() {
+    return a_position;
+}
+
+vec4 worldPosition() {
+    vec4 worldPosition = a_position * u_model + vec4(u_map_position.xy, 0., 1.);
```
bcamper: Here's a gotcha: since polylines have vertex positions at the "zero width" position, the world position should probably reflect the position after the vertex is extruded according to its rendered width. At least that's what I'm doing in JS, and I remember us fixing this because it was causing all lines that used world position to be a big smear. Not sure how that should be handled here since it's a bit trickier to do within a function.

blair1618: Ohhh good one, I can see how that would muck up procedural textures. I'll see what I can do about that...

blair1618: Well now I've got it just returning `v_world_position`, which is now calculated after width extrusion. A necessary caveat: `worldPosition()` will return undefined values in the `width` block (just like if you accessed `v_world_position` in a `width` block in Tangram Play currently). We could initialize it to the un-extruded vertex position, but that'll duplicate some math that's usually not needed.

blair1618:
@tallytalwar can you check over this?

tallytalwar:
Sure, Reviewing now.

core/resources/shaders/polyline.vs
```diff
@@ -43,15 +52,17 @@ void main() {
...
     {
         float width = a_extrude.z;
         float dwdz = a_extrude.w;
-        float dz =  u_map_position.z - abs(u_tile_origin.z);
-        // interpolate between zoom levels
+        float dz = u_map_position.z - abs(u_tile_origin.z);
+        // Interpolate between zoom levels
         width += dwdz * dz;
-        // scale to screen-space
+        // Scale to screen-space
```
tallytalwar: I might be misunderstanding this, but width should be in model (tile) space here right (just like position is in tile/model space)?

Even in Line: 71 `position.xy += a_extrude.xy * width` is doing a operation in model space.

And if width is in model space, does that mean injection for width is in model space?

core/src/style/style.cpp
```diff
@@ -138,11 +140,16 @@ void Style::onBeginDrawFrame(const View& _view, Scene& _scene) {
...
     if (m_dirtyViewport) {
-        const auto& mapPos = _view.getPosition();
         m_shaderProgram->setUniformf("u_resolution", _view.getWidth(), _view.getHeight());
-        m_shaderProgram->setUniformf("u_map_position", mapPos.x, mapPos.y, _view.getZoom());
     }
 
+    const auto& mapPos = _view.getPosition();
+    m_shaderProgram->setUniformf("u_map_position", mapPos.x, mapPos.y, _view.getZoom());
+    m_shaderProgram->setUniformMatrix3f("u_normalMatrix", glm::value_ptr(_view.getNormalMatrix()));
+    m_shaderProgram->setUniformf("u_meters_per_pixel", _view.pixelsPerMeter());
```
tallytalwar: is it meters per pixel or pixels per meter? if the names are correct then shouldn't this should be inverse?
`1.0f/_view.pixelsPerMeter()`.

tallytalwar:
Mostly looks good to merge, couple of minor questions/issues.

core/resources/shaders/polyline.vs
blair1618: Based on [the shader used in tangram-js](https://github.com/tangrams/tangram/blob/master/src/styles/polygons/polygons_vertex.glsl#L61) it seems that `width` is indeed modified in model space.

tallytalwar: Can we fix the comments in that case?

core/src/style/style.cpp
blair1618: Fixed, thanks!

core/resources/shaders/polyline.vs
blair1618: I can see how that may be misleading, fixed!

core/resources/shaders/point.vs
```diff
@@ -26,11 +18,19 @@ uniform float u_meters_per_pixel;
...
-varying vec2 v_texcoords;
+attribute vec2 a_uv;
+attribute LOWP vec2 a_position;
+attribute LOWP vec2 a_screenPosition;
+attribute LOWP float a_alpha;
+attribute LOWP float a_rotation;
+attribute LOWP vec4 a_color;
+attribute LOWP vec4 a_stroke;
+
 varying vec4 v_color;
```
hjanetzek: Just wondering about all the LOWP, 2^-8 precision is not sufficient for vertices, but most drivers will give you more precision anyway. Have you actually tried it on a device that uses the minimal requested precision?  Galaxy S6 seems to use the hint (maybe only for one particular driver version though)

https://www.khronos.org/registry/gles/specs/2.0/GLSL_ES_Specification_1.0.17.pdf (Precision qualifiers)

blair1618: I just left these qualifiers unaltered from the existing shader. LOWP always makes sense for colors, the others I assumed were in some relatively constrained domain, but I'm not completely sure. Maybe @karimnaaji knows more?

karimnaaji: Yeah we should remove the two first qualifiers, and add `lowp` hint to `a_uv`, this could give bad behavior when driver uses this hint on the position attribute, even if they are in a restricted range in that case.

PR #334, opened 2015-10-08
Require either color parameter, color block, or filter block to build feature

blair1618:


bcamper:
I believe this should also check for 'filter' blocks.

blair1618:
Cool, thanks!

PR #335, opened 2015-10-09
Size stops

blair1618:
Stops could maybe do with some refactoring, but should do the job for now. 

Icon sizes don't re-scale for continuity the way that roads do, but icons are also placed differently at each zoom level so it doesn't seem like continuity should be a concern.

### DIFF
--- a/core/src/scene/drawRule.cpp
+++ b/core/src/scene/drawRule.cpp
@@ -89,6 +89,8 @@ void DrawRule::eval(const StyleContext& _ctx) {
         if (param.stops) {
             if (StyleParam::isColor(param.key)) {
                 param.value = param.stops->evalColor(_ctx.getGlobalZoom());
+            } else if (StyleParam::isWidth(param.key)) {
+                param.value = param.stops->evalWidth(_ctx.getGlobalZoom());
             } else {
                 param.value = param.stops->evalFloat(_ctx.getGlobalZoom());
             }
--- a/core/src/scene/styleParam.cpp
+++ b/core/src/scene/styleParam.cpp
@@ -390,6 +390,7 @@ bool StyleParam::isWidth(StyleParamKey _key) {
     switch (_key) {
         case StyleParamKey::width:
         case StyleParamKey::outline_width:
+        case StyleParamKey::size:
             return true;
         default:
             return false;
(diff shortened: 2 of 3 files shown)


blair1618:
@karimnaaji Can you check this out? In particular I'd like to make sure this matches the JS behavior.

karimnaaji:
Sure

karimnaaji:
There is some more work on the shader side concerning the `size` params for icons. To smoothly interpolate between zooms and update a scale factor or we would have fractional size stops only (no interpolation in between).

But this already fixes all of the eraser-map stylesheet icons sizes so I'll merge this and create a separate task for interpolation of sizes between zooms.

karimnaaji:
One nitpick, not sure JS side handles this correctly (not even sure this has a potential usage), but
```yaml
icons:
    size: [[13, [12px, 20px]], [18, 20px]] # 2 dimensional stop size for first stop, 1 dim for second one
```
is not handled right now.

blair1618:
I was also wondering about the case of 2-dimensional sizes in stops, this
change won't handle that correctly but I don't know whether it's officially
supported

karimnaaji:
I just tried and it's not! The only follow-up on this would be the interpolation.

bcamper:
Hmm, right, currently in JS it's assumed that the stop values all have the
same dimension as the first one. How would we generalize this to
interpolate between stops of different dimension?

karimnaaji:
Can't we interpolate separately on each of those?

karimnaaji:
If we have `[[13, [12px, 20px]], [18, 20px]]` we simply extend the second one to be `[20px, 20px]` and interpolate `[12px, 20px] to [20px, 20px]`

bcamper:
I mean in the general case, if you have stops like this with different dimension lengths:

`[[14, [x1, y1, z1]], [15, [x2, y2]]]`

How do you interpolate `z1`? What happens to values that are added or lost between stops? For the `size` case, you just duplicate the first value to be for both x and y sizes, but I'm wondering if there is a general rule.

karimnaaji:
Yeah interpolating with `y1` doesn't make sense in that general case, 0 would probably be more generic. We can default to 0 and warn whenever a stop would be interpolated with inconsistent value.

PR #336, opened 2015-10-09
Add device pixel ratio to shaders

karimnaaji:


tallytalwar:
This looks good to me and also fixes an issue. Can merge when travis says go.

core/src/style/style.h
```diff
@@ -142,7 +142,7 @@ using StyleUniform = std::pair< std::string, UniformValue >;
     virtual void onEndBuildTile(Tile& _tile) const;
 
     /* Perform any setup needed before drawing each frame */
-    virtual void onBeginDrawFrame(const View& _view, Scene& _scene);
+    virtual void onBeginDrawFrame(const View& _view, Scene& _scene, int _textureUnit = 0);
```
blair1618: Is `_textureUnit` the next available texture unit?  The last-bound texture unit? A comment might be good.

nvkelso:
Has there been discussion about making this visible to the scene file for normal properties (outside of fragment shaders)?

blair1618:
@nvkelso You mean for things that are sized in pixels? Those should all be scaled by tangram such that they span a uniform number of "logical pixels" across devices (e.g. a "12px" font size will span 24 physical pixels on a retina screen). In the future we could consider some form of resource selection based on display density (the way that iOS selects @2x and @3x resources) but we don't have anything like that now.

bcamper:
@nvkelso yes we can do that. Are you mainly thinking of icon size adjustments, or other cases? Trying to think of the best way to integrate and/or expose (e.g. in some cases we may just want to apply retina adjustment automatically, in other cases people may want to use the value in a function() property?).

nvkelso:
@blair1618 and @bcamper Yes, things sized in "px" are really points (alla CSS) and a retina screen can hold a finer line stroke than a 1x screen – and font size, and icon size as Brett alludes to. Color changes are usually associated with a specific display, but is also related if the line weight needs to be thicker on a @1x so the color needs to be lighter.

tallytalwar:
I am going to merge this, as it also fixes an issue.

We can continue discussions on exposing device pixel ratio to style sheet offline/in trello?

PR #337, opened 2015-10-12
Reset state of dynamically updated data for cached tiles

karimnaaji:
Straight fix concerning inconsistent behaviors on cached tile data.


tallytalwar:
A definite improvement for label visibility. Will merge it as it right now.

PR #338, opened 2015-10-13
Enable dropping yaml file to load stylesheet

blair1618:
Got tired of changing the scene file path all the time :P

style-drop mov


nvkelso:
+1

tallytalwar:
Nice.

PR #340, opened 2015-10-13
Handle texture atlas resize

karimnaaji:
Handle texture atlas resize when texture size is reached.

- [ ] ~~find reasonable raster ratio for retina screens~~ (adjusting with device pixel ratio would be ok until we find atlas size bottleneck on it)
- [x] fix degenerate UVs, or use a backup texture atlas

### DIFF
--- a/core/src/gl/texture.cpp
+++ b/core/src/gl/texture.cpp
@@ -154,18 +154,16 @@ void Texture::update(GLuint _textureUnit) {
 
         m_shouldResize = false;
 
-        m_subData.clear();
+    }
 
-    } else {
-        // process queued sub data updates
-        while (m_subData.size() > 0) {
-            TextureSubData& subData = m_subData.front();
+    // process queued sub data updates
+    while (m_subData.size() > 0) {
+        TextureSubData& subData = m_subData.front();
 
-            glTexSubImage2D(m_target, 0, subData.m_xoff, subData.m_yoff, subData.m_width, subData.m_height,
-                            m_options.m_format, GL_UNSIGNED_BYTE, subData.m_data.data());
+        glTexSubImage2D(m_target, 0, subData.m_xoff, subData.m_yoff, subData.m_width, subData.m_height,
+                        m_options.m_format, GL_UNSIGNED_BYTE, subData.m_data.data());
 
-            m_subData.pop_front();
-        }
+        m_subData.pop_front();
     }
 
     m_dirty = false;
--- a/core/src/text/fontContext.cpp
+++ b/core/src/text/fontContext.cpp
@@ -69,7 +69,7 @@ FontID FontContext::addFont(const std::string& _family, const std::string& _weig
             !(data = bytesFromFile(bundledFontPath.c_str(), PathType::internal, &dataSize))) {
             const std::string sysFontPath = systemFontPath(_family, _weight, _style);
             if ( !(data = bytesFromFile(sysFontPath.c_str(), PathType::absolute, &dataSize)) ) {
-                
+
                 LOGE("Could not load font file %s", fontKey.c_str());
                 m_fonts.emplace(std::move(fontKey), INVALID_FONT);
                 goto fallback;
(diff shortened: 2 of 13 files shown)


hjanetzek:
Reviewing

core/src/text/fontContext.cpp
```diff
@@ -173,21 +173,40 @@ void FontContext::renderUpdate(void* _userPtr, int* _rect, const unsigned char*
...
+    case FONS_ATLAS_FULL: {
+        const auto& tex = fontContext->m_atlas;
+        unsigned int nw = tex->getWidth() * 2;
+        unsigned int nh = tex->getHeight() * 2;
 
+        if (nw > TANGRAM_MAX_TEXTURE_WIDTH || nh > TANGRAM_MAX_TEXTURE_HEIGHT) {
+            LOGE("Full font texture atlas size reached!");
+        } else {
+            if (fonsExpandAtlas(fontContext->m_fsContext, nw, nh, 1)) {
+                tex->resize(nw, nh);
```
hjanetzek: The resize needs synchronization with atlas->setSubData and atlas->bind

core/src/text/fontContext.cpp
```diff
@@ -173,21 +173,40 @@ void FontContext::renderUpdate(void* _userPtr, int* _rect, const unsigned char*
...
     default:
-        LOGE("Unexpected error in Fontstash %d:%d!", error, val);
+        LOGE("Unexpected error in Fontstash %d:%d!", _error, _val);
         break;
     }
 }
 
+glm::vec2 FontContext::getAtlasResolution() const {
+    return glm::vec2(m_atlas->getWidth(), m_atlas->getHeight());
+}
```
hjanetzek: It is unlikely, but we should return the texture size from when the atlas was bound since the resize could occur between the call to bind and getAtlasResolution

core/src/gl/texture.cpp
```diff
@@ -155,18 +155,16 @@ void Texture::update(GLuint _textureUnit) {
 
         m_shouldResize = false;
 
-        m_subData.clear();
+    }
```
hjanetzek: The previous subdata should probably be cleared in the call to resize then since its stride will be wrong when the texture width has changed (even though the data will most likely be overwritten by following subdata)

core/src/text/textBuffer.cpp
```diff
@@ -91,15 +96,28 @@ bool TextBuffer::addLabel(const TextStyle::Parameters& _params, Label::Transform
     uint32_t stroke = (_params.strokeColor & 0x00ffffff) + (strokeWidth << 24);
 
     for (auto& q : quads) {
+        // the atlas resolution at which the quad uvs have been generated onto
+        glm::vec2 atlasResolution(q.atlasWidth, q.atlasHeight);
+        glm::vec2 uvScaleFactor(1.f, 1.f);
+
+        if (atlasResolution != m_atlasRes) {
+            uvScaleFactor = atlasResolution / m_atlasRes;
+        }
```
hjanetzek: The uvScaleFactor of previously added vertices to this mesh will be invalid when the texture size changes during the build of a tile. I think a better approach would be use the pixel offset in the texture as uv which stays constant after resize, so that one only has to divide uv/textureSize in the shader.

core/src/text/fontContext.cpp
karimnaaji: Good point

core/src/text/fontContext.cpp
karimnaaji: Yeah very unlikely to happen, but I see the point, adding synchronisation should fix that.

core/src/text/textBuffer.cpp
karimnaaji: They won't because the current quad stores the atlas width and height at which it has been generated, if a resize happens while a mesh is built, the quads within a mesh would have stored different atlas sizes; the exact size at which the UVs have been generated against.
A good test case is to start with an atlas size of 2 pixels.

core/src/gl/texture.cpp
karimnaaji: Yeah, wasn't sure about what, isn't clearing subdata makes us loss information? We should probably fix the stride instead right?

hjanetzek: In this case we can just drop them since the old region will be set dirty and copied to subbuffers on the next call to FontContext::renderUpdate. Though I think with a slight change to texture and fontstash we only need one buffer and only copy instead of three + x (since also the subdata buffers can become quite large as one rasterize call can cause an update texture of the whole texture when the new glyphs are inserted at the top and bottom)

core/src/text/textBuffer.cpp
hjanetzek: Ah right. Still this looks too complicated to me :)

hjanetzek: Another advantage of using fixed size (and non-fractional) for uv values would be that we could easily switch to ushort instead of float and save 4 bytes per vertex

core/src/gl/texture.cpp
hjanetzek: Actually we dont need the subdata at all since the updated area is copied to to m_data in setSubData. It only needs to keep track of the  y-start and y-end and copy that region to the texture in Texture::update. This will also avoid copying overlapping subdata regions multiple time to the gl texture

core/src/text/textBuffer.cpp
karimnaaji: I went the straight way since it's a blocker for eraser-map content, and I think this could be merged with non-fixed size UV first. I will create a task for it, and follow up on it. But yeah I definitely agree that it will help us for packing vertex data.

karimnaaji:
Closing in favor of #356

PR #341, opened 2015-10-13
Swig mapdata

hjanetzek:
continuing #331

- [x] make LngLat a pure java object by adding in/out-typemaps for conversion 
- [ ] add cmake command for bindings generation


karimnaaji:
How easily the generation of those files could be integrated in the build system?

hjanetzek:
@karimnaaji seems to be well supported: http://www.swig.org/Doc3.0/Introduction.html#Introduction_build_system 
I'll look into it

blair1618:
Can SWIG use 4-space indentation? It's not a big deal but if it's easy that would be nice

blair1618:
And should we start looking at the code here yet? Is it nearly done?

tallytalwar:
Will get to reviewing this tomorrow and try to get it merged by tomorrow.

@hjanetzek do you want the cmake command by swig binding generation to this PR?

hjanetzek:
@tallytalwar one can now run 'make swig-bindings' when hacking on the bindings. I think it's better for now to ship the generated files than to require a swig installation for building the library.

tallytalwar:
@hjanetzek cool, Thanks.

Reviewing this now.

android/demo/src/com/mapzen/tangram/android/MainActivity.java
```diff
@@ -39,17 +45,62 @@ public void onCreate(Bundle savedInstanceState) {
         mapController.setMapZoom(16);
         mapController.setMapPosition(-74.00976419448854, 40.70532700869127);
 
+        final MapData touchMarkers = new MapData("touch");
+        Tangram.addDataSource(touchMarkers);
+
+        // tags.set("color", "magenta");
```
tallytalwar: I guess we can remove this testing (commented) code.

core/include/mapbox/geojsonvt_custom_tags.hpp
```diff
@@ -0,0 +1,22 @@
...
+
+#include "../../../src/data/properties.h"
+
+#include <memory>
+
+namespace mapbox {
+namespace util {
+namespace geojsonvt {
+
+struct Tags {
```
tallytalwar: Is this to not use geojson-vt's `std::map` implementation of `Tags`? Or there are other reasons to defining our own `Tags` structure?

core/src/data/clientGeoJsonSource.cpp
```diff
@@ -73,38 +78,28 @@ void ClientGeoJsonSource::addPoint(double* _coords) {
...
 
-void ClientGeoJsonSource::addLine(double* _coords, int _lineLength) {
-    
-    std::vector<geojsonvt::LonLat> line(_lineLength, { 0, 0 });
-    for (int i = 0; i < _lineLength; i++) {
-        line[i] = { _coords[2 * i], _coords[2 * i + 1] };
-    }
-    
+void ClientGeoJsonSource::addLine(const Properties& _tags, const Coordinates& _line) {
+    auto& line = reinterpret_cast<const std::vector<geojsonvt::LonLat>&>(_line);
```
tallytalwar: Why are you doing a `const` with `reinterpret_cast`. I remember reading, its not recommended to do `const` with `reinterpret_cast`, but it might be for a specific use case. Just mentioning if you had checked for its usage!

hjanetzek: It was probably about removing constness - which is wrong in almost any case.

Only the usual danger of reinterpret_cast applies here, that one has to be sure that the struct layout of LonLat and LngLat dont change independently.

core/include/mapbox/geojsonvt_custom_tags.hpp
hjanetzek: yes, Tags are just attached to a geometry - which is by default a std::map<string,string>. To reduce conversion and copies we can use our Properties instead

core/src/tile/tileTask.h
```diff
@@ -1,14 +1,16 @@
 #pragma once
 
-#include "tile.h"
-#include "data/tileData.h"
 
 #include <memory>
+#include <vector>
+#include <functional>
```
tallytalwar: Why `functional` include? Not sure where this is being used?

tallytalwar:
I am good with this. Some commented code can be removed before this gets merged.

@blair1618 is also giving a look.

core/src/tile/tileTask.h
blair1618: I think `std::function` is used nearby

tallytalwar: :+1:

blair1618:
Looks good, will remove some leftover comments and merge

hjanetzek:
Oh, that was too quick suddenly :) Some follow up cleanups coming later. I would like to move the generated java source to 'gen' instead of 'src' folder in android/tangram.

blair1618:
Sure, sounds like a reasonable follow-up. Just wanted to start making this
feature available to eraser-map!

PR #342, opened 2015-10-13
Allow passing scene file path as command line argument

blair1618:
Even faster than drag-n-drop!
```bash
./build/linux/bin/tangram ~/Development/styles/eraser-map/eraser-map.yaml
```
or on OS X:
```bash
open build/osx/bin/tangram.app --args  ~/Development/styles/eraser-map/eraser-map.yaml
```


hjanetzek:
and even without having to fire up a file manager! :+1:

PR #343, opened 2015-10-14
Speedup build

hjanetzek:
- more forward declarations
- move tile hash function to separate file so that functional is not included everywhere with tileID.h
- add gl_fwd.h for all used gl functions and defines (could be made nicer by providing enums instead of defines, and using those in RenderState)
- add gl mock for 'unit' tests to not have to link gl libraries
- mock platform functions to avoid linking curl in tests
- move raw tile data cache into DataSource (could also be moved to a separate file as it is only used by network sources)
- even more forward declarations

For a small sample it seems to be more than a minute faster for the linux build on travis.


hjanetzek:
Are there any objections to merge this?

tests/unit/fontTests.cpp
```diff
@@ -1,4 +1,10 @@
 
+int main(int argc, char *argv[]) {
```
karimnaaji: I guess we can simply remove that file.

core/src/gl/gl_fwd.h
```diff
@@ -0,0 +1,334 @@
+#pragma once
+
+#ifdef PLATFORM_ANDROID
```
karimnaaji: Why is this needed here? If this is not needed we can we move this file to `include/`.

karimnaaji:
I have nothing against forward for GL, at least for basic types, but for GL functions, don't you think would be a pain to maintain this `gl_fwd.h` file if we start using ES 3.0 or simply while using extensions? I kind of like how simple that is to maintain currently.

core/src/gl/gl_fwd.h
hjanetzek: The android build is picky about providing its own typedef for GLsizeiptr (due to building with  -pedantic I guess)


OK the last two commit could need a bit more thought - We can postpone them from this PR, though I really like to not have to link GL + GLFW to build the tests..

karimnaaji: Yeah I agree, tests should be able to be built without linking to GL or GLFW.

hjanetzek:
Yes, I started to move the function declarations first into RenderState because the gl.h include there made a lot of files include gl.h indirectly. (currently 34 of 58 source files require the gl headers) 

I don't think that it will be hard to maintain, compared to the effort it takes to implement something that uses a new function, lookup its documentation etc.

blair1618:
What's the reason for moving `filters.h` from `data/` into `scene/`? I hesitate to move files around because it becomes much much harder to track their history.

hjanetzek:
Filters is only used by other classes in scene/ - Is there a particular reason for it being in data? Git does an awesome job tracking files - it applies even changes to the moved file when rebasing

blair1618:
It's mostly that github has a hard time reasoning about the history of moved files, e.g. https://github.com/tangrams/tangram-es/commits/speedup-build/core/src/scene/filters.h

Filter was accessing mostly classes from data/ when it was first created, but I agree that it probably doesn't belong there now.

hjanetzek:
yes, github has some outstanding issues..

    git log --follow core/src/scene/filters.h

blair1618:
Wow `--follow` is handy.  

This seems good to me! I know you're anxious to speed up your builds but let's merge https://github.com/tangrams/tangram-es/pull/326 first because I bet the gl header changes will conflict a bit. Should have that one merged today!

tallytalwar:
Good with this to be merged, once rebased.

hjanetzek:
Checked the merge with vaos locally - the only conflicts are in gl.h which I will resolve next

core/CMakeLists.txt
```diff
@@ -49,7 +46,7 @@ if(NOT DEFINED CORE_LIB_NAME)
     set(CORE_LIB_NAME core)
 endif()
 
-add_library(${CORE_LIB_NAME} ${CORE_LIB_TYPE} ${FOUND_SOURCES} ${FOUND_HEADERS})
```
karimnaaji: This is actually useful to add the headers using cmake project generation.

hjanetzek: You mean for the headers to appear in xcode?

karimnaaji: Yeah I think we can revert that part, except if there was other reasons to remove it?

hjanetzek:
I think it's wrong to put headers into sources: a quick search for 'cmake add headers to sources' revealed the same

http://voices.canonical.com/jussi.pakkanen/2013/03/26/a-list-of-common-cmake-antipatterns/

karimnaaji:
Yeah the project generation is a little sketchy about it.

hjanetzek:
Maybe  https://cmake.org/cmake/help/v3.0/command/source_group.html ?

    source_group("My Headers" FILES ${MY_HDRS})

blair1618:
It's superfluous for project generation, but it's also the simplest and most concise way to add header files to IDEs and has no perceptible side effects

hjanetzek:
Hmm, so then. Adding them back.

blair1618:
I am curious whether `source_group` also works, I'll give it a try

karimnaaji:
We already use it https://github.com/tangrams/tangram-es/blob/master/toolchains/utils.cmake#L55

karimnaaji:
But just noticed that the macro assumes that ${FOUND_HEADERS} is defined.

PR #344, opened 2015-10-14
Line fixes

blair1618:
More nearly matches the line behavior in tangram-js

### DIFF
--- a/core/src/style/polylineStyle.cpp
+++ b/core/src/style/polylineStyle.cpp
@@ -88,7 +88,7 @@ bool evalStyleParamWidth(StyleParamKey _key, const DrawRule& _rule, const Tile& 
     double tileSize = _tile.getProjection()->TileSize();
 
     // NB: 0.5 because 'width' will be extruded in both directions
-    double tileRes = 0.5 / _tile.getProjection()->TileSize();
+    double tileRes = 0.5 / tileSize;
 
 
     auto& styleParam = _rule.findParameter(_key);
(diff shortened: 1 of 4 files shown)


blair1618:
@tallytalwar take a look when you get a chance, thanks!

tallytalwar:
Sure, will do near the end of day today.

core/src/style/polylineStyle.cpp
```diff
@@ -84,10 +87,8 @@ bool evalStyleParamWidth(StyleParamKey _key, const DrawRule& _rule, const Tile&
     int zoom  = _tile.getID().z;
     double tileSize = _tile.getProjection()->TileSize();
```
tallytalwar: `tileSize` redundant now!

core/src/style/polylineStyle.cpp
```diff
@@ -177,12 +178,10 @@ void PolylineStyle::buildLine(const Line& _line, const DrawRule& _rule, const Pr
         } else { height = extrude[1]; }
     }
 
-    GLfloat layer = _props.getNumeric("sort_key") + params.order;
```
tallytalwar: So not using sort_key values for ordering of roads now?

blair1618: Nope, tangram-js doesn't use "sort_key" at all unless specified in the
scene file

core/src/style/polylineStyle.cpp
blair1618: Right you are; will remove

PR #345, opened 2015-10-14
[rfc] - fast drawrule merging 

hjanetzek:
There was a lot of copying of StyleParams in the SceneLayer matching. This patch splits the duplicate use of DrawRule into a static container class and a new DrawRule optimized for storing the match results. 

- DrawRule became StaticDrawRule as a container for the actual scene data.
  (a better would be welcome :)
- DrawRule is now a reusable result container with direct access to keys
  for fast merging and retrieval. For non-dynamic parameters it only holds
  the pointer the static DrawRule, so no copying there. 

- [ ] Update tests

### DIFF
--- a/core/src/scene/sceneLayer.cpp
+++ b/core/src/scene/sceneLayer.cpp
@@ -5,72 +5,73 @@
 
 namespace Tangram {
 
-SceneLayer::SceneLayer(std::string _name, Filter _filter, std::vector<DrawRule> _rules,
+SceneLayer::SceneLayer(std::string _name, Filter _filter,
+                       std::vector<StaticDrawRule> _rules,
                        std::vector<SceneLayer> _sublayers) :
     m_filter(_filter),
     m_name(_name),
     m_rules(_rules),
     m_sublayers(_sublayers) {
 
-    // Rules must be sorted to merge correctly
+    // Rules must be sorted to merge correctly - not anymore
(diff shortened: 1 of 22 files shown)


core/src/scene/sceneLayer.cpp
```diff
@@ -1,70 +1,100 @@
...
+    for (auto& rule : rules) {
+        auto& map = _styling.styles;
 
-    // Sublayers must be sorted by the depth of their deepest leaf in order to
-    // correctly traverse them in match()
-    std::sort(m_sublayers.begin(), m_sublayers.end(),
-              [&](const auto& a, const auto& b) {
-                  return a.m_depth > b.m_depth;
-              });
+        auto it = std::find_if(map.begin(), map.end(), [&](auto& m){
```
tallytalwar: Both and me @karimnaaji, found that doing a manual iteration and checking is faster than using std::find_if ... odd but true :|.

hjanetzek: In Release builds - not default builds, and was the test with integers? 

I thought of changing this part by using pre-allocated DrawRules for each style name so that one has a direct lookup.

tallytalwar:
@hjanetzek can you rebase this!. We should also get some eyes on this. This has some good perf improvements.

tallytalwar:
Closing in favor of https://github.com/tangrams/tangram-es/pull/372

PR #347, opened 2015-10-16
Only use draw rule name as style name after rules are merged

blair1618:
This fixes situations like this abridged example from eraser-map:
```yaml
    buildings:
        data: { source: osm }
        filter:
            not: { location: underground }
        draw:
            polygons:
                visible: false
                order: 50
                color: *building2
        footprints:
            filter:
                ...
            draw:
                polygons:
                    visible: true
        extrude:
            filter:
                ...
            draw:
                polygons:
                    visible: true
                    style: building-grid
                    extrude: true
                    color: [.870, .870, .870]
```
Previously, the `polygons` rule under `footprints` had the style `polygons` assigned to it at the time that it was parsed from the scene file. Then, when a feature matched both `footprint` and `extrude`, the rules had conflicting styles (`polygons` and `building-grid`, respectively) and the style that prevailed was indeterminate. 

The correct behavior, as implemented now, is for the merged rule to take the `style` parameter from `extrude`, where it is explicitly set.


hjanetzek:
This seem correct to me now. I was a bit baffled while porting the previous logic in the fast-drawrule-merging branch 

... still not that happy with the collateral merging. Maybe another way to state the intended merging order explicitly could be:

        extrude:
            mix: footprints
            filter:


this way one could even allow to mix values from a different layer, like 

    buildings:
        amenities:
             mix: /landuse/amenities
             draw:
                  polygons:
                      color: darken(10%)

tallytalwar:
Good catch @blair1618.
This looks good to merge.

blair1618:
Yeah, this was probably an oversight on my part from a while ago, when I had some misconceptions about the relationship between draw rule name and style name. 

Maybe this will simplify some logic in fast-drawrule-merging! Those changes make sense to me as well, I'm just going for correctness first.

PR #348, opened 2015-10-16
double precision for View::screenToGroundPlane

tallytalwar:
- Also provides an overloaded method which takes float values as screen position (for platform API Compatibility, since
some platform API calls require screen positions to be in float)


tallytalwar:
I like small PRs :D.

PR #349, opened 2015-10-16
Speed up your build pt.II - Run Travis!

hjanetzek:
This will be the pretty much the last of this series of non-requested infrastructure and refactoring branches for a while. I promise :) 

This branch adds the option to build tests and benchmarks together with the application and uses it for travis builds. It is now also possible to reconfigure the project to add build tests and benchmarks afterwards without having to build core and its dependencies again, using:
```
TANGRAM_CMAKE_OPTIONS="-DUNIT_TESTS=1 -DBENCHMARK=1" make -j4 linux
```

-  Added option to build glfw, yaml-cpp and benchmark as cmake-external projects. 
This should also come in handy when we add more libraries and to use prebuilt external libs for CI. You can try it with:
```
mkdir -p build/ext; cd build/ext; cmake ../../external -G Ninja -DUSE_EXTERNAL_LIBS=1 \
 -DPLATFORM_TARGET=linux -DPLATFORM_LINUX=1; ninja; cd - 
mkdir -p build/linux; cd build/linux; cmake ../.. -G Ninja -DUSE_EXTERNAL_LIBS=1 \
-DUNIT_TESTS=1  -DBENCHMARK=1  -DPLATFORM_TARGET=linux; ninja; 
```
- Moved git submodules that can be build separately to external directory


toolchains/linux.cmake
```diff
@@ -1,51 +1,48 @@
...
+# set for test in other cmake files
+set(PLATFORM_LINUX ON)
+
+# global compile options
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -std=c++1y")
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-unknown-pragmas")
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wparentheses")
-#set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wuninitialized")
-set(CXX_FLAGS_DEBUG "-g -O0")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fno-omit-frame-pointer")
```
hjanetzek: TODO: remove

tallytalwar: perf :D.

tallytalwar: I know we have it for perf here, but I think it make sense to not have this for our master builds.

tallytalwar:
Reviewing this next.

karimnaaji:
Reviewed this and looks good! Will push a fix to not build benchmarks as the default build behavior.

hjanetzek:
Thanks! Right, I also notice that benchmark got configured without bein enabled

blair1618:
Is there a reason for leaving some of the compiled dependencies in `core/dependencies` (css-color-parser, duktape, geojson-vt-cpp)?

karimnaaji:
I think the intent was to split between what we maintain or not.

karimnaaji:
But actually it might not be since `yaml-cpp` is also forked in tangrams.

blair1618:
That doesn't seem like it should affect how libraries are built; also `yaml-cpp` is our own fork so that would be inconsistent

hjanetzek:
@blair1618 the reason to move only some things to external was to test the approach (with the libraries that took longest to build). The other libraries could be moved over as well.

blair1618:
I see! There isn't much practical difference to moving `core/dependencies` to `external` (as you said, those dependencies aren't very slow to build), but it might be confusing for newcomers to the repository to see these two categories separated without a clear reason. 

I think it would be simpler and more consistent to move all compiled dependencies to the `external` folder. 

Other than that I'm happy with this PR!

PR #350, opened 2015-10-19
Cameras can be perspective, isometric, or flat

blair1618:
Camera type can now be specified in scene file as `perspective`, `isometric`, or `flat` (currently `flat` is equivalent to `isometric` with an oblique axis of `[0, 0]`)

`isometric` with default axis of `[0, 1]`:

screen shot 2015-10-19 at 12 55 34 pm

### DIFF
--- a/core/src/scene/sceneLoader.cpp
+++ b/core/src/scene/sceneLoader.cpp
@@ -958,9 +958,10 @@ void SceneLoader::loadCameras(Node _cameras, Scene& _scene) {
 
         }
 
-        double x = -74.00976419448854;
-        double y = 40.70532700869127;
-        float z = 16;
+        // Default is world origin at 0 zoom
+        double x = 0;
+        double y = 0;
+        float z = 0;
 
         if (Node position = camera["position"]) {
             x = position[0].as<double>();
--- a/core/src/scene/sceneLoader.cpp
+++ b/core/src/scene/sceneLoader.cpp
@@ -20,6 +20,7 @@
 #include "scene/spriteAtlas.h"
 #include "scene/stops.h"
 #include "util/yamlHelper.h"
+#include "view/view.h"
 
 #include "yaml-cpp/yaml.h"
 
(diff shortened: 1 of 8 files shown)


blair1618:
@tallytalwar can you check this over when you get a chance?

@bcamper see if the behavior in this build makes sense: https://www.dropbox.com/s/0z5b61itayn4uno/tangram-es.dmg?dl=0

bcamper:
Isometric behavior looks good! As mentioned, there's some odd clipping and LOD behavior, but that can be considered a separate issue.

core/src/scene/sceneLoader.cpp
```diff
@@ -916,43 +917,52 @@ void SceneLoader::loadCameras(Node _cameras, Scene& _scene) {
...
         } else if (type == "flat") {
-            // TODO
+
+            view->setCameraType(CameraType::flat);
+
         }
 
         double x = -74.00976419448854;
         double y = 40.70532700869127;
         float z = 16;
```
tallytalwar: Should we take these position values out from here and set these in the scene.yaml for tangram-es?

bcamper: Map starting location can be defined by the camera, we should follow up
with that if we don't support already. And yeah having a NYC baked into
core engine code is weird :) Can put a default in the demo app itself.

tallytalwar:
screenshot from 2015-10-21 12 53 53


Don't think LOD is working with isometric camera projection.

blair1618:
Yes, I turned it off intentionally for orthographic projections because
every tile takes up the same screen space, so reducing LOD for some doesn't
really make sense

tallytalwar:
Yup, just had a chat with @bcamper about this. Makes sense.

bcamper:
Actually, now that I'm using this and seeing that screenshot, it occurs to me that we might want to apply an LOD based on tilt? Like that is a *lot* of tiles to load, which is fairly slow. We might experiment with down-zooming a level of two with more extreme zooms.

blair1618:
Yeah, that's a possibility. All tiles in view take up the same screen
space, but it's a smaller screen space than usual, so we could uniformly
reduce the detail on all of  them.

core/src/view/view.cpp
```diff
@@ -134,7 +135,12 @@ void View::setRoll(float _roll) {
 
 void View::setPitch(float _pitch) {
 
-    m_pitch = glm::clamp(_pitch, 0.f, (float)HALF_PI);
+    float max_pitch = HALF_PI;
+    if (m_type != CameraType::perspective) {
+        // Prevent projection plane from intersecting ground plane
+        max_pitch = atan2(m_pos.z, m_height * .5f);
```
tallytalwar: I think this might not be fully working.. Check the following image, I got to this by moving from landscape view to portrait view. Hence maybe ... this should be `max(m_height, m_width)`?

screenshot_2015-10-21-15-41-20

tallytalwar: Doing a `std::max(m_width, m_height)` seems to fix this. Will push the commit.

blair1618: Oh yeah, I hadn't tried changing the orientation - that's an interesting problem. `m_height` is still the value to check against here, but its value changes when the screen is re-oriented and the pitch won't be checked for validity until the next time its set. A quick fix here might be to call `setPitch(m_pitch)` in the `resize` call - I'll try that.

blair1618: Checking the max won't fix this in the general case: Consider resizing a portrait-oriented window by hand and making it taller with the same width. Height is still the larger dimension but you've just created the same problem with pitch.

tallytalwar: Why is `m_height` the correct thing to check against in landscape view? I would believe it should be `m_width` right when in landscape mode?.. because you will be doing the shove gesture with respect to that orientation?

tallytalwar: Sure, and anyhow calling setPitch on a setSize is a better solution. If you havn't I can push it. Tried it, works!

blair1618: Our `m_width` and `m_height` aren't fixed device dimensions, they're screen dimensions with respect to the current orientation, so they change when the screen is re-oriented.

tallytalwar: Yes I know :D.

blair1618: Calling `setPitch` in `setSize` doesn't seem to cover every case (at least it doesn't cover resizing the window to be taller), so I'm gonna play with things a bit.

PR #352, opened 2015-10-19
VAOs

karimnaaji:
rebase of #326

### DIFF
--- a/core/src/gl/extension.cpp
+++ b/core/src/gl/extension.cpp
@@ -22,16 +22,12 @@ bool isAvailable(std::string _extension) {
 
 void printAvailableExtensions() {
     std::string exts(s_glExtensions);
-    std::istringstream iss(exts);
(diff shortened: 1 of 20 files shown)


core/src/gl/extension.cpp
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    std::string exts(s_glExtensions);
+    std::stringstream ss(exts);
+
+    ss >> std::noskipws;
+    std::string extension;
+    LOG("GL Extensions available: ");
+    while (std::getline(ss, extension, ' ')) {
+        LOG("\t - %s", extension.c_str());
```
blair1618: Definitely handy to have this for debugging, but it's a bit noisy to log these for every run

tallytalwar: True. We were talking about this yesterday. Maybe it makes sense to put it
under debug log method.

PR #357, opened 2015-10-20
Support for $geometry filter

hjanetzek:


core/src/util/fastmap.h
```diff
@@ -0,0 +1,115 @@
...
+
+namespace Tangram {
+
+// see also:
+// http://www.boost.org/doc/libs/1_59_0/doc/html/boost/container/flat_map.html
+// https://realm.io/assets/news/binary-search/blog.cpp
+
+template<typename K, typename T>
+struct fastmap {
+    std::vector<std::pair<K, T>> map;
```
blair1618: If we're getting picky about performance, do you think separate vectors of keys and values would provide better cache-locality for the binary search?

hjanetzek: Yes, we should try that. It probably depends on the size of the map. This is is a first start which showed clear improvement in manual perf testing. But we could also postpone fastmap until we have a proper benchmarks.

tallytalwar: I would stick with fastmap for now as it showed good improvements in perf.
Yes benchmarks are needed, but I don't see any point in postponing these
changes until benchmarks have been written.

Matt's suggestion could definitely be further improvement on this.

core/src/util/fastmap.h
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    const_iterator find(const K& key) const {
+        const_iterator it = std::lower_bound(
+            map.begin(), map.end(), key,
+            [&](const auto& item, const auto& key) {
+                return item.first < key;
+            });
+
+        if (it == map.end() || it->first.k == key) {
```
tallytalwar: we do not need to have `it == map.end()` in the id condition.

core/src/scene/filters.h
```diff
@@ -26,15 +26,23 @@ enum class FilterType : int {
     undefined
 };
 
+enum class FilterGlobal : int {
```
tallytalwar: We can have this as a `char` instead of an `int`.

nvkelso: +1, this is problem necessary for eraser map styling after some scene and tile refactor.

tallytalwar:
This is good. Once the discussions are answered/resolved, will merge.

core/src/util/fastmap.h
hjanetzek: this handles the case when the map is empty

tallytalwar: Even if you do not have this condition,  it's returning `map.end()`.

hjanetzek: but where does it->first look into when it == end? ;)

tallytalwar: Ohh I get it. Dumb thing. :+1:

hjanetzek:
waiting for travis

PR #359, opened 2015-10-20
Wrap world

tallytalwar:
This is mostly working, but I am going to change the implementation logic a bit to better caching for wrapped tiles.

- [x] better use of tilecache for wrapped tiles.

### DIFF
--- a/core/src/tile/tileManager.cpp
+++ b/core/src/tile/tileManager.cpp
@@ -370,7 +370,7 @@ bool TileManager::addTile(TileSet& tileSet, const TileID& _tileID) {
         m_tiles.push_back(tile);
 
         // Update tile origin based on wrap (set in the new tileID)
-        tile->updateTileOrigin();
+        tile->updateTileOrigin(_tileID.wrap);
         // Reset tile on potential internal dynamic data set
         tile->reset();
 
--- a/core/src/tile/tileManager.cpp
+++ b/core/src/tile/tileManager.cpp
@@ -369,6 +369,8 @@ bool TileManager::addTile(TileSet& tileSet, const TileID& _tileID) {
     if (tile) {
         m_tiles.push_back(tile);
 
+        // Update tile origin based on wrap (set in the new tileID)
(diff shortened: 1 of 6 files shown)


blair1618:
Reviewing

core/src/tile/tileManager.cpp
```diff
@@ -369,6 +369,8 @@ bool TileManager::addTile(TileSet& tileSet, const TileID& _tileID) {
     if (tile) {
         m_tiles.push_back(tile);
 
+        // Update tile origin based on wrap (set in the new tileID)
+        tile->updateTileOrigin(_tileID.wrap);
```
blair1618: Is this call necessary? I don't see any effect when I comment it. It seems strange that this function would even be public; it seems that a `Tile` could potentially get an origin that doesn't correspond to the `wrap` value in its `TileID`.

blair1618:
Hit a segfault when zooming out very rapidly, crash report: http://pastebin.com/UHbqCksZ

tallytalwar:
Checking for the segfault issue.

core/src/tile/tileManager.cpp
tallytalwar: Yes. Its when a tile data (corresponding to some xyz tileID) has been cached, and a new tile is being added with the wrapped IDs same as the cached tile's IDs. In this case you would want to have the tile grab the data from the cache and update the tile origin with the new wrap.

hjanetzek:
Which reminds me that we need better error handling in mvt sources pbf parser - at least the runtime_error should be catched there :)

hjanetzek:
Could it be that a tile is loaded twice, the first tile is added to the raw cache and starts processing - then the second comes in and is added to the cache overwriting the first that is currently processed by mvt parser?

tallytalwar:
@hjanetzek thats very very possible.

blair1618:
Turns out those crashes were totally my fault for making a mistake in the wrapping math and trying to fix nonexistent tiles - I think it's all corrected now.

PR #360, opened 2015-10-20
Fix mixed sign comparison

hjanetzek:
Could we turn on signed comparison warnings for clang? :)


blair1618:
Thanks for this! I think clang does have signed comparison warnings turned on - I know I see them at least when I build for OS X.

PR #361, opened 2015-10-22
Anchor points for sprites/icons

karimnaaji:
Add anchor points for sprites (fixes some issues related to *you are here* icon when client data source was added on eraser-map)

Anchor points work as the following:
- **(0.5, 0.5)**: center of the sprite (*default*)
- **(0.0, 0.0)**: lower left corner of the sprite
- **(1.0, 1.0)**: upper right corner of the sprite

It would be nice to add a handle on this from the stylesheet.

### DIFF
--- a/core/src/labels/spriteLabel.cpp
+++ b/core/src/labels/spriteLabel.cpp
@@ -1,19 +1,35 @@
 #include "labels/spriteLabel.h"
+#include "platform.h"
 
 namespace Tangram {
 
 SpriteLabel::SpriteLabel(Label::Transform _transform, glm::vec2 _size, LabelMesh& _mesh,
-                         int _vertexOffset, Label::Options _options, float _extrudeScale) :
+    int _vertexOffset, Label::Options _options, float _extrudeScale, glm::vec2 _anchor) :
     Label(_transform, _size, Label::Type::point, _mesh, {_vertexOffset, 4}, _options),
     m_extrudeScale(_extrudeScale)
 {}
 
 void SpriteLabel::updateBBoxes(float _zoomFract) {
+    glm::vec2 halfSize = m_dim * 0.5f;
     glm::vec2 sp = m_transform.state.screenPos;
     glm::vec2 dim = m_dim + glm::vec2(m_extrudeScale * 2.f * _zoomFract);
-
-    m_obb = OBB(sp.x, sp.y, m_transform.state.rotation, dim.x, dim.y);
+    m_obb = OBB(sp.x + halfSize.x, sp.y - halfSize.y, m_transform.state.rotation, dim.x, dim.y);
     m_aabb = m_obb.getExtent();
+}
+
+void SpriteLabel::align(glm::vec2& _screenPosition, const glm::vec2& _ap1, const glm::vec2& _ap2) {
+
+    switch (m_type) {
+        case Type::debug:
+        case Type::point:
+            _screenPosition.x -= m_dim.x * m_anchor.x;
+            _screenPosition.y += m_dim.y * m_anchor.y;
+            break;
+        case Type::line:
+            LOGW("Line sprite labels not implemented yet");
+            break;
+    }
+
 }
 
 }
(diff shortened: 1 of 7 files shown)


nvkelso:
+1. In terms of scene file controls, would it be part of / similar to the sprite sheet definition? Or individual overrides for 1 of icons?

bcamper:
It's all centered for now (to match JS) and not exposed in scene file. We'll decide how to expose after we incorporate this for labels a la https://github.com/tangrams/tangram/pull/182.

core/src/labels/spriteLabel.cpp
```diff
@@ -1,15 +1,35 @@
...
 namespace Tangram {
 
-SpriteLabel::SpriteLabel(Label::Transform _transform, glm::vec2 _size,
-                         LabelMesh& _mesh, int _vertexOffset, Label::Options _options) :
-    Label(_transform, _size, Label::Type::point, _mesh, {_vertexOffset, 4}, _options) {}
+SpriteLabel::SpriteLabel(Label::Transform _transform, glm::vec2 _size, LabelMesh& _mesh,
+                         int _vertexOffset, Label::Options _options, glm::vec2 _anchor) :
+    Label(_transform, _size, Label::Type::point, _mesh, {_vertexOffset, 4}, _options)
+ {
+    m_anchor = _anchor;
```
tallytalwar: Why not in constructor's initializer list?

karimnaaji: If you call the constructor of the base class initializer lists don't apply (c++11), except if this is a c++14 feature.

karimnaaji: Updated with initializer lists

tallytalwar:
Waiting on travis, will merge when travis goes shrek.

tallytalwar:
@karimnaaji can you rebase this and make it mergeable?

karimnaaji:
It'd be nice to settle a keyword soon to access this sprite property. 
An icon like this one:
screen shot 2015-10-29 at 10 31 35 am
Would need anchor set to `(0.5, 0.0)` to properly be placed (since you probably don't want to use an `offset` property as a hack to place it up, since you would need to know the height of the sprite to properly offset it of half the icon size, pretty annoying! Especially when you start changing icons), the main issue is to solve confusion between this and the cardinal placement of https://github.com/tangrams/tangram/pull/182.
I think that anchor (or any other name we settle) should be a sprite definition, defined in the `sprites` object  property:
```yaml
textures:
    pois:
        url: images/poi_icons_18@2x.png
        filtering: mipmap
        sprites:
            # define sprites: [x origin, y origin, width, height]
            adult-boutique: 
                sprite: [802, 1116, 36, 38]
                anchor: [0.5, 0.0]
            airport: [631, 105, 36, 38]
            animal-shelter: [380, 1364, 36, 38]
```
cc @bcamper @nvkelso

bcamper:
For https://github.com/tangrams/tangram/pull/182, I'm updating to use `anchor`. It makes sense to me to be able to position sprites relative to an anchor, just like labels. However, I think (similar to labels) this belongs at the style rules level, rather than being hard-coded in the sprite definition for the texture. You could potentially place the same sprite at different anchor positions (unlikely but possible), and I see the anchor as part of the *placement* of the feature, rather than part of its inherent definition.

karimnaaji:
Agreed with that, so we could use the same keyword `anchor` for this too.
Meaning that anchor could be `up`, `down`, `left`... Or a vector of two positions describing the placement considering the bounding box of the object being placed (a text label could use `(0.5, 0.5)` to be centered, or `(0.0, 0.0)`, to be positioned at the lower left corner of the original feature position), and `anchor` would be a keyword applying for `point` and `text` style.

bcamper:
Right, `anchor` will already apply to text once that other PR is merged,
all we need to do is extend it to points (and add both to ES :)  I'm hoping
to merge that PR today.

nvkelso:
The anchor is most often know at artwork authoring time, and rarely (but we should support it) changed later. The spritesheet definition is closest to authoring time, it makes most sense to me to define it there, and override if necessary later.

bcamper:
I suppose a sprite could define a default `anchor` and it could be
overriden by the style rules (I do want to have this set in style rules
because that's where it's done for text so it should be consistent). This
feels a little overkill to me so I think we should add it to match text
first, then we can add a sprite definition default if needed.

karimnaaji:
I was mostly bringing that up because this is something that came up on the last android meeting, to place the start and end sprite of the route..

bcamper:
Yep, it makes sense to add, I just want it to parallel the (almost merged)
text anchors as closely as possible (and presumably share code as well).

blair1618:
Reviewing!

tallytalwar:
This is good to be merged. I was waiting for the discussion to possible result in some code change.
@karimnaaji any new code changes coming in this?

karimnaaji:
No changes to be done here.

PR #362, opened 2015-10-22
Interpolate stops size

karimnaaji:
Linearly interpolate between stops size from the vertex shader at fractional zoom levels.

### DIFF
--- a/core/resources/shaders/point.vs
+++ b/core/resources/shaders/point.vs
@@ -57,7 +57,8 @@ void main() {
         float de = a_extrude.y;
 
         if (de != 0.0) {
-            vertexPos.xy += unitExtrude[int(a_extrude.x)].xy * fract(u_map_position.z) * de;
+            float dz = u_map_position.z - abs(u_tile_origin.z);
+            vertexPos.xy += dz * unitExtrude[int(a_extrude.x)].xy * de;
         }
 
         // rotates first around +z-axis (0,0,1) and then translates by (tx,ty,0)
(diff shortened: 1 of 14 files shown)


hjanetzek:
:+1: Works! 

Since the label transformations are computed for each frame I wonder if it wouldn't make sense to just have a single screen-position attribute that represents the actual position on screen instead of computing it in the shader from screen-position + position + extrusion attributes.

karimnaaji:
@hjanetzek , yeah I thought of this at some point; but since our overhead is mostly on the CPU right now it might be good to let this to the vertex shader. If it's for memory consumption though, it would make sense to move this out, to reduce the buffer size.

We can address that once we start reducing buffer sizes while packing attributes.

hjanetzek:
ok then, ready to merge?

karimnaaji:
:+1:

core/resources/shaders/point.vs
```diff
@@ -46,10 +53,18 @@ void main() {
         float st = sin(a_rotation);
         float ct = cos(a_rotation);
 
+        vec2 vertexPos = a_position;
+        float de = a_extrude.y;
+
+        if (de != 0.0) {
+            float dz = u_map_position.z - abs(u_tile_origin.z);
+            vertexPos.xy += dz * unitExtrude[int(a_extrude.x)].xy * de;
```
karimnaaji: Accessing an array element with a non-constant or non-loop variable makes some drivers not receive this statement. Will follow-up on a fix related to this.

PR #363, opened 2015-10-23
Handle undefined js return value

hjanetzek:


tallytalwar:
@blair1618.

Small fixes, can you review this quickly. Should not take a lot of time.

core/src/scene/styleContext.cpp
```diff
@@ -310,8 +315,10 @@ bool StyleContext::parseStyleResult(StyleParamKey _key, StyleParam::Value& _val)
             default:
                 break;
         }
+    } else if (duk_is_null_or_undefined(m_ctx, -1)) {
+        // Ignore setting value
```
blair1618: Can we put a `LOGD` here so logging can be switched back on when we're feeling paranoid?

tallytalwar: :+1:

core/src/tile/tile.cpp
```diff
@@ -73,7 +73,7 @@ void Tile::build(StyleContext& _ctx, const Scene& _scene, const TileData& _data,
                     auto* style = _scene.findStyle(rule.getStyleName());
 
                     if (style) {
-                        rule.eval(_ctx);
+                        if (!rule.eval(_ctx)) { continue; }
```
blair1618: This isn't actually the current behavior of tangram-js, the trello card I created about this was somewhat mis-informed. As I understand it, tangram-js does not check whether the "required" rules of order, priority, or width are set to valid values. We might want that to be the future behavior, but it may cause discrepancies right now. cc @bcamper

blair1618: Note for posterity: This *is* the desired future behavior (except for priority, which isn't actually required) and we'll move forward with this check in tangram-es

hjanetzek:
But why is < -EPSILON accepted? Thinking about it, width == 0 or dwdz == 0 are valid when the other is greater 0. only when both are <= 0  (or epsilon) the line can be ommited

bcamper:
Let's **not** support `String.includes()`, because it's an ES6 method that is not guaranteed to be supported on all browsers or environments. Rather than polyfilling a hodgepodge of ES6 features, we should simply limit ourselves to ES5 for our JS properties/filters. I will also file an issue for the Eraser Map style to replace this with simple `indexOf()` functionality, which is easy in this case (string `includes()` is really just some simple syntactic sugar).

blair1618:
I think this covers the functionality we need right now. 

Conceptually, it would be nice to have all of these validity checks happen in one place like `DrawRule::eval`, instead of being sort of scattered.

blair1618:
Will follow-up with some re-organization later, merging for now.

PR #367, opened 2015-10-27
Word wrapping

karimnaaji:
as in [tangram-js](https://github.com/tangrams/tangram/pull/205)

- [x] alignments (center/left/right)
- [x] add stylesheet support

screen shot 2015-10-27 at 10 23 17 am

### DIFF
--- a/core/src/labels/textLabel.cpp
+++ b/core/src/labels/textLabel.cpp
@@ -24,7 +24,7 @@ void TextLabel::updateBBoxes(float _zoomFract) {
     // move down on the perpendicular to estimated font baseline
     obbCenter -= tperp * m_dim.y * 0.5f;
     // ajdust with baseline
-    obbCenter.y += m_metrics.lineHeight * m_nLines * 0.5f;
+    obbCenter += tperp * m_metrics.lineHeight * (float) m_nLines * 0.5f;
 
     m_obb = OBB(obbCenter.x, obbCenter.y, m_transform.state.rotation, m_dim.x, m_dim.y);
     m_aabb = m_obb.getExtent();
--- a/core/src/labels/textLabel.cpp
+++ b/core/src/labels/textLabel.cpp
@@ -22,7 +22,9 @@ void TextLabel::updateBBoxes(float _zoomFract) {
     // move forward on line by half the text length
     obbCenter += t * m_dim.x * 0.5f;
     // move down on the perpendicular to estimated font baseline
-    obbCenter -= m_dim.y * 0.5f + m_metrics.descender * (float) m_nLines;
+    obbCenter -= tperp * m_dim.y * 0.5f;
+    // ajdust with baseline
+    obbCenter.y += m_metrics.lineHeight * m_nLines * 0.5f;
 
     m_obb = OBB(obbCenter.x, obbCenter.y, m_transform.state.rotation, m_dim.x, m_dim.y);
     m_aabb = m_obb.getExtent();
--- a/core/src/labels/textLabel.cpp
+++ b/core/src/labels/textLabel.cpp
@@ -2,9 +2,10 @@
 
 namespace Tangram {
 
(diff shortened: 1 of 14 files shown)


core/src/text/textBuffer.cpp
```diff
@@ -15,6 +15,22 @@ TextBuffer::TextBuffer(std::shared_ptr<VertexLayout> _vertexLayout)
 TextBuffer::~TextBuffer() {
 }
 
+std::vector<TextBuffer::WordBreak> TextBuffer::findWordBreaks(const std::string& _text) {
+    std::vector<WordBreak> breaks;
+
+    for (int i = 0, quad = 0; i < _text.size(); ++i) {
+        if (std::isspace(_text[i])) {
+            int lookAHead = i;
```
bcamper: very minor nitpick, the H in lookAHead is probably not meant to be capitalized :)

core/src/text/textBuffer.cpp
```diff
@@ -15,6 +15,22 @@ TextBuffer::TextBuffer(std::shared_ptr<VertexLayout> _vertexLayout)
 TextBuffer::~TextBuffer() {
 }
 
+std::vector<TextBuffer::WordBreak> TextBuffer::findWordBreaks(const std::string& _text) {
```
bcamper: Looks like we also need to handle explicit line breaks (`\n` chars) in the text, to match JS (unless I missed it somewhere). For example, the string `One\nTwo\nThree` should render as:

```
One
Two
Three
```

core/src/text/textBuffer.cpp
karimnaaji: Oh yes :)

core/src/text/textBuffer.cpp
karimnaaji: Line breaks `\n` is something that can come from the string data?

bcamper: Yep, either if it was in the source data, or added in the stylesheet by a JS function, e.g. `text_source: function(){ return feature.name.replace(' ', '\n') }`. When there are explicit line breaks, we assume that they were desired, so we always break on them.

karimnaaji: Ok makes sense, I didn't assume that, but will update considering this.

core/src/labels/textLabel.cpp
```diff
@@ -20,9 +21,10 @@ void TextLabel::updateBBoxes(float _zoomFract) {
     obbCenter = m_transform.state.screenPos;
     // move forward on line by half the text length
     obbCenter += t * m_dim.x * 0.5f;
-    // TODO: use real font metrics
     // move down on the perpendicular to estimated font baseline
-    obbCenter -= tperp * (m_dim.y / 8);
+    obbCenter -= tperp * m_dim.y * 0.5f;
+    // ajdust with baseline
+    obbCenter.y += m_metrics.lineHeight * (float) m_nLines * 0.5f;
```
tallytalwar: Don't think we need to have `float` typecasting here.

core/src/text/fontContext.h
```diff
@@ -64,10 +73,12 @@ class FontContext {
     void initFontContext(int _atlasSize);
 
     std::map<std::string, int> m_fonts;
+    std::map<int, FontMetrics> m_fontMetrics;
```
tallytalwar: I don't see this being used anywhere.

tallytalwar:
Missing `word_wrap` parameter parsing to have styling control for text wrapping. Currently only default values will be respected.

bcamper:
@tallytalwar it's actually `text_wrap`, but it looks like @karimnaaji got it.

tallytalwar:
@bcamper yup. Sorry meant that :D.

tallytalwar:
This looks good to me now.

PR #369, opened 2015-10-27
Fix dependent array access

karimnaaji:
As mentioned in #362


tallytalwar:
Looks good and also fixes the issue.

PR #370, opened 2015-10-28
Tilemanager handle failure

hjanetzek:
Updated th last PR: an empty tile is now considered to be 'canceled'. This way it stays in the list of current tiles but will not be further processed or tried to become reloaded (as long as it stays in the list of current tiles).

### DIFF
--- a/core/src/tile/tileManager.cpp
+++ b/core/src/tile/tileManager.cpp
@@ -24,7 +24,14 @@ TileManager::TileManager()
     m_tileCache = std::unique_ptr<TileCache>(new TileCache(DEFAULT_CACHE_SIZE));
 
     m_dataCallback = TileTaskCb{[this](std::shared_ptr<TileTask>&& task){
-        if (setTileState(*task->tile, TileState::processing)) {
+        if (!task->rawTileData || task->rawTileData->empty()) {
+            // Set 'canceled' state when no data was received,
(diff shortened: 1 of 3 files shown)


core/src/tile/tileManager.cpp
```diff
@@ -97,61 +104,60 @@ std::shared_ptr<ClientGeoJsonSource> TileManager::getClientSourceById(int32_t _i
...
 
-void TileManager::tileProcessed(std::shared_ptr<TileTask>&& task) {
+void TileManager::tileProcessed(std::shared_ptr<TileTask>&& _task) {
     std::lock_guard<std::mutex> lock(m_readyTileMutex);
-    m_readyTiles.push_back(std::move(task));
+    m_readyTiles.push_back(std::move(_task));
 }
 
-bool TileManager::setTileState(Tile& tile, TileState state) {
+bool TileManager::setTileState(Tile& _tile, TileState _newState) {
```
tallytalwar: We should have unit tests for this, will help in any future modifications and will catch issues we catch the hard way in this branch.

tallytalwar:
Tested, looks good.

tallytalwar:
This is good. Discussed with @hjanetzek about making unit tests for tileState changes. But currently its a bit convoluted to make it testable. Will create a task for this and merge this PR as it is right now.

PR #371, opened 2015-10-28
Tilemanager tweaks

hjanetzek:
- Cancel download of tiles that are 3 or more zoom-level above the current
- when a tile becomes available that can be used as proxy, update proxy tiles of pending tiles
- give higher priority to tiles from lower zoom-levels as they cover a larger visible area

### DIFF
--- a/core/src/tile/tileManager.cpp
+++ b/core/src/tile/tileManager.cpp
@@ -57,19 +57,18 @@ void TileManager::setScene(std::shared_ptr<Scene> _scene) {
                 sources.begin(), sources.end(),
                 [&](auto& s){ return tileSet.source->equals(*s); });
 
-            if (sIt != sources.end()) {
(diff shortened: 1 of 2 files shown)


hjanetzek:
rfc - seems to be an unpopular label :)

The question is at which level pending tiles from higher zoom-level should be canceled - It doesnt make sense to keep loading and processing tiles from z15 while having zoomed out to z0. I think max 3 levels above the current is a reasonable limit.

blair1618:
Nice! This makes a noticeable improvement when changing scale rapidly. Cancelling at 3 zoom levels away sounds like a reasonable to me.

Will merge if there's no more to discuss here.

bcamper:
I think we could cancel even more aggressively then 3 levels (I think I am
canceling more than 1 away in JS), but this can easily be tuned later.

PR #372, opened 2015-10-29
Improve Drawrule matching performance

hjanetzek:
continuing #345

### DIFF
--- a/core/src/scene/drawRule.cpp
+++ b/core/src/scene/drawRule.cpp
@@ -9,9 +9,10 @@ namespace Tangram {
 
 const StyleParam NONE;
 
-DrawRule::DrawRule(const std::string& _name, const std::vector<StyleParam>& _parameters,
+DrawRule::DrawRule(std::string _styleName, int _styleId, const std::vector<StyleParam>& _parameters,
                    bool _sorted) :
-    name(_name),
+    styleName(_styleName),
+    styleId(_styleId),
     parameters(_parameters) {
 
     if (!_sorted) {
@@ -40,7 +41,7 @@ DrawRule DrawRule::merge(DrawRule& _other) const {
     while (myIt != myEnd) { merged.push_back(*myIt++); }
     while (otherIt != otherEnd) { merged.push_back(std::move(*otherIt++)); }
 
-    return { name, merged, true };
+    return { styleName, styleId, merged, true };
 }
 
 std::string DrawRule::toString() const {
@@ -76,7 +77,7 @@ bool DrawRule::isJSFunction(StyleParamKey _key) const {
 }
 
 bool DrawRule::operator<(const DrawRule& _rhs) const {
-    return name < _rhs.name;
+    return styleId < _rhs.styleId;
 }
 
 bool DrawRule::eval(const StyleContext& _ctx) {
@@ -109,7 +110,7 @@ const std::string& DrawRule::getStyleName() const {
     if (style) {
         return style.value.get<std::string>();
     } else {
-        return name;
+        return styleName;
     }
 }
 
--- a/core/src/scene/drawRule.cpp
+++ b/core/src/scene/drawRule.cpp
@@ -33,8 +33,8 @@ DrawRule DrawRule::merge(DrawRule& _other) const {
         } else if (*otherIt < *myIt) {
             merged.push_back(std::move(*otherIt++));
         } else {
-            merged.push_back(*otherIt++);
-            myIt++;
+            merged.push_back(*myIt++);
+            otherIt++;
         }
     }
     while (myIt != myEnd) { merged.push_back(*myIt++); }
--- a/core/src/scene/drawRule.cpp
+++ b/core/src/scene/drawRule.cpp
@@ -42,16 +42,14 @@ void DrawRule::logGetError(StyleParamKey _expectedKey, const StyleParam& _param)
     LOGE("wrong type '%d'for StyleParam '%d'", _param.value.which(), _expectedKey);
 }
 
-void Styling::apply(const Feature& _feature, const Scene& _scene, const SceneLayer& _layer,
-                    StyleContext& _ctx, Tile& _tile) {
+bool Styling::match(const Feature& _feature, const SceneLayer& _layer, StyleContext& _ctx) {
 
     _ctx.setFeature(_feature);
-
-    // Match layers
-    if (!_layer.filter().eval(_feature, _ctx)) { return; }
-
     styles.clear();
     processQ.clear();
+
+    // Match layers
+    if (!_layer.filter().eval(_feature, _ctx)) { return false; }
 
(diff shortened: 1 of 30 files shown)


hjanetzek:
Some tile build times for 10/301/384.mvt (on a xeon 2560). This is a pretty heavy tile, about 750kb:
For eraser-map style build time (style matching and geometry creation, without parsing) went down from 110ms to 65ms. For the default style from 68ms to 45ms.

Unfortunately one cannot build google benchmark on android so I will try it there with simple chrono counters.

```
Benchmark                             Time(ns)    CPU(ns) Iterations

# base: eraser
TileLoadingFixture/BuildTest_mean    110602791  110466660          5
TileLoadingFixture/BuildTest_stddev    1931617    2001095          0

# perf: eraser
TileLoadingFixture/BuildTest_mean     73978121   73952386          7
TileLoadingFixture/BuildTest_stddev    2108898    2108727          0

# perf: eraser - property key size as primary sort key
TileLoadingFixture/BuildTest_mean     65133547   65083325          8
TileLoadingFixture/BuildTest_stddev    2477973    2491654          0


# base: default
TileLoadingFixture/BuildTest_mean     67846241   67833338          8
TileLoadingFixture/BuildTest_stddev    3576580    3650531          0

# perf: default
TileLoadingFixture/BuildTest_mean     48632142   48027775         12
TileLoadingFixture/BuildTest_stddev    2522254    2348654          0

# perf: default - property key size as primary sort key
TileLoadingFixture/BuildTest_mean     44576521   44564100         13
TileLoadingFixture/BuildTest_stddev    2983526    2947713          0
```

blair1618:
Alright, I'm convinced; I'll try to review this today :)

core/src/scene/drawRule.cpp
```diff
@@ -1,120 +1,152 @@
...
-    return true;
 }
 
-const std::string& DrawRule::getStyleName() const {
+void Styling::mergeRules(const std::vector<StaticDrawRule>& rules) {
+    for (auto& rule : rules) {
 
-    const auto& style = findParameter(StyleParamKey::style);
+        auto it = std::find_if(styles.begin(), styles.end(), [&](auto& m) {
+                return rule.styleId == m.styleId; });
```
blair1618: This confuses me. It looks from the code as if rules are being merged according to their `style`, but they *should* be merged by the names of the rules (which can be the same as their `style`, but in general are not). 

The complete rule application process seems to produce the correct behavior somehow (at least in some brief tests), but at a minimum I'd like to refactor this function to make the intent more clear.

blair1618: Ok, having read through more of the changes I understand why this works. 

I think this logic is a little too clever for it's own good - I'm certain that I will be confused by this again if left as-is. I'll see if it reads any better with some renaming.

hjanetzek:
:+1: for tackling the hard problem of variable naming :)

blair1618:
:+1: for fixing syntax in those tests

core/src/scene/drawRule.cpp
```diff
@@ -1,120 +1,147 @@
...
+
+        if (valid) {
+            style->buildFeature(_tile, _feature, rule);
         }
     }
-    return true;
 }
 
-const std::string& DrawRule::getStyleName() const {
+void Styling::mergeRules(const std::vector<StaticDrawRule>& rules) {
```
tallytalwar: This is nice, way simpler than before.

core/src/scene/sceneLayer.cpp
```diff
@@ -4,67 +4,37 @@
...
-    m_rules(std::move(_rules)),
-    m_sublayers(std::move(_sublayers)),
-    m_depth(0) {
+    m_filter(_filter),
+    m_name(_name),
+    m_rules(_rules),
+    m_sublayers(_sublayers) {
 
-    // Rules must be sorted to merge correctly
+    // Rules must be sorted to merge correctly - not anymore
```
tallytalwar: yes not anymore, 1 more less sorting :D.

core/src/scene/styleParam.h
```diff
@@ -40,15 +42,16 @@ enum class StyleParamKey : uint8_t {
...
     width,
-    centroid,
-    collide,
-    transition_show_time,
-    transition_hide_time,
-    transition_selected_time
 };
 
+// UPDATE WITH StyleParamKey CHANGES!
+constexpr size_t StyleParamKeySize = static_cast<size_t>(StyleParamKey::width)+1;
```
tallytalwar: A potential candidate for future bug, if someone modifying the Key, forgets to update this.

core/src/tile/tile.cpp
```diff
@@ -138,14 +141,18 @@ void Tile::draw(const Style& _style, const View& _view) {
 }
 
 std::unique_ptr<VboMesh>& Tile::getMesh(const Style& _style) {
-    return m_geometry[_style.getName()];
+    static std::unique_ptr<VboMesh> NONE = nullptr;
+    if (_style.getID() >= m_geometry.size())
```
tallytalwar: Can we have braces here? Conforming to our code styling. #nitpick

core/src/scene/drawRule.h
```diff
@@ -38,15 +53,39 @@ struct DrawRule {
...
+     */
+    void apply(const Feature& _feature, const Scene& _scene,
+               const SceneLayer& _sceneLayer,
+               StyleContext& _ctx, Tile& _tile);
+
+    // internal
+    bool match(const Feature& _feature, const SceneLayer& _layer, StyleContext& _ctx);
+
+    // internal
+    void mergeRules(const std::vector<StaticDrawRule>& rules);
```
tallytalwar: `apply`, `match` and `mergeRules`: potential candidate for inline?

Don't think it will lead a lot of performance improvement, but something to think about based on their usage.

core/src/scene/styleParam.h
hjanetzek: yeah.. even though it's just one line below the referenced value -- What about a last item for StyleParamKey::NUM_ELEMENTS to make it more obvious?

core/src/scene/drawRule.h
hjanetzek: I would let the compiler decide in this case

blair1618: I believe the compiler decides regardless ;)

tallytalwar: true.

core/src/scene/styleParam.h
blair1618: Yeah that seems better.

tallytalwar: Yes, that's better.

core/src/tile/tile.cpp
hjanetzek: fixed

PR #374, opened 2015-11-02
Polyline extrusion fix

karimnaaji:
Really small fix on polyline extrusion, over extrusion was happening when a proxy tile was still being seen at really high zoom level while being on low zoom level:
screen shot 2015-11-02 at 12 01 58 pm


tallytalwar:
looks good to me.

PR #375, opened 2015-11-02
World normals

karimnaaji:
Add `worldNormal()` access for both fragment and vertex shader. To easily detect up-pointing faces like roofs on tilted view.
Following of an issue related in https://github.com/tangrams/eraser-map/pull/56.
Once merged, update eraser-map stylesheet.

screen shot 2015-11-02 at 1 34 32 pm

- [x] revert stylesheet once reviewed

### DIFF
--- a/core/resources/shaders/polygon.fs
+++ b/core/resources/shaders/polygon.fs
@@ -16,6 +16,7 @@ uniform vec2 u_resolution;
 uniform float u_time;
 uniform float u_meters_per_pixel;
 uniform float u_device_pixel_ratio;
+uniform mat3 u_inverseNormalMatrix;
 
 #pragma tangram: uniforms
 
@@ -32,6 +33,10 @@ varying vec2 v_texcoord;
 #pragma tangram: material
 #pragma tangram: lighting
 #pragma tangram: global
+
+vec3 worldNormal() {
+    return u_inverseNormalMatrix * v_normal;
+}
 
 void main(void) {
 
--- a/core/resources/shaders/polygon.vs
+++ b/core/resources/shaders/polygon.vs
@@ -43,6 +43,10 @@ vec4 worldPosition() {
     return v_world_position;
 }
 
+vec3 worldNormal() {
+    return a_normal;
+}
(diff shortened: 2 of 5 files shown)


hjanetzek:
What speaks against passing the untransformed vertex normal  to the polygon fragment shader?
The trade-off seems to be spending one uniform + a vector-matrix multiplication per fragment vs spending one varying (which is likely to be optimized out when not used)

karimnaaji:
I don't know, for me both solutions are fine, especially since one reduce the matrix mult, but this varying optimization is unclear since it seems to be driver specific, even if a warning is generated it's really unsure whether it'd be thrown away.

bcamper:
@hjanetzek either works, but @blair1618 and I were thinking that this feature is not used so often that pre-allocating an entire varying is worth it. If someone wanted to, they could create a varying themselves? If we find it's needed commonly, then we could decide to give up a varying for it, but since we may start to bump up against device limits for those, we wanted to be cautious.

blair1618:
Many mobile GPUs only have 8 interpolators and right now our shaders use 6 of those with built-in varying vectors. The cost of passing the un-transformed vertex normal into the fragment shader as a varying is that it further reduces the varyings available to a user. I agree that we don't know the exact trade-off of this yet, but since the user only sees the `worldNormal()` function, we're free to swap out the matrix multiplication for a varying in the future.

hjanetzek:
ok, the per fragment multiplication is probably not the expensive part of shaders using that feature. 

Just had an optimization thought for our use: If one could set colors for the wall and flat top separately then the fragment shader could just check the vertex color :)

bcamper:
@hjanetzek, we've seen that used a few times, so may be worth putting in options for different vertex color. Some cases (including current Eraser Map style) actually still need the normal because they do more complex changes like change texture/material properties... which wouldn't be possible within the bounds of the current single-VBO architecture (could be considered in the future though).

blair1618:
@karimnaaji This looks good to go after that test style code is reverted.

Issue #377, opened 2015-11-02
Proxy geometry drawn at misleading depth


blair1618:
Geometry from proxy tiles appears to sometimes draw over geometry that should occlude it in 3D space, especially in tilted views. 

I suspect that it's a consequence of this line: https://github.com/tangrams/tangram-es/blob/master/core/resources/shaders/polygon.vs#L93 (and the corresponding line in the polyline shader). 
```glsl
// Proxy tiles have u_tile_origin.z < 0, so this re-scaling will place proxy tiles deeper in
// the depth buffer than non-proxy tiles by a distance that increases with tile zoom
gl_Position.z /= 1. + .1 * (abs(u_tile_origin.z) - u_tile_origin.z);
```
Clearly this was deliberate, and solves (or used to solve) some problem, but now when I comment out these lines the visual result is *much* more pleasant. 

@tallytalwar @karimnaaji Thoughts?

PR #378, opened 2015-11-02
More gentle proxy tile depth adjustment

blair1618:
To address https://github.com/tangrams/tangram-es/issues/377

### DIFF
--- a/core/resources/shaders/polygon.vs
+++ b/core/resources/shaders/polygon.vs
@@ -88,9 +88,9 @@ void main() {
 
     gl_Position = u_proj * v_position;
 
-    // Proxy tiles have u_tile_origin.z < 0, so this re-scaling will place proxy tiles deeper in
-    // the depth buffer than non-proxy tiles by a distance that increases with tile zoom
-    gl_Position.z /= 1. + .1 * (abs(u_tile_origin.z) - u_tile_origin.z);
+    // Proxy tiles have u_tile_origin.z < 0, so this adjustment will place proxy tiles 
+    // deeper in the depth buffer than non-proxy tiles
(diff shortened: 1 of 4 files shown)


tallytalwar:
This seems to behave mostly right.

We saw a this case, when all the children tile of a proxy parent are loaded, there is a flash of proxy parent shown before all children (visible) get displayed.

Its weird because proxy tile should never be thrown on top of the visible tiles, which is happening here.

hjanetzek:
I think another visible interference happens when a distant tile is not a proxy, but the near tiles are proxies

tallytalwar:
Hannes Shouldn't tile sorting make sure that near tiles are loaded first.
That will avoid this issue.

hjanetzek:
I mean e.g. the case where the view is tilted and shows only non-proxy tiles, then when zooming out the near tiles switch to a lower zoom-level first (or maybe only the near tiles) and use the previous tiles as proxies while the distant tiles are still on the old zoom-level. Then you have near tiles that use proxies while distant tiles are shown in front. 
Another case where this may happen is when the distant tiles are cached but the near ones aren't

blair1618:
Yes, I think this has been happening in tilted views - it looks pretty odd when it happens. 

I believe that should be fixed with this change. The depth adjustment that proxy geometry gets is so small that it should no longer impact geometries that are actually separated in 3d space.

hjanetzek:
@tallytalwar wouldn't the proxy tile commit better fit into the tilemanager-tweaks branch?

tallytalwar:
@hjanetzek It was fixing, something were seeing for proxy so I added it over here.

Also I did not notice the tilemanager tweak PR (I think it was named something else before).

PR #380, opened 2015-11-03
Easing for view parameters

blair1618:
Allows interpolated transitions between view states

TODO:
 - [x] Document eased setters in `tangram.h`
 - [x] Document eased setters in `MapController.java`
 - [x] Last call to setter (with or without easing) determines final parameter value

### DIFF
--- a/android/demo/src/com/mapzen/tangram/android/MainActivity.java
+++ b/android/demo/src/com/mapzen/tangram/android/MainActivity.java
@@ -77,6 +77,9 @@ public class MainActivity extends Activity {
                 lastTappedPoint.set(tapPoint);
 
                 mapController.pickFeature(event.getX(), event.getY());
+
+                mapController.setMapPosition(tapPoint.longitude, tapPoint.latitude, 1);
+
                 return true;
             }
         });
--- a/android/tangram/jni/jniExports.cpp
+++ b/android/tangram/jni/jniExports.cpp
@@ -7,6 +7,10 @@ extern "C" {
         Tangram::setPosition(lon, lat);
     }
 
+    JNIEXPORT void JNICALL Java_com_mapzen_tangram_MapController_setPositionEased(JNIEnv* jniEnv, jobject obj, jdouble lon, jdouble lat, jfloat duration, jint ease) {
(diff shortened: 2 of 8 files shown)


blair1618:
Hey @ecgreb can you do an API thought experiment for me? If you have a function `setMapPosition(x, y)` that immediately moves to `(x, y)` and a function `setMapPosition(x, y, t)` that moves to `(x, y)` over `t` seconds, what would you expect to be the final positions of these sequences:
_A_
```
setMapPosition(x0, y0, 2);
setMapPosition(x1, y1, 1);
```
_B_
```
setMapPosition(x0, y0, 1);
setMapPosition(x1, y1);
```

ecgreb:
@blair1618 I would expect the final position to be (x1, y1) in both scenarios. Do I win a prize?

blair1618:
@ecgreb Cool, just wanted to check my reasoning about expectations. 

You may collect your high five next time you're in the office :)

blair1618:
Hey @hjanetzek can you look over this and merge if it seems good to you?

hjanetzek:
Looks good!
  
I wonder though why quint is the default? - It feels to start a bit too delayed for the move-on-tap. 

A few additions that would be nice to have, if you want to continue with this :)  
- more ease types
- A function to cancel all ongoing eases. 
- Have the pinch gesture use eases (on double tap) 
- chaining of eases for fly-to animations for zoom-out and back in while moving. (Or generally for longer running animations it should be split into a start-, going- and end-phase) 

I experienced with vtm that one often wants to coordinate between position changes from interactions and from animations. So maybe InputHandler and easings could be combined into a CameraController at some point.

blair1618:
Cool, thanks!

I just picked quint because I like the feel of the continuous 2nd derivative, but we should reconsider once we have some feedback from the app. I'm open to more ease types too!

Users can stop an ease right now just by calling the setter again without easing; I had imagined that this would cover most needs but an explicit cancel could be convenient too. 

I had forgotten about double-tap zooming! That would be a good candidate for easing, but I'm also not sure we want to keep that as built-in library behavior :\

I had actually been working on a fly-to animation, but I stashed it for later because it got a bit complicated and I wanted to get the other easing functions merged first.

PR #382, opened 2015-11-04
Fix global initializers in shaders

blair1618:
In the GLSL ES 1.0 specification, section 4.3:
>Declarations of globals without a storage qualifier, or with just the const qualifier, may include initializers, in which case they will be initialized before the first line of main() is executed. Such initializers must be a constant expression.

And according to section 5.1:  
>The following may not be used in constant expressions:
>• User-defined functions
>• Uniforms, attributes and varyings

(https://www.khronos.org/registry/gles/specs/2.0/GLSL_ES_Specification_1.0.17.pdf)

Turns out the shaders in Tangram have been violating this part of the specification for a long time: global variables for materials and lights are initialized with uniforms (e.g. [this line](https://github.com/tangrams/tangram-es/blob/master/core/resources/shaders/material.glsl#L57))

It seems that the vast majority of GL ES implementations don't enforce this, but some Adreno drivers will not only fail to compile shaders that violate this, but will actually crash the entire program with an internal error. 

Fortunately, it isn't too difficult to circumvent this issue by refactoring our shaders just a bit.

### DIFF
--- a/core/resources/shaders/lights.glsl
+++ b/core/resources/shaders/lights.glsl
@@ -2,6 +2,7 @@
 vec4 calculateLighting(in vec3 _eyeToPoint, in vec3 _normal, in vec4 _color) {
 
     // Do initial material calculations over normal, emission, ambient, diffuse and specular values
+    material = u_material;
     calculateMaterial(_eyeToPoint,_normal);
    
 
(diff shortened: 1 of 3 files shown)


blair1618:
cc @bcamper @patriciogonzalezvivo This issue affects tangram-js as well, it seems to be reliably reproducible on Adreno 420/430 GPUs with Android 5.

patriciogonzalezvivo:
:+1:  thanks!

bcamper:
Cool, thanks. @patriciogonzalezvivo can you make the change in JS?

patriciogonzalezvivo:
sure!

PR #383, opened 2015-11-05
Dev proxy cache

hjanetzek:
Just a super simple tile cache for development for offline work and (at least for me) much faster load time for the initial view  

- [ ] Add to Readme

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,9 @@ Icon
 # Thumbnails
 ._*
 
+# proxy.py cache
+.tiles
+
 # Files that might appear on external disk
 .Spotlight-V100
 .Trashes
--- a/osx/resources/tangram-Info.plist
+++ b/osx/resources/tangram-Info.plist
@@ -22,6 +22,11 @@
 	<string>????</string>
 	<key>CFBundleVersion</key>
 	<string>1.0</string>
+	<key>NSAppTransportSecurity</key>
+	<dict>
+		<key>NSAllowsArbitraryLoads</key>
+		<true/>
+	</dict>
 	<key>NSHighResolutionCapable</key>
 	<true/>
 </dict>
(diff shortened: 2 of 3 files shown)


blair1618:
Nice! This will come in handy when I'm on bus trips too.

Can't quite get it to work on OS X though. I started the python server and then ran tangram with my source url set to `http://localhost:8000/vector.mapzen.com/osm/all/{z}/{x}/{y}.mvt` and tiles were loading, but nothing was written in the `.tiles` folder.

I would test on Ubuntu but the latest OS X update removed my GRUB config :\

tallytalwar:
Good to know about GRUB config. Will update cautiously or do some hacking
around before I upgrade to El Capitan..

hjanetzek:
Not sure what may be go wrong there. My python is a bit rusty though. Could you try again with the updates?

blair1618:
I think the problem here is that the `NSURL` library we use on OS X now (as of 10.11) blocks cleartext HTTP loading unless you make a security exception for it - and the proxy server doesn't seem to work with HTTPS. I'll see if there's an easy fix, if not I'll just merge this as-is.

blair1618:
I think the only way to make this work on OS X is to add
```
<key>NSAppTransportSecurity</key>
<dict>
	<key>NSAllowsArbitraryLoads</key>
	<true/>
</dict>
```
To our `Info.plist`. As documented [here](https://developer.apple.com/library/mac/documentation/General/Reference/InfoPlistKeyReference/Articles/CocoaKeys.html#//apple_ref/doc/uid/TP40009251-SW33), this key is really only intended for debugging purposes. There isn't much harm in enabling this on our desktop demo app since it's not a real end-product, though it may end up masking networking issues that we'll encounter when running on iOS (without this flag).

PR #384, opened 2015-11-05
All in one build - follow ups

hjanetzek:
- Fix segv when loading geojson data. (Tags constructor did not initialize Properties shared_ptr)

- On linux make sure to build with CXX11_ABI for newer gcc to enable SSO strings,
  use libc++ when using clang (on linux it defaults to libstdc++, so this should make it better more comparable with the android build)


hjanetzek:
merging soon - if no one objects

tallytalwar:
go go go.

karimnaaji:
I noticed some small things that might be good to address in this PR, `make tests` is not available anymore, also headers not showing up in xcode project.

tallytalwar:
Yup headers have not been showing up in xcode for a while now. (I rarely use xcode these days so forgot about it).

hjanetzek:
@karimnaaji could you try the source_path approach? It's hard to test without xcode :)

Maybe this macro could be adapted and added to utils.cmake
http://cmake.3232098.n2.nabble.com/cmake-xcode-c-header-files-td7020444.html

karimnaaji:
Yep I can look into that!

core/src/scene/styleParam.cpp
```diff
@@ -159,7 +160,7 @@ StyleParam::Value StyleParam::parseString(StyleParamKey key, const std::string&
     case StyleParamKey::style:
         return _value;
     case StyleParamKey::font_size: {
-        float fontSize;
+        float fontSize = 10.f;
```
karimnaaji: I remember that discussing with @tallytalwar to remove those here to keep the default style value parameters in the struct they belong to (just to look up in the same place to edit the defaults). Is there any other reason to add them here?

hjanetzek: This should just makes sure that fontSize is not used with an arbitrary (undefined) value

karimnaaji: Yeah anyway a warning if that value is used a warning would be generated so 0.f would be good I guess.

karimnaaji:
Good to go!

PR #387, opened 2015-11-06
Android fonts aliases.

tallytalwar:
Will most likely avoid bunding a font file.

Fixes aliased font family for android.
~~Note for eraser map style sheet: Use all lower case font family name, i.e., `helvetica` and not `Helvetica`.~~

screenshot_2015-11-06-10-57-04


bcamper:
This relates to a conversation we were having with @karimnaaji. We should not mandate people use one capitalization over the other for fonts, the vast majority of web/CSS examples are capitalized, and a huge portion of people will choose the wrong one and it will be confusing and hard to debug. We can perhaps check for both capitalized and un-capitalized patterns.

tallytalwar:
@bcamper Sure, was not aware of it. Can include it with this PR. Thanks

android/tangram/src/com/mapzen/tangram/FontFileParser.java
```diff
@@ -47,6 +52,15 @@ private void processDocument(XmlPullParser parser) throws XmlPullParserException
 
                         String key = name + "_" + weightStr + "_" + styleStr;
                         fontDict.put(key, fullFilename);
+                        String name2 = Character.toUpperCase(name.charAt(0)) + name.substring(1);
```
karimnaaji: I think the uppercase step should happen at style parsing, so that we always ask the operating system with a valid font file name. This would also be more straightforward while implementing this on iOS.

android/tangram/src/com/mapzen/tangram/FontFileParser.java
```diff
@@ -20,6 +21,8 @@
 
     private void processDocument(XmlPullParser parser) throws XmlPullParserException, IOException {
 
+        Vector<String> weights = new Vector<String>();
```
blair1618: This list should be cleared whenever a new "family" is encountered. 

I'd also recommend `ArrayList` here instead of `Vector`; the difference between the two is that `Vector` synchronizes for thread safety on every operation, making it generally slower than `ArrayList`

android/tangram/src/com/mapzen/tangram/FontFileParser.java
blair1618: I agree, let's just normalize to font names to lowercase strings.

android/tangram/src/com/mapzen/tangram/FontFileParser.java
```diff
@@ -54,11 +68,33 @@ private void processDocument(XmlPullParser parser) throws XmlPullParserException
...
+                Vector<String> aliasWeights = new Vector<String>();
+                String fontFilename;
+
+                if (weightStr == null) {
+                    aliasWeights = weights;
+                } else {
+                    aliasWeights.add(weightStr);
+                }
+
+                for (int i = 0; i < aliasWeights.size(); ++i) {
```
blair1618: Range-based iterators are generally preferable to indexing:
```
for (String weight : aliasWeights)
```

android/tangram/src/com/mapzen/tangram/FontFileParser.java
```diff
@@ -54,11 +68,33 @@ private void processDocument(XmlPullParser parser) throws XmlPullParserException
             } else if ("alias".equals(parser.getName())) {
                 // Parse this alias to font to fileName
                 String aliasName = parser.getAttributeValue(null, "name");
+                String aliasName2 = Character.toUpperCase(aliasName.charAt(0)) + aliasName.substring(1);
                 String toName = parser.getAttributeValue(null, "to");
                 String weightStr = parser.getAttributeValue(null, "weight");
-                weightStr = (weightStr == null) ? "400" : weightStr;
-                String fontFilename = fontDict.get(toName + "_" + weightStr + "_normal"); // alias style is default: normal (always)
-                fontDict.put(aliasName + "_" + weightStr, fontFilename);
+                Vector<String> aliasWeights = new Vector<String>();
```
blair1618: Recommend `ArrayList` here as well

android/tangram/src/com/mapzen/tangram/FontFileParser.java
```diff
@@ -54,11 +68,33 @@ private void processDocument(XmlPullParser parser) throws XmlPullParserException
...
+                    fontDict.put(aliasName + "_" + aliasWeights.get(i) + "_normal", fontFilename);
+                    fontDict.put(aliasName2 + "_" + aliasWeights.get(i) + "_normal", fontFilename);
+                    fontDict.put(aliasName + "_" + aliasWeights.get(i) + "_Normal", fontFilename);
+                    fontDict.put(aliasName2 + "_" + aliasWeights.get(i) + "_Normal", fontFilename);
+                    // Italic style
+                    fontFilename = fontDict.get(toName + "_" + aliasWeights.get(i) + "_italic");
+                    fontDict.put(aliasName + "_" + aliasWeights.get(i) + "_italic", fontFilename);
+                    fontDict.put(aliasName2 + "_" + aliasWeights.get(i) + "_italic", fontFilename);
+                    fontDict.put(aliasName + "_" + aliasWeights.get(i) + "_Italic", fontFilename);
+                    fontDict.put(aliasName2 + "_" + aliasWeights.get(i) + "_Italic", fontFilename);
```
blair1618: These "normal" and "Normal" should also be normalized to lowercase "normal", etc.

sensescape:
Let me know if we are not bundling font. Font sizes will have to be adjusted for Roboto. Does that mean 2 stylesheets?

blair1618:
We plan to bundle at least one font with the app, just not in this pull
request :)

tallytalwar:
Thanks, all addressed.

sensescape:
Ok! Thanks!

android/tangram/src/com/mapzen/tangram/MapController.java
```diff
@@ -50,7 +50,7 @@
      */
     public MapController(Activity mainApp, MapView view) {
 
-        this(mainApp, view, "scene.yaml");
+        this(mainApp, view, "eraser-map.yaml");
```
karimnaaji: Don't forget to revert.

tallytalwar: :+1:

tallytalwar:
Good to go?

tallytalwar:
Is this good to be merged?

PR #391, opened 2015-11-10
Fix material textures on some devices

blair1618:
Accessing the texcoord varying in a function in the vertex shader seems to cause the varying to not be interpolated, even in the fragment shader later

### DIFF
--- a/core/resources/shaders/lights.glsl
+++ b/core/resources/shaders/lights.glsl
@@ -1,4 +1,4 @@
-
+#if (defined(TANGRAM_VERTEX_SHADER) && defined(TANGRAM_LIGHTING_VERTEX)) || (defined(TANGRAM_FRAGMENT_SHADER) && defined(TANGRAM_LIGHTING_FRAGMENT))
 vec4 calculateLighting(in vec3 _eyeToPoint, in vec3 _normal, in vec4 _color) {
 
     // Do initial material calculations over normal, emission, ambient, diffuse and specular values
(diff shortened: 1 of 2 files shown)


blair1618:
The symptoms of this are documented here: https://github.com/tangrams/eraser-map/issues/65

blair1618:
@hjanetzek Any idea why varyings stop varying when they're referenced from a global function?

hjanetzek:
@blair1618 something is broken in the driver probably :) I would guess that the function of the vertex shader leaks into the fragment shader's namespace (and the second definition gets discarded), which should not be possible. 
For testing that assumption: Have you tried to use a function that is only defined in vertex shader within the fragment shader?

blair1618:
Yeah I suspected something like that, but it must be more subtle - this test style produces the correct output (blue) on an affected device: 
```
    test-style:
        base: polygons
        shaders:
            blocks:
                global: |
                    #ifdef TANGRAM_VERTEX_SHADER
                    vec4 myfun() {
                        return vec4(1., 0., 0., 1.);
                    }
                    #endif
                    #ifdef TANGRAM_FRAGMENT_SHADER
                    vec4 myfun() {
                        return vec4(0., 0., 1., 1.);
                    }
                    #endif
                filter: |
                    color = myfun();
```
But this style produces a flat gray on the device, while producing the expected gradient on desktop:
```
    test-style:
        base: polygons
        shaders:
            blocks:
                global: |
                    varying vec4 v_coord;
                    vec4 myfun() {
                        return v_coord;
                    }
                filter: |
                    color = myfun();
                position: |
                    v_coord = modelPosition();
```
Whatever the problem is, varyings seem to be the key ingredient.

tallytalwar:
I am going to merge this, and we can keep on looking on whether its a driver bug or a glsl spec thingy.

PR #392, opened 2015-11-10
Uniform caching

karimnaaji:
Something interesting to have is shader uniform caching, which cleans up some part of the code. Before dispatching the uniform command to the GPU, we simply check if the uniform value has already been sent, on context loss, we simply clear the cache. 
On the current scene, this reduces the GL trace by ~40 GL calls per frame once tiles have been fetched, on the CPU side I didn't notice any increasing time on any of the `setUniform*` methods, and `getFromCache` takes a negligible amount of app time.

Current trace:
screen shot 2015-11-10 at 6 38 44 pm

With uniform caching:
screen shot 2015-11-10 at 6 37 33 pm

### DIFF
--- a/core/src/util/variant.h
+++ b/core/src/util/variant.h
@@ -8,7 +8,6 @@
 #define NDEBUG
 #endif
 
-#include "glm/glm.hpp"
 #include "variant/variant.hpp"
 
 #ifndef KEEP_NDEBUG
@@ -40,7 +39,5 @@ class Value : public detail::Value {
     using Base::Base;
 };
 
-/* Style Block Uniform types */
-using UniformValue = variant<none_type, bool, std::string, float, int, glm::vec2, glm::vec3, glm::vec4, glm::mat2, glm::mat3, glm::mat4>;
 
 }
--- a/core/src/util/variant.h
+++ b/core/src/util/variant.h
@@ -8,6 +8,7 @@
 #define NDEBUG
 #endif
 
+#include "glm/glm.hpp"
 #include "variant/variant.hpp"
 
 #ifndef KEEP_NDEBUG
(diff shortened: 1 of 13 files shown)


hjanetzek:
Reviewing now

What do you think of adding a useCache = true default parameter to the uniform setters to be able to not have to check uniform values for which we know that they change with each call? This is pretty much the case for all Tile uniforms I guess.

In my opinion the 'transpose' param for setUniformMatrix* could be removed until we use it (to avoid two default params) . I haven't ran into any usecase for it yet.

karimnaaji:
Yeah I'm ok for that! It would be really useful for uniforms like time. For Tile uniforms, I don't think we should default to useCache = true for all of them though, since some matrices are not modified if you don't tilt (like the normalMatrix for example, and a lot of other ones never change too, plus the style incoming uniforms that should be cached).
Also to prevent the request render animation which was triggering a lot of uniforms updates for nothing (when labels are fading out and you don't touch the screen at all for example).

Yep transpose could be removed too!

blair1618:
Make sure to clear the uniform cache when the shader is invalidated

core/src/util/variant.h
```diff
@@ -39,4 +39,5 @@ class Value : public detail::Value {
     using Base::Base;
 };
 
+
```
blair1618: If you're amending other changes, this newline can go as well :)

karimnaaji:
@hjanetzek actually I'm fine with letting this as is and not using an option to use the cache or not, it's only really relevant for the uniform time so far, otherwise most of the uniforms can't really be predicted to change, especially in the cases I mentioned or while using styles with `animated = true`.

hjanetzek:
yes, ok. it's now just u_model and u_tile_origin that change with every drawn tile -- earlier we had more in this place :)

blair1618:
Is someone working on restoring the `setUniformi` methods? If not I can do that.

karimnaaji:
It should be good from https://github.com/tangrams/tangram-es/commit/4a69cee287cdb569fc62e86b6fcbc5ad0a8cc8b7

PR #393, opened 2015-11-11
Remove now unused modelViewProj uniform

karimnaaji:
`u_modelViewProj` was only used in the debug shader, so we can easily remove it now.


blair1618:
Yep, looks good

PR #396, opened 2015-11-13
Reduce flickering

karimnaaji:
Reduce flickering of labels already visible on the screen. This smoothen transition between zoom levels and gives better visual experience while zooming in or out.

This relies on the labels of the tile cache (the _parent_ cached tile for a _zoom in_, 4 _children_ cached tiles for a _zoom out_) to determine whether a label is already visible. Since labels are first compared with a hash based on the style parameters of the label, and this hash has no concept of location, a new label would be considered to be the same based on the hash first, and whether it lies or not within a screen space circle defined by the bounding box of the already visible label. The label would then be flagged to skip transitions whenever this test passes.

This adds complexity, but can be improved; I'm planning to improve labels iteration and sparse their memory locations to be contiguous for fast iterations.

### DIFF
--- a/core/src/labels/labels.cpp
+++ b/core/src/labels/labels.cpp
@@ -8,6 +8,7 @@
 #include "style/material.h"
 #include "style/style.h"
 #include "tile/tile.h"
+#include "tile/tileCache.h"
 
 #include "glm/glm.hpp"
 #include "glm/gtc/matrix_transform.hpp"
@@ -26,7 +27,7 @@ int Labels::LODDiscardFunc(float _maxZoom, float _zoom) {
 
 
 void Labels::update(const View& _view, float _dt, const std::vector<std::unique_ptr<Style>>& _styles,
-                    const std::vector<std::shared_ptr<Tile>>& _tiles) {
+                    const std::vector<std::shared_ptr<Tile>>& _tiles, std::unique_ptr<TileCache>& _cache) {
 
     m_needUpdate = false;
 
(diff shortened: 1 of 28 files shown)


core/src/labels/labels.cpp
```diff
@@ -104,6 +105,54 @@ void Labels::update(const View& _view, float _dt, const std::vector<std::unique_
         }
     }
 
+    /// Mark labels to skip transitions
+
+    for (const auto& t0 : _tiles) {
```
hjanetzek: Shouldn't the transition state be only checked when _view.getNextZoom() != currentZoom? 

Hmm, still this only works for a non-tilted view: I think we could also handle the case where zoom-level changes of tiles are not equivalent to zoom-level changes of the view. I.e when proxy tiles were temporarily rendered and for tilted views. 
To do this we need to add a way to determine the first time when a tile is rendered and check the label state of the previously active tile(s) for its area.

karimnaaji: Yeah right! We should find a way to only trigger this loop when a tile transition happens.

tallytalwar:
Reviewing this now.

karimnaaji:
Did you have a chance to review this @tallytalwar ?

tallytalwar:
I did partially before I left. Will finish reviewing  this along with #400 today.

karimnaaji:
Cool thanks! I think @hjanetzek also went over it.

tallytalwar:
Sure, if @hjanetzek is cool with this, then go ahead and merge it :D.

PR #397, opened 2015-11-13
Reset render state for vbos

karimnaaji:
Found this bug while debugging with @tallytalwar .


tallytalwar:
very nice catch :D.

blair1618:
Cool, looks good

PR #400, opened 2015-11-16
Client data update

tallytalwar:
* Updates the API to handle client DataSource update
 * `DataSource::update()` call will clear the old data and give a hint to the core library to retain the previous tiles (stale), till new data (to be added with `DataSource::addData()`) tiles have been processed.

* Tile Management:
 *  Creates 2 new states for a tile: `stale` and `updating`.
  * `stale` tiles
    * Legal tile state change: from `ready` state to `stale` state.
    * contains geometry constructed from old data
    * results in creation of a tile building task with a new tile (consisting of new data).
  * `updating` tiles
    * once a task to build a new tile corresponding to the `stale` tile is created, this `stale` tile's state is set to `updating`.
    * `updating` tiles will part of the `m_tiles` (tiles to be drawn).
    * `updating` tile gets updated by its corresponding new tile when the tile building task finishes, marking this new tile as `ready`.

### DIFF
--- a/core/src/data/dataSource.h
+++ b/core/src/data/dataSource.h
@@ -59,8 +59,11 @@ public:
      */
     void setCacheSize(size_t _cacheSize);
 
+    /* ID of this DataSource instance */
     int32_t id() const { return m_id; }
 
+    /* Generation ID of DataSource state (incremented for each update, e.g. on clearData()) */
+    int64_t generation() const { return m_generation; }
 
 protected:
 
@@ -80,6 +83,10 @@ protected:
 
     // Unique id for DataSource
     int32_t m_id;
+
+    // Generation of dynamic DataSource state (incremented for each update)
+    int64_t m_generation = 1;
+
     // URL template for requesting tiles from a network or filesystem
     std::string m_urlTemplate;
 
--- a/core/src/tile/tileManager.cpp
+++ b/core/src/tile/tileManager.cpp
@@ -109,7 +109,8 @@ bool TileManager::setTileState(Tile& _tile, TileState _newState) {
             m_loadPending++;
             return true;
         }
-        if (_newState == TileState::processing) {
+        if (_newState == TileState::processing ||
+            _newState == TileState::stale) {
             _tile.setState(_newState);
             return true;
         }
@@ -120,7 +121,8 @@ bool TileManager::setTileState(Tile& _tile, TileState _newState) {
 
     case TileState::loading:
         if (_newState == TileState::processing ||
-            _newState == TileState::canceled) {
+            _newState == TileState::canceled ||
+            _newState == TileState::stale) {
             _tile.setState(_newState);
             m_loadPending--;
             return true;
@@ -128,7 +130,8 @@ bool TileManager::setTileState(Tile& _tile, TileState _newState) {
         break;
 
     case TileState::processing:
-        if (_newState == TileState::ready) {
+        if (_newState == TileState::ready ||
+                _newState == TileState::stale) {
             _tile.setState(_newState);
             return true;
         }
@@ -137,6 +140,16 @@ bool TileManager::setTileState(Tile& _tile, TileState _newState) {
     case TileState::canceled:
         // Ignore any state change when tile was canceled
         return false;
+
+    case TileState::stale:
+        _tile.setState(_newState);
+        return true;
+
+    case TileState::ready:
+        if (_newState == TileState::stale) {
+            _tile.setState(_newState);
+            return true;
+        }
 
     default:
         break;
@@ -178,6 +191,15 @@ void TileManager::clearTileSet(int32_t _id) {
     m_tileSetChanged = true;
 }
 
+void TileManager::markStale(int32_t _id) {
+    for (auto& tileSet : m_tileSets) {
+        if (tileSet.source->id() != _id) { continue; }
+        for (auto& tile : tileSet.tiles) {
+            setTileState(*tile.second, TileState::stale);
+        }
+    }
+}
+
 void TileManager::updateTileSets() {
 
     m_tileSetChanged = false;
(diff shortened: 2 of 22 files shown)


tallytalwar:
@blair1618 cam you review this please?  Thanks

hjanetzek:
I wonder if it wouldn't be easier to have a 'stale' tile flag instead of adding complexity to tile state changes with two additional states.

tallytalwar:
@hjanetzek yes, thats a nice catch.

I am not sure if the serial approach will fix it. Need to try it.

tallytalwar:
Thinking more..

In addition to what you described in your comment above:
We will need to update tile always from the tileTask tile only when the serial value from the tileTask is greater.

hjanetzek:
Stress testing seems to be handled correctly now - still needs some cleanup.

There was also an issue with the clearData approach: It triggers recreation of tiles when the features are cleared - not when all new features were added. 
This should be fixed by the 'atomic updates' commit, but this part could be moved to a separate task. In my opinion it would be good to have a way to trigger the update of m_store once when all features are set.

core/src/data/dataSource.h
```diff
@@ -80,6 +83,10 @@ class DataSource {
 
     // Unique id for DataSource
     int32_t m_id;
+
+    // Serial of dynamic DataSource state (incremented for each update)
+    int64_t m_serial = 1;
```
blair1618: "generation" might be a clearer name for this concept than "serial"

hjanetzek:
yes,  'generation' is good as well and also used in other places. To me the term implies some more complex inheritance of previous state than just incrementing a number though :)

hjanetzek:
@tallytalwar, @blair1618 could you have a look? client data updates should work allright now.

blair1618:
Cool, will do!

tallytalwar:
Sure. Will get to this most likely when I am back.

hjanetzek:
I should add that the tile.reloading flag should there be only temporarily - as it surely isn't that elegant. I'm almost ready with a larger rewrite of the tile state logic which avoids explicit state variables

tallytalwar:
Reviewing this now.

hjanetzek:
ok, except of removing the stress-testing code the cleanup is completed :)

tallytalwar:
I agree on the `reloading` flag. But its better than my original implementation of having 2 states: `stale` and `updating`. Will also check your revamp of the tile state logic. Thanks

core/src/tile/tileManager.cpp
```diff
@@ -339,38 +371,54 @@ void TileManager::enqueueTask(const TileSet& tileSet, const TileID& tileID,
...
+
+        auto id = *std::get<2>(item);
         auto& tileSet = *std::get<1>(item);
-        auto it = tileSet.tiles.find(id);
-        if (it == tileSet.tiles.end()) { continue; }
 
-        auto& tile = it->second;
-        auto& source = tileSet.source;
+        auto tileIt = tileSet.tiles.find(id);
+        if (tileIt == tileSet.tiles.end()) {
```
tallytalwar: I am not sure if this is required here. According to me there is no case when a `tile` in a `loadTask` will be removed from the `tileSet` (based on how `updateTileSet` works).

core/src/tile/tileWorker.cpp
```diff
@@ -67,6 +67,9 @@ void TileWorker::run() {
                     if (a->tile->isVisible() != b->tile->isVisible()) {
                         return a->tile->isVisible();
                     }
+                    if (a->tile->sourceGeneration() != b->tile->sourceGeneration()) {
+                        return a->tile->sourceGeneration() < b->tile->sourceGeneration();
```
tallytalwar: This should be when both `a` and `b` tiles belong to the same `source`, else `priority` should be checked.

Can you also update the comment above (line 64) to reflect this change in logic? Thanks

core/src/tile/tileWorker.cpp
```diff
@@ -94,10 +97,10 @@ void TileWorker::run() {
 
             // float loadTime = (float(clock() - begin) / CLOCKS_PER_SEC) * 1000;
             // LOG("loadTime %s - %f", task->tile->getID().toString().c_str(), loadTime);
-
-            m_tileManager.tileProcessed(std::move(task));
```
tallytalwar: Moving this outside the `if (tileData)` scope, may result in a task tile set to `ready` state, even though its tileData was not processed!

tallytalwar: And `requestRender` should be called when tile is finished building, and hence moved inside the scope of the above if condition.

hjanetzek: Hmm yes, the problem here is that the task must be passed back to TileManager to unset the 'reloading' flag of the original tile (by replacing it in tileset) otherwise future updates of the DataSource would not reload the tile.

core/src/tile/tileWorker.cpp
hjanetzek: Good point!

core/src/tile/tileWorker.cpp
tallytalwar: So in a case when a tile's `tileData` is not processed and is marked `ready` by `TileManager::tileProcessed`, line: 227 in `tileManager.cpp` will clear the corresponding proxy tiles, even though the actual ready tile is not processed!! :(

hjanetzek: Yes, but this is good! :) What if the source does not have any data anymore - I mean the old tile that had data should be replaced by the empty tile.

tallytalwar: Sure, but what is dataSource->parse fails and returns a nullptr?

hjanetzek: Then TileWorker passes back an empty tile to the TileManager. That could be reloaded once the DataSource changes again. Or am I missing something?

tallytalwar: Hmm, in that case there will be holes from clearing of proxy tiles?

hjanetzek: Not sure what you mean by that. The tile state logic is certainly not perfect. That's why I started the rather large rework to bring it into a shape that shall be easier to reason about.

PR #402, opened 2015-11-23
[WIP] Inlay blend mode

karimnaaji:
The technique to render inlay blend mode is here different than in tangram-js (https://github.com/tangrams/tangram/pull/161), since in tangram-es labels are rendered with no depth, in an orthographic camera (a different virtual camera than the one used to render usual world placed polygons), and have no information about the world position or actual depth at which they are attached to (a projection to the viewport, client-side, happens at an early stage before drawing in order to manage collisions in screen space). 
An alternative way to render the *inlay* blend mode uses the stencil buffer, which makes it a potentially more intensive blending mode to use.

The stencil render pass sets 1s in the stencil buffer whenever an occluding geometry for this blend mode appears (geometry _off_ the ground, should consider the *order* property), and 0s whenever there is no occluding geometry, which would output this in the stencil buffer (debug colors):
screen shot 2015-11-23 at 2 46 46 pm
The result of drawing *inlay* labels with this stencil technique is then the following:
screen shot 2015-11-23 at 5 01 07 pm
Related design issue: https://github.com/tangrams/eraser-map/issues/74
Todo:
- [ ] Add stencil polyline style
- [ ] Benchmark on devices

Known issue, unfortunately (for now), this blend mode is incompatible with other transparent styles:

screen shot 2015-11-23 at 5 07 33 pm

### DIFF
--- a/core/resources/scene.yaml
+++ b/core/resources/scene.yaml
@@ -7,14 +7,14 @@ cameras:
         position: [-74.00976419448854, 40.70532700869127, 16]
         type: isometric
         axis: [0, 1]
-        active: false
-    perspective-camera:
-        # Manhattan
-        position: [-74.00976419448854, 40.70532700869127, 16]
-        type: perspective #currently ignored
-        focal_length: [[16, 2], [20, 6]] # currently ignored
-        vanishing_point: [-250, -250] # currently ignored
         active: true
+    #perspective-camera:
+    #    # Manhattan
+    #    position: [-74.00976419448854, 40.70532700869127, 16]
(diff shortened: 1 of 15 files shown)


karimnaaji:
Closing for the time being, the technique used here solves the issue but is not a feature priority even if diverging from -js, feedback would be appreciated though.
Improvements to be considered: 
- as the technique was more of a proof of concept, inlay geometry was a direct copy from the initial geometry, it can be either reduced or point to the same buffer.
- inlay shader should get injection points, otherwise shader injection from stylesheet won't have effect on the inlay blend mode

PR #403, opened 2015-11-23
Apply multipolygon ring decoding to MVT sources

tallytalwar:
Refer tangram web PR: https://github.com/tangrams/tangram/pull/217
Refer mapzen/mapbox-vector-tile PR: https://github.com/mapzen/mapbox-vector-tile/pull/21

### DIFF
--- a/core/src/util/pbfParser.cpp
+++ b/core/src/util/pbfParser.cpp
@@ -2,6 +2,7 @@
 
 #include "tile/tile.h"
 #include "platform.h"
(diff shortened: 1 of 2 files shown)


blair1618:
Reviewing. Is there a test case for this? A tile that demonstrates the issue?

tallytalwar:
Nopes, nothing in the current set of tiles we have. But @rmarianski, got a test tileset setup, which I used for my testing.

Issue #404, opened 2015-11-24
Crash when updating client data source


blair1618:
The crash results from a particular timing of multi-threaded activity and is therefore difficult to reproduce. 

The suspected cause is that the "store" of data for a geojsonvt data source is cleared while a worker thread is parsing data from that source, leading the worker to access freed memory.

hjanetzek:
Yes, it looks like the assignment to m_store needs to be synchronized with m_store->getTile

tallytalwar:
Yup. That's what I found yesterday evening.

PR #406, opened 2015-11-24
Locks to synchronize write for m_store

tallytalwar:
Fixes: https://github.com/tangrams/tangram-es/issues/404

### DIFF
--- a/core/src/data/clientGeoJsonSource.cpp
+++ b/core/src/data/clientGeoJsonSource.cpp
@@ -40,6 +40,8 @@ ClientGeoJsonSource::~ClientGeoJsonSource() {}
 void ClientGeoJsonSource::addData(const std::string& _data) {
 
     m_features = geojsonvt::GeoJSONVT::convertFeatures(_data);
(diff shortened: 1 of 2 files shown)


tallytalwar:
This should fix #404. @blair1618 can you try eraser-map with this. I was not able to make it crash in my testing. Thanks

tallytalwar:
Hold on this. Found an async read!!

blair1618:
Sure. Ping when you want me to test.

tallytalwar:
This should be good now. 

However, had to make `DataSource::parse` non-const, so as to allow `ClientDataSource::parse` to lock `m_storeMutex`. Could not think of a better solution, to avoid this, at the moment.

tallytalwar:
WOW! Herb Sutter for help: https://channel9.msdn.com/posts/C-and-Beyond-2012-Herb-Sutter-You-dont-know-blank-and-blank

WIll rebase with a better way of handling synchronous read on `m_store`.

Issue #390, opened 2015-11-09
GL context loss make labels disappear


karimnaaji:
Related to issue https://github.com/mapzen/eraser-map/issues/116
Labels disappear after a GL context loss, all GL resources are re-uploaded, but labels/sprites need to be updated at least once in order to go into their animation/request-render loop. 
To fix this we should require Tangram to perform at least one `update` to get labels back to their visible states.
Easy to reproduce after destroying the context on OS X.

PR #407, opened 2015-11-30
Clear vertex layout caching after context loss

karimnaaji:
The vertex layout cache of vbos not using vertex array objects needs to be cleared on context loss.

For the vbos not using vaos (labelMesh), the current flow makes the mesh not visible after a context loss (the enabled attributes are not being enabled again until a switch between two different shader program happens, since the caching of vertex layout is storing the enabled attributes for the last shader used), this should fix some of the issues being seen in https://github.com/tangrams/tangram-es/issues/390.

### DIFF
--- a/core/src/gl/vertexLayout.cpp
+++ b/core/src/gl/vertexLayout.cpp
@@ -76,6 +76,10 @@ void VertexLayout::enable(const fastmap<std::string, GLuint>& _locations, size_t
 
 }
 
+void VertexLayout::clearCache() {
+    s_enabledAttribs.clear();
+}
+
 void VertexLayout::enable(ShaderProgram& _program, size_t _byteOffset, void* _ptr) {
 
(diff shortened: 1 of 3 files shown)


tallytalwar:
Definitely a step closer to fix #390.

core/src/gl/vertexLayout.cpp
```diff
@@ -76,6 +76,10 @@ void VertexLayout::enable(const fastmap<std::string, GLuint>& _locations, size_t
 
 }
 
+void VertexLayout::clearCache() {
+    s_enabledAttribs.clear();
```
blair1618: For modifying a static member, it makes more sense to me to use a static function. Then we can just call it once on invalidation, perhaps inside `VboMesh::invalidateAllVbos()`

karimnaaji: Yep makes sense.

PR #408, opened 2015-11-30
Delete index buffer

karimnaaji:
Currently, this index buffer is not deleted either on app destroy or context loss. 
We should probably have a `Tangram::teardown` to deallocate statically instantiated resources like this shared index buffer.


blair1618:
Seems to work. This is probably a reasonable way to handle the shared buffer; the only other approach I could think of is to create a static object representing the shared buffer whose destructor calls gl delete, but I don't think that's strictly better. 

If you're done with this I'll merge it!

karimnaaji:
Yep, will do a quick rebase on this.

PR #413, opened 2015-12-01
non-proxy tiles labels win over proxy tile label

tallytalwar:
fixes an issue when proxy tiles labels were causing non proxy tiles label to be not drawn on collision


karimnaaji:
This is good and fixes some inconsistent occlusions.

PR #415, opened 2015-12-03
Be more diligent in setting animate for label

tallytalwar:
every time a label enters a fading_in or fading_out or a visible state, animate should be set to true. This results in a
renderRequest, and hence also fixes the context loss issue.


karimnaaji:
Fix of #390

PR #417, opened 2015-12-03
Add EGL configuration chooser

blair1618:
Will prevent crashes at launch on older Android devices, including 2012 Nexus 7 (thanks @msmollin!)

Implementation affectionately stolen from libGDX <3


blair1618:
@hjanetzek you've probably written one of these before, does this make sense to you? It doesn't cover all possible EGL options but it takes care of tangram's needs.

hjanetzek:
This looks good. If I recall correct we used a custom config chooser to support 16bit color - but it will be hard to find one of these devices today :)

PR #419, opened 2015-12-04
Ease rotation over minimum angular distance

blair1618:
Since the view rotation is stored in radians wrapped to [0, 2 pi), the current implementation of rotation easing causes rotations over larger angular distances than necessary. For example, if the current view rotation is 6 radians and an ease is requested to 0 radians, the current implementation will rotate negative 6 radians to reach zero, when it is more desirable to rotate positive ~0.28 radians. This change ensures that rotation easing uses the smallest possible angular distance to reach the final rotation.


blair1618:
@hjanetzek take a quick look and merge, if this seems alright

hjanetzek:
Looks good!

Issue #412, opened 2015-12-01
Road labels should be centered on geom (like WebGL)


nvkelso:
Tangram ES should match the WebGL implementation and center road labels on the line geometry.

![img_7730-2](https://cloud.githubusercontent.com/assets/853051/11494801/c484ad8e-97ba-11e5-9c1b-1f2eb7121494.PNG)
![screenshot_2015-11-29-19-04-17](https://cloud.githubusercontent.com/assets/853051/11494802/c4a1cc48-97ba-11e5-9101-7ef927b4f7a5.png)

karimnaaji:
After a quick look it seems that it's a bounding box issue (being bigger than they should be) and not a centering issue.
![screen shot 2015-12-01 at 11 40 50 am](https://cloud.githubusercontent.com/assets/7061573/11507096/70a8e6ec-9820-11e5-8a39-f6513230b028.png)
Fixing the bouding boxes should fix that.

PR #420, opened 2015-12-04
Adjust bounding boxes

karimnaaji:
Adjust bounding boxes, and make a better adjustment on line labels on their geometry (issue-#412).
Before/After
screen shot 2015-12-04 at 4 35 01 pm
screen shot 2015-12-04 at 4 33 10 pm

### DIFF
--- a/core/src/text/textBuffer.cpp
+++ b/core/src/text/textBuffer.cpp
@@ -4,6 +4,8 @@
 #include "gl/texture.h"
 #include "gl/vboMesh.h"
 
+#include <limits>
+
 namespace Tangram {
 
 TextBuffer::TextBuffer(std::shared_ptr<VertexLayout> _vertexLayout)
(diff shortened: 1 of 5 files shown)


bcamper:
The "before" is definitely too high, but the "after" looks like might actually be low?? Did you compare to JS? It's possible we need to adjust to account for the text baseline?

karimnaaji:
Yep, it uses the metrics descender to adjust the label position.
Comparing JS/ES:
screen shot 2015-12-07 at 11 23 46 am
screen shot 2015-12-07 at 11 23 39 am

blair1618:
I was just comparing to JS too, looks pretty good to me!

core/src/text/textBuffer.cpp
```diff
@@ -248,8 +239,17 @@ bool TextBuffer::addLabel(const TextStyle::Parameters& _params, Label::Transform
...
+
+        yMin = std::min(yMin, q.y0);
+        bbox.x = std::max(bbox.x, q.x1);
+        bbox.y = std::max(bbox.y, std::abs(yMin - q.y1));
     }
 
+    // Adjust the bounding boxes
+    bbox.y += metrics.descender * nLine;
+    // Approximate the first left side bearing
+    bbox.x += quads[0].x0;
```
tallytalwar: Hmm, so right now you are shifting the bounding box by left side bearing of the very first quad (of the first line).
But we should be shifting the bounding box by the lowest left side bearing of first quad in all the lines. Should result in a small improvement.

PR #421, opened 2015-12-07
Fix zero area polygon handling

hjanetzek:
Assume that zero area polygons are inner rings and skip them instead of starting a new polygon. Otherwise subsequent inner rings could be missed (i.e. be considered as inner rings of an empty polygon). And if they were outer rings they wouldn't be visible anyway. 

This fixes flooded Manhattan at zoom 9.


tallytalwar:
Nice catch. Looks good to me.

PR #422, opened 2015-12-07
Client data atomic updates

hjanetzek:
Continuing #400


blair1618:
@hjanetzek What's the relationship between this and #400? Should We look into merging #400 first or just look at this one?

hjanetzek:
The atomic-updates commit should come later and probably need some fixing

PR #423, opened 2015-12-07
TileManager cleanup + use make_unique

hjanetzek:
GetInstance did nothing but create new instances

### DIFF
--- a/core/src/tile/tileManager.h
+++ b/core/src/tile/tileManager.h
@@ -32,11 +32,7 @@ class TileManager {
 
 public:
 
-    /* Returns the single instance of the TileManager */
-    static std::unique_ptr<TileManager> GetInstance() {
-        static std::unique_ptr<TileManager> instance (new TileManager());
-        return std::move(instance);
-    }
+    TileManager();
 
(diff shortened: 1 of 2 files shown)


core/src/tile/tileManager.h
```diff
@@ -32,11 +32,7 @@ class TileManager {
 
 public:
 
-    /* Returns the single instance of the TileManager */
-    static std::unique_ptr<TileManager> GetInstance() {
```
tallytalwar: Ohh I think we needed a private static Tile manager instance for it to be truely singleton.

hjanetzek: And would it still be useful? This implementation does create an instance and then directly move it out as return value, so each call to GetInstance returns a new instance.

blair1618: You know, in retrospect I don't think we need to jump through hoops to get the compiler to enforce that `TileManager` is a singleton :)

blair1618:
Looks good to me, cheers to c++14!

tallytalwar:
I agree.

PR #424, opened 2015-12-07
Mult android arch

tallytalwar:
Deploy tangram library for multiple architectures for android.

### DIFF
--- a/travis/script_deploy_android_snapshot.sh
+++ b/travis/script_deploy_android_snapshot.sh
@@ -8,9 +8,9 @@ if [ "${PLATFORM}" = "android" ] && [ "${TRAVIS_PULL_REQUEST}" = "false" ] && [ 
     # Build all android architectures (armeabi-v7a already build)
     make android ANDROID_ARCH=x86 
     make android ANDROID_ARCH=x86_64
-    #make android ANDROID_ARCH=armeabi 
+    make android ANDROID_ARCH=armeabi
     make android ANDROID_ARCH=arm64-v8a
-    #make android ANDROID_ARCH=mips
+    make android ANDROID_ARCH=mips
     make android ANDROID_ARCH=mips64
     cd "$TRAVIS_BUILD_DIR"/android; ./gradlew uploadArchives -PsonatypeUsername="$SONATYPE_USERNAME" -PsonatypePassword="$SONATYPE_PASSWORD"
     cd "$TRAVIS_BUILD_DIR"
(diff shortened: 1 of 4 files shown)


tallytalwar:
This is mostly done. However not able to figure out `__atomic_*_8` linking issues for `mips` and `armeabi` architectures. Both of these are old and very rarely used architectures.

tallytalwar:
Marking this for review.
Following works:
* armeabi-v7a
* arm64-v8a
* x86
* x86_64
* mips64


Still not able to figure out how to get `armeabi` and `mips` architecture working.  Based on https://code.google.com/p/android/issues/detail?id=182094, providing `-latomic` should have done the trick, but does not seem to resolve the issue. I might not be looking at some wrong place in android toolchain.

I am actually fine with not having support for these 2 architecture, if we can not find a way to fix the builds for these, since there are not many devices based on these architectures.

tallytalwar:
@hjanetzek Since you already on this, can you review this and merge?

Thanks

travis/script_deploy_android_snapshot.sh
```diff
@@ -3,7 +3,16 @@
 set -e
 set -o pipefail
 
-if [ "${PLATFORM}" = "android" ] && [ "${TRAVIS_PULL_REQUEST}" = "false" ] && [ "${TRAVIS_BRANCH}" = "master" ]; then
+if [ "${PLATFORM}" = "android" ] && [ "${TRAVIS_PULL_REQUEST}" = "false" ] && [ "${TRAVIS_BRANCH}" = "mult-android-arch" ]; then
+
```
hjanetzek: @tallytalwar can we rename the branch to android-deploy? Since we will have to push to this branch to deploy the android libs.

blair1618:
Glad all the architectures are working now! Now that I'm looking at all of them in the CI builds, we may want to pare down the number of architectures that we build - 25 minutes is just too long to wait for a CI build.

I think a reasonable set of architectures to support for now would be:

 - ARMv7
 - x86
 - ARM64

If someone requests a MIPS build, then we can think about those ;)

travis/script_deploy_android_snapshot.sh
tallytalwar: I had it there for testing, but I guess for now we want to build all (or selected architectures) on every master build. So before merging this, we can revert this change back to `master`.

hjanetzek: I see. ok

hjanetzek:
So.. ready to merge once you decided which platforms to build :)

tallytalwar:
I think we will wait on this at least until tomorrow to discuss in the mobile meeting for what architectures to build and what should be the appropriate strategy to have multiple architecture deployment, considering it costs us ~4mins per architecture on travis.

tallytalwar:
So based on the discussion in the mobile meeting today, I have modified the travis deployment script to do builds for `armeabi`, `armeabi-v7a` (default build), `arm64-v8a` and `x86`.

Deployment will only happen for the `master` branch builds.

Will merge once travis goes green.

PR #427, opened 2015-12-09
Fix and refactor extrusion logic

blair1618:
Extrusion options are somewhat more complicated than one might expect. They're handled in a much clearer way now!

 - JS functions for `extrude` can now return single numbers as well as an array of two numbers
 - Numbers specified for `extrude` are now correctly interpreted as meters and scaled to match
 - Extrude parsing and evaluation is now contained in a separate file

This will fix the craziness demonstrated in https://github.com/tangrams/eraser-map/issues/109

### DIFF
--- a/core/src/util/extrude.cpp
+++ b/core/src/util/extrude.cpp
@@ -1,0 +1,82 @@
+#include "extrude.h"
+#include <cmath>
+#include <cstdlib>
+
+namespace Tangram {
+
+Extrude parseExtrudeString(const std::string& _str) {
+
+    // Values specified from the stylesheet are assumed to be meters with no unit suffix
+    float first = 0, second = 0;
+
+    if (_str == "true") {
+        // "true" means use default properties for both heights, we indicate this with NANs
+        return Extrude(NAN, NAN);
+    }
+
+    if (_str == "false") {
+        // "false" means perform no extrusion
+        return Extrude(0, 0);
+    }
+
+    // Parse the first of two possible numbers
+    const char* pos = _str.c_str();
+    char* end = nullptr;
+
+    // Get a float if possible & advance pos to the end of the number
+    first = std::strtof(pos, &end);
+    if (pos == end) {
+        // No numbers found, return zero extrusion
+        return Extrude(0, 0);
+    }
+
+    // Advance and skip the delimiter character
+    pos = end + 1;
+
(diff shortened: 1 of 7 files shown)


blair1618:
@karimnaaji Take a look and merge, when you get a chance

karimnaaji:
Sure, will review this.

core/src/util/extrude.cpp
```diff
@@ -0,0 +1,82 @@
...
+
+    // Get a float if possible & advance pos to the end of the number
+    first = std::strtof(pos, &end);
+    if (pos == end) {
+        // No numbers found, return zero extrusion
+        return Extrude(0, 0);
+    }
+
+    // Advance and skip the delimiter character
+    pos = end + 1;
```
karimnaaji: From the docs (https://mapzen.com/documentation/tangram/draw/#extrude), it seems that the extrude parameter can have `m`, `px` units associated with the value, defaulting to `m` by default. On stop offset interpolation I just added some allowed unit check based on the draw key https://github.com/tangrams/tangram-es/blob/interpolate-offsets/core/src/scene/styleParam.cpp#L56_L60, we can probably review and merge #429, then parse and add checks for allowed units for sanity here.

karimnaaji:
Looking good, we can merge this after we add parsing for units.

core/src/util/extrude.cpp
bcamper: Yeah, turns out those docs are wrong :)  Only meters are allowed.

blair1618: Right, I based this functionality on observed behavior in tangram-play. It would actually be incorrect for us to accept units here, since tangram-js doesn't. 

Looking more closely, the extrusion behavior in tangram-js is... nuanced. 

`10m` behaves like `true`
`blah` behaves like `true`
`[10, 20m]` behaves like `[10, 0]`
`[10m, 20]` behaves like `[0, 20]`

It also seems that `extrude: 10` extrudes from `min_height` to 10 (which results in some inside-out walls). I had expected it to extrude from 0 to 10.

This PR won't precisely match some of these behaviors; I'm not sure that these are all desirable anyway.

bcamper: Agreed, not all of those are desirable. That code is very old, one of the
first configurable style parameters. We can revisit soon.

PR #429, opened 2015-12-09
Interpolate offsets

karimnaaji:
- Interpolation for offset
- Add unit (meters/pixel/...) validity check for widths

### DIFF
--- a/core/src/scene/stops.cpp
+++ b/core/src/scene/stops.cpp
@@ -74,7 +74,7 @@ auto Stops::FontSize(const YAML::Node& _node) -> Stops {
     return stops;
 }
 
-auto Stops::Widths2D(const YAML::Node& _node, const std::vector<Unit>& _units) -> Stops {
+auto Stops::Offsets(const YAML::Node& _node, const std::vector<Unit>& _units) -> Stops {
     Stops stops;
     if (!_node.IsSequence()) {
         return stops;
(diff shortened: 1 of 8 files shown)


core/src/scene/stops.cpp
```diff
@@ -74,7 +74,52 @@ auto Stops::FontSize(const YAML::Node& _node) -> Stops {
...
+            }
+        }
+    }
+
+    return stops;
+}
+
+auto Stops::Widths(const YAML::Node& _node, const MapProjection& _projection, const std::vector<Unit>& _units) -> Stops {
     Stops stops;
     if (!_node.IsSequence()) { return stops; }
```
tallytalwar: I don't think we need this check again, since its already checked in `sceneLoader.cpp`.

core/src/scene/styleParam.cpp
```diff
@@ -456,9 +464,18 @@ bool StyleParam::isColor(StyleParamKey _key) {
...
         case StyleParamKey::outline_width:
         case StyleParamKey::size:
+        case StyleParamKey::font_stroke_width:
+            return true;
+        default:
+            return false;
+    }
+}
+
+bool StyleParam::is2DWidth(StyleParamKey _key) {
```
tallytalwar: I am not sure if `2DWidth` or `Widths2D` are appropriate names here. I can see it being confusing as its not exactly a width parameter! How about something more generic like, `2DValue` and `Values2D` respectively.

Apart from this all logic looks good here.

hjanetzek: I would prefer Offset(s), just like the style param name, if I'm not missing something.

tallytalwar: :+1:

PR #431, opened 2015-12-09
Consistently store YAML and JS numbers as doubles

blair1618:
Different pieces of code have been inconsistently interpreting numbers from the YAML stylesheet, from tile data (JSON and protobuff), and from JS functions. This was enabled by the specific variant type used for interchange between these pieces of code:
```
using Value = variant<none_type, float, int64_t, std:string>
```
Inconsistent encoding caused problems with filtering, where `Value`s are compared for equality. For example:

 - The scene specifies a filter: `{ scalerank: [0, 1] }`. The numbers are encoded as `float`s. 
 - A feature in a tile contains the tag: `scalerank: 1`. The number is encoded as an `int64_t`. 
 - The filter is applied to the feature and tries to compare the values by `==`, which fails because they are different types within the `variant`. 

This can be fixed in two ways:

1. During filtering, check whether values being compared are of convertible types and then attempt conversion and comparison. This adds complexity and is likely to cause problems in the future if someone forgets to cover a possible conversion case. 

2. Store numbers in these contexts as `double`. Always.  

I've opted for the second solution, for simplicity and robustness to future code changes. The variant type is now:
```
using Value = variant<none_type, double, std::string>
```
This change substantially improves the correctness of rule matching behavior in the eraser-map style!


tallytalwar:
Looks good to me. Can you add some appropriate test cases?

blair1618:
@tallytalwar Tests are a good idea, though as you found earlier today this problem won't manifest if you just instantiate filters and properties through code - this was an artifact of the interaction between the tile parsing process and the filtering process.

hjanetzek:
Looks good to me as well. I've added the integer type to properly handle osm ids for the initial feature selection implementation for which just having float Values was insufficient.

A third option would have been to use a binary visitor for Value to check equality. There is an (a bit more elaborate than necessary) example in  https://github.com/mapbox/variant/blob/master/test/binary_visitor_test.cpp

PR #432, opened 2015-12-10
Floating Point comparison fix in filter equality

tallytalwar:
We should still do this along with the fixes from https://github.com/tangrams/tangram-es/pull/431.

### DIFF
--- a/core/src/scene/filters.cpp
+++ b/core/src/scene/filters.cpp
@@ -2,6 +2,8 @@
 #include "scene/styleContext.h"
 #include "data/tileData.h"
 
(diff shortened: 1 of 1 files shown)


blair1618:
Sure, but let's put the simple equality check first so we can skip the epsilon check if it isn't needed.

tallytalwar:
Good call, covered by a179272

blair1618:
Think you can squash these and remove that extra space before "v" on line 41? Thanks :)

tallytalwar:
Ha sure. Had already squashed.

PR #433, opened 2015-12-11
Resolve ambiguous draw rules by lexicographic order of layer names

blair1618:
See: https://github.com/tangrams/tangram/pull/225

For normal builds, ambiguous rules will be resolved with no warning messages. 

With the compile flag `-DTANGRAM_WARN_ON_RULE_CONFLICTS` warnings will be logged for each unique rule conflict. These warnings include some conflicts that do not result in true ambiguity. If two matched rules assign the same parameter at depth *d* and another matched rule assigns the parameter at a greater depth *d+n*, then the result is unambiguously the parameter from depth *d+n* but the conflict at depth *d* may still be logged. In most cases it is still desirable to remove these conflicts.


blair1618:
@hjanetzek thanks for the tweaks! Is this code making sense to you?

hjanetzek:
yeah, looks good to me!

tallytalwar:
Can I give this a look before it's merged?

blair1618:
@tallytalwar Yeah let me know when you're good with it. We want it merged today but it's also important to get this right.

tallytalwar:
Sure, I am on it.

tallytalwar:
This looks good, Merging.

PR #434, opened 2015-12-11
Fix bbox (Final round!)

karimnaaji:
After setting a background for the fragments of the glyph quads, the issue was more easy to solve. The bounding box is actually based on the quad positions, thus can have negative values on the lower left corner in order for the first glyph to be perfectly aligned with (0,0) based on its left side bearing and other font parameters. But using this as the origin of the bounding box makes it relatively wrong compared to the quad pixels:

screen_shot_2015-12-10_at_12 29 30_pm

Shifting the bounding box to its origin (first glyph lower left corner, min of the first glyph of each line for multi-line labels) fixes the issue and makes the bounding box perfectly aligned with the quads:

screen_shot_2015-12-10_at_2 28 51_pm

In this case the bounding boxes are still too big compared to the visible pixels. 
A *buffer* is being used (half of the font size gives relatively good result), which makes the bounding box closer to the actual visible pixels:

screen shot 2015-12-11 at 9 58 06 am

See https://trello.com/c/uPgO2dEr/761-label-bounding-boxes-are-wrong for more details.

### DIFF
--- a/core/resources/scene.yaml
+++ b/core/resources/scene.yaml
@@ -97,7 +97,6 @@ layers:
             polygons:
                 order: 0
                 color: '#f0ebeb'
-
     landuse:
         data: { source: osm }
         filter:
--- a/core/resources/scene.yaml
+++ b/core/resources/scene.yaml
@@ -327,7 +327,6 @@ layers:
                 polygons:
                     style: polygons
                     color: '#bddec5'
-
     poi_labels:
         data: { source: osm, layer: pois }
         filter: { name: true, not: { kind: [peak, viewpoint, bicycle_rental, car_sharing] }, $zoom: { min: 15 } }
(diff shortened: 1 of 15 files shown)


bcamper:
Thanks, this will make the observed visual much closer to JS, and we can
then further tweak either as needed (could add more buffer on JS too if we
want).

karimnaaji:
Yeah glad this is finally fixed.. and looking much more accurate.
I based this on #429 by for testing, so let's merge #429 first.
@tallytalwar can you give a look to this when you get a chance, that's mostly the fix we were speaking about.

blair1618:
Final Round. FINISH HIM

tallytalwar:
Yup will fire the final bullet for this.

tallytalwar:
@karimnaaji this should be based off master, right?
Currently it seems its based off `interpolate-offsets`

karimnaaji:
@tallytalwar, yes it made things easier to test eraser-map stylesheet which uses offset interpolation features to iterate on this fix.
I can close this and rebase master.

tallytalwar:
No worries will review https://github.com/tangrams/tangram-es/pull/429, along with this PR. Thanks

PR #435, opened 2015-12-11
Expose debugging api

tallytalwar:
Provides swig bindings to enable Tangram debugging from android apps.


tallytalwar:
@hjanetzek can you check this and merge. Thanks

android/tangram/src/com/mapzen/tangram/DebugFlags.java
```diff
@@ -0,0 +1,57 @@
...
+ * Version 3.0.7
+ *
+ * Do not make changes to this file unless you know what you are doing--modify
+ * the SWIG interface file instead.
+ * ----------------------------------------------------------------------------- */
+
+package com.mapzen.tangram;
+
+public final class DebugFlags {
+    public final static DebugFlags freeze_tiles = new DebugFlags("freeze_tiles", 0);
```
ecgreb: Java Language Specification suggests putting `static` before `final` by convention but in reality the modifier order does not matter. Its mostly for readability.

http://checkstyle.sourceforge.net/config_modifier.html

blair1618:
It's weird that swig generated a set of static class instances instead of an enum for DebugFlags; a Java enum would have more useful functionality here. This is one of those cases where hand-written bindings would turn out better, but this is probably faster and requires less maintenance.

android/tangram/src/com/mapzen/tangram/DebugFlags.java
ecgreb: Also public static final fields (constants) are typically expressed in ALL_CAPS.

blair1618: You should tell [SWIG](http://www.swig.org/) ;)

ecgreb: Oh ha I see now that this file is generated. In that case all codestyle bets are off ;)

tallytalwar:
You can read more here: http://www.swig.org/Doc1.3/Java.html#typesafe_enums

hjanetzek:
@blair1618 iirc swig could use enums but doesn't do by default. I'll have a look again

tallytalwar:
We could do this though to use `proper java enums`. http://www.swig.org/Doc1.3/Java.html#proper_enums

tallytalwar:
@hjanetzek and @blair1618 Check this:  cce9b13

blair1618:
Hey nice!

Now I'll be even more impressed if we can force the names to follow the Java ALL_CAPS convention :)

tallytalwar:
We can probably do that with `clang_format` and `astyle`

blair1618:
The naming convention isn't critical. We could always tweak the SWIG code by hand but I'm ok with this right now.

ecgreb:
:+1:

PR #436, opened 2015-12-14
WIP Rework tile states

hjanetzek:
Now based on master - continue #428

This branch removes the explicit use of TileState by directly working with the actual state of tiles. The Tile class now only holds the data that is used in rendering, state related to tile management is kept within the TileEntries of a TileSet.

### DIFF
--- a/bench/src/tileLoading.cpp
+++ b/bench/src/tileLoading.cpp
@@ -8,6 +8,7 @@
 #include "scene/styleContext.h"
 #include "util/mapProjection.h"
 #include "tile/tile.h"
+#include "tile/tileTask.h"
 
 #include <vector>
 #include <iostream>
@@ -69,7 +70,13 @@ struct TestContext {
 
     void parseTile() {
         Tile tile({0,0,0}, s_projection);
-        tileData = source->parse(tile.getID(), s_projection, rawTileData);
+        source = scene->dataSources()[0];
+        auto task = source->createTask(tile.getID());
+        auto& t = dynamic_cast<DownloadTileTask&>(*task);
+        t.rawTileData = std::make_shared<std::vector<char>>();
+        std::swap(*t.rawTileData, rawTileData);
+
+        tileData = source->parse(*task, s_projection);
     }
 };
 
@@ -126,7 +133,12 @@ public:
 
         if (!scene->dataSources().empty()) {
             source = scene->dataSources()[0];
-            tileData = source->parse(tile.getID(), s_projection,rawTileData);
+            auto task = source->createTask(tile.getID());
+            auto& t = dynamic_cast<DownloadTileTask&>(*task);
+            t.rawTileData = std::make_shared<std::vector<char>>();
+            std::swap(*t.rawTileData, rawTileData);
+
+            tileData = source->parse(*task, s_projection);
         }
 
         LOG("Ready");
--- a/bench/src/tileLoading.cpp
+++ b/bench/src/tileLoading.cpp
@@ -69,7 +69,7 @@ struct TestContext {
 
     void parseTile() {
         Tile tile({0,0,0}, s_projection);
-        tileData = source->parse(tile, rawTileData);
+        tileData = source->parse(tile.getID(), s_projection, rawTileData);
     }
 };
 
@@ -124,9 +124,10 @@ public:
 
         Tile tile({0,0,0}, s_projection);
 
-        source = scene->dataSources()[0];
-
-        tileData = source->parse(tile, rawTileData);
+        if (!scene->dataSources().empty()) {
+            source = scene->dataSources()[0];
+            tileData = source->parse(tile.getID(), s_projection,rawTileData);
+        }
 
         LOG("Ready");
     }
--- a/core/src/data/clientGeoJsonSource.cpp
+++ b/core/src/data/clientGeoJsonSource.cpp
@@ -21,6 +21,10 @@ const uint8_t indexMaxZoom = maxZoom;
 const uint32_t indexMaxPoints = 100000;
 double tolerance = 1E-8;
 
+std::shared_ptr<TileTask> ClientGeoJsonSource::createTask(TileID _tileId) {
+    return std::make_shared<TileTask>(_tileId, shared_from_this());
+}
+
 // Transform a geojsonvt::TilePoint into the corresponding Tangram::Point
 Point transformPoint(geojsonvt::TilePoint pt) {
     return { pt.x / extent, 1. - pt.y / extent, 0 };
@@ -51,11 +55,6 @@ bool ClientGeoJsonSource::loadTileData(std::shared_ptr<TileTask>&& _task, TileTa
 
     _cb.func(std::move(_task));
 
-    return true;
-}
-
-bool ClientGeoJsonSource::getTileData(std::shared_ptr<TileTask>& _task) {
-    _task->loaded = true;
     return true;
 }
 
@@ -120,8 +119,8 @@ void ClientGeoJsonSource::addPoly(const Properties& _tags, const std::vector<Coo
     m_generation++;
 }
 
-std::shared_ptr<TileData> ClientGeoJsonSource::parse(const TileID& _tileId, const MapProjection& _projection,
-                                                     std::vector<char>& _rawData) const {
+std::shared_ptr<TileData> ClientGeoJsonSource::parse(const TileTask& _task,
+                                                     const MapProjection& _projection) const {
 
     auto data = std::make_shared<TileData>();
 
@@ -129,7 +128,7 @@ std::shared_ptr<TileData> ClientGeoJsonSource::parse(const TileID& _tileId, cons
     {
         std::lock_guard<std::mutex> lock(m_mutexStore);
         if (!m_store) { return nullptr; }
-        tile = m_store->getTile(_tileId.z, _tileId.x, _tileId.y);
+        tile = m_store->getTile(_task.tileId().z, _task.tileId().x, _task.tileId().y);
     }
 
     Layer layer(""); // empty name will skip filtering by 'collection'
--- a/core/src/data/clientGeoJsonSource.cpp
+++ b/core/src/data/clientGeoJsonSource.cpp
@@ -120,17 +120,16 @@ void ClientGeoJsonSource::addPoly(const Properties& _tags, const std::vector<Coo
     m_generation++;
 }
 
-std::shared_ptr<TileData> ClientGeoJsonSource::parse(const Tile& _tile, std::vector<char>& _rawData) const {
+std::shared_ptr<TileData> ClientGeoJsonSource::parse(const TileID& _tileId, const MapProjection& _projection,
+                                                     std::vector<char>& _rawData) const {
 
     auto data = std::make_shared<TileData>();
-
-    auto id = _tile.getID();
 
     geojsonvt::Tile tile;
     {
         std::lock_guard<std::mutex> lock(m_mutexStore);
         if (!m_store) { return nullptr; }
-        tile = m_store->getTile(id.z, id.x, id.y); // uses a mutex lock internally for thread-safety
+        tile = m_store->getTile(_tileId.z, _tileId.x, _tileId.y);
     }
 
     Layer layer(""); // empty name will skip filtering by 'collection'
--- a/core/src/data/dataSource.cpp
+++ b/core/src/data/dataSource.cpp
@@ -31,16 +31,16 @@ struct RawCache {
     int m_usage = 0;
     int m_maxUsage = 0;
 
-    bool get(std::shared_ptr<TileTask>& _task) {
+    bool get(TileTask& _task) {
 
         if (m_maxUsage <= 0) { return false; }
 
         std::lock_guard<std::mutex> lock(m_mutex);
-        auto it = m_cacheMap.find(_task->tile->getID());
+        auto it = m_cacheMap.find(_task.tileId);
         if (it != m_cacheMap.end()) {
             // Move cached entry to start of list
             m_cacheList.splice(m_cacheList.begin(), m_cacheList, it->second);
-            _task->rawTileData = m_cacheList.front().second;
+            _task.rawTileData = m_cacheList.front().second;
 
             return true;
         }
@@ -121,7 +121,7 @@ void DataSource::constructURL(const TileID& _tileCoord, std::string& _url) const
 }
 
 bool DataSource::getTileData(std::shared_ptr<TileTask>& _task) {
-    if (m_cache->get(_task)) {
+    if (m_cache->get(*_task)) {
         _task->loaded = true;
         return true;
     }
@@ -129,7 +129,7 @@ bool DataSource::getTileData(std::shared_ptr<TileTask>& _task) {
 }
 
 void DataSource::onTileLoaded(std::vector<char>&& _rawData, std::shared_ptr<TileTask>& _task, TileTaskCb _cb) {
-    TileID tileID = _task->tile->getID();
+    TileID tileID = _task->tileId;
 
     if (!_rawData.empty()) {
         _task->loaded = true;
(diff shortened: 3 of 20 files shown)


core/src/tile/tile.h
```diff
@@ -51,7 +49,7 @@ class Tile {
     float getScale() const { return m_scale; }
 
     /* Returns the reciprocal of <getScale()> */
-    float getInverseScale() const { return m_inverseScale; }
+    float getInverseScale() const { return 1.0f / m_scale; }
```
tallytalwar: I would still store the `inverse scale` as a member. Doing an division repeatedly is usually not recommended, specially a floating-point division.

core/src/data/dataSource.h
```diff
@@ -14,7 +16,7 @@ struct RawCache;
 class TileTask;
 struct TileTaskCb;
 
-class DataSource {
+class DataSource : public std::enable_shared_from_this<DataSource> {
```
tallytalwar: This is nice, Didn't know about this. :+1:

core/src/data/dataSource.cpp
```diff
@@ -120,23 +128,15 @@ void DataSource::constructURL(const TileID& _tileCoord, std::string& _url) const
...
+    TileID tileID = _task->tileId();
 
     if (!_rawData.empty()) {
-        _task->loaded = true;
 
         auto rawDataRef = std::make_shared<std::vector<char>>();
         std::swap(*rawDataRef, _rawData);
-        _task->rawTileData = rawDataRef;
+
+        dynamic_cast<DownloadTileTask*>(_task.get())->rawTileData = rawDataRef;
```
tallytalwar: I would first check if the dynamic_cast is legal. I know it will never be in this case, but proper is proper :D. So something like this:

```
auto t = dynamic_cast<DownloadTileTask*>(_task.get());
if (t) {
    t->rawTileData = rawDataRef;
}
```

core/src/tile/tile.h
blair1618: I must have believed this was worthwhile at one point, but it's almost certainly much slower to go to memory and fetch this member than to do a floating point divide.

core/src/data/dataSource.cpp
karimnaaji: In this case it's also safe to use a `static_cast` since it happens to be only for downloadable tiles. We should think of reducing usage of `dynamic_cast` for no RTTI in general anyway.

core/src/tile/tile.h
tallytalwar: There might be some cache locality taking into account.
Considering when we call this in https://github.com/tangrams/tangram-es/blob/rework-tile-states/core/src/style/polygonStyle.cpp#L94, there is already some reference to `_tile.getID()`, which would have brought in neighboring memory in cache/ram!

Anyhow, irrespective of the guess work of cache locality, what I am suggesting here are just best practices.

core/src/data/dataSource.cpp
tallytalwar: :+1:  to what @karimnaaji suggested.

core/src/data/clientGeoJsonSource.cpp
```diff
@@ -20,6 +20,20 @@ const uint8_t indexMaxZoom = 18;
 const uint32_t indexMaxPoints = 100000;
 double tolerance = 1E-8;
 
+class Task : public TileTask {
```
tallytalwar: Instead of creating this temp `Task` class for `clientGeoJsonSource`, how about not making the base `TileTask` class abstract and make it implement `hasData` method to return `true`.

Other data source classes are already overriding this method.

hjanetzek: Yes, makes sense to have the default implementation for hasData() now. The initial version of that commit did also override TileTask.process() which is now gone.

core/src/tile/tile.h
hjanetzek: I agree with avoiding division where possible for values that are used in a loop. But we are discussing here about avoiding one division per feature. In this case my rule of thumb would say to rather keep the struct lean to fit into one cacheline. 
In any case the effect is likely to be below a measurable threshold.

core/src/data/geoJsonSource.cpp
```diff
@@ -18,14 +19,17 @@ GeoJsonSource::GeoJsonSource(const std::string& _name, const std::string& _urlTe
     DataSource(_name, _urlTemplate) {
 }
 
-std::shared_ptr<TileData> GeoJsonSource::parse(const Tile& _tile, std::vector<char>& _rawData) const {
+std::shared_ptr<TileData> GeoJsonSource::parse(const TileTask& _task,
+                                               const MapProjection& _projection) const {
+
+    auto& task = dynamic_cast<const DownloadTileTask&>(_task);
```
tallytalwar: I think you can safely do `static_cast` here also. http://en.cppreference.com/w/cpp/language/static_cast.

Similarly in `mvtSource.cpp` and in ` bench/src/tileLoading.cpp`.

core/src/tile/tileWorker.cpp
```diff
@@ -64,43 +63,51 @@ void TileWorker::run() {
...
-                        return a->tile->isVisible();
+                    if (a->isProxy() != b->isProxy()) {
+                        return !a->isProxy();
                     }
-                    if (a->tile->sourceID() == b->tile->sourceID() &&
-                        a->tile->sourceGeneration() != b->tile->sourceGeneration()) {
-                        return a->tile->sourceGeneration() < b->tile->sourceGeneration();
+                    if (a->source().id() == b->source().id() &&
+                        a->sourceGeneration() != b->sourceGeneration()) {
+                        return a->sourceGeneration() < b->sourceGeneration();
```
tallytalwar: Should a higher sourceGeneration get priority to be build first?

core/src/tile/tileWorker.cpp
```diff
@@ -64,43 +63,51 @@ void TileWorker::run() {
...
-        auto tileData = task->process();
-
-        // NB: Save shared reference to Scene while building tile
-        auto scene = m_tileManager.getScene();
+        if (task->isCanceled()) { continue; }
 
+        // Save shared reference to Scene while building tile
+        // FIXME: Scene could be released on Worker-Thread and
+        // therefore call unsafe glDelete* functions...
+        auto scene = m_scene;
```
tallytalwar: We can solve this by having a weak pointer to scene here!.

core/src/tile/tileManager.cpp
```diff
@@ -303,204 +215,244 @@ void TileManager::updateTileSet(TileSet& tileSet) {
...
+
+            auto& entry = it->second;
+
+            LOG("REMOVE %s - ready:%d proxy:%d/%d",
+                it->first.toString().c_str(),
+                entry.isReady(),
+                entry.getProxyCounter(),
+                entry.m_proxies);
+
+            if (visibleTiles.find(it->first) != visibleTiles.end()) {
```
tallytalwar: How is this possible? If a tile was in visibleTiles set then it should not be a part of `removeTiles` at the first place!

core/src/tile/tileManager.cpp
```diff
@@ -303,204 +215,244 @@ void TileManager::updateTileSet(TileSet& tileSet) {
...
-            } else if (!tileSet.source->loadTileData(std::move(task),
-                                                     m_dataCallback)) {
-                LOGE("Loading failed for tile [%d, %d, %d]", id.z, id.x, id.y);
-                m_loadPending--;
+        } else if (m_loadPending < int(MAX_DOWNLOADS)) {
+            entry.task = task;
+            if (tileSet.source->loadTileData(std::move(task), m_dataCallback)) {
+                m_loadPending++;
+            } else {
+                // TODO: How to handle this case?
```
tallytalwar: We can introduce a `tries` member to `DownloadTask`. And increment the tries. If it reaches a max, then we get rid of the task?

core/src/tile/tileManager.cpp
```diff
@@ -303,204 +215,244 @@ void TileManager::updateTileSet(TileSet& tileSet) {
...
     {
-        const auto& it = tiles.find(_proxy);
+        const auto& it = tiles.find(_proxyTileId);
         if (it != tiles.end()) {
-            auto& proxyTile = it->second;
-            if (proxyTile && _tile.setProxy(_proxyID)) {
-                proxyTile->incProxyCounter();
-                if (proxyTile->isReady()) {
-                    m_tiles.push_back(proxyTile);
+            // FIXME: this uses indirect proxy tiles
```
tallytalwar: Can you explain this situation?

core/src/tile/tileManager.cpp
```diff
@@ -303,204 +215,244 @@ void TileManager::updateTileSet(TileSet& tileSet) {
...
+            // even if a cached tile for the direct proxy
+            // may be available
+            auto& entry = it->second;
+            if (_tile.setProxy(_proxyId)) {
+                entry.incProxyCounter();
+
+                if (entry.isReady()) {
+                    m_tiles.push_back(entry.tile);
                 }
-                return true;
```
tallytalwar: And why remove this?

hjanetzek: Yeah, I see now there is no good reason to remove this..

If the tile is available in TileCache then it would have been added to TileSet, so no need to check the cache again.

core/src/tile/tileManager.cpp
hjanetzek: The general problem is that this logic does not find the nearest available proxy. I try to give an example: Say, tiles from z=0 and z=3 are ready; The view is zooming in from z=0 to z=2 quickly, so that no new tiles get loaded in between. 
At z=1 the direct parent tile is used as proxy(this is ok). But at z=2 the direct parent is used as proxy which is not ready and still references the proxy from z=0, even though the direct proxies from z=3 would be available.

core/src/tile/tileManager.cpp
hjanetzek: Yes, there may be a better way to solve this. Currently it is possible to have circles in the proxies where a parent references its children and the children reference their parent as proxies. This can lead to adding current tiles to the removal list.

core/src/tile/tileWorker.cpp
hjanetzek: This order ensures that the earlier requests are processed first. Otherwise when there are continuous updates to the source the lower prioritized tiles would never be updated.

core/src/tile/tileWorker.cpp
hjanetzek: Unfortunately this would only move the problem: To use the scene it is necessary to acquire a shared reference.

core/src/tile/tileManager.cpp
hjanetzek: Added a test case for circular references between proxies 
https://github.com/tangrams/tangram-es/commit/5aa00eb3db71d8c6b26ab0e5ecf9f24bbf271e23

hjanetzek: @tallytalwar  sorry, that was my fault. I wanted to remove the tile entry 'visible' flag. But it was required for the isVisible() check in clearProxyTiles. It's now back in..

tallytalwar:
I am mostly good with this. There are some `fixme`s in there. Plus I would like to have some testing done with this before this gets merged. Its a big a change to the very core of tangram-es.

hjanetzek:
No hurry, though if you want to work on parts touched by this PR, please see if you can implement it on top of tilemanager-testing branch. 
One fixme was obsolete and the other for proxy improvements was rather a todo.

bench/src/tileLoading.cpp
```diff
@@ -69,7 +70,13 @@ struct TestContext {
...
     void parseTile() {
         Tile tile({0,0,0}, s_projection);
-        tileData = source->parse(tile, rawTileData);
+            source = scene->dataSources()[0];
+            auto task = source->createTask(tile.getID());
+            auto& t = dynamic_cast<DownloadTileTask&>(*task);
+            t.rawTileData = std::make_shared<std::vector<char>>();
+            std::swap(*t.rawTileData, rawTileData);
+
+            tileData = source->parse(*task, s_projection);
```
blair1618: Extra indentation

blair1618:
With these changes, I'm seeing the update/render loop running continuously even after tiles are done loading and the view is static. Initial debugging shows that `m_workers->checkPendingTiles()` is constantly returning true, but I'm not sure of the root cause yet.

hjanetzek:
@blair1618  hmm, yes I see it now too. It seems to happen when the client data source has no data. When you click once to add a point it stops rerunning. I'll look into it.

core/src/tile/tileWorker.h
```diff
@@ -5,42 +5,57 @@
...
 
     void run();
-    
-    TileManager& m_tileManager;
-    
+
     bool m_running;
-    
+
+    std::atomic<bool> m_pendingTiles;
```
tallytalwar: Maybe `std::atomic_flag` has a better usage here: http://en.cppreference.com/w/cpp/atomic/atomic_flag.

tallytalwar:
@hjanetzek is it because https://github.com/tangrams/tangram-es/blob/rework-tile-states/core/src/tile/tileManager.cpp#L343, always results in a callback for TileTask for a clientDataSource?

core/src/tile/tileWorker.h
hjanetzek: I think atomic<bool> is appropriate here. atomic_flag seems to be useful if one wants to implement a spinlock

hjanetzek:
@tallytalwar yes, in this case that's the issue. I thought of adding a function to DataSource to check if it can provide data at all (e.g. to check when clientDataSource m_store is not empty). Then one could skip updates for these sources and remove all of its tiles in TileManager

core/src/tile/tileManager.cpp
```diff
@@ -246,248 +155,293 @@ void TileManager::updateTileSet(TileSet& tileSet) {
...
+            if (!task->hasData()) {
+                m_loadPending++;
+            }
+        }
+
+        if (entry.isReady()) {
+            // Mark as proxy
+            entry.tile->setProxyState(entry.getProxyCounter() > 0);
+        }
+        ++it;
```
blair1618: Can this increment go in the loop declaration?

core/src/tile/tileManager.cpp
```diff
@@ -246,248 +155,293 @@ void TileManager::updateTileSet(TileSet& tileSet) {
...
-        auto& tile = entry.second;
-        auto tileCenter = m_view->getMapProjection().TileCenter(tile->getID());
-        double scaleDiv = exp2(tile->getID().z - m_view->getZoom());
-        // prefer parent tiles
-        if (scaleDiv < 1) { scaleDiv = 0.1/scaleDiv; }
-        tile->setPriority(glm::length2(tileCenter - viewCenter) * scaleDiv);
+    for (auto it = tiles.begin(); it != tiles.end();) {
+        auto& entry = it->second;
+
+        LOG("> %s - ready:%d proxy:%d/%d loading:%d",
```
blair1618: Let's make these `LOGD` before merging

hjanetzek: Sure, if we are near merging :)  I'll go over it for cleanup until Monday

core/src/tile/tileManager.cpp
blair1618: As I see it, the meaning of `loadTileData` returning `false` is a non-recoverable error, so loading should not be re-tried. Network failures and re-tries should be handled within the networking backend itself, `TileManager` shouldn't have to reason about networking like that.

core/src/tile/tile.h
hjanetzek: I'll remove this change as it does not really belong into this PR.

Thinking about the meaning of scale vs inverse scale it's not quite clear to me why tile width in projection-meters is considered to be the 'scale' of a tile - Is this an arbitrary decision or is there a good reasoning for it? 

If the distiction is arbitrary I would suggest to remove the m_scale member instead of m_inverseScale and rename inverseScale to scale. Since m_scale and tile.getScale() are not used anywhere, except for initialization of the tile's modelmatrix  :)

blair1618: I believe the distinction is fairly arbitrary, it would be reasonable to reconsider this member now that we have a broader view of the tile pipeline.

hjanetzek:
ping @tallytalwar, @blair1618

tallytalwar:
I am good with this. @blair1618 good to go?

blair1618:
Good timing, I was about to say that I'm ready to merge this :+1:

PR #437, opened 2015-12-15
Use JS Proxy object for feature

hjanetzek:
Duktape supports ES6 Proxy objects which enables use of generic handler functions for all property access.
This avoids creating (and checking the existence of) explicit property accessors for each feature.

Found this option by looking through duk_hobject_props.c - Writing the proxy code oneself would have rather been a full month project I guess :)

### DIFF
--- a/core/src/scene/styleContext.cpp
+++ b/core/src/scene/styleContext.cpp
@@ -15,8 +15,7 @@
 
 namespace Tangram {
 
-const static char DATA_ID[] = "\xff""\xff""data";
-const static char ATTR_ID[] = "\xff""\xff""attr";
+const static char INSTANCE_ID[] = "\xff""\xff""obj";
 const static char FUNC_ID[] = "\xff""\xff""fns";
 
 static const std::string key_geometry = "$geometry";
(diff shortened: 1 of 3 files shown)


blair1618:
Nice. That was fast :)

Looks like this essentially sends feature properties to duktape on-demand? Makes way more sense to me than populating lots of unused accessors.

hjanetzek:
Yes, it's definitely the right way to wrap hash-like objects. All property accesses are redirected to the 'get(obj,key,..)' handler which can be a native function 

https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Proxy

core/src/scene/styleContext.cpp
```diff
@@ -25,20 +24,8 @@ static const std::string key_zoom("$zoom");
...
-        LOGE("Ctx not assigned");
-    }
-
-    // put object in global scope
-    if (!duk_put_global_string(m_ctx, "feature")) {
-        LOGE("Feature not assigned");
-    }
-
+    //// Create global geometry constants
+    // TODO make immutable
```
blair1618: Maybe these immutable globals could be proxy objects with a no-op setter?

blair1618: It seems that these globals are not strictly immutable in tangram-js, so to most nearly reproduce its behavior we should technically allow these to be re-assigned within a function. However tangram-js also confines these globals to a closure for each function evaluation, so re-assignment would not impact their value in other functions. It seems that we would need to assign the global variables before every evaluation in order to mimic this. 

Since this is a relatively unlikely case, I'm fine postponing its resolution until we run out of more important issues to work on, but now at least we know about this potential discrepancy.

hjanetzek: Yeah, this can be done later - there is also Object.freeze() which sounds related

bcamper: Object.freeze may be slow, but worth trying both (agree low priority)

blair1618: Yep, that too - though it's not clear that freeze would work on primitives (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/freeze).

blair1618:
This is looking good to me - gonna merge unless anyone has objections.

tallytalwar:
Haven’t looked at it. But I take your word for it :D.

Will check it out though.

PR #438, opened 2015-12-15
Blend order

tallytalwar:
Blend order to tie-break for different non-opaque tiles.

Equivalent JS PR: https://github.com/tangrams/tangram/pull/223

### DIFF
--- a/core/src/scene/sceneLoader.cpp
+++ b/core/src/scene/sceneLoader.cpp
@@ -98,15 +98,7 @@ bool SceneLoader::loadScene(Node& config, Scene& _scene) {
 
     // Styles that are opaque must be ordered first in the scene so that
     // they are rendered 'under' styles that require blending
-    std::sort(_scene.styles().begin(), _scene.styles().end(),
-              [](std::unique_ptr<Style>& a, std::unique_ptr<Style>& b) {
-                  if (a->blendMode() != b->blendMode()) {
-                      return a->blendMode() == Blending::none;
-                  }
-                  // Just for consistent ordering
-                  // anytime the scene is loaded
-                  return a->getName() < b->getName();
-              });
+    std::sort(_scene.styles().begin(), _scene.styles().end(), Style::compare);
 
     // Post style sorting set their respective IDs=>vector indices
     // These indices are used for style geometry lookup in tiles
@@ -450,6 +442,15 @@ void SceneLoader::loadStyleProps(Style& style, Node styleNode, Scene& scene) {
         else { LOGW("Invalid blend mode '%s'", str.c_str()); }
     }
 
+    if (Node blendOrderNode = styleNode["blend_order"]) {
+        try {
+            auto blendOrder = blendOrderNode.as<int>();
+            style.setBlendOrder(blendOrder);
+        } catch (const BadConversion& e) {
+            LOGE("Integral value expected for blend_order style parameter.\n");
+        }
+    }
+
     if (Node texcoordsNode = styleNode["texcoords"]) {
         LOGW("'texcoords' style parameter is currently ignored");
 
@@ -666,7 +667,7 @@ Node SceneLoader::mixStyles(const std::vector<Node>& styles) {
         }
     }
 
-    for (auto& property : {"base", "lighting", "texture", "blend", "material", "shaders"}) {
+    for (auto& property : {"base", "lighting", "texture", "blend", "material", "shaders", "blend_order"}) {
         Node node = propMerge(property, styles);
         if (!node.IsNull()) {
             result[property] = node;
--- a/core/src/style/style.h
+++ b/core/src/style/style.h
@@ -31,8 +31,8 @@ enum class Blending : char {
     none,
     add,
     multiply,
-    overlay,
     inlay,
+    overlay,
 };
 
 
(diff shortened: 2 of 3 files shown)


core/src/scene/sceneLoader.cpp
```diff
@@ -450,6 +458,11 @@ void SceneLoader::loadStyleProps(Style& style, Node styleNode, Scene& scene) {
         else { LOGW("Invalid blend mode '%s'", str.c_str()); }
     }
 
+    if (Node blendOrderNode = styleNode["blend_order"]) {
+        auto blendOrder = blendOrderNode.as<int>();
```
blair1618: With new additions to the yaml parsing we should try to be robust to unexpected content (i.e. wrap conversions in a try-catch)

core/src/scene/sceneLoader.cpp
```diff
@@ -101,7 +101,15 @@ bool SceneLoader::loadScene(Node& config, Scene& _scene) {
     std::sort(_scene.styles().begin(), _scene.styles().end(),
               [](std::unique_ptr<Style>& a, std::unique_ptr<Style>& b) {
                   if (a->blendMode() != b->blendMode()) {
-                      return a->blendMode() == Blending::none;
+                      if (a->blendMode() == Blending::none || b->blendMode() == Blending::none) {
+                          return a->blendMode() == Blending::none;
```
blair1618: Tangram-JS compares the style names as a tie-breaker [here](https://github.com/tangrams/tangram/pull/223/files#diff-e90eeb4f0c5a50cd03b669245e7e0724R428); we may as well do the same.

blair1618: Also, this sorting logic could be nicely encapsulated into a static method of `Style`, for legibility and potential re-use.

blair1618: A more compact way to encode this logic might be something like:
```
size_t hash_a = (a->blendMode() << 16) + (a->blendOrder() + 1);
size_t hash_b = (b->blendMode() << 16) + (b->blendOrder() + 1);

return hash_a < hash_b
```

tallytalwar: With respect to tie breaking, that is already happening for us (https://github.com/tangrams/tangram-es/blob/blend-order/core/src/scene/sceneLoader.cpp#L116).

Yes, I agree on the other 2 suggestions. However, will have to use c style enum, instead of c++11 strongly typed enum here, for `<<` operator to work.

tallytalwar:
Once reviewed, will squash last 2 commits, revert the style sheet test commit.

core/src/scene/sceneLoader.cpp
blair1618: Yeah, you can static cast to an integer type but after thinking about it a
bit more, the integer comparison might be a bit more clever than necessary
:P
On Dec 15, 2015 6:45 PM, "Varun" <notifications@github.com> wrote:

In core/src/scene/sceneLoader.cpp
<https://github.com/tangrams/tangram-es/pull/438#discussion_r47714848>:

> @@ -101,7 +101,15 @@ bool SceneLoader::loadScene(Node& config, Scene& _scene) {
>      std::sort(_scene.styles().begin(), _scene.styles().end(),
>                [](std::unique_ptr<Style>& a, std::unique_ptr<Style>& b) {
>                    if (a->blendMode() != b->blendMode()) {
> -                      return a->blendMode() == Blending::none;
> +                      if (a->blendMode() == Blending::none || b->blendMode() == Blending::none) {
> +                          return a->blendMode() == Blending::none;

With respect to tie breaking, that is already happening for us (
https://github.com/tangrams/tangram-es/blob/blend-order/core/src/scene/sceneLoader.cpp#L116
).

Yes, I agree on the other 2 suggestions. However, will have to use c style
enum, instead of c++11 strongly typed enum here, for << operator to work.

—
Reply to this email directly or view it on GitHub
<https://github.com/tangrams/tangram-es/pull/438/files#r47714848>.

tallytalwar: Refer to newer commits. I think your suggestion  definitely makes the code more concise.

blair1618:
Looks good! Those test styles could be reproduced in code to make a good unit test case.

core/src/style/style.h
```diff
@@ -111,9 +112,25 @@ using StyleUniform = std::pair< std::string, UniformValue >;
...
+        auto aBlend = static_cast<int>(a->blendMode());
+        auto bBlend = static_cast<int>(b->blendMode());
+
+        if ( !(aBlend * bBlend) && (aBlend ^ bBlend) ) { // When either is opaque
+            return !aBlend;
+        } else if (aBlend * bBlend) { // All non opaque styles, order takes precedence
+            size_t hash_a = (a->blendOrder() << 16) + aBlend;
+            size_t hash_b = (b->blendOrder() << 16) + bBlend;
+            if (hash_a != hash_b) { return hash_a < hash_b; }
+        }
```
blair1618: The arithmetic operations are quite clever, but they do make this rather difficult to read. Something like this would also work:
```
if (modeA != Blending::none && modeB != Blending::none) {
    if (orderA != orderB) {
        return orderA < orderB;
    }
}
if (modeA != modeB) {
    return static_cast<uint8_t>(modeA) < static_cast<uint8_t>(modeB);
}
return a->getName() < b->getName();
```

PR #439, opened 2015-12-16
Enable posting Runnables to the render thread

blair1618:
This is pretty useful for any render-loop-based application, but in particular this will help with instances where a client application needs to make multiple synchronous changes to the map view. For example, this code might cause stuttering in the view:
```java
mapController.setMapPosition(a.longitude, a.latitude);
// Render might start here...
mapController.setMapRotation(a.bearing);
// ... then render again after producing one frame with the wrong rotation!
```
But this code has no such risk:
```java
mapController.queueEvent(new Runnable() {
    public void run() {
        // This runs on the render thread!
        mapController.setMapPosition(a.longitude, a.latitude);
        mapController.setMapRotation(a.bearing);
    }
});
```

### DIFF
--- a/android/tangram/src/com/mapzen/tangram/MapController.java
+++ b/android/tangram/src/com/mapzen/tangram/MapController.java
@@ -282,6 +282,14 @@ public class MapController implements Renderer, OnTouchListener, OnScaleGestureL
         featureTouchListener = listener;
     }
 
+    /**
+     * Enqueue a Runnable to be executed synchronously on the rendering thread
+     * @param r Runnable to run
+     */
+    public void queueEvent(Runnable r) {
+        mapView.queueEvent(r);
+    }
+
     // Native methods
     // ==============
 
--- a/android/tangram/src/com/mapzen/tangram/MapController.java
+++ b/android/tangram/src/com/mapzen/tangram/MapController.java
@@ -22,7 +22,6 @@ import com.squareup.okhttp.Callback;
 import com.squareup.okhttp.Request;
 import com.squareup.okhttp.Response;
 
-import java.io.File;
 import java.io.IOException;
 
 import javax.microedition.khronos.egl.EGLConfig;


blair1618:
@hjanetzek Does this look alright to you?

tallytalwar:
Nice. Its exactly http://developer.android.com/reference/android/view/View.html#post%28java.lang.Runnable%29 right? Can the View::post not be used from the app?

This is also one of the suggestions here: http://stackoverflow.com/a/8976579

blair1618:
In theory we could use `View.post()` but then the client application needs to have a reference to the `MapView` as well as the `MapController`, the latter of which handles all other interaction. Using our own method also allows us to ensure that `requestRender()` is called so that the `Runnable` is actually run.

hjanetzek:
For posting a Runnable on the render thread an trigger a redraw there is a GLSurfaceView.queueEvent method already http://grepcode.com/file/repository.grepcode.com/java/ext/com.google.android/android/5.1.1_r1/android/opengl/GLSurfaceView.java#GLSurfaceView.GLThread.queueEvent%28java.lang.Runnable%29

Since the native methods of MapController are synchronized on the class itself, I think one could also use:

    synchronized(MapController.class) {
        mapController.setMapPosition(a.longitude, a.latitude);
        mapController.setMapRotation(a.bearing);
    }

blair1618:
True, it seems there's no need to store our own ArrayList of Runnables - we could post directly to that of the GLSurfaceView. It is still a useful convenience to be able to do this through a MapController and not only a MapView. 

I'll check whether synchronization achieves the same result in Eraser Map, but I'm still figuring out how concurrency works in Kotlin :\

blair1618:
Ok, so a `synchronized` block doesn't quite do the trick. The render call itself (`onDrawFrame`) isn't synchronized so it can still occasionally produce incorrect frames. 

I've edited the changes such that runnables are sent directly to the queue in our GLSurface view, and it seems to work nicely!

blair1618:
Any further feedback before merging?

tallytalwar:
This looks good as per my limited Java knowledge.

I would like @ecgreb to also have a look at this to make sure the user of our API acknowledges this change.

blair1618:
Yep, I'll be demonstrating this update with a suggested change to eraser-maps.

PR #441, opened 2015-12-17
Label repeat

karimnaaji:
Refer to https://github.com/tangrams/tangram/pull/222

### DIFF
--- a/core/src/labels/labels.cpp
+++ b/core/src/labels/labels.cpp
@@ -279,7 +279,7 @@ void Labels::update(const View& _view, float _dt,
     }
 
     // Ensure the labels are always treated in the same order in the visible set
-    std::sort(m_visibleTextSet.begin(), m_visibleTextSet.end(), [](TextLabel* _a, TextLabel* _b) {
+    std::sort(visibleSet.begin(), visibleSet.end(), [](TextLabel* _a, TextLabel* _b) {
         return glm::length2(_a->transform().modelPosition1) < glm::length2(_b->transform().modelPosition1);
     });
 
--- a/core/src/labels/labels.cpp
+++ b/core/src/labels/labels.cpp
@@ -5,7 +5,6 @@
 #include "gl/shaderProgram.h"
 #include "gl/primitives.h"
 #include "view/view.h"
-#include "style/material.h"
 #include "style/style.h"
 #include "tile/tile.h"
 #include "tile/tileCache.h"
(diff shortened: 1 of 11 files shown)


hjanetzek:
reviewing!

core/src/labels/labels.cpp
```diff
@@ -118,64 +108,185 @@ void Labels::update(const View& _view, float _dt, const std::vector<std::unique_
...
+            std::vector<GroupElement>& group = repeatGroups[seed];
+
+            if (std::find(group.begin(), group.end(), element) == group.end()) {
+                std::vector<GroupElement> newGroup(group);
+                bool add = true;
+                newGroup.push_back(element);
+                float threshold2 = pow(element.threshold, 2);
+
+                for (size_t i = 0; i < newGroup.size() - 1; ++i) {
+                    const GroupElement& ge = newGroup[i];
```
hjanetzek: Why do you need a copy of group here instead of using range for-loop over group?

core/src/labels/labels.cpp
```diff
@@ -118,64 +108,185 @@ void Labels::update(const View& _view, float _dt, const std::vector<std::unique_
...
                     }
                 }
+
+                if (add) {
+                    group.push_back(element);
+                    textLabelIt++;
+                } else {
+                    textLabel->setOcclusion(true);
+                    _visibleSet.erase(textLabelIt);
+                }
```
hjanetzek: Since _visibleSet is not used afterwards, you could avoid the erase and then use a range for-loop over _visibleSet for the outer loop.

core/src/labels/labels.cpp
```diff
@@ -118,64 +108,185 @@ void Labels::update(const View& _view, float _dt, const std::vector<std::unique_
...
+
+        GroupElement element;
+        element.position = textLabel->transform().state.screenPos;
+        element.threshold = textLabel->options().repeatDistance;
+        element.hash = seed;
+
+        auto it = repeatGroups.find(seed);
+        if (it != repeatGroups.end()) {
+            std::vector<GroupElement>& group = repeatGroups[seed];
+
```
hjanetzek: This should use the found iterator directly, instead of looking up seed again:
std::vector<GroupElement>& group = *it;

tallytalwar:
Can we close this in favor of #465?

core/src/labels/labels.cpp
karimnaaji: :+1:

core/src/labels/labels.cpp
karimnaaji: Right as done in #465 , this saves an erase and a copy of the group.

karimnaaji:
Closing in favor or rebased version.

PR #442, opened 2015-12-17
Check for known-unsupported compilers

blair1618:
The CMake toolchain won't know everything about every compiler, so we'll only make checks for compiler versions that we *know* to be unsupported; any compilers that we don't have checks for will just receive a warning that compilation is untested in their configuration. Known-unsupported compilers are currently:

 - GCC 4.8 or lower
 - Clang 3.3 or lower
 - AppleClang (i.e. Xcode) 5.1 or lower

The README has also been update to reflect this.
 
This also *formally* raises our minimum CMake version to 3.0 (required to correctly identify compilers). To install this version on Ubuntu 12.04 another PPA was needed in the Travis config.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ matrix:
           env: PLATFORM=linux CXX=g++-4.9 CC=gcc-4.9
           addons:
             apt:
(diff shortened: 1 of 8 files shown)


tallytalwar:
Looks good to me. Merging.

PR #444, opened 2015-12-21
fix: use correct outlineWidth when outline needs to be build

hjanetzek:


core/src/style/polylineStyle.cpp
```diff
@@ -205,8 +205,9 @@ void PolylineStyle::buildLine(const Line& _line, const DrawRule& _rule, const Pr
         if (evalStyleParamWidth(StyleParamKey::outline_width, _rule, _tile, widthOutline, dWdZOutline) &&
             ((widthOutline > 0.0f || dWdZOutline > 0.0f)) ) {
 
-            widthOutline += width;
-            dWdZOutline += dWdZ;
+            // Note: this must update width and dWdZ as they are captured (and used) by builder
+            width += widthOutline;
```
tallytalwar: should `width` and `dWdZ` be zeroed-out before this?

And this is only when `Builder::buildPolyLine` is again called for re-triangulation.

hjanetzek: This should be correct - width is added to outlineWidth. Before the sum was only used in the common case where the vertices are just copied and modified for the outline. 
The problem was the the builder instance did not use outlineWidth.

PR #445, opened 2015-12-21
Vertex packing

hjanetzek:


hjanetzek:
Vertices reduced to almost half the size :) In bytes that is: 
```
polygon    48 => 24
line       40 => 24
text/point 48 => 28
```

core/src/style/pointStyle.cpp
```diff
@@ -115,10 +115,16 @@ PointStyle::Parameters PointStyle::applyRule(const DrawRule& _rule, const Proper
...
                           const glm::vec2& _uvTR, unsigned int _color, float _extrudeScale) const {
 
-    _vertices.push_back({{    0.0,       0.0}, {_uvBL.x, _uvTR.y}, {-1.f,  1.f, _extrudeScale}, _color});
-    _vertices.push_back({{_size.x,       0.0}, {_uvTR.x, _uvTR.y}, { 1.f,  1.f, _extrudeScale}, _color});
-    _vertices.push_back({{    0.0,  -_size.y}, {_uvBL.x, _uvBL.y}, {-1.f, -1.f, _extrudeScale}, _color});
-    _vertices.push_back({{_size.x,  -_size.y}, {_uvTR.x, _uvBL.y}, { 1.f, -1.f, _extrudeScale}, _color});
+    float es = _extrudeScale;
+
+    // Attribute will be normalized - scale to max short;
+    glm::vec2 uvTR = _uvTR * 32767.f;
```
tallytalwar: Instead of explicitly using the value, we can use std::numeric_limits<short>::max().

core/src/style/polygonStyle.cpp
```diff
@@ -12,34 +12,38 @@
...
+#define NORMAL_SCALE 127.0f
+
 namespace Tangram {
 
 struct PolygonVertex {
-    glm::vec3 pos;
-    glm::vec3 norm;
-    glm::vec2 texcoord;
+    glm::i16vec4 pos; // pos.w contains layer (params.order)
+    glm::i8vec4 norm;
```
tallytalwar: Why `i8vec4` and not `i8vec3`, am I missing something here?

core/src/style/pointStyle.cpp
hjanetzek: Yea, but not all scale factors use max() values. The factor should have a #defined name though

tallytalwar: Oh yes true. And yes define will work for me. Thanks

core/src/style/polygonStyle.cpp
hjanetzek: Just for padding. I'll change it to have an explicit padding byte

core/src/style/polylineStyle.cpp
```diff
@@ -10,16 +10,36 @@
...
 #include "util/mapProjection.h"
 #include "util/extrude.h"
+
 #include "glm/vec3.hpp"
+#include "glm/gtc/type_precision.hpp"
+
+#define EXTRUSION_SCALE 4096.0f
+#define POSITION_SCALE 4096.0f
+#define TEXTURE_SCALE 16384.0f
+#define ORDER_SCALE 2.0f
```
tallytalwar: Also, don't get ORDER_SCALE, can you elaborate? Thanks

hjanetzek: We put the outline between the usual integer orders: 

    float outlineOrder = std::min(params.outlineOrder, params.order) - 0.5f;

So to keep the modified order, when packed as integer, it needs to be scaled up - and down again in the shader.
 I'm not completely sure about this whole logic either

core/src/labels/label.h
```diff
@@ -42,18 +42,26 @@ class Label {
...
-        glm::vec3 extrude;
+        Vertex(glm::vec2 pos, glm::vec2 uv, glm::vec2 extrude, uint32_t color, uint32_t stroke = 0)
+            : pos(pos * 4.f), uv(uv),
+              extrude(extrude * 256.f),
+              color(color), stroke(stroke) {}
+
+        glm::i16vec2 pos;
+        glm::i16vec2 uv;
+        glm::i16vec2 extrude;
         uint32_t color;
```
tallytalwar: will it better for padding if `color` and `stroke` is ordered before `pos`, `uv' and `extrude`?

hjanetzek: It may be beneficial to start 4byte attributes at 4byte boundaries.. on a second look this is also the case here :)

core/src/labels/label.h
```diff
@@ -42,18 +42,26 @@ class Label {
...
         uint32_t stroke;
         struct State {
-            glm::vec2 screenPos;
-            float alpha = 0.f;
-            float rotation = 0.f;
+            State() {}
+            State(glm::vec2 pos, float alpha, float rotation)
+                : screenPos(pos * 4.f),
+                  alpha(alpha * 32767.f),
+                  rotation(rotation * 4096.f) {}
```
tallytalwar: Can we have `DEFINES` for these also?

tallytalwar:
Small queries left to be answered. Good otherwise. Once answered/fixed, this can be merged.

core/src/labels/label.h
hjanetzek: I didn't want to use defines in a header (as they should be undefined at the end to not leak to the outside). constexpr with anynomous namespace seem to be a good alternative for local defines generally.

core/src/style/pointStyle.cpp
```diff
@@ -15,6 +15,8 @@
 #include "data/propertyItem.h" // Include wherever Properties is used!
 #include "scene/stops.h"
 
+constexpr float texture_scale = static_cast<float>(std::numeric_limits<int16_t>::max());
```
blair1618: Feels a little odd to use a literal for this scale in one place and not in the other (even though I know it's the same value)

hjanetzek: yes, changed this one back to number literals. Hope that's ok with everyone

tallytalwar: Yup, I just don't like magic numbers. But this is fine with me. My initial comment was to get rid of magic number :D.

core/src/style/pointStyle.cpp
```diff
@@ -115,10 +117,16 @@ PointStyle::Parameters PointStyle::applyRule(const DrawRule& _rule, const Proper
...
 
-    _vertices.push_back({{    0.0,       0.0}, {_uvBL.x, _uvTR.y}, {-1.f,  1.f, _extrudeScale}, _color});
-    _vertices.push_back({{_size.x,       0.0}, {_uvTR.x, _uvTR.y}, { 1.f,  1.f, _extrudeScale}, _color});
-    _vertices.push_back({{    0.0,  -_size.y}, {_uvBL.x, _uvBL.y}, {-1.f, -1.f, _extrudeScale}, _color});
-    _vertices.push_back({{_size.x,  -_size.y}, {_uvTR.x, _uvBL.y}, { 1.f, -1.f, _extrudeScale}, _color});
+    float es = _extrudeScale;
+
+    // Attribute will be normalized - scale to max short;
+    glm::vec2 uvTR = _uvTR * texture_scale;
+    glm::vec2 uvBL = _uvBL * texture_scale;
```
blair1618: UV wrapping behavior may differ here. Before, and input UV of 1.5 would result in sampling a texture at 1.5 (and then be either clamped or wrapped). Now, an input UV of 1.5 would result in sampling a texture at 1.0. I don't think we currently have any need to generate UVs that wrap, but this is a potential point of future confusion.

hjanetzek: yes, though for point sprites I find it hard to imagine for what wrapping uv coordinates could be useful. For anyone who wants to implement a point style that relies on wrapped uv it should also be easy to figure out what's wrong with the texture coordinates :)

blair1618: I believe this would be the case in every style that uses uv coordinates, not just the Point style. However I agree that this doesn't seem like a practical limitation right now.

blair1618: Along those lines: Why are uv attributes normalized in these vertices but not in those for other styles? The uv attributes seem to behave the same in every style here so it's a bit confusing to treat them differently. Am I misunderstanding something?

hjanetzek: Right, uv attributes for point, line and polygon styles could be treated the same at the moment. The text style is different. There we need to do the mapping from pixel to 0..1 in the shader since the texture size changes when the glyph atlas grows.

Do you prefer a custom scale factor as it's done for lines and polys (which provides enough precision for the largest textures and would allow proper wrapping) or use the normalize-on-the-go option of point style?

blair1618:
As expected, we've now run into the problem that the webgl renderer has had for a while with the Highline. At high zooms, features whose vertices extend several tile-lengths beyond a tile will end up doing silly things.  
screen shot 2016-01-07 at 3 24 21 pm
@bcamper I can't remember, did you settle on a fix for this in tangram-js?

tallytalwar:
Hmm, I am not sure its addressed yet: https://trello.com/c/lr1SrM6a/633-rogue-high-line-geo

hjanetzek:
One simple workaround that might work would be to decrease the precision until highline fits into the available range. As we don't use over-zoom tiles (except proxies) the current precision of 8192 points within a tile is quite a lot.

In long term we should clip the geometries of features extending way beyond tile boundaries

core/src/style/pointStyle.cpp
blair1618: I don't see why the text style should treat UVs differently. The text fragment shader has to scale the texture coordinates by a factor that changes when the atlas resizes, but every step before that can be the same, no?

hjanetzek: The difference is that for the text style the scaling *must* be done in the shader as the scaling factor can change. For the point style we can request to do the normalization automatically for the vertex attribute which could be slightly more efficient.

blair1618: Maybe what I'm really asking is: why isn't automatic normalization applied for texture coordinates in *all* styles?

bcamper:
@nvkelso @zerebubuth this issue with the High Line and needing to clip features server-side rears its head again (see screen above) :) I thought there might be an issue for this somewhere but I can't find it now. I know we've also discussed potentially clipping all buildings to the tile boundaries, but this has greater ramifications, so could we just clip within a +/- 1 tile border for now (so like within a 3x3 tile box)? That will fix this issue in both Tangram JS and ES and then we can address clipping policy further in the future.

core/src/style/pointStyle.cpp
hjanetzek: It would just be superfluous to use it for text as the scaling must be done in the shader anyway - so it would just be an additional multiplication when using automatic normalization (I assume that for automatic normalization the values are transformed during upload to a float representation). One may also loose a bit of precision during the back and forth transformation

If we decide that uv coordinates must be in the range -1..1 we can use normalization for the other styles

nvkelso:
@bcamper the vector tiles issue is: https://github.com/mapzen/vector-datasource/issues/197

core/src/style/pointStyle.cpp
blair1618: Well, texture coordinates would naturally use `GL_UNSIGNED_SHORT` since [0, 1] is the range we'd like.

hjanetzek: And what about wrapping or procedural textures? :)

What would be your proposal? I'm ok with using normalization and unsigned shorts. Just for text I would prefer to have it as it is currently done.

blair1618: The packing of texture coordinates in polylines and polygons here only allows a UV range of [-2.0, 2.0], so it's not as though this approach really allows wrapping the way floating point UVs would. 

What I forgot though is that texture coordinates for text atlases are in *actual texels* 0_0 So now I understand why it makes sense to treat them differently.

blair1618:
Ok, last comment I promise :)

When attributes are unpacked in shaders, it might be more clear to use a macro, e.g.
```
#define UNPACK_POSITION(x) (x / 4096.0)
```

blair1618:
Looking good to me, and Travis is behaving nicely again! Merge when ready :)

PR #446, opened 2015-12-22
android: remove superfluous synchronized modifier

hjanetzek:
onUrlSuccess/Failure call back into a thread-safe method. No need for additional syncs with the rest of MapController. 

This seems also to fix most of the jagged fling animations :)


blair1618:
Good find

PR #449, opened 2015-12-24
Tilemanager testing

hjanetzek:


core/src/tile/tileManager.cpp
```diff
@@ -473,11 +460,12 @@ void TileManager::updateProxyTiles(TileSet& _tileSet, const TileID& _tileID, Til
         return;
     }
     // Try children
-    if (m_view->s_maxZoom > _tileID.z) {
-        for (int i = 0; i < 4; i++) {
-            updateProxyTile(_tileSet, _tile, _tileID.getChild(i), static_cast<ProxyID>(1 << i));
-        }
+    // FIXME maxZoom also depends on DataSource
```
tallytalwar: Yup, good catch, view maxZoom and dataSource maxZoom can be different. A legit FIXME :D.

tests/unit/tileManagerTests.cpp
```diff
@@ -0,0 +1,250 @@
...
+    tileManager.updateTileSets(viewState, visibleTiles);
+
+    REQUIRE(tileManager.getVisibleTiles().size() == 1);
+    REQUIRE(source->tileTaskCount == 1);
+    REQUIRE(worker.processedCount == 1);
+
+}
+
+
+TEST_CASE( "Use proxy Tile", "[TileManager][updateTileSets]" ) {
```
tallytalwar: Can we add a test case for use of grand parent proxy?

hjanetzek: Sure! This is just a start to make TileManager testable. Optimally we should add tests for all the logic we currently expect and then for every change we should first add some tests.

tallytalwar: Agreed!

tallytalwar:
I am good with this also. Ofcourse we can have more tests for tilemanager functionality,  but this is a start. Once we are all green from Travis, can merge this. @blair1618 you should also have a look at this. 

Thanks

hjanetzek:
The test already already catched a regression in https://github.com/tangrams/tangram-es/commit/b3d7c8996519607bf4778cac6988675e28a34c04

I'm working on it

Issue #447, opened 2015-12-23
"make clean" shouldn't download things it doesn't need to


msmollin:
If you run make clean, it will go through the hoops of downloading and installing gradle and a bunch of apache projects in an attempt to clean a nonexistent android build (in my case, I had run `make xcode`). See attached screenshot.

![screen shot 2015-12-23 at 11 52 40 am](https://cloud.githubusercontent.com/assets/1163880/11980830/b67bfc7a-a96b-11e5-9b3c-9a7ed34f88a6.png)

tallytalwar:
Ahaa... so `make clean` is a uber clean. If you have had done `make android`, `make osx`, `make xcode`, etc... `make clean` takes care of all.

For `xcode` you can do `make clean-xcode`.

msmollin:
I see. Is there a reason for it? Other than it's expected of projects using the CMake build system to have a "clean" step? Just seems like the amount of side-effects involved with the clean step is out of line with what clean is supposed to do, i.e. remove files and dependencies - not install new & unnecessary ones?

msmollin:
Specifically with gradle, installing a new, very old gradle build chain (latest version is 2.10 - this is installing 2.2.1) has the distinct possibility of breaking any existing gradle configurations. Making an assumption that ~/.gradle/ is unused and available to be overwritten is a Bad Idea™ (regardless of whether or not the installer will actually do that).

tallytalwar:
Hmm.. for all other targets, clean just does a `rm -rf` on the respective build directory.

But for `android`, it also runs a `gradlew clean` which takes care of cleaning the android project.

tallytalwar:
Ohh, didn't know its installing an old version, I am not sure where thats happening, if you refer the makefile, it only does a `./gradlew clean`. [here](https://github.com/tangrams/tangram-es/blob/master/Makefile#L154-L158)

msmollin:
That's about where my knowledge of gradle ends :) No idea why it would attempt an install then. I would just expect that command to fail. @ecgreb may have a better idea. But yeah, 2.2.1 is from November of 2014.

tallytalwar:
Probably this is the reason: https://docs.gradle.org/current/userguide/gradle_wrapper.html.

tallytalwar:
Based on the gradle wrapper documentation, probably we should update our `build.gradle`, with:

```
task wrapper(type: Wrapper) {
    gradleVersion = '2.0'
}
```

blair1618:
[...]

PR #450, opened 2015-12-27
Clarify Makefile default tasks

blair1618:
Addresses https://github.com/tangrams/tangram-es/issues/447

`make clean` will currently target all platforms, which includes the Gradle steps for Android, leading to confusion for new users when Gradle starts installing and running. 

This changes `make clean` to be a no-op that simply lists the likely tasks that the user *actually* wants to run. Similarly, `make` now lists all of the platform targets instead of going straight into the android build.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,17 @@
+list:
+	@echo "Specify a target to build with one of the following tasks:"
+	@echo "  android"
+	@echo "  osx"
+	@echo "  xcode"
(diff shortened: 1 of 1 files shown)


tallytalwar:
I am fine with this, because we never build/clean all target at once.

blair1618:
Closed in favor of https://github.com/tangrams/tangram-es/pull/452

PR #451, opened 2015-12-29
Client data source uses max zoom set by View

blair1618:
Fixes: https://github.com/mapzen/eraser-map/issues/221


tallytalwar:
:+1:

PR #452, opened 2015-12-29
Clean Android build files without invoking Gradle

blair1618:
2nd attempt to address https://github.com/tangrams/tangram-es/issues/447

With this change, `make clean` will still remove build files for all target platforms, but the sub-task `clean-android` now has no Gradle invocation; instead the Makefile manually removes the same directories that `gradle clean` would remove. Consequently, running `make clean` will no longer require a Gradle installation.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -80,9 +80,9 @@ else
 	ANDROID_TOOLCHAIN = arm-linux-androideabi-clang3.6
 endif
 
-#$(info ANDROID_ARCH is ${ANDROID_ARCH}) 
-#$(info ANDROID_TOOLCHAIN is ${ANDROID_TOOLCHAIN}) 
-#$(info ANDROID_BUILD_DIR is ${ANDROID_BUILD_DIR}) 
+#$(info ANDROID_ARCH is ${ANDROID_ARCH})
+#$(info ANDROID_TOOLCHAIN is ${ANDROID_TOOLCHAIN})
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -152,10 +152,12 @@ LINUX_CMAKE_PARAMS = \
...
 
 clean-android:
-	@cd android/ && \
-	./gradlew clean
 	rm -rf ${ANDROID_BUILD_DIR}
+	rm -rf android/build
+	rm -rf android/tangram/build
 	rm -rf android/tangram/libs
+	rm -rf android/demo/build
+	rm -rf android/demo/libs
```
tallytalwar: We should also remove files created for `android-debug` target.

I think the following:
```
android/demo/jni
android/*/libs
android/*/obj
*.mk
```

I do not have a device, which can run gdb (samsung galaxy s5 :(), so can not test this. @blair1618 your personal phone however can be used for debugging, can you check and update?

Thanks

blair1618: Turns out the CMake configuration for Android native debugging is broken now that we've upgraded to 3.0 (due to this policy https://cmake.org/cmake/help/v3.3/policy/CMP0026.html). 

Since we effectively don't have an android-debug target right now, I'm going to leave that as a separate issue and merge this as-is.

tallytalwar: Sure, can you make an issue for the same? Thanks

PR #454, opened 2016-01-05
Refactor touch input management

blair1618:
This is an overhaul of the code that responds to input events on all platforms, with a particular focus on Android touch input. 

 - Android input is now handled by the new class ~~`TouchManager`~~ `TouchInput`
 - Gesture exclusivity can be managed at run-time via `TouchInput` (i.e. we can freely set which gestures are allowed to be recognized while another gesture is in progress)
 - Callbacks can be set for all available gesture types via `MapController`
 - A logic error in the calculation of scaling "fling" has been fixed
 - Input units for "shove" are now consistent with other input functions
 - 3rd-party gesture recognizer code is now consistently styled
 - Pan flinging is now handled by an explicit callback at the end of a pan gesture, instead of being inferred
 - Flings now correctly interpolate with the time step between frames

This touches many lines of code, but the result is quite a major improvement in code quality and input behavior. 

The most important feedback for this is comments on "feel". Try out every combination of gestures you can think of and see if there are any edge cases that feel wrong.

### DIFF
--- a/android/tangram/src/com/mapzen/tangram/MapController.java
+++ b/android/tangram/src/com/mapzen/tangram/MapController.java
@@ -5,7 +5,6 @@ import android.content.res.AssetManager;
 import android.opengl.GLSurfaceView;
 import android.opengl.GLSurfaceView.Renderer;
 import android.util.DisplayMetrics;
-import android.view.View;
 
 import com.squareup.okhttp.Callback;
 import com.squareup.okhttp.Request;
@@ -18,7 +17,7 @@ import javax.microedition.khronos.opengles.GL10;
 
 import okio.BufferedSource;
 
-import com.mapzen.tangram.TouchManager.Gestures;
+import com.mapzen.tangram.TouchInput.Gestures;
 
 public class MapController implements Renderer {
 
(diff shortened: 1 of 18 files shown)


blair1618:
@tallytalwar @hjanetzek Please review but do not merge yet!

This is an API-breaking change for eraser-map. I will comment here again after I submit a PR with the necessary changes to the eraser-map repo.

hjanetzek:
Feels quite good and TouchManager looks well designed! 

One thing I've just noticed is that rotation is always around screen center which might not be intended. Though otherwise if the view is tilted the change is a bit abrupt (does too much movement) when rotating in the upper screen area in my opinion.

The conflict between the animated move on tap and input handling needs to be resolved, but this belongs not to this task I guess.

blair1618:
Ah, I see now that rotation *is* always around the screen center, which is a regression. I'll look into that.

Is the conflict you refer to the fact that animated movement disallows manual map movement? I would agree that addressing this is better left to another PR; it's a more isolated issue with different design implications.

blair1618:
Seems that the rotate-around-center behavior, though indeed undesired, is actually _not_ a regression from the current build. I suggest that this be addressed in a subsequent PR so that this one is not further complicated, let's track it as https://github.com/tangrams/tangram-es/issues/456

blair1618:
https://github.com/mapzen/eraser-map/pull/232 is now ready to update eraser-map for these changes.

hjanetzek:
I'm ok with merging. @tallytalwar ?

tallytalwar:
Have not got a chance to check this out. Will try to get to this today.

core/src/util/inputHandler.cpp
```diff
@@ -54,20 +71,36 @@ void InputHandler::handlePanGesture(float _startX, float _startY, float _endX, f
...
+}
+
+void InputHandler::handleFlingGesture(float _posX, float _posY, float _velocityX, float _velocityY) {
+
+    if (glm::length(glm::vec2(_velocityX, _velocityY)) / m_view->pixelScale() <= THRESHOLD_START_PAN) {
+        return;
     }
 
-    m_view->translate(dx, dy);
+    const static float epsilon = 0.0167f;
```
tallytalwar: I would call it delta (like time delta).

tallytalwar:
Seems we never had flinging for rotation gesture, should be good to add, but as part of another PR.

android/tangram/src/com/mapzen/tangram/MapController.java
```diff
@@ -75,16 +66,20 @@ public MapController(Activity mainApp, MapView view, String sceneFilePath) {
...
-        gestureDetector.setOnDoubleTapListener(this);
+        touchManager = new TouchManager(mainApp);
+        setPanResponder(null);
+        setScaleResponder(null);
+        setRotateResponder(null);
+        setShoveResponder(null);
+
+        touchManager.setSimultaneousDetectionAllowed(Gestures.SHOVE, Gestures.ROTATE, false);
+        touchManager.setSimultaneousDetectionAllowed(Gestures.ROTATE, Gestures.SHOVE, false);
+        touchManager.setSimultaneousDetectionAllowed(Gestures.SHOVE, Gestures.SCALE, false);
```
tallytalwar: `SHOVE` and `SCALE` should both be mutually exclusive.

Right now first gesture as `SCALE` and second gesture `SHOVE` is allowed, which according to me should not be.

tallytalwar: Another thing I noticed was panning was not allowed post long press gesture. I think we can do panning and longpress together! (Refer gmaps)

tallytalwar: And another thing, `LONG_PRESS` gesture should be disallowed after `SCALE` gesture has started.

blair1618:
As we're thinking more specifically about the interactions of gestures, another thing to consider is whether simultaneous gesture recognition should be exposed for app developers to configure. Exposing all of `TouchManager` would probably make it too easy to break things, but exposing just the exclusivity rules might be a nice option for an SDK.

tallytalwar:
Very nice implementation of `TouchManager` class. Good to be merged once couple of issues mentioned in the comments are addressed. Good work. Thanks

tallytalwar:
@blair1618 Agree on exposing simultaneous gesture recognition API, application should be able to override the defaults in the MapController.

Issue #411, opened 2015-12-01
Building feature at 2 zoom levels shown at once


nvkelso:
This is [McKinleyville High School](https://www.openstreetmap.org/#map=17/40.95529/-124.10899) in McKinleyville California. I only expected geometry from a single zoom level to be visible at once.

![screenshot_2015-11-28-20-22-14](https://cloud.githubusercontent.com/assets/853051/11494489/b9004f48-97b7-11e5-9118-04e27f7bfe9e.png)

hjanetzek:
This may happen when a parent tile is still drawn while some of its children (the tiles matching the current zoom-level) are already loaded. The coarser pattern in the upper-right corner seems to still use the parent tile here. If the view remains as on the screenshot it could be that the tile is missing or failed to load

tallytalwar:
@hjanetzek correctly described this behavior.

We have something called as `proxy` tiles in es, which are basically already loaded (parent or children) tiles from last update. 
These proxy tiles are drawn till the new actual visible tiles (respective children or parent tile for these proxies) are loaded and ready to be drawn.

Possible in the above instance, one of the children tile fail to load and you were seeing the parent tile along with other 3 children tiles, hence mixed geometries.

bcamper:
But my understanding of desired proxy tile behavior is that the proxy
should be drawn underneath, so that the desired zoom level is always shown.
They should not be mixed.

nvkelso:
That's also my expected behavior (proxy tiles drawn under).

tallytalwar:
Yup proxy tiles are drawn under. Parent proxy tile is only visible for the not loaded child tile. Rest all loaded children tile should have overlapping proxy parent.

hjanetzek:
Ah, yes, the proxy building should be drawn below the earth polygon of the current tile - but to enforce this we would need to know offset range to sweep the proxy under the carpet :) I guess for now we need to find a working building 'order' and should not tweak position.z in the custom building shader

tallytalwar:
Proxy tiles are drawn deeper in depth buffer based on their z id.
https://github.com/tangrams/tangram-es/blob/master/core/resources/shaders/polygon.vs#L98-L100

I am checking as to why in this specific case, it appears that proxy tiles are drawn above visible tile set, and hence overlapping geometry.

PR #458, opened 2016-01-08
Stencil proxy

tallytalwar:
Stencil Buffer usage to draw proxy tiles:

Approach:

1. Draw Non-Proxy Tiles
 - Enable stencil writing and draw all non-proxy tiles marking stencil fragments to 1

2. Draw Proxy Tiles
 - Pass 1:
   - Disable color buffer writing
    - enable stencil writes and depth tests
    - set stencil to invert when depth and stencil test pass.
    - enable GL_POLYGON_OFFSET_FILL
    - set glPolygonOffset (this is required to avoid z fighting between same depth proxy and non-proxy fragments)
 - Pass 2: Enable color buffer writing, disable stencil writes and draw proxy tiles with stencil tests enabled

Issues to work on:
- [x] Text and Point/Icon style drawing for proxy tiles. (Because these have their own z depth).
- [x] Update RenderStates with new gl functions.
- [ ] Possible better values for `glPolygonOffset` for 16bit depth buffer or another approach?

### DIFF
--- a/core/src/tangram.cpp
+++ b/core/src/tangram.cpp
@@ -172,14 +172,21 @@ void render() {
 
     // Set up openGL for new frame
     RenderState::depthWrite(GL_TRUE);
+    RenderState::stencilTest(GL_TRUE);
+    RenderState::stencilWrite(0xFF);
     auto& color = m_scene->background();
     RenderState::clearColor(color.r / 255.f, color.g / 255.f, color.b / 255.f, color.a / 255.f);
-    glClear(GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT);
+    glClear(GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT | GL_STENCIL_BUFFER_BIT);
 
     {
         std::lock_guard<std::mutex> lock(m_tilesMutex);
 
         // Loop over all styles
+        // 1. Draw all non proxy tiles, and filling the stencil buffer
+        RenderState::stencilTest(GL_TRUE);
+        RenderState::stencilWrite(0xFF);
+        RenderState::stencilOp(GL_KEEP, GL_KEEP, GL_REPLACE);
+
         for (const auto& style : m_scene->styles()) {
 
             // Set time uniforms style's shader programs
@@ -187,11 +194,62 @@ void render() {
 
             style->onBeginDrawFrame(*m_view, *m_scene);
 
-            // Loop over all tiles in m_tileSet
             for (const auto& tile : m_tileManager->getVisibleTiles()) {
-                tile->draw(*style, *m_view);
-            }
 
+                const auto& tileID = tile->getID();
+
+                if (tile->getProxyCounter() == 0) {
+                    RenderState::stencilFunc(GL_ALWAYS, 1, 0xFF);
+                    tile->draw(*style, *m_view);
+                }
+            }
+            style->onEndDrawFrame();
+        }
+
+        // 2. Disable Color buffer writting and invert stencil value wherever depth from proxy wins
+        RenderState::colorWrite(GL_FALSE, GL_FALSE, GL_FALSE, GL_FALSE);
+        RenderState::stencilTest(GL_TRUE);
+        RenderState::stencilWrite(0xFF);
+        RenderState::stencilOp(GL_KEEP, GL_KEEP, GL_INVERT);
+
+        glEnable(GL_POLYGON_OFFSET_FILL);
+        glPolygonOffset(1.0, 1.0);
+
+        for (const auto& style : m_scene->styles()) {
+
+            // Set time uniforms style's shader programs
+            style->getShaderProgram()->setUniformf("u_time", g_time);
+            style->onBeginDrawFrame(*m_view, *m_scene);
+
+            for (const auto& tile : m_tileManager->getVisibleTiles()) {
+                const auto& tileID = tile->getID();
+                if (tile->getProxyCounter() > 0) {
+                    RenderState::stencilFunc(GL_EQUAL, 1, 0xFF);
+                    tile->draw(*style, *m_view);
+                }
+            }
+            style->onEndDrawFrame();
+        }
+
+        // 3. Draw all proxy tiles, with stencil check and not no writes to stencil buffer
+        RenderState::colorWrite(GL_TRUE, GL_TRUE, GL_TRUE, GL_TRUE);
+        RenderState::stencilTest(GL_TRUE);
+        RenderState::stencilWrite(0x00);
+        glDisable(GL_POLYGON_OFFSET_FILL);
+
+        for (const auto& style : m_scene->styles()) {
+
+            // Set time uniforms style's shader programs
+            style->getShaderProgram()->setUniformf("u_time", g_time);
+            style->onBeginDrawFrame(*m_view, *m_scene);
+
+            for (const auto& tile : m_tileManager->getVisibleTiles()) {
+                const auto& tileID = tile->getID();
+                if (tile->getProxyCounter() > 0) {
+                    RenderState::stencilFunc(GL_NOTEQUAL, 1, 0xFF);
+                    tile->draw(*style, *m_view);
+                }
+            }
             style->onEndDrawFrame();
         }
     }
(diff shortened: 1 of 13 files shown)


blair1618:
Now that you've started using polygon offset, I wonder.. Can we just

1. Render non-proxy tiles
2. Set glPolygonOffset to add a small depth
3. Render proxy tiles

?

tallytalwar:
Nopes that will give the same effect (https://github.com/tangrams/tangram-es/issues/411), where the buildings from both proxy and non-proxy tiles will be visible.

blair1618:
Ah right, I guess the trick is that we can't just offset polygons by a small delta because depth is already being modified by 'order'. 

Does the approach here resolve the case in #411 correctly? It seems like we would end up with the same problem because fragments from the proxy building geometry would pass a depth test against the non-building geometry.

tallytalwar:
I have not tested the school case in #411. But based on the logic used in the drawing, it should not draw buildings from both the tiles (one of the building from the 2 tiles (proxy and non-proxy) should/will win).

tallytalwar:
With respect to #411, I do not seem to notice the issue, with this PR.

tallytalwar:
Its more discussion than review. If discussion is all positive we can merge.

karimnaaji:
I'm not seeing the right behavior here. A good test case is slowing down the tile worker, did you get it to work as you expect?

core/src/tile/tile.cpp
```diff
@@ -135,7 +135,7 @@ void Tile::draw(const Style& _style, const View& _view) {
         float zoomAndProxy = m_proxyCounter > 0 ? -m_id.z : m_id.z;
 
         shader->setUniformMatrix4f("u_model", m_modelMatrix);
-        shader->setUniformf("u_tile_origin", m_tileOrigin.x, m_tileOrigin.y, zoomAndProxy);
+        shader->setUniformf("u_tile_origin", m_tileOrigin.x, m_tileOrigin.y, m_id.z);
```
karimnaaji: Can we remove this last param usage from the shader for no confusion with previous technique.

tallytalwar: I did some git history cleaning yesterday. Will test it again today and
confirm. Thanks

core/src/tangram.cpp
```diff
@@ -187,11 +194,65 @@ void render() {
...
+            style->onEndDrawFrame();
+        }
+
+        // 2. Disable Color buffer writting and invert stencil value wherever depth from proxy wins
+        RenderState::colorWrite(GL_FALSE, GL_FALSE, GL_FALSE, GL_FALSE);
+        RenderState::stencilTest(GL_TRUE);
+        RenderState::stencilWrite(0xFF);
+        RenderState::stencilOp(GL_KEEP, GL_KEEP, GL_INVERT);
+        RenderState::polygonOffsetFill(GL_TRUE);
+        RenderState::polygonOffset(1.0, 1024.0);
```
karimnaaji: Can you explain the value `1024` for the units param; I'm not sure to understand. It seems that `RenderState::polygonOffset(1.0, 1.0);` should be fine for all depth slope (polygon facing and not facing the camera) refering to http://www.glprogramming.com/red/chapter06.html#name4.

hjanetzek:
As I understand it, this approach effectively only changes the use of custom depth modification via projection matrix to the use of polygon offset. (And uses a larger value for offsetting) 
If the gist of the rendering passes is:
 - pass 1: write original depth, set stencil mask to 1 for each drawn fragment (of non-proxy tiles)

proxy tiles:
- pass 2: set stencil mask to 0 where stencil mask is 1 for fragments that pass the depth test (offsetting incoming fragments with polygon_offset)
- pass 3: draw to color buffer where stencil mask is 0

I wonder why is it necessary to split pass 2 and 3. How is this different from directly drawing where the depth test passes?

The problem still remains that polygon offset must be chosen large enough to hide all flat-ground features of proxy tiles below the ground (earth-layer) of non-proxy tiles. So I would think #411 is not resolved yet.

For the general case we also need to handle overlapping proxy tiles from multiple zoom-levels. But that for later..

karimnaaji:
@hjanetzek  (correct me if I'm wrong @tallytalwar), but he intent of pass 2 and pass 3 is what we were dealing with with @tallytalwar on our tests, the blue part in the screenshot:
screen shot 2016-01-12 at 2 21 54 pm
Where the building should be discarded even if coming from a non-proxy tile, for fragment where stencil test failed (so flipping the bits will put 0s in stencil buffer, and in pass 3 continue after stencil test and let the depth test fail, so proxy would be drawn in blue area).

core/src/tangram.cpp
tallytalwar: Yup, I went with 1.0 for both `slope` and `units` parameter. But it was working well for a 16bit depth buffer on my s5. 1024 was chosen as it was working best for a 16bit depth buffer (as well for a 32bit depth buffer on osx/linux). I am open to suggestions here though.

tallytalwar:
@karimnaaji you are correct in explaining the usage of 2 passes to render the proxy tiles.

hjanetzek:
This looks like the proxy was drawn with just checking the stencil mask (for where no stencil bit is set).

As I see it, the second pass does solely unset the stencil bit for fragments where the proxy tile passes the depth test, in order to be able to draw the proxy tile on top of the previous tiles (in the third pass).

So if the only factor that determines whether the proxy tile will be drawn on top of non-proxies is passing of the depth test in the second pass - then I don't see the difference of directly using polygon-offset for proxy tiles (with the usual depth testing) in a single pass.

tallytalwar:
You get some z-fighting when you do not have polygon offset, specially when fragments from both proxy and non-proxy are at same depth. glPolygonOffset is used to get rid of such cases.

hjanetzek:
yes, but one could use polygon-offset to draw proxy tiles in the first (and single) pass likewise

core/src/tangram.cpp
karimnaaji: :+1:

tallytalwar:
Also stencil test is done before depth testing in the opengl pipeline. If we do not do the passes separately, it will simply discard the fragments failing stencil test (potentially having lower depth than non-proxy fragments) without even doing the depth test. (Hence the image from @karimnaaji comment).

hjanetzek:
How is the order of test relevant when the stencil op in the second pass requires both tests to pass to have an effect?

tallytalwar:
@hjanetzek, do you mean this? https://github.com/tangrams/tangram-es/pull/458/files#diff-7681bca1a2db71358101887f729b7b1dR213

`glStencilOp` only tells on what condition stencil buffer is written.

wrt stencil testing (glStencilFunc), https://github.com/tangrams/tangram-es/pull/458/files#diff-7681bca1a2db71358101887f729b7b1dR228, is only set to pass when the value of the stencil buffer is equal to 1.

blair1618:
I too am not clear on how the stencil operations are making any difference here. The only fragments that are culled by the stencil test are those that did not pass a depth test against present geometry, and those fragments would be culled by a depth test anyway. Is this not the case?

karimnaaji:
@tallytalwar , the current issue I'm seeing is on the video, it's a 100% reproducible bug; it happens sometimes and everything is cleared in between the zooms.
[stencil-rec.mov.zip](https://github.com/tangrams/tangram-es/files/87559/stencil-rec.mov.zip)

tallytalwar:
Had a chat with @blair1618 about this.

The confusing behavior we are discussing here is (why use stencil buffer at all if, depth test is checked with proxy and non-proxy tiles), an alternative approach being to:

```
1. Draw non-proxy tiles
2. polygon offset
3. Draw proxy tiles
```

Which conceptually seems to be the same thing as being done with the stencil buffer. However on trying the same, it results in slight different behavior (more z-fighting). I will try to reason on what the differences are which results in different behavior.

tallytalwar:
@karimnaaji I have not been able to reproduce the issue you have in the video, can anyone else repro it?

blair1618:
@tallytalwar Yep that's the essence of my comment earlier this morning.

tallytalwar:
Yup, as mentioned, stencil approach in this PR and doing just polygonOffset for proxy tiles, have the same result (practically and conceptually).

Closing this PR, until we find a better approach to do the same.

tallytalwar:
Does anyone has any other suggestions for this?

Issue #460, opened 2016-01-11
Text halo collisions for POI labels


nvkelso:
Noticed this in Eureka, California but present just about everywhere.

Looks like each character's text halo is being calculated sequentially instead of all drawing at once before the text fill pass.

`yXyXyX` where `X` is fill, `y` is halo and the halo draws first, then the fill. Instead it' usually `yyy` then `XXX`.

It's not as bad for **not bold** text, letter spacing of bold text seems off, which is compounding this?

![screenshot_2016-01-11-12-49-02](https://cloud.githubusercontent.com/assets/853051/12246287/9517bf78-b862-11e5-9192-12fb3f854594.png)
![screenshot_2016-01-11-12-49-17](https://cloud.githubusercontent.com/assets/853051/12246289/95393248-b862-11e5-9ce5-a35c51a20424.png)
![screenshot_2016-01-11-12-49-34](https://cloud.githubusercontent.com/assets/853051/12246288/95342a64-b862-11e5-9aab-f8ec35b3f9f0.png)

blair1618:
Resolved: https://github.com/tangrams/tangram-es/pull/463

PR #463, opened 2016-01-12
two-pass label shader

hjanetzek:
Draw fill and stroke in separate passes to prevent drawing strokes over other glyphs fill.

Should resolve issue https://github.com/tangrams/tangram-es/issues/460

### DIFF
--- a/core/resources/shaders/sdf.fs
+++ b/core/resources/shaders/sdf.fs
@@ -26,6 +26,7 @@ uniform vec2 u_resolution;
 uniform float u_time;
 uniform float u_meters_per_pixel;
 uniform float u_device_pixel_ratio;
+uniform int u_pass;
 
 #pragma tangram: uniforms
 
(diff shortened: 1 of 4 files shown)


karimnaaji:
This fixes the issue, we can ask for a two pass only when the style has stroke params on.

core/resources/shaders/sdf.fs
```diff
@@ -68,16 +69,20 @@ float sampleAlpha(in vec2 uv, float distance, float threshold) {
...
-
-    float alpha_fill = pow(sampleAlpha(v_texcoords, distance, threshold_fill), 0.4545);
-    float alpha_stroke = pow(sampleAlpha(v_texcoords, distance, threshold_stroke), 0.4545);
-
-    vec4 color = mix(v_strokeColor, v_color, alpha_fill);
-    color.a = max(alpha_fill, alpha_stroke) * v_alpha * v_color.a;
+    vec4 color;
+
+    if (u_pass == 1) {
+      float threshold_fill = 0.5;
```
karimnaaji: nitpick: can we keep the 4 spaces tabs.

hjanetzek:
@karimnaaji do you mean to check whether any of the added labels has stroke?

karimnaaji:
Yeah exactly what's done with https://github.com/tangrams/tangram-es/commit/0e9e57ad2824cd43d33f14b61316c054947d5c9d

core/resources/shaders/sdf.fs
```diff
@@ -68,16 +69,20 @@ float sampleAlpha(in vec2 uv, float distance, float threshold) {
...
     float distance = texture2D(u_tex, v_texcoords).a;
-
-    float alpha_fill = pow(sampleAlpha(v_texcoords, distance, threshold_fill), 0.4545);
-    float alpha_stroke = pow(sampleAlpha(v_texcoords, distance, threshold_stroke), 0.4545);
-
-    vec4 color = mix(v_strokeColor, v_color, alpha_fill);
-    color.a = max(alpha_fill, alpha_stroke) * v_alpha * v_color.a;
+    vec4 color;
+
+    if (u_pass == 1) {
```
bcamper: I know we're not filling a lot of fragments with SDF shaders, but do we know the trade-off in a branching shader vs. compiling a separate shader for each pass?

hjanetzek: Branching on uniforms is quite cheap as far as I know, as it should be easy to predict which branch to take. Both branches must only be computed when both can be reached within a shader unit (those that are processing e.g. 16x16 pixel in parallel with one process counter). That's just guessing though, but may the reason why apple states that branching on uniforms is acceptable: https://developer.apple.com/library/ios/documentation/3DDrawing/Conceptual/OpenGLES_ProgrammingGuide/BestPracticesforShaders/BestPracticesforShaders.html

I'll see if it is possible to move the branching to the vertex shader

bcamper: Thanks, just curious.

blair1618:
Nice, that was :zap: fast.

One very very minor optimization if it makes sense to you: We can prevent text without stroke from changing any uniforms. Let `u_pass == 0` correspond to the fill pass, then if `m_strokePass` we can set `u_pass` to 1, draw, then reset it to 0. 

Draw calls on text buffers are actually quite an expensive part of our scenes right now, presumably because the buffers contain all occluded labels as well as the visible ones. Whatever the cause, text draw calls consume upwards of 70% of render time on low-end devices, so optimizations here are not premature.

hjanetzek:
@blair1618, yes, the uniform optimization makes sense.

The number of discarded labels that are sent with each draw call are also my concern. Generally I would propose to use a single mesh (possibly double buffered) to collect all active label vertices. The updates could also be done first in a render-loop iteration so that upload happens during static meshes are drawn.

blair1618:
Cool, this is looking good to me. I'll merge in an hour if there are no further comments.

PR #465, opened 2016-01-13
rebased: label repeat

hjanetzek:


nvkelso:
:+1:

core/src/labels/label.h
```diff
@@ -167,6 +167,12 @@ class Label {
     const OBB& obb() const { return m_obb; }
     const Transform& transform() const { return m_transform; }
     const State& state() const { return m_currentState; }
+    bool occludedLastFrame() const { return m_occludedLastFrame; }
+
+    glm::vec2 xAxis() const;
+    glm::vec2 yAxis() const;
```
blair1618: Looks like these two don't have an implementation?

karimnaaji: And we actually don't need those accessors, removing

blair1618:
When I run this branch on desktop, the render loop runs continuously without any user input. Is something making continuous render requests?

karimnaaji:
Yeah I'm on it, it's a bug related to inconsistent state changes, we need to differentiate how the occlusion has been triggered in order to solve this.

tallytalwar:
We should also make more unit tests for label state changes..

karimnaaji:
@tallytalwar It's hard to test since it's not related to internal state changes which are already tested, but mainly because of what's in the label set sent to repeat groups; what's going on is:
```
//// frame n
l.state == wait_occ
//// frame n+1
l.state == fading_in
visible_set.add(l)
repeat_groups(visible_set)
l.occluded(true)
//// frame n+2
l.state == sleep
visible_set.contains(l) == false
repeat_groups(visible_set)
l.occluded(false)
//// frame n+3 == frame n+1, hence the looping and requestRender()
```

karimnaaji:
There is following work to be done here; I am noticing some visual unpleasant artifacts on label transitions (direct transitions are consistent, but a 3 step transition can be considered _inconsistent_, we can improve by storing a 3 steps history of transitions, for example a `fading_in`->`visible`->`fading_out` shouldn't be allowed by the state machine), which is somehow induced by this feature, but would be better if post-bone to another pr. 
@hjanetzek it seems that you reviewed this? If anyone else is happy with it then we can merge.

core/src/labels/labels.cpp
```diff
@@ -84,98 +69,218 @@ void Labels::update(const View& _view, float _dt, const std::vector<std::unique_
...
+        auto it = repeatGroups.find(hash);
+        if (it == repeatGroups.end()) {
+            repeatGroups[hash].push_back(element);
+            continue;
+        }
+
+        std::vector<GroupElement>& group = it->second;
+        if (std::find(group.begin(), group.end(), element) != group.end()) {
+            // TBD: when is this case possible?
+            // -> When two tiles contain the same label?
```
tallytalwar: how is this possible? Will the hash of the label not be same in that case!!! which is not possible, or is it?

core/src/labels/labels.cpp
```diff
@@ -84,98 +69,218 @@ void Labels::update(const View& _view, float _dt, const std::vector<std::unique_
...
-    //// Update label meshes
+        size_t hash = std::hash<std::string>()(textLabel->options().repeatGroup);
+        GroupElement element;
+
+        element.position = textLabel->center();
+        element.threshold = textLabel->options().repeatDistance;
+        element.group = &textLabel->options().repeatGroup;
+
+        auto it = repeatGroups.find(hash);
+        if (it == repeatGroups.end()) {
```
tallytalwar: Probably a future performance consideration:

"If" we can store the modification of a repeatGroup (that is, no label is added or removed from that repeatGroup) then we can skip checking that repeatGroup completely.

core/src/labels/labels.cpp
hjanetzek: That's why I added the comment - I'm not sure for what this is for except it could be for filtering the same label from two different tiles, i.e. labels with same properties and same position in screen space. 

Sure the hash would be the same for the suggested case - as the hash is for a string that contains "[repeatGroup]/[labelText]"

core/src/labels/labels.cpp
hjanetzek: I'm not sure what you mean. This case adds the first item of a repeat group. (So there is no need to check if it conflicts with another labels of the same group)

core/src/style/textStyle.cpp
```diff
@@ -111,6 +111,18 @@ auto TextStyle::applyRule(const DrawRule& _rule, const Properties& _props) const
...
+        if (_rule.get(StyleParamKey::repeat_distance, repeatDistance)) {
+            p.labelOptions.repeatDistance = repeatDistance.value * m_pixelScale;
+        }
+    } else {
+        // TODO: default to 'draw.key'
+    }
+
+    // TBD: should avoid allocation for combined string
+    p.labelOptions.repeatGroup += "/" + p.text;
+
```
hjanetzek: If the label text need to be kept in TextLabel like this, I think it would be better to have separate members for repeatGroup and text. This will also increase the chance to avoid heap allocated string when both fit into 'small-strings' (16 or 24 chars)

core/src/labels/labels.cpp
karimnaaji: I think what @tallytalwar means is keeping track of addition/removal from previous frames, if nothing is changing, there's not reason to go and check again.

hjanetzek: Ah, yes, this could be investigated later. so the comment was not for this particular line :)

core/src/style/textStyle.cpp
hjanetzek: @karimnaaji  Is it necessary that all text labels have a repeatGroup? Most point features have a unique position and label. For these it would be good to (at least be able to) avoid the per frame check.

tallytalwar: @karimnaaji yup you are right .. by what I meant :).
@hjanetzek yup that was a generic comment not on that line and also a
future potential optimization.

core/src/labels/labels.cpp
tallytalwar: @karimnaaji thanks :D

core/src/style/textStyle.cpp
karimnaaji: >Is it necessary that all text labels have a repeatGroup

Since there is a default behavior on it I would say yes.

core/src/style/textStyle.cpp
```diff
@@ -111,6 +111,25 @@ auto TextStyle::applyRule(const DrawRule& _rule, const Properties& _props) const
...
 
+    if (!_rule.get(StyleParamKey::repeat_group, p.labelOptions.repeatGroup)) {
+        // TODO: default to 'draw.key'
+        // TODO: Optimize - this is kind of heavy on allocations
+        for (auto* name : _rule.getLayerNames()) {
+            p.labelOptions.repeatGroup += name;
+            p.labelOptions.repeatGroup += "/";
+        }
+        //LOG("rg: %s", p.labelOptions.repeatGroup.c_str());
+    }
```
hjanetzek: I would suggest that we only keep the hash of the repeatGroup to avoid the huge amount of large strings to keep around. There is a slight chance of hash collisions of course, so that some labels may fall into the same group.

karimnaaji: Are you suggesting a collision handling within labels.cpp with a smaller hash for the repeatGroup?

hjanetzek: The used hash value would be the same. I mean we currently keep the whole repeatGroup string in labelOptions, calculate the hash for the string in checkRepeatGroup and then only match groups based on the hash value there. This could be considered a bug or an optimization (by only comparaing the hash value instead of the repeatGroup strings) 

My suggestion is to only store the repeatGroup hash value in labelOptions

core/src/labels/labels.cpp
```diff
@@ -84,98 +69,221 @@ void Labels::update(const View& _view, float _dt, const std::vector<std::unique_
...
+            // -> When two tiles contain the same label?
+            continue;
+        }
+
+        bool add = true;
+        float threshold2 = pow(element.threshold, 2);
+
+        for (const GroupElement& ge : group) {
+            float d2 = distance2(ge.position, element.position);
+            if (d2 < threshold2) {
```
hjanetzek: Here one would need to check here if group.group == element.group due to possible hash collisions

core/src/style/textStyle.cpp
karimnaaji: Makes sense to me; as you said except the fact that we would have a slight chance of hash collision I'm good with it.

core/src/labels/labels.cpp
karimnaaji: :+1:

hjanetzek:
Everyone ready to merge? :)

karimnaaji:
Creating an issue to follow the discussion relative to the label state machine that needs to be smarter about un-allowing 3 steps state transitions.
Otherwise merging!

Issue #466, opened 2016-01-13
Easing position at high zoom causes "wiggle"


blair1618:
This can be observed in the desktop demo by double-clicking on a position in the map at zoom 19.

tallytalwar:
I would call this a snoopdogg issue :P. (sorry, couldn't help myself when I saw "wiggle").

PR #467, opened 2016-01-13
Better numerical stability for easing

blair1618:
Fixes https://github.com/tangrams/tangram-es/issues/466


tallytalwar:
good thought/catch.

PR #475, opened 2016-01-15
Harfbuzz integration

karimnaaji:
See https://github.com/tangrams/fontstash-es/pull/18 for fontstash-es API update; mainly script/bidi detection.

No font fallbacks, if the font doesn't support any codepoint for the shaping plan, it will simply fail for now (fixing https://github.com/mapzen/eraser-map/issues/101).

Build system integration can be followed here: https://github.com/tangrams/harfbuzz-icu-freetype, where Harfbuzz RAGEL generated files are directly inside the repo, we would need to regenerate those everytime we update the version of Harfbuzz.

screen shot 2016-01-15 at 5 33 40 pm

~~ICU dependency will be removed for a smaller library (UCDN).~~ still used for multiple bidi detection
Sorry for the lots of diffs in CMake, we should consistently use 4 spaces tabs in it.

- [ ] ~~Solve cases with multiple bidi detected (ICU might be useful in that case)~~ (separate task for this)
- [ ] ~~Update README to note gcc-5 instructions~~ (not necessary anymore)


tallytalwar:
YAY :metal:

karimnaaji:
@hjanetzek can you review https://github.com/tangrams/fontstash-es/pull/18 for this pull request.

karimnaaji:
@hjanetzek , can you review please. 

What is done by this pull request:
- build system integration
- bidi detection (single direction detection, LTR/RTL)
- extension of fontstash-es api to detect drawable strings, along with harfbuzz functionalities

What should still be done:
- multiple bidi handling within a string which involves line break detection with [libunibreak](http://vimgadgets.sourceforge.net/libunibreak/) to break shaping into different runs
- font fallbacks through a font manager (which would somehow have a platform-dependent font resolution) for unresolved codepoints

karimnaaji:
Based on a discussion with @hjanetzek, implementation is ready but we decided to hold onto merging this branch for now for two reasons:
- using harfbuzz is a no-op until font-fallback is added
- shaping is a time-consuming step, which is aggravated by unseen labels

bcamper:
Just to clarify, in terms of extra computation, we only apply shaping when
needed though, right?

PR #476, opened 2016-01-16
Data sources can set maximum zoom for tile requests

blair1618:
Beyond `max_zoom`, data from the highest available zoom is scaled up and re-styled for the current zoom level.

### DIFF
--- a/core/src/tile/tileID.h
+++ b/core/src/tile/tileID.h
@@ -21,7 +21,9 @@ struct TileID {
     int32_t s; // Styling zoom
     int32_t wrap;
 
-    TileID(int32_t _x, int32_t _y, int32_t _z, int32_t _s, int32_t _wrap = 0) : x(_x), y(_y), z(_z), s(_s), wrap(_wrap) {}
+    TileID(int32_t _x, int32_t _y, int32_t _z, int32_t _s, int32_t _wrap) : x(_x), y(_y), z(_z), s(_s), wrap(_wrap) {}
+
+    TileID(int32_t _x, int32_t _y, int32_t _z) : TileID(_x, _y, _z, _z, 0) {}
 
     TileID(const TileID& _rhs) = default;
 
--- a/core/src/tile/tileID.h
+++ b/core/src/tile/tileID.h
@@ -17,9 +17,9 @@ struct TileID {
 
     int32_t x; // Index from left edge of projection space
     int32_t y; // Index from top edge of projection space
-    int32_t z; // Data zoom
-    int32_t s; // Styling zoom
-    int32_t wrap;
+    int8_t  z; // Data zoom
+    int8_t  s; // Styling zoom
+    int16_t wrap;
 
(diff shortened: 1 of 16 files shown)


core/src/tile/tileID.h
```diff
@@ -41,14 +44,30 @@ struct TileID {
         return isValid() && z <= _maxZoom;
     }
 
+    TileID withMaxSourceZoom(int32_t _max) const {
```
tallytalwar: Not sure `withMaxSourceZoom` is the appropriate name here?
how about `tileIDUnzoomed` or something on those lines.

tallytalwar:
Can we now reset `View::s_maxZoom` back to `18.5`? [here](https://github.com/tangrams/tangram-es/blob/overzoom/core/src/view/view.h#L138)

tallytalwar:
Simple and succinct. Looks to me.

Will merge once couple of small questions are answered/addressed.

hjanetzek:
Wouldn't it be simpler to just modify the visibleTileSet per data source when the view zoom is greater than  max zoom of the datasource? I.e replace all tiles of visibleSet with their respective [grand*]parents at max zoom.

blair1618:
@tallytalwar I believe we still want the maximum view zoom to be 20 - this is effectively the maximum zoom for all sources. Later, we should probably allow configuring this through the scene file but we'll need to consult with the tangram-js team on that. 

@hjanetzek The ID re-mapping step didn't have a measurable time impact when I profiled, but applying it conditionally certainly won't hurt - thanks for adding that.

bcamper:
Yes, 20 is the default max **view** zoom (this should also probably be configurable both as a scene preference, and directly through the API).

The max zoom at which we **fetch tiles** for a given source defaults to 18 in Tangram JS, but is configurable through the `max_zoom` parameter on the data source definition.

When the view zoom is higher than the max tile source zoom, the tile data from the max zoom is scaled -- but importantly, still re-styled. 

In the future, we may also allow for tile sources to serve arbitrary zoom levels and thus also perform "overzooming" at other intermediate zooms, e.g. serve tiles for zooms 12 and 14, and overzoom z12 tiles at z13.

blair1618:
@bcamper so it sounds like the default value for `max_zoom` is 18. I'll make that the case here as well, and we may want to update [the docs](https://github.com/tangrams/tangram-docs/blob/gh-pages/pages/sources.md#max_zoom) to reflect that default.

core/src/tile/tileID.h
blair1618: I think this name makes usage at the call site pretty clear, e.g. 
```c++
requestTileId = tileId.withMaxSourceZoom(source.maxZoom());
```
"unzoomed" seems misleading since tiles below the max zoom are unchanged by the call.

tallytalwar: Yes, from a callee side it make sense.

PR #477, opened 2016-01-19
Add TopoJSON data sources

blair1618:
- Separates common Json types and functions into their own file
 - Refactors data parsing functions to maximize return-value-optimization and readability
 - Adds `TopoJsonSource` class and `TopoJson` namespace of decoding functions


blair1618:
I wonder if it would be of utility to anyone else to maintain separate repos for just topojson/geojson decoding into native types? If something like that existed it would have saved me much of the trouble of writing this :\

tallytalwar:
Sure, if geojson/topojson parser existed as is, it would have saved us a bunch of time.

There is always a possibility that someone might be interested in using these geojson/topojson utilities, so I vote for making separate repos for these.

karimnaaji:
Yeah I agree on that. It would be really useful to separate this decoding.

blair1618:
Cool. I'll make separate decoders an 'enhancement' issue to follow-up on later.

tallytalwar:
Sure, one thing we should discuss is, should the person using our future geojson/topojson project be restricted to use our `TileData`, since our `TileData` is tightly coupled with our implementations of geojson and topojson parsers.

blair1618:
Here's the new issue: https://github.com/tangrams/tangram-es/issues/480

I think we can turn these into header-only template libraries that are reasonably type-agnostic, so others wouldn't be bound to using our specific data types.

tallytalwar:
Yup. Thanks for creating this issue.

hjanetzek:
TileData has been replaced in the tile+style-builder branch with a callback interface, which could also be made more fine grained:
```
struct TileDataSink {
    virtual bool beginLayer(const std::string& _layer) = 0;
    virtual bool matchFeature(const Feature& _feature) = 0;
    virtual void addFeature(const Feature& _feature) = 0;
};
```
beginLayer and matchFeature can return false to skip processing of these. (The feature passed to matchFeature has only Properties and GeometryType available to be able to avoid extracting unused geometries)

https://github.com/tangrams/tangram-es/blob/e763483d7606cff5adc7b9bcf591686e2b90b7ef/core/src/data/dataSource.h#L27

core/src/data/topoJsonSource.h
```diff
@@ -0,0 +1,20 @@
...
+class TopoJsonSource : public DataSource {
+
+protected:
+
+    virtual std::shared_ptr<TileData> parse(const TileTask& _task,
+                                            const MapProjection& _projection) const override;
+
+public:
+
+    TopoJsonSource(const std::string& _name, const std::string& _urlTemplate);
```
tallytalwar: can you add `maxZoom` argument to the constructor?

core/src/util/topoJson.h
```diff
@@ -0,0 +1,35 @@
...
+
+Point getPoint(const JsonValue& _coordinates, const Topology& _topology, glm::ivec2& _cursor);
+
+Line getLine(const JsonValue& _arcs, const Topology& _topology);
+
+Polygon getPolygon(const JsonValue& _arcs, const Topology& _topology);
+
+Feature getFeature(const JsonValue& _geometry, const Topology& _topology, int32_t _sourceId);
+
+Layer getLayer(JsonValue::MemberIterator& _object, const Topology& _topology, int32_t _sourceId);
```
tallytalwar: `pbfParser` methods should be renamed to follow the `get*` convention we are going with `GeoJson` and `TopoJson` parsers.

blair1618: Addressed by https://github.com/tangrams/tangram-es/commit/c4a119473defa9b79deaa6b5718888793aac3705!

core/src/data/topoJsonSource.h
blair1618: Addressed by https://github.com/tangrams/tangram-es/commit/82a5e7306b75b2757d3f845e9bc7844ae7c73086!

blair1618:
@tallytalwar any further comments here?

tallytalwar:
Sorry, got distracted with some other eraser-map work. Will again pick reviewing this.

PR #483, opened 2016-01-20
Respond to only the final 'up' event for double-taps

blair1618:


blair1618:
@tallytalwar This should allow the double-tap zooming you wanted to add.

tallytalwar:
LGTM.

PR #484, opened 2016-01-21
Filter optimization

hjanetzek:
The main addition here is to sort filters so that cheaper tests are performed first:

- check globals (zoom, geom) before properties
- check functions last
- for list of zoom+area pairs and the zoom ranges overlap (have min zoom but no max) start to check the highest zoom-level first, the other way round one would fail more often on the area property:
```
-> original
- { $zoom: { min: 14 }, area: { min: 50000 } }
- { $zoom: { min: 15 }, area: { min: 2000 } }
- { $zoom: { min: 16 } }
-> use shorter group when both have the same number of global tests
- { $zoom: { min: 16 } }
- { $zoom: { min: 14 }, area: { min: 50000 } }
- { $zoom: { min: 15 }, area: { min: 2000 } }
-> when both match on the same range property use the one with smaller range first
- { $zoom: { min: 16 } }
- { $zoom: { min: 15 }, area: { min: 2000 } }
- { $zoom: { min: 14 }, area: { min: 50000 } }
```

TODO
- determine possible zoom-range for layers, to be able to skip whole layers directly (needs TileDataSink)
- comparison of filter groups can be further extended
- one could e.g. also analyze sublayer groups further for things like switch case matching on one property. Or propagate required properties up by inserting Existence filters on the lowest common layer

### DIFF
--- a/core/src/data/properties.cpp
+++ b/core/src/data/properties.cpp
@@ -31,14 +31,22 @@ void Properties::setSorted(std::vector<Item>&& _items) {
 const Value& Properties::get(const std::string& key) const {
     const static Value NOT_FOUND(none_type{});
 
-    const auto it = std::lower_bound(props.begin(), props.end(), key,
-                                     [](const auto& item, const auto& key) {
-                                         return keyComparator(item.key, key);
-                                     });
-
-    if (it == props.end() || it->key != key) {
+    const auto it = std::find_if(props.begin(), props.end(),
+                                 [&](const auto& item) {
+                                     return item.key == key;
+                                 });
+    if (it == props.end()) {
         return NOT_FOUND;
     }
+
+    // auto it = std::lower_bound(props.begin(), props.end(), key,
+    //                            [](auto& item, auto& key) {
+    //                                return keyComparator(item.key, key);
+    //                            });
+    // if (it == props.end() || it->key != key) {
+    //     return NOT_FOUND;
+    // }
+
     return it->value;
 }
 
--- a/core/src/scene/filters.cpp
+++ b/core/src/scene/filters.cpp
@@ -1,10 +1,211 @@
 #include "filters.h"
 #include "scene/styleContext.h"
 #include "data/tileData.h"
+#include "platform.h"
 
 #include <cmath>
 
 namespace Tangram {
+
+void countTypes(const std::vector<Filter>& filters, int& global, int& function, int& property) {
+
+}
+
+void Filter::print(int _indent) const {
+
+    switch (data.get_type_index()) {
+
+    case Data::type<OperatorAny>::value: {
+        logMsg("%*s any\n", _indent, "");
+        for (const auto& filt : data.get<OperatorAny>().operands) {
+            filt.print(_indent + 2);
+        }
+        break;
+    }
+    case Data::type<OperatorAll>::value: {
+        logMsg("%*s all\n", _indent, "");
+        for (const auto& filt : data.get<OperatorAll>().operands) {
+            filt.print(_indent + 2);
+        }
+        break;
+    }
+    case Data::type<OperatorNone>::value: {
+        logMsg("%*s none\n", _indent, "");
+        for (const auto& filt : data.get<OperatorNone>().operands) {
+            filt.print(_indent + 2);
+        }
+        break;
+    }
+    case Data::type<Existence>::value: {
+        auto& f = data.get<Existence>();
+        logMsg("%*s existence - key:%s\n", _indent, "", f.key.c_str());
+        break;
+    }
+    case Data::type<Equality>::value: {
+        auto& f = data.get<Equality>();
+        if (f.values[0].is<std::string>()) {
+            logMsg("%*s equality - global:%d key:%s val:%s\n", _indent, "",
+                   f.global != FilterGlobal::undefined,
+                   f.key.c_str(),
+                   f.values[0].get<std::string>().c_str());
+        }
+        if (f.values[0].is<double>()) {
+            logMsg("%*s equality - global:%d key:%s val:%f\n", _indent, "",
+                   f.global != FilterGlobal::undefined,
+                   f.key.c_str(),
+                   f.values[0].get<double>());
+        }
+        break;
+    }
+    case Data::type<Range>::value: {
+        auto& f = data.get<Range>();
+        logMsg("%*s range - global:%d key:%s min:%f max:%f\n", _indent, "",
+               f.global != FilterGlobal::undefined,
+               f.key.c_str(), f.min, f.max);
+        return;
+    }
+    case Data::type<Function>::value: {
+        logMsg("%*s function\n", _indent, "");
+        break;
+    }
+    default:
+        break;
+    }
+
+}
+
+
+int Filter::matchCost() const {
+    // Add some extra penalty for set vs simple filters
+    int sum = -100;
+
+    switch (data.get_type_index()) {
+    case Data::type<OperatorAny>::value:
+        for (auto& f : operands()) { sum -= f.matchCost(); }
+        return sum;
+
+    case Data::type<OperatorAll>::value:
+        for (auto& f : operands()) { sum -= f.matchCost(); }
+        return sum;
+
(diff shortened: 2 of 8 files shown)


blair1618:
Awesome! I'll do some profiling and look over the code.

tallytalwar:
It will be good to profile on eraser-map style sheet, since we have a lot of filtering going on in that.

nvkelso:
:100:

bcamper:
I started filtering first on zoom in JS recently and got a
significant improvement (~15%) for the EM scene.

blair1618:
For the eraser-map style, at zoom 14, viewing a dense region, on an iPhone 5S:

Filtering time is reduced by 36%

Total tile build time is reduced by 23%

tallytalwar:
:metal: :+1:

hjanetzek:
@blair1618 yay! that are roughly the numbers I got on desktop as well :)

core/src/scene/filters.h
```diff
@@ -57,19 +62,29 @@ struct Filter {
...
     }
     inline static Filter MatchNone(const std::vector<Filter>& filters) {
-        return { OperatorNone{ filters }};
+        return { OperatorNone{ sort(filters) }};
     }
     // Create an 'equality' filter
     inline static Filter MatchEquality(const std::string& k, const std::vector<Value>& vals) {
+        // if (vals.size() == 1) {
+        //     return { EqualityOne{ k, vals[0], globalType(k) }};
+        // } else {
```
blair1618: Looks like `EqualityOne` isn't being used. Did you try it and find any difference?

core/src/scene/filters.cpp
```diff
@@ -1,11 +1,214 @@
 #include "filters.h"
 #include "scene/styleContext.h"
 #include "data/tileData.h"
+#include "platform.h"
 
 #include <cmath>
 
 namespace Tangram {
 
+void countTypes(const std::vector<Filter>& filters, int& global, int& function, int& property) {
```
blair1618: Leftover?

core/src/scene/filters.h
hjanetzek: Couldn't see a difference. I guess it wouldn't  hurt to add it back though - just was one case more to handle during the cost evaluation..

I'll go over it for cleanups tomorrow
 over it for cleanups tomorrow


I'll go over it for cleanups tomorrow
, just was one more case to handle during 

I'll go over it for cleanups tomorrow

core/src/scene/filters.h
```diff
@@ -57,19 +62,29 @@ struct Filter {
     Filter() : data(none_type{}) {}
     Filter(Data _data) : data(std::move(_data)) {}
 
+    static std::vector<Filter> sort(const std::vector<Filter>& filters);
+    void print(int _indent = 0) const;
+    int matchCost() const;
```
blair1618: This name might be confusing with all the other "matchX" methods. `getCost()`? `cost()`?

core/src/scene/filters.cpp
```diff
@@ -1,11 +1,214 @@
...
+
+    if (oa[0].data.is<Filter::Range>() &&
+        ob[0].data.is<Filter::Range>() &&
+        oa[0].key() == ob[0].key()) {
+        // take the one with more restrictive range
+        auto ra = oa[0].data.get<Filter::Range>();
+        auto rb = ob[0].data.get<Filter::Range>();
+
+        if (ra.max == std::numeric_limits<double>::infinity() &&
+            rb.max == std::numeric_limits<double>::infinity()) {
```
blair1618: Using `std::isinf()` could be more compact and type-safe.

core/src/scene/filters.cpp
```diff
@@ -1,11 +1,214 @@
...
+
+            return rb.min - ra.min;
```
blair1618: It seems that this will prefer the more restrictive range in the case where neither range has a maximum, but in other cases there is no preference. Is this intentional? Feels like we could reasonable handle all cases.

blair1618: Actually, I'm now uncertain that we can reliably prefer some ranges over others. For an "any" filter we should prefer *less* restrictive ranges so we can return as soon as possible, but for an "all" filter we should prefer *more* restrictive ranges.

core/src/scene/filters.cpp
```diff
@@ -1,11 +1,214 @@
...
+    default:
+        break;
+    }
+
+}
+
+
+int Filter::matchCost() const {
+    // Add some extra penalty for set vs simple filters
+    int sum = -100;
```
blair1618: The negative costing is feeling very hard to reason about, and seems like it could have bad edge cases; e.g. "set" filters are intended to have negative costs, but couldn't a "set" filter whose operands are "set" filters return a positive cost? Maybe I'm mistaken, but that would evidence my first point :)

core/src/scene/styleContext.cpp
```diff
@@ -465,4 +412,50 @@ duk_ret_t StyleContext::jsGetProperty(duk_context *_ctx) {
     return 1;
 }
 
+/* This function is only used by tests - Remove? */
```
blair1618: It would be nice to remove this, if possible. Same with the one below.

core/src/scene/filters.cpp
hjanetzek: But for an 'any' filter more restrictive ranges make sense when it first checks on a global to avoid failing multiple times on a property. As with the zoom/area combination.

So far I only considered the patterns found in eraser-map style. I would say a more in-depth analysis of possible patterns can be done later. There are really a *lot* of things that could be considered.

core/src/scene/filters.cpp
hjanetzek: Right! changed it to check for isOperator separately

blair1618:
This looks super solid now. Merging!

core/src/data/properties.cpp
```diff
@@ -31,14 +31,22 @@ void Properties::setSorted(std::vector<Item>&& _items) {
 const Value& Properties::get(const std::string& key) const {
     const static Value NOT_FOUND(none_type{});
 
-    const auto it = std::lower_bound(props.begin(), props.end(), key,
-                                     [](const auto& item, const auto& key) {
-                                         return keyComparator(item.key, key);
-                                     });
-
-    if (it == props.end() || it->key != key) {
+    const auto it = std::find_if(props.begin(), props.end(),
```
tallytalwar: Just a quick comment.

In our work with `isect2d`, we found that doing a find explicitly by manually iterating was faster than using `std::find_if`.
Not sure if you tried that?

tallytalwar:
Before this gets merged, @hjanetzek can you remove some not used commented code. Thanks

blair1618:
Holding merge till all comments are addressed..

core/src/data/properties.cpp
hjanetzek: Were these tests with Release builds and on std::vector? In my tests find_if was faster or just as fast than manual iteration. This is really a thing to let the compiler can figure out :)

core/src/data/properties.cpp
```diff
@@ -31,14 +31,22 @@ void Properties::setSorted(std::vector<Item>&& _items) {
...
+    // auto it = std::lower_bound(props.begin(), props.end(), key,
+    //                            [](auto& item, auto& key) {
+    //                                return keyComparator(item.key, key);
+    //                            });
+    // if (it == props.end() || it->key != key) {
+    //     return NOT_FOUND;
+    // }
+
     return it->value;
 }
```
hjanetzek: @tallytalwar If you mean this one. Can we leave it in as a reminder to test which one is faster on different platforms?

core/src/data/properties.cpp
tallytalwar: I know and agree that these things compiler should figure out, just sharing what we found :D.

Yes it was with std::vector. Do not remember if it was release/debug build.

core/src/data/properties.cpp
tallytalwar: :+1:

PR #487, opened 2016-01-21
Omit all tile edges for polylines

blair1618:
- Fixes polyline builder never checking line segments for coincidence with tile edges (not sure when this was broken)
 - Modifies tile edge check to match edges of any tile, not just the current tile


blair1618:
@tallytalwar give this a quick look if you please.

tallytalwar:
Sure on it. Thanks

PR #488, opened 2016-01-21
Ignore picking labels not currently visible

tallytalwar:


tallytalwar:
@karimnaaji can you check this please, thanks.

core/src/labels/labels.cpp
```diff
@@ -313,6 +313,8 @@ const std::vector<TouchItem>& Labels::getFeaturesAtPoint(const View& _view, floa
 
             for (auto& label : labelMesh->getLabels()) {
 
+                if (label->state() != Label::State::visible) { continue; }
```
karimnaaji: You can directly use https://github.com/tangrams/tangram-es/blob/master/core/src/labels/label.cpp#L169

tallytalwar: :+1:

karimnaaji:
Also, this option should still be considered in the test https://github.com/tangrams/tangram-es/blob/pickFeature-fix/core/src/labels/labels.cpp#L289

tallytalwar:
@karimnaaji thanks.

karimnaaji:
Good to be merged!

PR #489, opened 2016-01-22
Pick feature prop update

tallytalwar:
- Adds a position member to `TouchItem`, which is used to determine the position of the client geometry.
- Ability to update a `PropertyItem::value`, if a `value` exists corresponding to a `PropertyItem::key`

### DIFF
--- a/android/tangram/jni/platform_android.cpp
+++ b/android/tangram/jni/platform_android.cpp
@@ -66,7 +66,7 @@ void setupJniEnv(JNIEnv* _jniEnv, jobject _tangramInstance, jobject _assetManage
     getFontFilePath = jniEnv->GetMethodID(tangramClass, "getFontFilePath", "(Ljava/lang/String;)Ljava/lang/String;");
     requestRenderMethodID = _jniEnv->GetMethodID(tangramClass, "requestRender", "()V");
     setRenderModeMethodID = _jniEnv->GetMethodID(tangramClass, "setRenderMode", "(I)V");
-    featureSelectionCbMID = _jniEnv->GetMethodID(tangramClass, "featureSelectionCb", "(Lcom/mapzen/tangram/Properties;)V");
+    featureSelectionCbMID = _jniEnv->GetMethodID(tangramClass, "featureSelectionCb", "(Lcom/mapzen/tangram/Properties;FF)V");
 
     jclass propertiesClass = jniEnv->FindClass("com/mapzen/tangram/Properties");
 
@@ -302,7 +302,7 @@ void featureSelectionCallback(JNIEnv* jniEnv, const std::vector<Tangram::TouchIt
     jresult = reinterpret_cast<jlong>(new std::shared_ptr<Tangram::Properties >(items[0].properties));
     jobject object = jniEnv->NewObject(propertiesClass, propertiesConstructorMID, jresult, true);
 
-    jniEnv->CallVoidMethod(tangramInstance, featureSelectionCbMID, object);
+    jniEnv->CallVoidMethod(tangramInstance, featureSelectionCbMID, object, items[0].position[0], items[0].position[1]);
 }
 
 void initGLExtensions() {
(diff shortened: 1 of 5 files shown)


tallytalwar:
@hjanetzek can you check this please. Thanks

blair1618:
Might be simpler to add the label's obb or aabb to TouchItem and return that in the selection callback instead of just the properties.

tallytalwar:
I did consider that option but had already done this work. I can modify that.
However part of this PR is still useful for updating `PropertyItem` in `Feature` `Properties`.

tallytalwar:
Also, just the centroid of the label will do, instead of `obb` or `aabb`. Will update, but tomorrow :D.

core/src/data/properties.cpp
```diff
@@ -104,11 +109,32 @@ void Properties::sort() {
...
+    auto it = std::lower_bound(props.begin(), props.end(), key,
+                                 [](auto& item, auto& key) {
+                                     return keyComparator(item.key, key);
+                                 });
+
+    if (it == props.end() || it->key != key) {
+        props.emplace_back(std::move(key), Value{value});
+    } else {
+        it->value = Value{value};
+    }
```
hjanetzek: This is good as it also ensures the intended behavior of having unique keys. Could you rename the functions to 'set' ?

tallytalwar: sure.

tallytalwar: This rename resulted in touching of a bunch of files!!

tallytalwar:
Updated..

core/src/tangram.h
```diff
@@ -7,6 +7,8 @@
 #include <vector>
 #include <string>
 
+#include "glm/vec2.hpp"
```
blair1618: unused?

tallytalwar: updated. Thanks

android/tangram/jni/platform_android.cpp
```diff
@@ -302,7 +302,10 @@ void featureSelectionCallback(JNIEnv* jniEnv, const std::vector<Tangram::TouchIt
     jresult = reinterpret_cast<jlong>(new std::shared_ptr<Tangram::Properties >(items[0].properties));
     jobject object = jniEnv->NewObject(propertiesClass, propertiesConstructorMID, jresult, true);
 
-    jniEnv->CallVoidMethod(tangramInstance, featureSelectionCbMID, object);
+    jdoubleArray itemCenter = jniEnv->NewDoubleArray(2);
```
karimnaaji: Do you need a `ReleaseDoubleArrayElements` following this?

tallytalwar: Nopes, I think if you use the `*Region` call then you do not need to do a Release.

http://developer.android.com/training/articles/perf-jni.html#region_calls

blair1618: I am curious too - it's the "New" call that makes me think this needs to be released, the "Region" call should be fine as it is.

tallytalwar: Also referring to stamford course notes:
http://statweb.stanford.edu/~naras/java/course/lec5/lec5.html

tallytalwar: Based on what I have read and understood, `Get<PrimitiveType>ArrayRegion` functions require the `Release<PrimitiveType>ArrayElements()` call.

More documentation: http://docs.oracle.com/javase/1.5.0/docs/guide/jni/spec/functions.html#wp17314

blair1618: Looking at this and other examples, I think the current code is probably ok. Of course, it would be even simpler to pass two separate numbers :)

tallytalwar: I think its more proper to pass an array of 2 than separate numbers.

blair1618: Not sure about that - array doesn't guarantee any number of values to the receiver. [PointF](https://developer.android.com/reference/android/graphics/PointF.html) is another option.

blair1618:
Rebasing, hang tight!

PR #490, opened 2016-01-22
Opt view scan function

hjanetzek:


blair1618:
Makes sense to me. It might be nice later to move the scanning code to a separate file, since it it quite self-contained.

PR #491, opened 2016-01-22
Update instead of adding multiple key:value props

hjanetzek:
- Renamed add to set

Thought it would make a good PR on its own :)

### DIFF
--- a/core/src/data/properties.cpp
+++ b/core/src/data/properties.cpp
@@ -114,31 +114,29 @@ void Properties::sort() {
 void Properties::set(std::string key, std::string value) {
 
     auto it = std::lower_bound(props.begin(), props.end(), key,
-                                 [](auto& item, auto& key) {
-                                     return keyComparator(item.key, key);
-                                 });
+                               [](auto& item, auto& key) {
+                                   return keyComparator(item.key, key);
+                               });
 
     if (it == props.end() || it->key != key) {
-        props.emplace_back(std::move(key), Value{std::move(value)});
+        props.emplace(it, std::move(key), std::move(value));
     } else {
-        it->value = Value{std::move(value)};
+        it->value = std::move(value);
     }
-    sort();
 }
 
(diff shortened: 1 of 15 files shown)


tallytalwar:
:+1:

tallytalwar:
This is a breaking change on eraser-map builds. 
So, #489 and #491 should be merged in parallel and then https://github.com/mapzen/eraser-map/pull/264 on eraser-map.

core/src/data/properties.cpp
```diff
@@ -50,6 +50,8 @@ const Value& Properties::get(const std::string& key) const {
     return it->value;
 }
 
+
+
```
blair1618: ?
?

blair1618: :)

blair1618:
Looks good to me. Let's merge this!

This could substantially improve parse time for JSON tile formats - right now feature properties are collected by making repeated calls to `Properties::add()` which is needlessly re-sorting the list over and over.

blair1618:
Not necessarily a concern for this PR, but [this bit in Properties](https://github.com/tangrams/tangram-es/blob/master/core/src/data/properties.cpp#L12-L18) seems like it can (should) be replaced with `std::swap(props, _items)`

tallytalwar:
Agreed!

hjanetzek:
This constructor was used specifically for mvt source. The idea was to reuse the allocated vector on the caller site and just clobber the values from it.

tallytalwar:
Rather `props.swap(_items)`

blair1618:
Ah, ok. I'll leave that constructor alone for now. I can do what I wanted with the slightly uglier
```
properties.setSorted(unsortedItems);
properties.sort();
```

tallytalwar:
Can we do this outside this PR?

blair1618:
Yeah I'm just thinking about changes elsewhere. Carry on!

PR #492, opened 2016-01-22
Line parameter refactor

blair1618:
- Cleanly separates polyline rule interpretation from mesh building process
 - Handles `tile_edges` parameter and provides correct default behavior when absent

Fixes https://github.com/mapzen/eraser-map/issues/265


tallytalwar:
screen shot 2016-01-22 at 5 32 06 pm

- Left: Current branch behavior
- Right: master behavior


Code looks good to me.
Merge when travis building issues revolve.

PR #493, opened 2016-01-22
Debug text display

karimnaaji:
Render debug display using `stb_easy_font` for debug.
screen shot 2016-01-22 at 4 40 35 pm
More features will be added (frame time, log to screen...)


blair1618:
It's beautiful. I don't even know why we bother with fonts :P

nvkelso:
+1

Issue #478, opened 2016-01-19
Refactor label state machine


karimnaaji:
As induced by https://github.com/tangrams/tangram-es/pull/465, there is some labels appearing and disappearing instantly; the label state machine should disallow some three steps transitions (which can be considered inconsistent as `fading_in`->`visible`->`fading_out` for example)

hjanetzek:
Looking into this next.

PR #497, opened 2016-01-25
Refactor label state machine

hjanetzek:
https://github.com/tangrams/tangram-es/issues/478

- label occlusion *should* have stopped running into loops
- separate updateState() and evalState().
  Before the state was applied for the previous frame (m_occludedLastFrame)
- this also fixes first frame flickering (non-drawn labels) when
  switching zoom-level
- refactor: skipTransitions (move nested loop into separate function)
- remove temporary 'occlusions' set. The result of broadphase is already a unique set

### DIFF
--- a/core/src/labels/label.cpp
+++ b/core/src/labels/label.cpp
@@ -17,17 +17,14 @@ Label::Label(Label::Transform _transform, glm::vec2 _size, Type _type, LabelMesh
     if (!m_options.collide || m_type == Type::debug){
         enterState(State::visible, 1.0);
     } else {
-        m_currentState = State::wait_occ;
+        m_state = State::wait_occ;
         m_transform.state.alpha = 0.0;
     }
 
     m_occludedLastFrame = false;
-    m_occlusionSolved = false;
-    m_occlusionType = OcclusionType::none;
     m_updateMeshVisibility = true;
     m_dirty = true;
     m_proxy = false;
-    m_skipTransitions = false;
     m_xAxis = glm::vec2(1.0, 0.0);
     m_yAxis = glm::vec2(0.0, 1.0);
 }
@@ -63,11 +60,13 @@ bool Label::updateScreenTransform(const glm::mat4& _mvp, const glm::vec2& _scree
         }
         case Type::line:
         {
-            // project label position from mercator world space to clip coordinates
+            // project label position from mercator world space to clip
+            // coordinates
             glm::vec4 v1 = worldToClipSpace(_mvp, glm::vec4(m_transform.modelPosition1, 0.0, 1.0));
             glm::vec4 v2 = worldToClipSpace(_mvp, glm::vec4(m_transform.modelPosition2, 0.0, 1.0));
 
-            // check whether the label is behind the camera using the perspective division factor
+            // check whether the label is behind the camera using the
+            // perspective division factor
             if (_testVisibility && (v1.w <= 0 || v2.w <= 0)) {
                 return false;
             }
@@ -126,15 +125,6 @@ bool Label::updateScreenTransform(const glm::mat4& _mvp, const glm::vec2& _scree
     return true;
 }
 
-bool Label::update(const glm::mat4& _mvp, const glm::vec2& _screenSize, float _dt, float _zoomFract) {
-    bool animate = false;
-
-    animate = updateState(_mvp, _screenSize, _dt, _zoomFract);
-    m_occlusionSolved = false;
-
-    return animate;
-}
-
 bool Label::offViewport(const glm::vec2& _screenSize) {
     const auto& quad = m_obb.getQuad();
 
@@ -148,13 +138,8 @@ bool Label::offViewport(const glm::vec2& _screenSize) {
     return true;
 }
 
-void Label::occlude(OcclusionType _occlusionType, bool _occlusion) {
-    if (!canOcclude()) {
-        return;
-    }
-
-    m_occlusionType = _occlusionType;
-    m_occludedLastFrame = _occlusion;
+void Label::occlude(bool _occlusion) {
+    m_occluded = _occlusion;
 }
 
 bool Label::canOcclude() {
@@ -162,23 +147,27 @@ bool Label::canOcclude() {
         return false;
     }
 
-    int occludeFlags = (State::visible | State::wait_occ | State::fading_in | State::sleep | State::out_of_screen);
-    return (occludeFlags & m_currentState) && !(m_type == Type::debug);
+    int occludeFlags = (State::visible |
+                        State::wait_occ |
+                        State::skip_transition |
+                        State::fading_in |
+                        State::sleep |
+                        State::out_of_screen);
+
+    return (occludeFlags & m_state) && !(m_type == Type::debug);
 }
 
 bool Label::visibleState() const {
-    int visibleFlags = (State::visible | State::fading_in | State::fading_out);
-    return (visibleFlags & m_currentState);
-}
+    int visibleFlags = (State::visible |
+                        State::fading_in |
+                        State::fading_out |
+                        State::skip_transition);
 
-void Label::occlusionSolved() {
-    m_occlusionSolved = true;
+    return (visibleFlags & m_state);
 }
 
 void Label::skipTransitions() {
-    if (!m_occlusionSolved) {
-        m_skipTransitions = true;
-    }
+    enterState(State::skip_transition, 0.0);
 }
 
 glm::vec2 Label::center() const {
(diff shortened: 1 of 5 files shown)


core/src/labels/label.cpp
```diff
@@ -228,32 +219,37 @@ void Label::pushTransform() {
...
+    bool ruleSatisfied = updateScreenTransform(_mvp, _screenSize, true);
 
     // one of the label rules has not been satisfied
     if (!ruleSatisfied) {
-
-        // go to dead state, this breaks determinism, but reduce potential label set since a lot
-        // of discarded labels are discared for line exceed (lots of tiny small lines on a curve
-        // for example), which won't have their rule satisfied
-        enterState(State::dead, 0.0);
+        if ((m_state & State::wait_occ) != 0) {
```
tallytalwar: Can you explain why all labels not satisfying rule are not put to dead state?

hjanetzek: Say a tile is initially shown at tile scale ~2. Line labels that just fit on the line 'satisfy the rule' and wait_occ may change to visible state. Then one slightly zooms out so that the rule is not satisfied - now when zooming back in again the label should also be able to show up again.

tallytalwar: :+1: nicely thought :D.

core/src/labels/labels.cpp
```diff
@@ -221,60 +208,121 @@ void Labels::update(const View& _view, float _dt,
 
     /// Collect and update labels from visible tiles
 
-    m_needUpdate = updateLabels(_styles, _tiles, _dt, dz, _view);
+    updateLabels(_styles, _tiles, _dt, dz, _view);
+
+
+    /// Mark labels to skip transitions
+
+    if (int(m_lastZoom) != int(_view.getZoom())) {
```
tallytalwar: Will it be better if we do `skipTransitions` after resolving collisions?

core/src/labels/label.h
```diff
@@ -184,25 +176,19 @@ class Label {
...
 
-    bool updateState(const glm::mat4& _mvp, const glm::vec2& _screenSize, float _dt, float _zoomFract);
-
     void setAlpha(float _alpha);
 
     bool m_proxy;
     // the current label state
-    State m_currentState;
+    State m_state;
+    State m_prevState;
```
tallytalwar: I can not see any use of this? Leftover?

core/src/labels/label.cpp
```diff
@@ -264,31 +260,40 @@ bool Label::updateState(const glm::mat4& _mvp, const glm::vec2& _screenSize, flo
         enterState(State::out_of_screen, 0.0);
     }
 
+    return true;
+}
+
+bool Label::evalState(const glm::vec2& _screenSize, float _dt) {
```
tallytalwar: Probably make sense to put asset statements here for purely illegal state transitions..

example, a label should not go from fading_in to wait_occ or dead!

core/src/labels/label.cpp
karimnaaji: Good catch on that!

It would easier to read:
```c++
if (m_state != State::wait_occ) { ... }
```

Also, what you explained is right, but I think it's still missing something since this transition would get the label stuck:
`State::wait_occ` -> `ruleSatisfied` -> `State::visible` -> `!ruleSatisfied` -> `State::dead` (no more output if `ruleSatisfied` again). 

We should handle the transition you mentioned by adding a transition from `State::dead` to `State::wait_occ` when `ruleSatisfied && m_state == State::dead` in the `evalState` switch case below.

core/src/labels/label.cpp
```diff
@@ -264,31 +260,40 @@ bool Label::updateState(const glm::mat4& _mvp, const glm::vec2& _screenSize, flo
...
+            if (m_occluded) {
+                // enterState(State::sleep, 0.0);
+                enterState(State::fading_out, m_transform.state.alpha);
+                animate = true;
                 break;
             }
             setAlpha(m_fade.update(_dt));
             animate = true;
-            if (m_fade.isFinished()) {
+            if (m_fade.isFinished() || (m_transform.state.alpha > 0.9f)) {
```
karimnaaji: Did you notice any transition not working properly while only testing `m_fade.isFinished()`?

core/src/labels/labels.cpp
```diff
@@ -56,54 +56,53 @@ bool Labels::updateLabels(const std::vector<std::unique_ptr<Style>>& _styles,
...
+                    continue;
+                }
+
+                float d2 = glm::distance2(l0->transform().state.screenPos,
+                                          l1->transform().state.screenPos);
+
+                // The new label lies within the circle defined by the bbox of l0
+                if (sqrt(d2) < std::max(l0->dimension().x, l0->dimension().y)) {
+
+                    auto* t = dynamic_cast<const TextLabel*>(l0.get());
```
karimnaaji: To be removed once ready?

core/src/labels/label.cpp
tallytalwar: @karimnaaji made me think more... actually this should be

If the state is `wait_occ` and rule is **not** satisfied, then label should go to `dead` state. In every other label state it should go to `sleep` state when a rule is not satisfied.. (run this with the example you mentioned).

So:
```
     if (!ruleSatisfied) {
        if (m_state == State::wait_occ) {
            // go to dead state, this breaks determinism, but reduce potential
            // label set since a lot of discarded labels are discared for line
            // exceed (lots of tiny small lines on a curve for example), which
            // won't have their rule satisfied
            enterState(State::dead, 0.0);
            pushTransform();
        } else {
            enterState(State::sleep, 0.0);
            pushTransform();
        }
        return false;

    }
```

tallytalwar:
there is a bunch of commented old code + some LOG messages. Should clean these before merging.

tallytalwar:
Also not sure if `skipTransition` is working fully.
skip-transition

core/src/labels/label.cpp
tallytalwar: It's exactly what you have,  I guess :).

core/src/labels/label.cpp
hjanetzek: It seems to work properly. But I noticed that label are much quicker fully visible than the fade effect that takes 1 second in the default theme. See in the debug mode how long the bbox stay yellow compared to when the label seems to be faded in. So one could stop the animation at the point when no difference is visible instead of the end of the animation duration.

hjanetzek:
@tallytalwar it works much better - see all the other labels :)
But there are a few things that could definitely be improved: Mainly the lookup of proxies only checks the cached, but currently active tiles are not in the TileCache - when a parent tile is still used as proxy for a neighbor tile it will not be used in skipTransitions at the moment. 
Also the check that all label properties must match is too specific in my opinion. In eraser-map style label properties change at almost each zoom-level, so no transitions could be skipped there.

core/src/labels/label.h
hjanetzek: Yes, used for debugging and visualization

core/src/labels/labels.cpp
hjanetzek: I wanted to give labels that were previously visible a slightly larger chance to stay visible with this order: Label in skip_transition state are also considered as in visibleState() for the conflict resolution in labels.cpp

tallytalwar: I feel that doing the reverse and slight tweaking to `evalState` might give better result with `skip transition`.

tallytalwar: Something like this might work better:

- checkSkipTransition after collision resolution.
 - this will make sure only "non-colliding/visible" labels go to skip-transition state
- in evalState, if a label's state is skipTransition... set its state to "visible".

hjanetzek: Give it a try :) But when doing the conflict resolution before checking possible skip transitions the chance is greater to hit the last case in conflict resolution (labels.cpp line:~275)

karimnaaji:
>Also the check that all label properties must match is too specific in my opinion. 
I agree, we should have another hash/way to compare that is more loose. Especially since when using stops this technique is not working.

karimnaaji:
@tallytalwar  I don't think there's any `skipTransition` active in the gif, otherwise the red lines would get splited for the next zoom level, the strange transition might be something else.

karimnaaji:
(Since I think what you're pointing out is Bridge St. label)

karimnaaji:
This was good to merge to me (except if @tallytalwar wanted to address what he mentioned). But I noticed some regression since yesterday.
Skipping the transition was still working but now seems broken?

tallytalwar:
Sure this was mostly good to me. So go for it. Will try my suggestions when
I am back. Thanks

hjanetzek:
@karimnaaji works just as good as before for me. 

Hold on - I'll add a method to lookup proxies that are currently in _tiles (and therefore not in _cache) for skipTransitions

karimnaaji:
Yes, not sure what I was looking at probably not focused on the right labels, works fine!
Merging once travis is good.

PR #498, opened 2016-01-25
add TileBuilder

hjanetzek:
This PR adds a TileBuilder to each TileWorker thread. 
So far the TileBuilder itself is only a refactoring of existing parts to have one place to hold all structures used for tile building. This is the groundwork for StyleBuilder and TileDataSink branches.

- Tile::build() moved to TileBuilder
- TileBuilder holds a Scene reference, and embeds a StyleContext and DrawRuleMergeSet.
- Finally save scene reloading is possible. Old TileBuilder instances are disposed on the main-loop.

[Review on Reviewable](https://reviewable.io/reviews/tangrams/tangram-es/498)

### DIFF
--- a/core/src/tangram.cpp
+++ b/core/src/tangram.cpp
@@ -27,12 +27,16 @@
 #include <cmath>
 #include <bitset>
 #include <mutex>
+#include <queue>
 
 
 namespace Tangram {
 
 const static size_t MAX_WORKERS = 2;
 
+std::mutex m_tilesMutex;
+std::mutex m_tasksMutex;
+std::queue<std::function<void()>> m_tasks;
 std::unique_ptr<TileManager> m_tileManager;
 std::unique_ptr<TileWorker> m_tileWorker;
 std::shared_ptr<Scene> m_scene;
@@ -40,7 +44,6 @@ std::shared_ptr<View> m_view;
 std::unique_ptr<Labels> m_labels;
 std::unique_ptr<Skybox> m_skybox;
 std::unique_ptr<InputHandler> m_inputHandler;
-std::mutex m_tilesMutex;
 
 std::array<Ease, 4> m_eases;
 enum class EaseField { position, zoom, rotation, tilt };
@@ -150,6 +153,14 @@ void update(float _dt) {
 
     m_view->update();
 
+    {
+        std::lock_guard<std::mutex> lock(m_tasksMutex);
+
+        while (!m_tasks.empty()) {
+            m_tasks.front()();
+            m_tasks.pop();
+        }
+    }
     {
         std::lock_guard<std::mutex> lock(m_tilesMutex);
         ViewState viewState {
@@ -546,6 +557,11 @@ void setupGL() {
     GLExtensions::printAvailableExtensions();
 
     while (Error::hadGlError("Tangram::setupGL()")) {}
+}
+
+void runOnMainLoop(std::function<void()> _task) {
+    std::lock_guard<std::mutex> lock(m_tasksMutex);
+    m_tasks.emplace(std::move(_task));
 }
 
 }
--- a/core/src/tile/tileBuilder.h
+++ b/core/src/tile/tileBuilder.h
@@ -1,0 +1,39 @@
+#pragma once
+
+#include "data/dataSource.h"
+#include "scene/styleContext.h"
+#include "scene/drawRule.h"
+
+namespace Tangram {
+
+class DataLayer;
+class DataSource;
+class Tile;
+struct TileData;
+
+class TileBuilder {
+
+public:
+
+    TileBuilder();
+
+    ~TileBuilder();
+
+    void setScene(std::shared_ptr<Scene> _scene);
+
+    std::shared_ptr<Tile> build(TileID _tileID, const TileData& _data, const DataSource& _source);
+
+    const Scene& scene() const { return *m_scene; }
+
+    Tile& tile() const { return *m_tile; }
+
+private:
+    std::shared_ptr<Scene> m_scene;
+
+    StyleContext m_styleContext;
+    DrawRuleMergeSet m_ruleSet;
+
+    std::shared_ptr<Tile> m_tile;
+};
+
+}
--- a/core/src/tile/tileWorker.cpp
+++ b/core/src/tile/tileWorker.cpp
@@ -5,9 +5,9 @@
 #include "tile/tile.h"
 #include "view/view.h"
 #include "scene/scene.h"
-#include "scene/styleContext.h"
 #include "tile/tileID.h"
 #include "tile/tileTask.h"
+#include "tile/tileBuilder.h"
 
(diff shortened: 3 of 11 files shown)


core/src/tile/tileWorker.cpp
```diff
@@ -30,11 +33,17 @@ TileWorker::~TileWorker(){
     }
 }
 
-void TileWorker::run() {
+void disposeBuilder(std::unique_ptr<TileBuilder> _builder) {
+    if (_builder) {
+        Tangram::runOnMainLoop([b = std::shared_ptr<TileBuilder>(std::move(_builder))](){});
```
blair1618: Let me make sure I understand this bit: The lambda here will create a `shared_ptr` that takes ownership of `_builder` and then immediately goes out of scope? I assume there's a reason for doing this instead of just resetting `_builder` - a comment here might be nice.

hjanetzek: The shared_ptr is stored in the function (to which this lamba gets converted). The function runs on the next mainloop iteration and so the last shared_ptr is destructed with it.
std::function must be copyable, that's why _builder is upgraded to a shared_ptr. 

I think a compiler is not allowed to optimize any of this away.

core/src/tile/tileBuilder.h
```diff
@@ -0,0 +1,44 @@
...
+
+    Tile& tile() const { return *m_tile; }
+
+private:
+    std::shared_ptr<Scene> m_scene;
+
+    StyleContext m_styleContext;
+    DrawRuleMergeSet m_ruleSet;
+
+    std::vector<const DataLayer*> m_activeLayers;
```
blair1618: Seems unused. Part of another branch's work?

core/src/tile/tileBuilder.h
```diff
@@ -0,0 +1,44 @@
...
+
+class TileBuilder {
+
+public:
+
+    TileBuilder();
+
+    ~TileBuilder();
+
+    StyleBuilder* getStyleBuilder(const std::string& _name);
```
blair1618: Also unused. For style-builder branch I would guess?

core/src/tile/tileBuilder.h
hjanetzek: yeah - I forgot to add tileBuilder.h for a while in the large tile+style-builder branch... This is belongs tile-data-sink, the other to style-builder

blair1618:
Hmm. This branch is now reliably crashing at exit on OS X. Looks like a `ShaderProgram` is being deleted on a worker thread.

blair1618:
Alright, everything looks good here now; I'll merge once Travis passes.

tallytalwar:
@blair1618 can you hold on this, want to give this a quick look. Thanks

core/src/tile/tileWorker.cpp
```diff
@@ -78,26 +108,17 @@ void TileWorker::run() {
...
 
-        context.initFunctions(*scene);
-
         if (tileData) {
-            auto tile = std::make_shared<Tile>(task->tileId(),
-                                               *scene->mapProjection(),
-                                               &task->source());
 
-            tile->build(context, *scene, *tileData, task->source());
+            auto&& tile = builder->build(task->tileId(), *tileData, task->source());
```
tallytalwar: I don't think you need to have a move reference on the lhs here.
Follow should suffice.
`auto tile = builder->build(task->tileId(), *tileData, task->source());`

core/src/tangram.cpp
```diff
@@ -548,4 +559,9 @@ void setupGL() {
     while (Error::hadGlError("Tangram::setupGL()")) {}
 }
 
+void runOnMainLoop(std::function<void()> _task) {
+    std::lock_guard<std::mutex> lock(m_tasksMutex);
+    m_tasks.emplace(std::move(_task));
```
tallytalwar: Shouldn't `m_tasks` be cleared on a new scene load or a context loss?

Currently that means, old scene references will still be available in previous scene's `TileBuilder` instances living in `m_tasks`.

blair1618: `m_tasks` is emptied on every main loop. Currently the only tasks that go into this queue are lambdas that delete `TileBuilder` instances. Enqueueing those instances for deletion on the main thread is the fastest way that we can safely delete them.

core/src/tile/tileWorker.cpp
blair1618: I would guess that this is a minor optimization to prevent the construction and destruction of a `shared_ptr`.

core/src/tangram.cpp
tallytalwar: :+1:

core/src/tile/tileWorker.cpp
tallytalwar: `TileBuilder::build` is doing a `std::move` on the `shared_ptr` of the `Tile`, should it should not destruct the previous instance or construct a new one.

tallytalwar: To the best of knowledge, a rvalue reference make sense on the rhs!! It has no affect on lhs.

blair1618: It is true that in either case the `Tile` being pointed to will be neither constructed nor destructed, but creating a `shared_ptr` can have noticeable cost in tight loops, since they are made to be thread-safe. 

`auto tile = builder->...` would create a local `shared_ptr`. `std::move` transfers ownership of contents, but there would still be a `shared_ptr` on the stack that needs to be destructed (consider that it would still be legal to re-assign `tile` after the move).

tallytalwar: Not if the function `TileBuilder::build` signature is made to return a rvalue!

hjanetzek: ok, I'll leave out this trickery in future unless a benchmark proves a difference :)

The intention was to not create a local shared_ptr variable but to hand it over directly

tallytalwar: Yup, I really do not think destructing of an empty shared_ptr will have any performance impact.
Based on scott meyers reference above, it should have no difference at all! I will be happy to be proven wrong by benchmarks :D.

hjanetzek: It's about the shared_ptr on the stack that needs to be constructed and moved out of (unless the compiler is clever and optimizes it out)

tallytalwar: IMO, RVO will take care of the trickery.

blair1618: Ok... since we at least agree that this line is not *worse* than an alternative, it sounds like we're ok to merge this.

tallytalwar: In a simple test example the assembly generated with this way is 2 instructions more, namely 1 `lea` and 1 `mov` :D. So compiler is doing some return value optimization.

But I am live with it :D.

PR #499, opened 2016-01-25
add StyleBuilder

hjanetzek:
continuing #496, split-up of #481

This PR adds the StyleBuilder interface to create builder instances for Styles per worker thread and be able to reuse resources used during building of meshes, labels, etc.

Please look at the changes for each commit as they are quite large.. sorry


hjanetzek:
@blair1618, @tallytalwar could you review TileBuilder and StyleBuilder PR first? As they change quite a lot of files it would be good to avoid further unmergable commits

tallytalwar:
Starting to review this now!

PR #501, opened 2016-01-25
Shutdown GLFW properly

hjanetzek:
Explicity stop UrlWorkers before other static globals may be destructed. This seems to fix occasional crashes on shutdown for linux. 

While at it I've added a shutdown method that ensures that TileWorker is stopped before TileManager gets destructed. It may be sufficient though to change the implicit order of destruction by switching the lines 

```
std::unique_ptr<TileManager> m_tileManager;
std::unique_ptr<TileWorker> m_tileWorker;
```

... no it was the global m_tasks queue introduced in TileBuilder branch which was destructed to early.
Still having an explicit order might be good. In case someone shuffles global declarations in tangram.cpp

### DIFF
--- a/linux/src/main.cpp
+++ b/linux/src/main.cpp
@@ -8,7 +8,6 @@
 #include <sys/types.h>
 #include <sys/stat.h>
 #include <cstdlib>
-
 #include <signal.h>
 
 
@@ -267,7 +266,16 @@ int main(int argc, char* argv[]) {
 
     static bool keepRunning = true;
 
-    signal(SIGINT, [](int) { keepRunning = false; });
+    // Give it a chance to shutdown cleanly on CTRL-C
+    signal(SIGINT, [](int) {
+            if (keepRunning) {
+                logMsg("shutdown\n");
+                keepRunning = false;
+                glfwPostEmptyEvent();
+            } else {
+                logMsg("killed!\n");
+                exit(1);
+            }});
 
     int argi = 0;
     while (++argi < argc) {
--- a/linux/src/main.cpp
+++ b/linux/src/main.cpp
@@ -332,6 +332,8 @@ int main(int argc, char* argv[]) {
         }
     }
 
+    finishUrlRequests();
+
     curl_global_cleanup();
     glfwTerminate();
     return 0;
--- a/linux/src/main.cpp
+++ b/linux/src/main.cpp
@@ -21,6 +21,7 @@ std::string sceneFile = "scene.yaml";
 GLFWwindow* main_window = nullptr;
 int width = 800;
 int height = 600;
+bool recreate_context;
 
 // Input handling
 // ==============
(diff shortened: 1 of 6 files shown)


tallytalwar:
Need to check this on linux, will reboot and try.

tallytalwar:
@hjanetzek though I had not seen these crashes on shutdown, however this does not fix random crashes on triggering context loss (which is triggered from `glfwWaitEvents()` in `linux/src/main.cpp`).

Logs:
```
#0  0x00007ffff746ea90 in Xutf8LookupString () from /usr/lib/x86_64-linux-gnu/libX11.so.6
#1  0x00000000008fa6a1 in processEvent (event=0x7fffffffddc0) at /home/varun/Development/tangram-es/external/glfw/src/x11_window.c:897
#2  _glfwPlatformPollEvents () at /home/varun/Development/tangram-es/external/glfw/src/x11_window.c:1784
#3  0x0000000000763d9e in main (argc=1, argv=0x7fffffffe1c8) at /home/varun/Development/tangram-es/linux/src/main.cpp:320

```

linux/src/main.cpp
```diff
@@ -329,6 +329,10 @@ int main(int argc, char* argv[]) {
         }
     }
 
+    finishUrlRequests();
+
+    Tangram::shutdown();
```
tallytalwar: We also do this to do a proper/explicit shutdown on osx.

hjanetzek: This shutdown method can be removed I guess. It only needs the stopping of urlWorkers before tileManager gets disposed to not call its tileData callback

tallytalwar: There is no hard having an explicit control on destroying individual tangram blocks!

hjanetzek:
The simulated context loss is a bit brutal. Killing the window while in a glfw function that might have pending events for that window :) 

I think it could help to set a flag 'recreate_context' and check it after glfwWait/PollEvents

hjanetzek:
hold on. I'll update this PR

tallytalwar:
I like that idea to avoid the brutality of simulated context loss. (Another PR though).

tallytalwar:
:+1: 

I am on linux now so let me know whenever you done and can try things out.

hjanetzek:
@tallytalwar hope it's ok to have context recreation fix in this PR.

tallytalwar:
@hjanetzek sure.

tallytalwar:
Looks good to me.

osx/src/main.cpp
```diff
@@ -3,7 +3,9 @@
 #include "data/clientGeoJsonSource.h"
 #include <cmath>
 #include <memory>
-
+#include <signal.h>
+#include <stdlib.h>
```
blair1618: ".h" system headers are technically deprecated, prefer `<csignal>`, `<cstdlib>`, etc. in the future.

blair1618:
For future reference, it is actually quite safe to destroy a GLFW window within a callback http://www.glfw.org/docs/latest/window.html#window_destruction

PR #504, opened 2016-01-26
Use rounding when encoding position attributes as integers

blair1618:
Fixes some small (but visible) discrepancies in vertex positions at the edges of tiles.


tallytalwar:
LGTM.

PR #505, opened 2016-01-26
Fixes #502

### LINKED ISSUE #502: Reverse VBO order for line and outline, so that outline is drawn before line.

tallytalwar:


tallytalwar:
@blair1618 can you review this please? Thanks

core/src/style/polylineStyle.cpp
```diff
@@ -214,38 +214,37 @@ void PolylineStyle::buildMesh(const Line& _line, Parameters& _params, VboMesh& _
...
-            for (size_t i = 0; i < offset; i++) {
-                vertices.push_back({ vertices[i], order, { width, slope }, color });
-            }
+    // Must update existing variables, they are captured (and used) by the builder
+    color = _params.fill.color;
+    width = _params.fill.width;
+    order = _params.fill.order;
+    slope = _params.fill.slope;
+
+    if (!_params.outlineOn || _params.outline.cap != _params.fill.cap || _params.outline.join != _params.fill.join) {
```
blair1618: We will also need to triangulate the fill if `_params.outline.width <= 0.f` and `_params.outline.slope <= 0.f`, to correctly negate the conditional on line 222. To do this concisely it may be nicer to set a boolean when the outline is built.

tallytalwar: That's a good catch. Will fix it.

hjanetzek: A simpler change would be to keep the original order. Before processing the outline move out the builder indices to, say, fillIndices and vertices to fillVertices. At the end add the meshes in the intended order

```
mesh.addVertices(std::move(vertices), std::move(builder.indices));
mesh.addVertices(std::move(fillVertices), std::move(fillIndices));
```

tallytalwar: Thanks @hjanetzek. Incorporated your suggestion here: fd297b0

blair1618:
All good here.

PR #506, opened 2016-01-27
[rfc] alfons style

hjanetzek:
Cooking text soup with alfons the friendly font chef.

- [x] Splitting of text into runs by direction and script, use harfbuzz shape on individual runs and merge resulting glyph shapes in visual order. 
- [x] Font fallbacks for missing glyphs
- [x] BIDI text
- [x] No limitation on max texture size since additional atlases are added when a texture fills up
- [x] Refcounting of GlyphAtlas to reuse those not referenced by any LabelMesh
- [x] Packing only visible labels into VBOs
- [x] Batch drawing per GlyphAtlas (and Style).  
- [x] Simple line wrapping

Missing ingredients:
- [x] Support loading fonts defined in scenes
- [x] Line-wrapping based on character count
- [x] Alignment of multi-line labels
- [ ] reduce distance between text-lines
- [ ] Create Glyphs with the original font size (no texture downscaling) to guarantee a given stroke width.
- [ ] Sprite shader extrude depends on tile zoom - move to attribute (there is room left in alpha)
   
TODO
- [ ] clean up the kitchen
- [ ] retarget merge to master 
- [x] adjust the bounding box as in master branch
- [x] debug text style


blair1618:
Care to explain what "alfons" is/does?

hjanetzek:
@blair1618  sorry, I should have added that it's something @karimnaaji and I were discussing to use as font library.

Alfons is the font utility/text-shaping/glyph-atlas library that I developed during my thesis. Over the holidays I got it finally into shape to share it publicly. 

At this point it seems to be still quite a bit of work to complete our fontstash shaping, font fallbacks, etc.
   
So the question is whether it is worth to put this effort into alfons which already has BIDI text and fallbacks working - and I'm overall quite happy with the interface and modularity of the components now. 

https://github.com/hjanetzek/alfons/tree/wip

karimnaaji:
@hjanetzek looking good from what I've seen so far, will look into more details the implementation next week!

core/src/style/alfonsStyle.cpp
```diff
@@ -0,0 +1,832 @@
...
+
+    uint32_t strokeWidth = (_params.strokeWidth / 3.f * 255. * .25) / fontScale;
+
+    m_scratch.stroke = (_params.strokeColor & 0x00ffffff) + (strokeWidth << 24);
+
+    {
+        auto ctx = m_style.context();
+
+        std::lock_guard<std::mutex> lock(ctx->m_mutex);
+        auto line = m_shaper.shape(ctx->m_font, _params.text);
```
karimnaaji: For a first integration using advanced C++ features for the alfons API makes sense. But just for a future note I'd rather not force the user of alfons use C++11/14 features, it's a nice to have within alfons for auto pointers and syntax sugar; but for a more portable usage of the library it would make sense to only expose a more early version of C++ (through headers / API), it might be hard right now because of the number of inclusion files we need for a standard shaping, but we can think of that more if it makes sense to you.

karimnaaji:
Can we keep 36f658d (part of https://github.com/tangrams/tangram-es/pull/497) out of this branch, so that it doesn't interfere too much with the integration.

hjanetzek:
@karimnaaji yes, the pulled in features branches should be merged in advance. This branch is Very WIP so far, so the alfons commits can be picked to a separate branch once we get nearer to merge it.

core/src/style/alfonsStyle.cpp
hjanetzek: Do you mean using shared_ptr in the API?

I think when using c++ one should not care about pre c+11. Otherwise I would rather write it in C :)

karimnaaji: I'm not convinced about "not care about pre c++11" honestly, it should be totally fine to be able to integrate alfons in some code that uses pre c++11. 
>Otherwise I would rather write it in C :)

I'm not saying to write aflons in C :) alfons could be totally written C++14. What I'm suggesting is to think more of what to expose as a more gentle API (restricting the features of modern C++ used in what is exposed) to be able to integrate alfons in a more straightforward way.

karimnaaji:
Cool, makes sense :+1:

core/src/style/alfonsStyle.cpp
blair1618: Some thoughts on this:

1. There's a reason we use C++11/14 features in our code, they make things more convenient or safe or sometimes just possible! If we don't expose that same level of niceness in the interface, that's a cost to everyone who uses the library. 

2. Limiting the interface to C++03 would essentially add another compatibility layer on top of the "natural" C++11 interface. I think if we encountered a case where we needed to meet an old standard (game console, smart fridge, etc.) *then* we could go to the effort of making an additional interface to meet those requirements. 

3. As part of a coding community, I think we should make our code clearly take the opinion that we are moving forward with the language, and thereby encourage everyone else to adopt the new standards as well.

karimnaaji: To clarify the confusion here, I'm only speaking about the alfons interface.

blair1618: Me too :)

karimnaaji: I agree on the point you mentioned, though I quite disagree for the interface; I don't see why someone would have to change compilation flags on his code while integrating something like alfons simply because the headers uses C++14, being clever on what to use in the headers the user has to include and what's being returned shouldn't be too much effort.

tallytalwar: As far as implementation goes, we should be pushing ourselves to use newer language features, as these help us in many ways (as pointed by @blair1618 above). 
But with respect to API design and exposing functionality to user, we should see from a user's perspective. It should be easy for "any" user to start using our API (which may not be the case if we expose c++11/14 features with our API). It will also help in adoptibility (if thats a word!!), if the API is flexible with what most users can use.

hjanetzek: @karimnaaji on which level do you suggest to hide c++11 features from the interface? If the public headers contain structs with share_ptr one has to compile the application using it with -std=c++11. 

As for alfons, please let us focus on how the api should work, how to make it safe and nice. And postpone considerations how to expose it to older c++ or c.

karimnaaji: Anything not supported by visual-c++ should be **out**.
@hjanetzek I know, this started as a future note for things we have to consider.

tallytalwar: @hjanetzek yes agreed, there are bigger fish to fry right now, so lets fry them first :D.

hjanetzek: @karimnaaji which feature exactly do you mean?

hjanetzek: @karimnaaji I think one could separate it into the current modular API and a simplified version provided by an Alfons class that keeps all the modules FontManager, TextShaper, GlyphAtlas, etc internal and only gives out ID handles to Fonts.

karimnaaji: @hjanetzek Makes sense, I like the inheritance way to handle callbacks. 

I guess what the only thing the user would need in that sense is two inheritance implementation `alf::TextureCallback` (handling texture updates) and `alf::MeshCallback` (handling vertex updates) and a usage of `alf::FontID` (managing the currently used font), all handled through an Alfons class/context. 

Which seems like a pretty reasonable starting API?

hjanetzek: Yes, something like this will be handy for simple use cases. Though my intention was to also enable modular use. Say, to have one textShaper per worker thread (synchronized per FontFace). Or to use the FontManager, TextShaper and GlyphAtlas with different FontFace implementations.

karimnaaji:
As we talked about and since this will become hard to review as we go further; to summarize some of the last changes on this branch, 
- Adding an option to load font from memory to alfons (previously using Freetype file reader)
- Use of the stylesheet described fonts.
- Merge the textMesh which has been added while integrating alfons for a single labelMesh class, to more easily trace the vertex data flow.
- Reduce copy for the sprites removing the mesh compilation step.
- Now uploading the full buffer at each frame (mapping buffer when available or glBufferData as it was done previously). We're not tracking anymore the dirty offsets since the set of initially created labels has been reduced, which makes the initial buffer much smaller. (this would probably need some benchmark to see the benefit of not tracking the dirty ranges).
Some things that still need to be addressed:
- More cleanup; trying to unify the concepts to get closer to the architecture currently in master (trying to remove labelSet and labelContainer if possible, remove multiple meshes from the style for a single mesh owned by the tile (wip https://github.com/tangrams/tangram-es/tree/wip-alfons-single-mesh).

hjanetzek:
@karimnaaji I think the approach with having meshes owned by the style is favorable to having meshes per tile. 
- It reduces the number of draw calls - i.e. one per texture (or less with multi-texturing) instead of one per tile and texture. 
- Much less VBOs used

LabelSet and LabelContainer can be merged again when SpriteStyle and TextStyle use the same rendering approach.

karimnaaji:
@hjanetzek sorry didn't mean to say that, I also think this is a better approach; what I meant is to still keep a back reference from the tile to the shared mesh. We can keep a flag that within the mesh to ensure it's drawn once per style (so that the draw call takes place and not in onEndDrawFrame), otherwise it can get quite confusing.

What's in progress [here](https://github.com/tangrams/tangram-es/tree/wip-alfons-single-mesh) is moving from one vbo per atlas to one vbo with multi texturing.

karimnaaji:
@hjanetzek if you're good with https://github.com/tangrams/tangram-es/tree/wip-alfons-single-mesh I'll rebase and add the changes on top of this branch. It's now using one mesh even if several texture atlases are used (up to 4 atlases for now but I'll make this more robust if you're good with this change).

hjanetzek:
@karimnaaji as the multitexturing is an optimization that adds complexity I would like to postpone this.
With just a single fontsize one can easily get 10 or more active 256px textures (saw this in testing with eraser-map style, and switching between various places having cjk script labels). Making the textures larger increases the chance of holding onto atlases with very few used glyphs. 
So one could still require multiple meshes even with multi-texturing.

karimnaaji:
Ok, I was a bit concerned because of eraser-map specifically, since panning around for a bit and you get to a lot of atlases already. But yeah let's postpone this and consolidate later.

hjanetzek:
@karimnaaji, can we discuss the need for StyleMesh in PR #523  that introduces it :)

karimnaaji:
@hjanetzek sure, that confused me since I thought this was introduced by this PR for the need of the integration. I'm reviewing #523 then.

TKJohn:
It would be nice to have a feature called  'Vertical text', which is a very common in CJK displays.
This is used to show the label of vertical line. 

It looks like ('Vertical text' in Chinese  '竖排文字')
|竖
|排
|文
|字

In English characters it looks like (of course it's ugly to show non CKJ in this way)
|V
|e
|r
|t
...

bcamper:
@TKJohn thanks for the suggestion, this makes sense as a future enhancement, I have created an issue for it here: https://github.com/tangrams/tangram-es/issues/556

hjanetzek:
@TKJohn yes, it should be possible to implement this as an additional line wrapping method for text with paragraph direction top-bottom in future. 
A quick solution might be to insert spaces between cjk codepoints (to make the string breakable) and set text_wrap to 1 :)
https://github.com/tangrams/tangram-es/commit/0a75cae0e5305eaf68caeb566b3dbe41e3a14827

PR #507, opened 2016-01-27
line shader - improve proxy scaling

hjanetzek:
Just a small tweak to prevent lines from proxy tiles scaling becoming too thick.

### DIFF
--- a/core/resources/shaders/polyline.vs
+++ b/core/resources/shaders/polyline.vs
@@ -69,9 +69,9 @@ void main() {
         vec4 extrude = UNPACK_EXTRUSION(a_extrude);
         float width = extrude.z;
         float dwdz = extrude.w;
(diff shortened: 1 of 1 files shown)


blair1618:
Yeah, line extrusion gets a little fishy when tiles become proxies. I don't know that there's currently a way to get the extrusion distances to perfectly interpolate across width stops, since each vertex is only aware of it's own segment of the piecewise function, but this seems to improve the behavior somewhat. +1

hjanetzek:
yes, the problem is that we can only encode a linear slope that may change depending on the zoom, so one cannot scale the 'slope' properly over more than one zoomlevel. I guess that's want you mean by the piecewise function, right? This shows especially when outlines have a larger slope when they appear like [13, 0px],[14, 1px],[18,1px], where a tile from z13 would have a 4px outline at z16.

PR #508, opened 2016-01-27
Check drawable codepoints

karimnaaji:
Since https://github.com/tangrams/tangram-es/pull/475 is on hold this adds the feature to detect non-renderable codepoints while not using Harfbuzz.
Addressing https://github.com/mapzen/eraser-map/issues/101.


karimnaaji:
@hjanetzek good to go?

PR #512, opened 2016-01-28
Drawrule optimizations

hjanetzek:
- The most expensive part in merging was to zero out the params-,layer- and depth arrays on DrawRule initialization. Using a bitset reduces this to 4 bytes vs previously 480 or 960 bytes (on 64bit arch)

- Instead of creating a set a all matched layer strings to calculate the label repeatGroup hash, drawRule got a new helper functions to calculate the hash without allocations.

### DIFF
--- a/core/src/scene/drawRule.h
+++ b/core/src/scene/drawRule.h
@@ -50,8 +50,13 @@ struct DrawRuleData {
 struct DrawRule {
 
     const StyleParam* params[StyleParamKeySize] = { nullptr };
-    const char*       layers[StyleParamKeySize] = { nullptr };
-    size_t            depths[StyleParamKeySize] = { 0 };
+
+    struct Layer {
+        const char* name = nullptr;
+        size_t depth = 0;
+    };
+
+    Layer layers[StyleParamKeySize];
 
     const std::string* name = nullptr;
 
--- a/core/src/scene/drawRule.h
+++ b/core/src/scene/drawRule.h
@@ -37,12 +37,15 @@ class StyleContext;
 
 struct DrawRuleData {
 
+    // https://github.com/tangrams/tangram-docs/blob/gh-pages/pages/draw.md#style-parameters
     std::vector<StyleParam> parameters;
+
+    // draw-rule name (and assigned id)
+    // https://github.com/tangrams/tangram-docs/blob/gh-pages/pages/draw.md#draw-rule
     std::string name;
     int id;
 
-    DrawRuleData(std::string _name, int _id,
-                 const std::vector<StyleParam>& _parameters);
+    DrawRuleData(std::string _name, int _id, std::vector<StyleParam> _parameters);
 
     std::string toString() const;
 
--- a/core/src/scene/drawRule.h
+++ b/core/src/scene/drawRule.h
@@ -94,8 +94,9 @@ private:
 
 };
 
-struct DrawRuleMergeSet {
+class DrawRuleMergeSet {
 
+public:
     /* Determine and apply DrawRules for a @_feature and add
      * the result to @_tile
      */
@@ -108,12 +109,15 @@ struct DrawRuleMergeSet {
     // internal
     void mergeRules(const SceneLayer& _layer);
 
+    auto& matchedRules() { return m_matchedRules; }
+
+private:
     // Reusable containers 'matchedRules' and 'queuedLayers'
-    std::vector<DrawRule> matchedRules;
-    std::vector<const SceneLayer*> queuedLayers;
+    std::vector<DrawRule> m_matchedRules;
+    std::vector<const SceneLayer*> m_queuedLayers;
 
     // Container for dynamically-evaluated parameters
-    StyleParam evaluated[StyleParamKeySize];
+    StyleParam m_evaluated[StyleParamKeySize];
 
 };
 
--- a/core/src/scene/drawRule.h
+++ b/core/src/scene/drawRule.h
@@ -50,17 +50,22 @@ struct DrawRuleData {
 
 struct DrawRule {
 
-    std::bitset<StyleParamKeySize> active = { 0 };
-
-    struct Param {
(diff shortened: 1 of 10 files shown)


core/src/scene/drawRule.h
```diff
@@ -37,11 +38,10 @@ class StyleContext;
 struct DrawRuleData {
 
     std::vector<StyleParam> parameters;
-    std::string name;
-    int id;
+    std::string styleName;
+    int styleId;
```
blair1618: In general, the name of a draw rule is **not** the same as the name of the style used by the rule. If a style is not explicitly specified then it will be inferred from the name of the rule, but these things are _necessarily distinct_ and the names of variables should clearly reflect that.

hjanetzek: Ok, I see, DrawRule.name for the rule name makes sense. reverting and adding a comment for my future self.. :)

tallytalwar: :+1: This also causes a great deal of confusion between js and es implementations so a comment is good for everyone :D.

blair1618: Thanks! I'll finish reviewing `style-builder` before I come back to this, just wanted to clear up any potential confusion here.

hjanetzek: It seems the tangram-docs need to be updated to state that ```draw-style``` may be an arbitrary string when a ```style``` parameter is provided.
In my opinion styleName is not that bad - Either draw-style references a style name or it implicitly defines a new style name (inheriting parameters from a predefined style)
Or ```name``` member  could be renamed to ```drawStyle``` for matching the docs 1-to-1

https://github.com/tangrams/tangram-docs/blob/gh-pages/pages/draw.md#draw-style
https://github.com/tangrams/tangram-docs/blob/gh-pages/pages/draw.md#style

blair1618: Oh wow the docs are suuuper misleading here. I'll bring up this nomenclature with the JS and docs folks so we can clear it up. I don't think we should adapt our code to mimic confusing documentation, we should rather clarify the documentation itself.

core/src/scene/sceneLoader.cpp
```diff
@@ -1341,10 +1341,10 @@ SceneLayer SceneLoader::loadSublayer(Node layer, const std::string& name, Scene&
                 std::vector<StyleParam> params;
                 parseStyleParams(ruleNode.second, scene, "", params);
 
-                auto name = ruleNode.first.as<std::string>();
-                int nameId = scene.addIdForName(name);
+                auto drawStyle = ruleNode.first.as<std::string>();
```
tallytalwar: Since https://github.com/tangrams/tangram-docs/pull/65 is in sight of making the documentation unambiguous, can we have this named to ruleName or something on those lines?

tallytalwar:
So I started to profile this PR on osx and found that its not drawing the right thing, all roads are missing.
@hjanetzek can you also see the same behavior?

hjanetzek:
Yeah.. hadn't checked the last rebase. fixed now

hjanetzek:
```
curl  https://vector.mapzen.com/osm/all/11/1100/671.mvt | gunzip > tile.mvt
perf record -g ./bench/tileLoading.out --benchmark_repetitions=10
```

```
   - 41.56% Tangram::DrawRuleMergeSet::apply                                                                                                                                                                                                                                                                                  ▒
      + 27.83% Tangram::StyleBuilder::addFeature                                                                                                                                                                                                                                                                              ▒
      - 8.56% Tangram::DrawRuleMergeSet::match                                                                                                                                                                                                                                                                                ▒
         + 5.81% Tangram::Filter::eval                                                                                                                                                                                                                                                                                        ▒
         + 1.32% Tangram::DrawRuleMergeSet::mergeRules                                                                                                                                                                                                                                                                        ▒
         + 0.41% Tangram::StyleContext::setFeature                                                                                                                                                                                                                                                                            ▒
         + 0.16% strlen
```

drawrule-optimization:
```
   - 40.57% Tangram::DrawRuleMergeSet::apply                                                                                                                                                                                                                                                                                  ▒
      + 27.10% Tangram::StyleBuilder::addFeature                                                                                                                                                                                                                                                                              ▒
      - 7.77% Tangram::DrawRuleMergeSet::match                                                                                                                                                                                                                                                                                ▒
         + 5.54% Tangram::Filter::eval                                                                                                                                                                                                                                                                                        ▒
           0.62% Tangram::DrawRuleMergeSet::mergeRules                                                                                                                                                                                                                                                                        ▒
         + 0.43% Tangram::StyleContext::setFeature                                                                                                                                                                                                                                                                            ▒
           0.18% Tangram::matcher::eval                                                                                                                                                                                                                                                                                       ▒
           0.12% __memcmp_sse4_1                                                                                                                                                                                                                                                                                              ▒
           0.11% std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::compare                                                                                                                                                                                                                     ▒
           0.10% __memcpy_sse2_unaligned                                                                                                                                                                                                                                                                                      ▒
           0.08% strlen
```

blair1618:
Reviewing, sorry for the delay! I anticipate merging this one today.

core/src/scene/drawRule.h
```diff
@@ -36,12 +37,15 @@ class StyleContext;
 
 struct DrawRuleData {
 
+    // https://github.com/tangrams/tangram-docs/blob/gh-pages/pages/draw.md#style-parameters
     std::vector<StyleParam> parameters;
+
+    // https://github.com/tangrams/tangram-docs/blob/gh-pages/pages/draw.md#draw-style
+    // draw-style name (and assigned id)
```
blair1618: This link is outdated now, probably best to replace it with https://github.com/tangrams/tangram-docs/blob/gh-pages/pages/draw.md#draw-rule

blair1618:
Lookin' good. Merging!

Issue #509, opened 2016-01-27
Tile debug graphics should not require tile rebuild


blair1618:
We should be able to just flip the switch and see the change immediately

PR #513, opened 2016-01-28
Rebuild tiles for debug modes that needs it

karimnaaji:
Adressing https://github.com/tangrams/tangram-es/issues/509


blair1618:
Oh! that's one solution haha. I was thinking more along the lines of drawing tile boundaries with `Primitives` and drawing tile ids with `TextDisplay`, but this is definitely a smaller diff.

karimnaaji:
Oh yes that could have been a solution too, removing some of the debug styles code. It seems to be easier this way though.

tallytalwar:
Merging this, future improvements can be handled later.

Thanks

PR #514, opened 2016-01-29
Style builder

hjanetzek:
continue #499

### DIFF
--- a/core/src/labels/labelMesh.cpp
+++ b/core/src/labels/labelMesh.cpp
@@ -31,6 +31,16 @@ LabelMesh::~LabelMesh() {
     }
 }
 
+void LabelMesh::addVertices(std::vector<Label::Vertex>&& _vertices,
+                            std::vector<uint16_t>&& _indices) {
+
+    m_nVertices += _vertices.size();
+    m_nIndices += _indices.size();
+
+    m_vertices.push_back(std::move(_vertices));
+    m_indices.push_back(std::move(_indices));
+}
+
 void LabelMesh::addLabel(std::unique_ptr<Label> _label) {
     m_labels.push_back(std::move(_label));
 }
--- a/core/src/labels/labelMesh.cpp
+++ b/core/src/labels/labelMesh.cpp
@@ -66,7 +66,7 @@ void LabelMesh::loadQuadIndices() {
                  reinterpret_cast<GLbyte*>(indices.data()), GL_STATIC_DRAW);
 }
 
-void LabelMesh::compileVertexBuffer() {
+void LabelMesh::compile() {
     size_t sumVertices = 0;
     int stride = m_vertexLayout->getStride();
     m_glVertexData = new GLbyte[stride * m_nVertices];
--- a/core/src/labels/labelMesh.cpp
+++ b/core/src/labels/labelMesh.cpp
@@ -31,20 +31,6 @@ LabelMesh::~LabelMesh() {
     }
 }
 
-void LabelMesh::addVertices(std::vector<Label::Vertex>&& _vertices,
-                            std::vector<uint16_t>&& _indices) {
-
-    m_nVertices += _vertices.size();
-    m_nIndices += _indices.size();
-
-    m_vertices.push_back(std::move(_vertices));
-    m_indices.push_back(std::move(_indices));
-}
-
-void LabelMesh::addLabel(std::unique_ptr<Label> _label) {
-    m_labels.push_back(std::move(_label));
-}
-
 void LabelMesh::reset() {
     for (auto& label : m_labels) {
         label->resetState();
@@ -76,30 +62,34 @@ void LabelMesh::loadQuadIndices() {
                  reinterpret_cast<GLbyte*>(indices.data()), GL_STATIC_DRAW);
 }
 
-void LabelMesh::compile() {
-    size_t sumVertices = 0;
+void LabelMesh::compile(std::vector<std::unique_ptr<Label>>& _labels,
+                        std::vector<Label::Vertex>& _vertices) {
+
+    constexpr size_t maxVertices = 16384;
+
+    typedef std::vector<std::unique_ptr<Label>>::iterator iter_t;
+
+    m_labels.reserve(_labels.size());
+    m_labels.insert(m_labels.begin(),
+                    std::move_iterator<iter_t>(_labels.begin()),
+                    std::move_iterator<iter_t>(_labels.end()));
+
+    // Compile vertex buffer directly instead of making a temporary copy
+    m_nVertices = _vertices.size();
+
     int stride = m_vertexLayout->getStride();
     m_glVertexData = new GLbyte[stride * m_nVertices];
+    std::memcpy(m_glVertexData,
+                reinterpret_cast<const GLbyte*>(_vertices.data()),
+                m_nVertices * stride);
 
-    for (auto& vertices : m_vertices) {
-        size_t nVertices = vertices.size();
-
-        std::memcpy(m_glVertexData + (sumVertices * stride),
-                    reinterpret_cast<GLbyte*>(vertices.data()), nVertices * stride);
-        sumVertices += nVertices;
-    }
-
-    for (size_t offset = 0; offset < sumVertices; offset += maxVertices) {
+    for (size_t offset = 0; offset < m_nVertices; offset += maxVertices) {
         size_t nVertices = maxVertices;
-        if (offset + maxVertices > sumVertices) {
-            nVertices = sumVertices - offset;
+        if (offset + maxVertices > m_nVertices) {
+            nVertices = m_nVertices - offset;
         }
         m_vertexOffsets.emplace_back(nVertices / 4 * 6, nVertices);
     }
-
-    // clear vertices...
-    std::vector<std::vector<Label::Vertex>>().swap(m_vertices);
-
     m_isCompiled = true;
 }
 
--- a/core/src/labels/labelMesh.cpp
+++ b/core/src/labels/labelMesh.cpp
@@ -69,8 +69,7 @@ void LabelMesh::compile(std::vector<std::unique_ptr<Label>>& _labels,
 
     typedef std::vector<std::unique_ptr<Label>>::iterator iter_t;
 
-    m_labels.reserve(_labels.size());
-    m_labels.insert(m_labels.begin(),
+    m_labels.insert(m_labels.end(),
                     std::move_iterator<iter_t>(_labels.begin()),
                     std::move_iterator<iter_t>(_labels.end()));
 
@@ -135,4 +134,3 @@ void LabelMesh::draw(ShaderProgram& _shader) {
 }
 
 }
-
--- a/core/src/style/polygonStyle.cpp
+++ b/core/src/style/polygonStyle.cpp
@@ -42,8 +42,7 @@ using Mesh = TypedMesh<PolygonVertex>;
 
 
 PolygonStyle::PolygonStyle(std::string _name, Blending _blendMode, GLenum _drawMode)
-    : Style(_name, _blendMode, _drawMode) {
-}
+    : Style(_name, _blendMode, _drawMode) {}
 
 void PolygonStyle::constructVertexLayout() {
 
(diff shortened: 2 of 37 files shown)


blair1618:
I think we need to take a closer look at the PolylineStyle Builder, I don't see any Polyline strokes :\

blair1618:
@hjanetzek Can you help me understand the purpose of the `compileIndices` and `compile` methods in the `Mesh` struct in `polylineStyle.cpp`? How is it different from using the methods of `TypedMesh`? We should prefer to not duplicate such similar code if at all practical.

core/src/style/polygonStyle.cpp
```diff
@@ -65,56 +96,114 @@ void PolygonStyle::constructShaderProgram() {
     m_shaderProgram->setSourceStrings(fragShaderSrcStr, vertShaderSrcStr);
 }
 
-VboMesh* PolygonStyle::newMesh() const {
-    return new Mesh(m_vertexLayout, m_drawMode);
-}
+namespace {
```
blair1618: One consequence of using anonymous namespaces for these structs is that I can't seem to differentiate them in a profiling context, which makes it hard to evaluate their individual performance.

hjanetzek:
The usage is different: Collecting all vertices/indices into large reusable buffers is not possible with the typemesh implementation.  

'compile' in TypesMesh takes over multiple parallel vertices/indices pairs for single meshes and packs them into batches (by shifting indices) to draw multiple meshes in one draw call.

'compile' in Polyline has an additional vertexOffsets vector to keep track of the offsets of single meshes within the vertices/indices vector pairs. 

The typedmesh compile is now only used by PointStyle and DebugStyle.

I would rather see if I can make the Polyline compile version more general, put it in typedmesh, and use it also for PolygonStyle (to avoid the indice shifiting in addPolygon) there

core/src/style/polygonStyle.cpp
hjanetzek: Yeah, saw that too - what do you suggest? Using named namespace from filename (minus 'Style') or prefix all structs with StyleX ?

blair1618: "PolygonStyle::Builder" would make sense to me - we have to make sure to
distinguish it from "PolygonBuilder" in builders.h too haha

blair1618:
Ah I see now. And yes, you read my mind: it would be better still to make
TypedMesh more efficient and use it for all styles.

core/src/style/polygonStyle.cpp
hjanetzek: Ya, but that would mean that we have to pull in all includes and internal structs into the headers which I would like to avoid.

core/src/text/textBuffer.cpp
```diff
@@ -213,26 +229,29 @@ bool TextBuffer::addLabel(const TextStyle::Parameters& _params, Label::Transform
...
+    int nLine = 1;
+
+    m_wordBreaks.clear();
+    nLine = applyWordWrapping(quads, _params, m_metrics);
 
     /// Generate the quads
     float yMin = std::numeric_limits<float>::max();
     float xMin = std::numeric_limits<float>::max();
 
+    m_scratchVertices.clear();
```
karimnaaji: We can reserve `m_scratchVertices(quads.size() * 4)` here; even if the previous adds on the scratch makes it probably big enough.

hjanetzek: Sure, would be ok to avoid possible over-allocations or multiple reallocations on push_back below. Though I generally took the approach to allow reused buffers to over-allocate and overall this should reduce reallocations to happen in relatively few runs.

blair1618: I would also tend to prefer the built-in reallocation in this case - `Vector::reserve` can only be certain to reduce allocations when we know the container's maximum size over its whole lifetime, otherwise we're just growing at a sub-optimal rate.

core/src/style/polygonStyle.cpp
blair1618: Yes you're right, maybe the best we can do is a prefix like "PolygonStyleBuilder"

hjanetzek: OK

I think it would be worth to consider using more namespaces in future - instead of using many inner classes for name spacing and having that long class names for internal classes.  like e.g. tangram::style::polyline::Builder

tallytalwar: I am in agreement of what you propose @hjanetzek

core/src/style/polygonStyle.cpp
```diff
@@ -65,56 +61,101 @@ void PolygonStyle::constructShaderProgram() {
...
+    if (m_params.minHeight != m_params.height) {
+        Builders::buildPolygonExtrusion(_polygon, m_params.minHeight,
+                                        m_params.height, m_builder);
     }
 
-    Builders::buildPolygon(_polygon, height, builder);
-    mesh.addVertices(std::move(vertices), std::move(builder.indices));
+    Builders::buildPolygon(_polygon, m_params.height, m_builder);
+
+    auto& m = m_meshData;
```
tallytalwar: Is this just for convenience?

hjanetzek: umm, no this is just a left-over not cleaned up during refactoring I think

core/src/style/polylineStyle.cpp
```diff
@@ -67,26 +66,160 @@ void PolylineStyle::constructShaderProgram() {
...
+        stroke.width *= 2.0f;
+        stroke.slope *= 2.0f;
+        stroke.slope -= stroke.width;
+
+        stroke.width += fill.width;
+        stroke.slope += fill.slope;
 
-void PolylineStyle::buildLine(const Line& _line, const DrawRule& _rule, const Properties& _props,
-                              VboMesh& _mesh, Tile& _tile) const {
+        stroke.order = std::min(stroke.order, fill.order);
```
tallytalwar: why not directly set the stroke order to be `0.5` less than the `min` value? #nitpick

core/src/style/polylineStyle.cpp
```diff
@@ -98,164 +231,135 @@ double widthMeterToPixel(int _zoom, double _tileSize, double _width) {
...
 
-    _rule.get(StyleParamKey::color, p.fill.color);
-    _rule.get(StyleParamKey::cap, cap);
-    _rule.get(StyleParamKey::join, join);
-    _rule.get(StyleParamKey::order, p.fill.order);
+    Parameters params = parseRule(_rule, _feat.props);
 
-    p.fill.cap = static_cast<CapTypes>(cap);
-    p.fill.join = static_cast<JoinTypes>(join);
+    // if (params.fill.width <= 0.0f && params.fill.slope <= 0.0f ) { return; }
```
tallytalwar: Is there a reason for taking this out? Or am I overlooking something?

core/src/style/polylineStyle.cpp
hjanetzek: stroke is uint in this place

Just noted again: Is everybody OK with using stroke as name instead of outline? I thought to have changed it back before committing..

tallytalwar: `outline` name has 1-1 mapping with the naming in the style sheet. Other than that I am fine with `stroke`.

core/src/style/polylineStyle.cpp
hjanetzek: It's back now!

blair1618:
This is looking far nicer now, I can almost taste that delicious green merge button. 

I'm wondering about that funny anonymous namespace in typedMesh.h that surrounds `compileIndices`. Did you make it separate from the `TypedMesh` class to prevent generating multiple template versions? Clang seems to think that `compileIndices` is "unused" and I can't help but assume it's related.

blair1618:
I was going to suggest moving `compileIndices` to the base mesh class, but from https://github.com/tangrams/tangram-es/tree/back-to-vbomesh it seems you've already thought of this :)

hjanetzek:
yea, I can move compileIndices over to vboMesh.cpp in this branch already.

core/src/labels/labelMesh.cpp
```diff
@@ -66,30 +62,34 @@ void LabelMesh::loadQuadIndices() {
...
-        size_t nVertices = vertices.size();
+    constexpr size_t maxVertices = 16384;
 
-        std::memcpy(m_glVertexData + (sumVertices * stride),
-                    reinterpret_cast<GLbyte*>(vertices.data()), nVertices * stride);
-        sumVertices += nVertices;
-    }
+    typedef std::vector<std::unique_ptr<Label>>::iterator iter_t;
+
+    m_labels.reserve(_labels.size());
```
tallytalwar: `insert` should take care of `reserve`. Or does it not when using a `move_iterator`?

hjanetzek: yes, insert should take care of reserve as well

core/src/style/polylineStyle.cpp
```diff
@@ -98,164 +227,135 @@ double widthMeterToPixel(int _zoom, double _tileSize, double _width) {
...
-        if (_params.outline.cap != _params.fill.cap || _params.outline.join != _params.fill.join) {
-            // need to re-triangulate with different cap and/or join
-            builder.cap = _params.outline.cap;
-            builder.join = _params.outline.join;
-            Builders::buildPolyLine(_line, builder);
-        } else {
-            // re-use indices and positions from original line
+    } else {
+        auto& fill = m_meshData[0];
+        auto& stroke = m_meshData[1];
```
tallytalwar: if we build the `outline`/`stroke` mesh before, can we avoid having a vector of `meshData`?

`line` `meshData` could just append to the `stroke` `meshData`?
@hjanetzek thoughts?

hjanetzek: From the discussion in #502 I understood that the goal was to draw all outlines first. 

Though thinking about it now, it would make sense to switch the order unless blending mode is overlay to reduce overdraw.

tallytalwar: Yup outline should be drawn first, thats why I was suggesting to put those before the line vertices in out mesh.

core/src/text/textBuffer.cpp
```diff
@@ -131,8 +129,8 @@ int TextBuffer::applyWordWrapping(std::vector<FONSquad>& _quads,
 }
 
 
-std::string TextBuffer::applyTextTransform(const TextStyle::Parameters& _params,
-                                           const std::string& _string) {
+std::string TextBuffer::Builder::applyTextTransform(const TextStyle::Parameters& _params,
+                                                    const std::string& _string) {
```
tallytalwar: White space :(

core/src/style/polylineStyle.cpp
hjanetzek: Ah I see :) 

yea, how about flipping meshData[0/1] before compile, when blend_mode is overlay (or inlay)?

core/src/text/textBuffer.cpp
hjanetzek: ?

core/src/text/textBuffer.h
```diff
@@ -19,31 +14,54 @@ class FontContext;
...
+    TextBuffer(std::shared_ptr<VertexLayout> _vertexLayout);
+    ~TextBuffer();
+
+    virtual void draw(ShaderProgram& _shader) override;
+
+protected:
 
     bool m_strokePass = false;
+
+    friend class Builder;
```
tallytalwar: I am not sure there is anything in textBuffer requiring Builder to be a friend. Is this a leftover, or am I missing something?

core/src/text/textBuffer.cpp
tallytalwar: I am seeing extra whitespaces on line 133.

core/src/style/polylineStyle.cpp
tallytalwar: That will be good.

I was proposing a way to avoid having 2 `meshData`s. 
But if we are reversing the order when blending mode is overlay/inlay (though inlay not supported for ES right now) **only** then I can not think of a way to avoid having 2 meshDatas (one for fill and one for stroke).

tallytalwar:
This is looking close to be merging. I am good with this, once couple of queries are answered/addressed.

This PR has some very good concepts both wrt class design/flow and c++ usage.

core/src/text/textBuffer.cpp
hjanetzek: Are you sure? I cannot see any

blair1618: This line is fine, maybe the github diff just looks weird.

tallytalwar: Yup, its the github diff making it look weird. Sorry about that!.

hjanetzek: Ahh ok, the indentation diff.

tallytalwar:
@hjanetzek all good with this? Or you still working on it?

blair1618:
I'm happy with this now - all good to merge when everyone is ready!

hjanetzek:
I'm ready. Thanks everyone for reviewing!

PR #515, opened 2016-01-29
Android font fallbacks

karimnaaji:
Android platform layer for font fallbacks
- Fallbacks are given by the platform by order of importance (determined by android order from `font_fallbacks.xml`)
- For a given _hint_ boldness `systemFontFallbackPath` would give the most appropriate font for the required bold.

For OS X, fallbacks can be found in `/System/Library/Frameworks/ApplicationServices.framework/Frameworks/CoreText.framework/Resources/DefaultFontFallbacks.plist`

### DIFF
--- a/android/tangram/src/com/mapzen/tangram/FontFileParser.java
+++ b/android/tangram/src/com/mapzen/tangram/FontFileParser.java
@@ -131,10 +131,9 @@ class FontFileParser {
         }
     }
 
-    public FontFileParser() {
-
+    public void parse(String fontXMLPath) {
         InputStream in = null;
-        final File fontFile = new File("/system/etc/fonts.xml");
+        final File fontFile = new File(fontXMLPath);
 
         // TODO: Handle system_fonts parsing which has a different xml layout as compared to fonts.xml
         //       system_fonts.xml also does not seem to have a good css style font parameter mapping as is the case with
--- a/android/tangram/src/com/mapzen/tangram/FontFileParser.java
+++ b/android/tangram/src/com/mapzen/tangram/FontFileParser.java
@@ -6,18 +6,20 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.FileNotFoundException;
 import java.util.HashMap;
+import java.util.Iterator;
 import java.util.Map;
 import java.util.ArrayList;
+import java.util.Vector;
 
 import org.xmlpull.v1.XmlPullParser;
 import org.xmlpull.v1.XmlPullParserException;
 
 import android.util.Xml;
-import android.util.Log;
 
 class FontFileParser {
 
     private Map<String, String> fontDict = new HashMap<String, String>();
+    private Map<Integer, Vector<String>> fallbackFontDict = new HashMap<Integer, Vector<String>>();
 
     private void processDocument(XmlPullParser parser) throws XmlPullParserException, IOException {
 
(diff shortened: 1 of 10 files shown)


tallytalwar:
Reviewing this now.

android/tangram/src/com/mapzen/tangram/FontFileParser.java
```diff
@@ -90,20 +122,19 @@ private void skip(XmlPullParser parser) throws XmlPullParserException, IOExcepti
         int depth = 1;
         while (depth > 0) {
             switch (parser.next()) {
-            case XmlPullParser.START_TAG:
-                depth++;
-                break;
-            case XmlPullParser.END_TAG:
-                depth--;
-                break;
+                case XmlPullParser.START_TAG:
```
tallytalwar: @karimnaaji can you remove these whitespaces?

blair1618:
I feel like leaving font fallback logic as the responsibility of platform-specific layers could invite unneeded difficulty. There are many considerations when choosing a font to fall back on: weight, italic, code point range, serif, maybe others that I can't think of. If a platform function is the way that these get resolved, then do we re-write the logic for all of these parameters on every platform? Is there any benefit to that? 

Instead we could make the font resolution logic part of the core library, then all that each platform needs to do is provide a list of all the fonts that are available and what properties they have.

tallytalwar:
I agree with @blair1618, it will also make it more consistent without having any code repetition.

karimnaaji:
@tallytalwar , @blair1618 In general a platform has a description of its own series of fallbacks ordered by importance. The core library can't make this ordering assumption for the platform, or the fallbacks would get really inconsistent and non-optimal since the ordering has its importance.

For italic, OS X has the family described in the fallbacks, not android, so the platform layer would need to take care of this.

For code point range; not sure either the platform nor the core library would be able to solve that without loading all the set of fonts in memory, which is not what should be happening in a fallback system.

blair1618:
@karimnaaji Yeah the code point resolution is going to be much harder than the others, not sure how to go about that :\

Regarding platform font preferences: Those absolutely make sense for creating general application UI elements, but I think the typography we're doing in tangram falls into a different paradigm - it's more like browser font resolution. If we use the preferences from the device then we're basically leaving the quality of our typographic design up to Android OEMs and quite frankly I don't trust them enough for that.

karimnaaji:
From my understanding of how a browser handles the specified css fonts description (might be wrong on that):

```
for a codepoint and css stylesheet
if (!codepoint drawable by described css-font)
    while (css has fallback) try to draw with next fallback // (1)
    while (platform has font fallback) try to draw with next platform dependent font fallback // (2)      
```
And if we compare to what we do in tangram-es, once https://github.com/tangrams/tangram/pull/235 is addressed it would be step (1), and this dependent platform-fallback would be step (2).
It doesn't mean that we leave typographic design to Android or the platform, since the fallback is already something we would handle within the stylesheet for the designer, but just getting _sure_ that something is drawn even if the user hasn't provided any fallback as in https://github.com/tangrams/tangram/pull/235 or if the fallbacks were not consistent enough.

Now, 
- do we want tangram-es to _always_ show something (going in the direction of a platform-dependent fallback + user specified fallback), 
- _or_ do we want tangram-es to _always_ show what is specified by the designer with the risk of not showing anything if the styling is not done properly (e.g. not doing any fallback that is not specified by the designer) which might differ from tangram-js 
- _or_ do we want the fallbacks given by the platform and managed within the core library (which is quite unclear to me, regarding how we can order the font fallback importance etc...).

blair1618:
I think we definitely want the overall font resolution system to end up with *some font* rather than *no font* in every case (if for no other reason than to more nearly match the behavior of tangram-js). As you pointed out, this implies that we need internally-resolved fallbacks when user-provided fallbacks fail. 

I think we're actually not suggesting conflicting things. On Android, `fonts.xml` provides a set of font *aliases* that map many names to one - we should still use that information for resolving fallbacks internally, I just think that the logic for choosing a fallback among all known fonts (including aliased fonts) would be best handled by one piece of code inside the core library.

karimnaaji:
>aliases that map many names to one - we should still use that information for resolving fallbacks internally

Well it's confusing, how aliases could help in that sense? Except if we extract ourself information (about shaping, codepoint ranges, kerning etc) among the set of all known fonts or anything like that, there's no way we can order a font fallback stack based on an aliases that match many names to one (aliases just help for you to write something like `sans-serif-thin` instead or `Roboto-Thin` for example, at least on Android, and this is not enough information for any fallback related thing), and I don't see ourself reading through all the set of fonts at runtime to know whether a font is better than the other one for a given shaping plan, since it's huge process.

I agree that this is probably not the best, and having this within core library is definitely nicer, but I don't see why it wouldn't be a first good first step towards that unresolved issue.

I'll leave this work as it is since moving that within core right now is not possible, and work on alfons might make us change our mind on a best way to handle that.

Issue #516, opened 2016-01-30
Full font texture atlas size reached error


softlxf:
### When I use Chinese font, my poi data which has a large number of Chinese characters would fill font texture atlas to full, then error occurred.

here is my log:
01-30 12:07:43.302 702-836/com.mapzen.tangram.android D/Tangram: WARNING fontContext.cpp:195: Texture Atlas resized to 1024 1024
01-30 12:07:43.769 702-836/com.mapzen.tangram.android D/Tangram: WARNING fontContext.cpp:195: Texture Atlas resized to 2048 2048
01-30 12:07:45.972 702-772/com.mapzen.tangram.android D/Tangram: ERROR fontContext.cpp:189: Full font texture atlas size reached!
...
01-30 12:07:46.368 702-772/com.mapzen.tangram.android D/Tangram: ERROR fontContext.cpp:189: Full font texture atlas size reached!
01-30 12:07:46.368 702-772/? D/Tangram: ERROR fontContext.cpp:189: Full font texture atlas size reached!
01-30 12:07:46.368 702-772/? D/Tangram: ERROR fontContext.cpp:189: Full font texture atlas size reached!

czxczq521:
I am also concerned about this issue

czxczq521:
If try to change TANGRAM_MAX_TEXTURE_WIDTH and TANGRAM_MAX_TEXTURE_HEIGHT size to 4096, that is OK？

karimnaaji:
We currently do not support atlas regeneration nor multiple-atlases, but we will add some mechanism around that. The limitation on `TANGRAM_MAX_TEXTURE_WIDTH/HEIGHT` of 2048 is a device-hardware compatibility/limitation, since most of mobile devices support this size as the maximum texture size; but it would make sense to make this size a per-hardware value.

PR #517, opened 2016-02-01
Load hardware max texture size

karimnaaji:
Related #516 
Instead of using a default value of 2048 (which is the most widely least supported texture size) load the hardware value.


blair1618:
Cool, this makes sense as a feature. It seems very similar in nature to the GL extension code, do you think it would make sense to consolidate them into a class/namespace that queries device features?

karimnaaji:
Yeah good call to reorganize this a bit.

blair1618:
Lookin' good. Merging!

czxczq521:
It's great.

PR #520, opened 2016-02-03
Fix vertex attribute enabling for non-buffered vertex arrays

blair1618:
This bug had me seriously questioning my sanity last night. In the current code, calling `enable` on a `VertexLayout` for non-buffered arrays of vertex data will only correctly enable the first vertex attribute; all subsequent attributes will be pointed to the same data as the first attribute because `attrib.offset` is not applied to the input pointer. This change just adds the offset to the input pointer whether or not it is `nullptr`.

### DIFF
--- a/core/src/gl/vertexLayout.cpp
+++ b/core/src/gl/vertexLayout.cpp
@@ -97,7 +97,7 @@ void VertexLayout::enable(ShaderProgram& _program, size_t _byteOffset, void* _pt
                 loc = glProgram;
             }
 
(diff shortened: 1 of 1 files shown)


blair1618:
@karimnaaji Have you encountered similar problems with debug drawing? Does this fix make sense to you?

karimnaaji:
@blair1618 , that's my bad when I added the non-buffered vertex arrays, this looks good to me. Merging!

PR #522, opened 2016-02-04
TileDataSink

hjanetzek:
Replacing TileData container with TileDataSink callbacks. With this a single Feature object can be reused for tile creation. Unmatched layers can be skipped directly, unmatched features (by props and geometry type) can be skipped without extracting the geometry.

### DIFF
--- a/core/src/data/clientGeoJsonSource.cpp
+++ b/core/src/data/clientGeoJsonSource.cpp
@@ -236,30 +236,30 @@ struct add_geometry {
     }
 };
 
-std::shared_ptr<TileData> ClientGeoJsonSource::parse(const TileTask& _task,
-                                                     const MapProjection& _projection) const {
+bool ClientGeoJsonSource::process(const TileTask& _task,
+                                  const MapProjection& _projection,
+                                  TileDataSink& _sink) const {
 
-    std::lock_guard<std::mutex> lock(m_mutexStore);
-
-    auto data = std::make_shared<TileData>();
-
-    if (!m_store->tiles) { return nullptr; }
+    if (!m_store->tiles) { return false; }
     auto tile = m_store->tiles->getTile(_task.tileId().z, _task.tileId().x, _task.tileId().y);
 
-    data->layers.emplace_back("");  // empty name will skip filtering by 'collection'
-    Layer& layer = data->layers.back();
+    // empty name will skip filtering by 'collection'
+    // TODO why not use DataSource name as layer name?
+    if (!_sink.beginLayer("")) {
+        return true;
+    }
 
     for (auto& it : tile.features) {
+
         Feature feature(m_id);
 
         if (geometry::geometry<int16_t>::visit(it.geometry, add_geometry{ feature })) {
             feature.props = m_store->properties[it.id.get<uint64_t>()];
-            layer.features.emplace_back(std::move(feature));
+            _sink.addFeature(feature);
         }
     }
 
-
-    return data;
+    return true;
 }
 
 }
--- a/core/src/data/mvtSource.cpp
+++ b/core/src/data/mvtSource.cpp
@@ -8,33 +8,52 @@
 
 namespace Tangram {
 
-std::shared_ptr<TileData> MVTSource::parse(const TileTask& _task, const MapProjection& _projection) const {
+#define TAG_TILE_LAYER 3
+#define TAG_LAYER_NAME 1
 
-    auto tileData = std::make_shared<TileData>();
+bool MVTSource::process(const TileTask& _task, const MapProjection& _projection,
+                        TileDataSink& _sink) const {
 
     auto& task = static_cast<const BinaryTileTask&>(_task);
 
-    protobuf::message item(task.rawTileData->data(), task.rawTileData->size());
+    protobuf::message tileMsg(task.rawTileData->data(), task.rawTileData->size());
     PbfParser::ParserContext ctx(m_id);
 
     try {
-        while(item.next()) {
-            if(item.tag == 3) {
-                tileData->layers.push_back(PbfParser::getLayer(ctx, item.getMessage()));
-            } else {
-                item.skip();
+        while (tileMsg.next()) {
+            if (tileMsg.tag != TAG_TILE_LAYER) {
+                tileMsg.skip();
+                continue;
+            }
+            protobuf::message layerMsg = tileMsg.getMessage();
+            protobuf::message layerItr = layerMsg;
+
+            while (layerItr.next()) {
+
+                if (layerItr.tag != TAG_LAYER_NAME) {
+                    layerItr.skip();
+                    continue;
+                }
+                auto layerName = layerItr.string();
+
+                if (!_sink.beginLayer(layerName)){
+                    break;
+                }
+
+                PbfParser::extractLayer(ctx, layerMsg, _sink);
+                break;
             }
         }
     } catch(const std::invalid_argument& e) {
         LOGE("Cannot parse tile %s: %s", _task.tileId().toString().c_str(), e.what());
-        return {};
+        return false;
     } catch(const std::runtime_error& e) {
         LOGE("Cannot parse tile %s: %s", _task.tileId().toString().c_str(), e.what());
-        return {};
+        return false;
     } catch(...) {
-        return {};
+        return false;
     }
-    return tileData;
-}
 
+    return true;
+}
 }
--- a/core/src/util/pbfParser.cpp
+++ b/core/src/util/pbfParser.cpp
@@ -79,20 +79,49 @@ void PbfParser::extractGeometry(ParserContext& _ctx, protobuf::message& _geomIn)
     }
 }
 
-// Reuse string allocations in Value. Variant is not yet smart enough
-// probably due to traits not being so well supported everywhere..
-struct copy_visitor {
+struct string_visitor {
     using result_type = bool;
-    template <typename T>
-    bool operator()(T& l, const T& r) const {
-        l = r;
+    Value& propValue;
+    std::string& str;
+
+    bool operator()(const PbfParser::StringView& v) const {
+        str.assign(v.second, v.first);
         return true;
     }
-    template <typename L, typename R>
-    bool operator()(L l, R r) const {
+    template<typename T>
+    bool operator()(const T& v) const {
+        propValue = v;
         return false;
     }
 };
+struct double_visitor {
+    using result_type = bool;
+    Value& propValue;
+
+    bool operator()(const PbfParser::StringView& v) const {
+        propValue = std::string(v.second, v.first);
+        return false;
+    }
+    template<typename T>
+    bool operator()(const T& v) const {
+        propValue = v;
+        return true;
+    }
+};
+struct update_visitor {
+    using result_type = bool;
+    const PbfParser::TagValue& tagValue;
+    Value& propValue;
+
+    template<typename T>
+    bool operator()(T& v) const {
+        return PbfParser::TagValue::visit(tagValue, double_visitor{propValue});
+    }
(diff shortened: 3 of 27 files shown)


blair1618:
I've been profiling this change on an iPhone and I only see rather modest improvements in load times:
```
Time spent in TileWorker::run loading Manhattan at zoom 16 with the eraser-map scene,
average of 3 runs

MVT
master:         1050ms
tile-data-sink: 990ms

GeoJSON
master:         1140ms
tile-data-sink: 1170ms

TopoJSON
master:         1130ms
tile-data-sink: 1150ms

Time spent in TileWorker::run loading Manhattan at zoom 16 with the default testing scene,
average of 3 runs

MVT
master:         1000ms
tile-data-sink: 895ms

GeoJSON
master:         1055ms
tile-data-sink: 1060ms

TopoJSON
master:         1005ms
tile-data-sink: 1045ms
```
@hjanetzek have you measured significant performance improvements in another context?

hjanetzek:
@blair1618 what is the ratio of tiledata loadtime vs style-building? 100ms less overall doesn't sound too bad. 
```
curl  https://vector.mapzen.com/osm/all/11/1100/671.mvt | gunzip > tile.mvt
 ./bench/tileLoading.out --benchmark_repetitions=10
```

```
TileLoadingFixture/BuildTest          43544713   41372529         17
TileLoadingFixture/BuildTest_mean     42006110   41607835         17
TileLoadingFixture/BuildTest_stddev    2937931     348561          0
```


```
TileLoadingFixture/BuildTest          32216103   29710174         23
TileLoadingFixture/BuildTest_mean     29815215   29753622         23
TileLoadingFixture/BuildTest_stddev    3732407     172089          0
```

Checking with larger tiles shows a good speedup from 42ms (24% parsing vs 43% building) to 30 ms (80%-66%  = 14% parsing vs 66% building) 

GeoJSON and TopoJSON source don't implement any ways to avoid allocations that became possible with tile-data-sink yet.

blair1618:
Hmm I don't remember the proportions, I can check again. I would be much more trusting of the benchmarks if we could run them on an Android or iOS device, otherwise they aren't really answering the question that matters to us.

hjanetzek:
The tests on RPI2 show also a 2x speedup for tile-sink vs current master. This measurement is only mvt source tile load time only (the scene contains empty rules that are always matching so that tile-data-sink does not skip anything) 

```
        emptyScene = std::make_shared<Scene>();
        emptyScene->layers().push_back(DataLayer{
                SceneLayer{"", Filter::MatchNone({}), {},{}},
                    "osm", {"buildings", "roads", "landuse", "pois", "water", "earth",
                            "landuse_labels", "boundaries", "places", "transit"}});
        tileBuilder.setScene(emptyScene);
```

All tests are compiled with '-mcpu=cortex-a7 -mfpu=neon-vfpv4 -mfloat-abi=hard', which is about 10ms faster on master. Same tile as above, but the above test also built the geometries. 

```
curl  https://vector.mapzen.com/osm/all/11/1100/671.mvt | gunzip > tile.mvt
 ./bench/tileLoading.out --benchmark_repetitions=10
```

Current master w/ TileData,
```
TileLoadingFixture/BuildTest         159988768  160000000          4
TileLoadingFixture/BuildTest         160751581  162500000          4
TileLoadingFixture/BuildTest         160282103  160000000          4
TileLoadingFixture/BuildTest         160656076  162500000          4
TileLoadingFixture/BuildTest_mean    160419632  161250000          4
TileLoadingFixture/BuildTest_stddev     304407    1250000          0
```

TileDataSink branch
```
TileLoadingFixture/BuildTest          70631931   71000000         10
TileLoadingFixture/BuildTest          69590250   70000000         10
TileLoadingFixture/BuildTest          69699014   70000000         10
TileLoadingFixture/BuildTest          70681391   71000000         10
TileLoadingFixture/BuildTest_mean     70150647   70500000         10
TileLoadingFixture/BuildTest_stddev     507775     500000          0
```

With tile-data-sink it's also possible to reuse the geometry object, 4x speedup on geometry-container branch
```
TileLoadingFixture/BuildTest          43855733   44166667         12
TileLoadingFixture/BuildTest          41045754   40833333         12
TileLoadingFixture/BuildTest          41061380   40833333         12
TileLoadingFixture/BuildTest          41944008   42500000         12
TileLoadingFixture/BuildTest_mean     41976719   42083333         12
TileLoadingFixture/BuildTest_stddev    1144149    1381927          0
```

blair1618:
Ok I'm reviewing this in earnest now, sorry for putting it off so long :)

core/src/data/clientGeoJsonSource.cpp
```diff
@@ -117,30 +117,52 @@ void ClientGeoJsonSource::addPoly(const Properties& _tags, const std::vector<Coo
...
         std::lock_guard<std::mutex> lock(m_mutexStore);
-        if (!m_store) { return nullptr; }
+        //if (!m_store) { return nullptr; }
+        if (!m_store) { return false; }
         tile = m_store->getTile(_task.tileId().z, _task.tileId().x, _task.tileId().y);
     }
 
-    Layer layer(""); // empty name will skip filtering by 'collection'
+    // empty name will skip filtering by 'collection'
+    // TODO why not use DataSource name as layer name?
```
blair1618: The scene file already filters data by source name, so it would be redundant to perform filtering against the same name for the layer.

core/src/tangram.cpp
```diff
@@ -29,6 +29,7 @@
 #include <bitset>
 #include <mutex>
 #include <queue>
+#include <array>
```
blair1618: Leftover?

core/src/data/mvtSource.cpp
```diff
@@ -7,30 +7,53 @@
 #include "util/pbfParser.h"
 #include "platform.h"
 
+#include <sstream>
+#include <fstream>
```
blair1618: These two includes look unused.

core/src/util/pbfParser.cpp
```diff
@@ -38,7 +35,7 @@ PbfParser::Geometry PbfParser::getGeometry(ParserContext& _ctx, protobuf::messag
 
     while(_geomIn.getData() < _geomIn.getEnd()) {
 
-        if(cmdRepeat == 0) { // get new command, length and parameters..
+        if(cmdRepeat == 0) { // get new command, lengh and parameters..
```
blair1618: typo

core/src/util/pbfParser.cpp
```diff
@@ -139,102 +246,111 @@ Feature PbfParser::getFeature(ParserContext& _ctx, protobuf::message _featureIn)
...
+                double area = signedArea(it, it + length);
+                if (area == 0.f) {
+                    offset += length;
                     continue;
                 }
-                int winding = area > 0 ? 1 : -1;
-                // Determine exterior winding from first polygon.
-                if (_ctx.winding == 0) {
-                    _ctx.winding = winding;
+                WindingOrder winding(area);
```
blair1618: So, the logic here related to winding direction is intentionally not generalized. You can't generally say whether a positive signed area is clockwise or counterclockwise because it depends on the coordinate system of the points in the ring (notably, whether y is up or down). It's really simpler to avoid talking in terms of "clockwise" and "counterclockwise" altogether, since all that matters to us is the direction of one ring relative to another and the English terms can (and do) lead to confusion. The `WindingOrder` type is really unrelated to the purpose of this PR so I suggest it be omitted.

blair1618:
Not sure why yet, but this branch doesn't correctly render buildings in MVT tiles from Mapbox (current master does). I've been testing with this scene:
```yaml
sources:
    mb:
        type: 'MVT'
        url: 'https://a.tiles.mapbox.com/v4/mapbox.mapbox-streets-v6/{z}/{x}/{y}.vector.pbf?access_token=pk.eyJ1IjoiYmNhbXBlciIsImEiOiJWUmh3anY0In0.1fgSTNWpQV8-5sBjGbBzGg'
        max_zoom: 15
scene:
    background: { color: 'lightgrey' }
lights:
    sun:
        type: 'directional'
        origin: 'world'
        direction: [1.0, 1.0, -1.0]
        diffuse: [0.3, 0.3, 0.3, 1.0]
        ambient: [0.7, 0.7, 0.7, 1.0]
layers:
    water:
        data: { source: mb }
        draw: { polygons: { order: 1, color: 'lightblue' } }
    building:
        data: { source: mb }
        draw: { polygons: { order: 10, color: 'grey', extrude: 20 } }
    landuse:
        data: { source: mb }
        draw: { polygons: { order: 8, color: 'honeydew' } }
    road:
        data: { source: mb, layer: [road, bridge, tunnel] }
        draw: { lines: { order: 9, color: 'darkgrey', width: 8 } }
```

core/src/data/clientGeoJsonSource.cpp
hjanetzek: I see

core/src/tangram.cpp
hjanetzek: std::array is used in tangram.cpp - could be that some header changes removed the indirect include

core/src/util/pbfParser.cpp
hjanetzek: ack

hjanetzek:
@blair1618, nice catch! mapbox buildings don't have tags. I'll add some tests for tile parsing.

core/src/util/pbfParser.cpp
hjanetzek: I agree that this would be part of a different PR. I'm not sure if it makes things simpler to use sign of area instead of winding. It's hard to not think of the default math coordinate system when talking about winding order. Whether one interprets the coordinates a y-down to visualize the geometry is not relevant to the algorithm working with them.  

IMO we just flip y-axis in too many places - and got confused by that :)

blair1618: Agreed! If we could reduce the need to flip the axis around I would be all for it. 

The basic problem is that a counter-clockwise ring has positive area in y-up coordinates but negative area in y-down coordinates, which adds a layer of complication to the MVT spec when it defines winding direction in terms of signed area. And on top of that, mapbox's tiles don't currently ([at least a little while ago](https://github.com/mapbox/vector-tile-spec/issues/61)) follow their own spec for winding! That's why it seems easiest to me to ignore everything except relative direction.

core/src/data/dataSource.h
```diff
@@ -41,8 +44,8 @@ class DataSource : public std::enable_shared_from_this<DataSource> {
     /* Stops any running I/O tasks pertaining to @_tile */
     virtual void cancelLoadingTile(const TileID& _tile);
 
-    /* Parse a <TileTask> with data into a <TileData>, returning an empty TileData on failure */
-    virtual std::shared_ptr<TileData> parse(const TileTask& _task, const MapProjection& _projection) const = 0;
+    /* FIXME doc /Parse a <TileTask> with data into a <TileData>, returning an empty TileData on failure */
```
blair1618: "Transform the data from a TileTask into Features, which are passed into a TileDataSink" something like that?

core/include/pbf/pbf.hpp
```diff
@@ -151,6 +152,14 @@ std::string message::string()
...
 }
 
+std::pair<size_t, const char*> message::chunk()
+{
+    uint64_t len = varint();
+    value_type string = static_cast<value_type>(data_);
+    skipBytes(len);
+    return { len, string };
+}
+
```
blair1618: This method is used exactly once. I think we should avoid generalizing code until it's used at least twice, but since it's already written, meh :P

core/src/data/dataSource.h
blair1618: Does it always return `true`? Would it be useful to return anything else? Maybe `true` if any features were processed, otherwise `false`

PR #523, opened 2016-02-05
merge TypedMesh into VboMesh

hjanetzek:
- change VboMesh to VboMeshBase to share commmon implementations
- add interface for StyledMesh to make only the functions visible
  that should be used externally once a Mesh is built. This removes
  the requirement that tile meshes must extend VboMesh (which is not
  needed for alfons LabelContainer)
- StyledMesh should also provide more flexibility when creating new Styles: A StyledMesh may be composed of multiple VboMeshes to be able to have different Vertex layouts per Style. As a follow up I would suggest to make Style more opaque by not exposing getShaderProgram() (and thereby restricting it to) a single shader. Instead Style::draw(Mesh) could keep all the draw logic internal. 
- [ ] better name for StyledMesh?


karimnaaji:
Reviewing this

core/src/gl/vboMesh.h
```diff
@@ -116,6 +105,194 @@ class VboMesh {
...
+     * Update _nVerts vertices in the mesh with the new attribute A
+     * _newAttributeValue starting after _byteOffset in the mesh vertex data
+     * memory
+     */
+    template<class A>
+    void updateAttribute(Range _vertexRange,
+                         const A& _newAttributeValue,
+                         size_t _attribOffset = 0);
+
+protected:
```
karimnaaji: Can we remove this empty protected block

core/src/gl/renderState.cpp
```diff
@@ -65,6 +72,14 @@ namespace RenderState {
         textureUnit.init(std::numeric_limits<unsigned int>::max(), false);
     }
 
+    bool isCurrentGeneration(int _generation) {
```
karimnaaji: I think we can use the same generation process through RenderState for Texture and Shaders.

core/src/style/style.h
```diff
@@ -36,6 +35,14 @@ enum class Blending : int8_t {
     overlay,
 };
 
+struct StyledMesh {
```
karimnaaji: I agree another name would be nice, what about `DrawBuffer`? And maybe declare this in `vboMesh.h`.

core/src/gl/renderState.cpp
hjanetzek: yes. what about renaming to isValidGeneration() ?

karimnaaji: yes that would be more explicit too.

core/src/style/style.h
hjanetzek: Collecting: DrawBucket, Drawable, Batch?

I would say it belongs to Style - it's what the StyleBuilder produces and Style should know how to render them (in my plan)

karimnaaji: What made me anxious is that `vboMesh.h` now has to include `style.h`, it would be nice to keep modules with dependencies only within its own module, but this makes sense.

`Drawable` sounds Java, but seems clear enough.

karimnaaji:
Some small changes to be addressed, otherwise looking good!

blair1618:
While we're rearranging these classes, we should remove `Vbo` from type names. They're actually VAOs wherever the hardware supports it, so the name is somewhat misleading.

karimnaaji:
Agreed on that.

hjanetzek:
I can move the last commit to a separate branch if wanted. It enables Styles to decide how to draw a StyledMesh. E.g. for Text- and PointStyle meshes this would be a NOOP with the changes in alfons branch.

hjanetzek:
Further naming suggestions?
For VboMesh, I actually like the name. It implicates that it holds GL resources to draw vertices + vertex attribute indices. (VAO just allows to save and restore the offsets of vertex attributes + buffer bindings index)

core/src/style/style.cpp
```diff
@@ -95,13 +95,20 @@ void Style::setupShaderUniforms(int _textureUnit, Scene& _scene) {
 
 void Style::onBeginDrawFrame(const View& _view, Scene& _scene, int _textureUnit) {
 
+    // TODO cache which uniforms are used by the shader!
```
karimnaaji: Is this something we need to address? A uniform not used would at worse involve a `glUseProgram` and a hash table access. Or maybe you mean something else?

hjanetzek: In my measurements the uniform hash table lookup became the most expensive part cpu-wise in Tangram::render() + the string allocations to pass the uniform name to setUniform(const std::string)

I think of a UniformHandle template could help to keep all infos on the caller site. Something like
```
template<T>
struct UniformHandle {
    int id;
    int generation;
    T value;
    std::string name;
    set(shader&, T&)
}
 ```

karimnaaji: I see, I saw that coming up in the profiler too at some point. We could declare all the string uniform like `u_time` etc statically in some files to address the string allocations at first and see how it goes, it would reduce the allocations.

blair1618:
I don't want to get pedantic about naming, but: 

1. I think that a class called `Mesh` or `MeshBase` in a folder called "gl" would imply that GL resources are involved. 
2. Class names should match file names. If I were reading unfamiliar code and saw a header with a different name than the single class it describes, I would find it sloppy (this is the case in [filters.h](https://github.com/tangrams/tangram-es/blob/master/core/src/scene/filters.h) and I myself have been confused by it before). 

And now I'll check out the actual substance of the changes here...

hjanetzek:
Mesh is OK :)

karimnaaji:
@blair1618 , did you have a chance to go over this? Other than the renaming this looks good.

blair1618:
Yes I'm good with this. The consolidated "generation" handling is elegant, nice work.

PR #525, opened 2016-02-05
Fix retina display scaling when using GLFW on OS X

blair1618:
The OS X application now correctly provides a pixel scale for Tangram and provides coordinates for resizing and input in pixels, not screen coordinates. 

This fixes the appearance of some screen-space shader effects, such as the water color gradient in the eraser-maps style, but it causes a new issue with labels when display density changes after launch: any labels that have already been rendered will display at an incorrect size when the new density is applied. This needs further investigation. 

These changes could probably be safely applied to the Linux application as well, but so far no Linux testing has discovered this pixel-to-screen-coordinate difference that OS X has. 

Update: The OS X application now rebuilds tiles completely when density changes, so incorrectly-scaled labels are no longer visible

Update update: hang on, I can make this just a little nicer

Update update update: Alright good to go

### DIFF
--- a/osx/src/main.cpp
+++ b/osx/src/main.cpp
@@ -212,17 +212,17 @@ void drop_callback(GLFWwindow* window, int count, const char** paths) {
 // Window handling
 // ===============
 
-void window_size_callback(GLFWwindow* window, int width, int height) {
+void framebuffer_size_callback(GLFWwindow* window, int fWidth, int fHeight) {
 
-    // The callback parameters are measure in screen coordinates, but Tangram
-    // should be given sizes in pixels, which can be measured differently on
-    // hi-dpi displays.
-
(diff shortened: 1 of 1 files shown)


karimnaaji:
Works great! And now we really have double density rendering on retina display (which is great for testing) which is fabulous.
Do we want the same on linux? Otherwise merging.

blair1618:
As far as I can tell, on Ubuntu the GLFW window size always equals the framebuffer size, so there's no need for this logic. I'm also hopeful that code like this can eventually be shared between all desktop clients in a single GLFW "platform" folder.

karimnaaji:
Yeah we had that here (https://trello.com/c/FbRwhr4v/590-common-platform-code), while thinking about a better structure for platform code which would be pretty handy for cases like that.

Issue #486, opened 2016-01-21
Loading large scene files should be faster


blair1618:
There are two major contributors to scene load time:

**YAML parsing**
In some instances this cost is unavoidable, such as when evaluating a modified stylesheet. In production use, however, the scene file will be the same nearly every time the app starts up so there's no need to parse it all over again: instead we can serialize the parsed structure as a protocol buffer or a flat buffer and then on the next launch, if the scene file timestamp is unchanged, deserialize from the buffer instead of the YAML file. 

**Scene initialization**
This part is should be more straightforward to optimize; somehow `SceneLoader::loadStyle` is taking _substantially more time_ than the entire YAML parsing process, multiple seconds for the eraser-map style on a phone. In particular, `SceneLoader::mixStyles` and `SceneLoader::getMixins` are far more time-consuming than they have any reason to be.

hjanetzek:
Yea, the yaml processing time is amazing indeed :D  I was also looking into this yesterday as while profiling with heaptrack. I have no clue yet why mixStyles/mergeBlocks does such an amount of allocations.. ~80MB of small nodes for the eraser-map Styles.  But also 50MB for just loading.
I think we should rewrite the merging without using yaml-cpp nodes. I.e. transferring the nodes to
 appropriate structs for Style properties and apply the merging on that.  

![shot-2016-01-21_18-14-36](https://cloud.githubusercontent.com/assets/1747/12488314/40f04170-c06b-11e5-8d06-884c3f52c1a2.jpg)

blair1618:
I'm going to see what I can do about the time and memory performance of `SceneLoader::loadStyle` first - should be some low-hanging fruit.

blair1618:
https://github.com/tangrams/tangram-es/pull/555 addresses part of this issue. Parsing large YAML files can still take an unacceptably long time on some devices, so a caching scheme for the parsed structure could still make sense.

PR #526, opened 2016-02-05
[rfc] Run time atlas packing for icons

karimnaaji:
Pack icon atlases at runtime; I'm not especially in favor of a runtime process on that, just pointing out an issue that needs to be addressed, either at runtime or offline, so let's discuss.

Current status on this shown below (benchmark results on S6). As @blair1618 mentioned, the data can also be reinterpreted at runtime for a better internal compressed texture format like PVRTC and others.

Packing takes ~1% of scene loading here, and can be improved.

screenshot_2016-02-05-14-59-17

screen shot 2016-02-05 at 2 30 20 pm

### DIFF
--- a/android/tangram/src/com/mapzen/tangram/MapController.java
+++ b/android/tangram/src/com/mapzen/tangram/MapController.java
@@ -52,7 +52,7 @@ public class MapController implements Renderer {
      */
     public MapController(Activity mainApp, MapView view) {
 
-        this(mainApp, view, "scene.yaml");
+        this(mainApp, view, "eraser-map.yaml");
     }
 
     /**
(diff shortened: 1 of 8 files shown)


blair1618:
Nice, repacking is quicker than I thought!

The total scene load time is bizarrely high for reasons that I hope to delve into soon, but tens of milliseconds is quite manageable for a one-time cost. Aside from actual texture uploads, could we run re-packing on a background thread if we wanted?

karimnaaji:
Yeah for reasons mentioned in https://github.com/tangrams/tangram-es/issues/486 the scene loading would be reduced.
Yep, packing doesn't have to be done at scene loading and could be done from a worker thread before any first use. I think that we can also make it more clever, don't repack anything that is already well packed, and don't use POT texture when the filtering is anything except mipmaps (which is not supported in core ES 2.0), so if filtering is say, linear, the packing can save even more memory.

PR #528, opened 2016-02-08
Clip line geometry past tile boundaries

### LINKED ISSUE #527: Fix outline behavior for tiles with a geometry "buffer"
The tile pipeline should correctly handle tile sources that provide geometry clipped to some boundary outside the tile area. The issue and proposed solution in tangram-js are described here: https://github.com/tangrams/tangram/pull/242

tallytalwar:
- Resolves handling of buffered tiles
Fixes #527 

Refer Tangram JS: https://github.com/tangrams/tangram/pull/242

### DIFF
--- a/core/src/util/builders.cpp
+++ b/core/src/util/builders.cpp
@@ -234,30 +234,19 @@ void addCap(const glm::vec3& _coord, const glm::vec2& _normal, int _numCorners, 
     addFan(_coord, nA, nB, nC, uA, uB, uC, _numCorners, _ctx);
 }
 
-bool nearlyEqual(float _a, float _b, float _tolerance = 0.001) {
-    return fabsf(_a - _b) < _tolerance;
-}
-
-// Tests if a line segment (from point A to B) is nearly coincident with the edge of a tile
-bool isOnTileEdge(const glm::vec3& _a, const glm::vec3& _b) {
+// Tests if a line segment (from point A to B) is outside the edge of a tile
+bool isOutsideTile(const glm::vec3& _a, const glm::vec3& _b) {
 
     float tolerance = 0.0005; // tweak this adjust if catching too few/many line segments near tile edges
     // TODO: make tolerance configurable by source if necessary
     glm::vec2 tile_min(0.0, 0.0);
     glm::vec2 tile_max(1.0, 1.0);
 
-    if (nearlyEqual(_a.x, _b.x, tolerance)) {
-        auto x = fmod(_a.x, tile_max.x);
-        if (nearlyEqual(x, tile_min.x, tolerance) || nearlyEqual(x, tile_max.x, tolerance)) {
-            return true;
-        }
-    }
-    if (nearlyEqual(_a.y, _b.y, tolerance)) {
-        auto y = fmod(_a.y, tile_max.y);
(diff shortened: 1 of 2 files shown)


bcamper:
Please test thoroughly!

core/src/util/builders.cpp
```diff
@@ -383,7 +372,7 @@ void Builders::buildPolyLine(const Line& _line, PolyLineBuilder& _ctx) {
         for (size_t i = 0; i < _line.size() - 1; i++) {
             const glm::vec3& coordCurr = _line[i];
             const glm::vec3& coordNext = _line[i+1];
-            if (isOnTileEdge(coordCurr, coordNext)) {
+            if (isOutsideTile(coordCurr, coordNext)) {
                 Line line = Line(&_line[cut], &_line[i+1]);
                 buildPolyLineSegment(line, _ctx);
```
hjanetzek: Is this correct? The range to cut-coordNext is added while the segment coordCurr-CoordNext should be clipped?

tallytalwar: Nopes, I do not think its correct. Will fix it.

blair1618: I believe the original is correct. It's invoking a constructor for `std::vector`, the 4th version here: http://en.cppreference.com/w/cpp/container/vector/vector. Note that `last` is excluded from the resulting vector.

tallytalwar: Ohh Yes, Line is a vector. Good call and my bad.

hjanetzek: ah, right! though &_line[cut + length] would be easier to comprehend I think. One could also skip the length == 1 case here.

hjanetzek: The good old cohen-sutherland trivial reject case :)

I would add the tolerance to slightly extend the tile bbox. By subtracting there is a chance that a real boundary line gets omitted

tallytalwar: @hjanetzek I am not sure what do you mean by `&_line[cut + length]`, if its that we can ignore all the coordinates past "this" edge which is outside the tile bounds then I do not think we should do that. Its possible some edges between `i+1` and `length` are within the tile bounds. [by edges I mean line segments formed by the line coordinates].

And `length == 1` will automatically be checked in `buildPolyLineSegment`.

I agree on extending the tile bbox to do the bound check. Will change this.

hjanetzek: @tallytalwar ```i+1``` should be clear enough. just nitpicking. using ```some_iterator + length``` looks to me to be explicitly an end iterator.

```
       auto cut = _line.begin();
       const glm::vec3& coordCurr = _line.front();
        for (auto it = cut + 1; it < _line.end(); it++) {
              const glm::vec3& coordNext = *it;
              if (isOutsideTile(coordCurr, coordNext)) {
                  size_t length = it - cut;
                  if (length > 1) {
                      Line line = Line(cut, cut + length);
                      buildPolyLineSegment(line, _ctx);
                  }
                  cut = it;
              }
              coordCurr = coordNext;
         }
         Line line = Line(cut, _line.end());
         buildPolyLineSegment(line, _ctx);
```

hjanetzek:
LGTM

PR #552, opened 2016-02-17
Let glfw focus monitor when reinitializing main window

karimnaaji:
This fixes something quite annoying when the main window is created from the main monitor and then dragging the window to the second, it sometimes keeps focusing the window on the main monitor which makes you unable to move it to the second.


blair1618:
Ah, I see the issue now - somehow I had never tried this with the GLFW window. The fix looks good, I'll merge once travis is done.

PR #554, opened 2016-02-18
Static uniform string alloc

karimnaaji:
Cache strings for uniforms and hash on the address of it while accessing the uniform hash table of the shader.


karimnaaji:
#577

PR #555, opened 2016-02-22
Improved style mixing

blair1618:
This is a refactoring and optimization of the style mixing process that removes several hundred lines of code, clarifies mixing behavior, and **improves style loading speed by roughly 100X**. 

 - Style mixing functions are now in a dedicated file with clearly separated and documented roles
 - Style mixing uses topological sorting to resolve dependencies
 - Tests related to style mixing are now in a single location and are more clean and concise


tallytalwar:
Very nicely done. 100x god speed.

blair1618:
@karimnaaji take a look when you're able. Thanks!

karimnaaji:
Sure, will review.

core/src/style/styleMixer.cpp
```diff
@@ -0,0 +1,222 @@
...
+            // Skip mixing built-in styles.
+            const auto& builtIn = Style::builtInStyleNames();
+            if (std::find(builtIn.begin(), builtIn.end(), styleNameToMix) != builtIn.end()) {
+                continue;
+            }
+
+            mixins.push_back(_styles[styleNameToMix]);
+        }
+
+        applyStyleMixins(style, mixins);
```
karimnaaji: When mixing with base style, `mixins` is most of the time empty, we can save some allocs and only apply the `shaders:blocks:` mixing in that case.

core/src/scene/sceneLoader.cpp
```diff
@@ -92,16 +93,21 @@ bool SceneLoader::loadScene(Node& config, Scene& _scene) {
...
     }
 
-    std::unordered_set<std::string> mixedStyles;
-
     if (Node styles = config["styles"]) {
-        for (const auto& style : styles) {
+        StyleMixer mixer;
+        try {
+            mixer.mixStyleNodes(styles);
+        } catch ( YAML::RepresentationException e) {
```
karimnaaji: nitpick: Extra space

core/src/style/styleMixer.h
```diff
@@ -0,0 +1,49 @@
...
+
+    void mergeBooleanFieldAsDisjunction(std::string _key, Node _target, const std::vector<Node>& _sources);
+
+    void mergeFieldTakingLast(std::string _key, Node _target, const std::vector<Node>& _sources);
+
+    void mergeMapFieldTakingLast(std::string _key, Node _target, const std::vector<Node>& _sources);
+
+};
+
+}
```
karimnaaji: nitpick: End of file, for :octocat:

core/src/scene/sceneLoader.cpp
blair1618: Ah thanks, I'll fix that.

core/src/style/styleMixer.h
blair1618: Yep, I'll get that!

core/src/style/styleMixer.cpp
```diff
@@ -0,0 +1,222 @@
...
+std::vector<std::string> StyleMixer::getMixingOrder(const Node& _styles) {
+
+    // Input must be a map of names to style configuration nodes.
+    if (!_styles.IsMap()) {
+        return {};
+    }
+
+    // Dependencies are pairs of strings that form a DAG.
+    // If style 'a' mixes style 'b', the dependency would be {'b', 'a'}.
+    std::vector<std::pair<std::string, std::string>> dependencies;
```
karimnaaji: We can probably declare this local static and clear it every time this get called, this would reduce allocations.

blair1618: I think for this allocation and the one you mentioned before, I could make them instance variables and clear them when appropriate. `StyleMixer` doesn't actually have any member variables right now so that would also help justify it as a class instead of just a namespace of functions.

hjanetzek:
Looking good overall. Speedup of style mixing is more 500x :) 

One thing: styleMixer belongs into scene directory imo: it's basically part of scene loading and has no dependencies on style.h (other than the list of base style names)

tallytalwar:
500x god speed.

Agree with Hannes, styleMixer should be a part of scene dir.

blair1618:
@hjanetzek Yeah you have a good point about the file location. The size of the "scene" directory is getting somewhat cumbersome, but the right solution is probably subdirectories or a new directory for scene configuration classes.

karimnaaji:
Agreed with @hjanetzek , I got confused looking at where styleMixing was at first, I was looking for it into scene folder.

core/src/style/styleMixer.cpp
karimnaaji: Yes that would make sense to do that too.

core/src/util/topologicalSort.h
```diff
@@ -0,0 +1,68 @@
...
+#include <vector>
+
+/**
+ * Given a list of edges describing a directed acyclic graph, returns a
+ * list of nodes sorted in topological order; that is, if there is a path
+ * from node 'a' to node 'b' in the graph, then 'a' will always precede
+ * 'b' in the sorted result.
+ */
+template<typename T>
+std::vector<T> topologicalSort(std::vector<std::pair<T, T>> edges) {
```
karimnaaji: It might make sense to add this to Tangram namespace.

blair1618: Oh yeah, I probably should since we've done that with all our other utility classes. Will do!

karimnaaji:
Looks really good to me! Code is cleaner and style mixing gets much more comprehensible.
The scene loads much faster now (drops from an overall of ~1500ms to ~500ms in my tests for eraser-map).

core/src/style/styleMixer.cpp
blair1618: Well normally I always prefer to avoid unneeded allocation, but I looked into the cost of just allocating these temporary vectors as needed and it's totally negligible - they aren't even visible in the iOS sampling profiler. I think we're good here.

core/src/util/topologicalSort.h
tallytalwar: Nice usage.

Issue #535, opened 2016-02-11
Handle arbitrary layers from a client geojson datasource


tallytalwar:
Currently tangram-es core architecture requires datasource to contain features in appropriate uber layers, which are then be styled/filtered upon. However it is totally fine for a geojson to provide a feature collection with no uber layer defined (eg: https://transit.land/api/v1/operators.geojson).

Tangram-es should provide a way for these feature sets to be filtered/styled and drawn.

bcamper:
Almost all GeoJSON sources will be a standard FeatureCollection. The presence of a JSON wrapper splitting these into "layers" is just an arbitrary format that Tangram implemented. In Tangram JS we're supporting both -- a single GeoJSON object, or a set of layers provided as GeoJSON objects -- for both tiled and non-tiled GeoJSON (it was initially implemented for tiled GeoJSON since that's how our tiles were formatted, and then carried over to non-tiled GeoJSON because the code is shared).

bcamper:
For styling/filtering purposes, single-layer data sources simply omit the `layer` property in the `data` block, e.g.:

`data: { source: geojson }` is used to reference the data

PR #565, opened 2016-02-24
Allow GeoJsonSource to parse a root object FeatureCollection

blair1618:
Fixes https://github.com/tangrams/tangram-es/issues/535

With this change, `GeoJsonSource` will accept a JSON input whose root object is either a GeoJSON FeatureCollection object or a mapping of layer names to FeatureCollection objects. If the root object is a FeatureCollection, the resulting data has an empty layer name and no top-level layer filtering can be applied.

### DIFF
--- a/core/src/data/geoJsonSource.cpp
+++ b/core/src/data/geoJsonSource.cpp
@@ -45,8 +45,15 @@ std::shared_ptr<TileData> GeoJsonSource::parse(const TileTask& _task,
     };
 
     // Transform JSON data into TileData using GeoJson functions
-    for (auto layer = document.MemberBegin(); layer != document.MemberEnd(); ++layer) {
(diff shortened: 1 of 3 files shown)


hjanetzek:
To support our GeoJSON layers format getLayer should get the layer name, passing empty string for raw GeoJSON.

blair1618:
Yep I could pass the name into `GeoJson::getLayer` - it seemed simpler at the time to just set the name after the layer is parsed ([here](https://github.com/tangrams/tangram-es/pull/565/files#diff-ee0322514faf83324ee97f8d2a78acd1R54)), but that would be a simple revision.

hjanetzek:
Ah it's ok like it is - didn't saw the update of layer.name.

Looks good to merge

PR #566, opened 2016-02-24
View cleanup

blair1618:
This is just a bit of tidying for `View`. Moving the rasterization functions into a separate file cleans up `view.cpp` nicely and also means we could use raster scanning in other contexts.


karimnaaji:
Looking good to me! @hjanetzek you can merge once you had a chance to give a look.

Issue #568, opened 2016-02-24
Mipmap filtering should only be applied to power-of-two textures


blair1618:
Mipmaps can only be generated for textures whose width and height dimensions are both a power of two. For all other textures, `glGenerateMipmap` will produce a GL error and filtering the texture with `GL_*_MIPMAP_*` will cause the sampler to behave incorrectly on some devices (notably on PowerVR GPUs). 

Tangram-es should check the dimensions of a texture before applying mipmapping and fall back to linear filtering if the texture does not have power-of-two dimensions.

PR #569, opened 2016-02-24
Npot texture fallback

karimnaaji:
Addressing #568 , fallback to linear filtering when repeat or mipmaps is used for npot textures.


blair1618:
Lookin' good with our eraser-map icons. 
screen shot 2016-02-25 at 11 59 06 am

PR #572, opened 2016-02-25
More generous view limits

### LINKED ISSUE #531: Map view should make a "best effort" to apply parameters that exceed world bounds
When you try to set a view parameter (position, zoom, rotation, or tilt) that would result in seeing area beyond the edge of the world, tangram-es stops the change completely and remains at the previous value. What I would expect as a developer and as a user would be for tangram-es to take the closest possible value to the input, rather than doing nothing.

blair1618:
This is a looser, more approximate, but more practical approach to limiting the map view to the bounds of the world. It constrains the view position and zoom according to a characteristic radius that approximates the extent of the view in world space on all sides; basically it's just colliding a circle against the specified boundaries in x and or y. This means that, in contrast with the previous implementation, you're always allowed to tilt and rotate the view - which means that it's possible to view areas beyond the world bounds. However, this also means that the view no longer gets "stuck" against the edge of the world and when a position/zoom is requested that can't be exactly satisfied, the view will now get as close as possible to that position/zoom instead of doing nothing.

The easiest way to see the difference here is just to try it yourself :)

Fixes #531

### DIFF
--- a/core/src/view/viewConstraint.cpp
+++ b/core/src/view/viewConstraint.cpp
@@ -1,0 +1,68 @@
+#include "viewConstraint.h"
+
+namespace Tangram {
+
+void ViewConstraint::setLimitsY(double yMin, double yMax) {
+
+    m_yMin = yMin;
+    m_yMax = yMax;
+
+}
+
+void ViewConstraint::setLimitsX(double xMin, double xMax) {
+
+    m_xMin = xMin;
+    m_xMax = xMax;
+
+}
+
+void ViewConstraint::setRadius(double r) {
+
+    m_radius = r;
+
+}
+
+auto ViewConstraint::getConstrainedX(double x) -> double {
+
+    return constrain(x, m_radius, m_xMin, m_xMax);
+
+}
+
+auto ViewConstraint::getConstrainedY(double y) -> double {
+
+    return constrain(y, m_radius, m_yMin, m_yMax);
+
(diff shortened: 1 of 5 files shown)


tallytalwar:
Reviewing this now.

core/src/view/viewConstraint.cpp
```diff
@@ -0,0 +1,68 @@
...
+    if (diameter > yRange) { yScale = yRange / diameter; }
+    if (diameter > xRange) { xScale = xRange / diameter; }
+
+    return std::fmin(xScale, yScale);
+
+}
+
+auto ViewConstraint::constrain(double _pos, double _radius, double _min, double _max) -> double {
+
+    double spaceAbove = _max - (_pos + _radius);
```
tallytalwar: Instead of Above and Below, can we be more generic here.

blair1618: You mean with naming? I was just thinking in terms of a number line. Do you think something else would be clearer?

blair1618: The STL uses "upper" and "lower" for this kind of thing, e.g. http://en.cppreference.com/w/cpp/algorithm/upper_bound

Is that better?

tallytalwar: I am not going to argue with STL :D.

PR #574, opened 2016-02-25
Upgrade to OkHTTP3

### LINKED ISSUE #573: Upgrade OkHTTP
Our version is behind and might have bugs / inefficiencies.

blair1618:
New version required some syntax changes and calls can no longer be cancelled by tag, so we now maintain a list of current calls and search that list when a cancel is made.

Resolves #573

### DIFF
--- a/android/tangram/build.gradle
+++ b/android/tangram/build.gradle
@@ -46,7 +46,7 @@ afterEvaluate {
 }
 
 dependencies {
-  compile 'com.squareup.okhttp:okhttp:2.5.0'
+  compile 'com.squareup.okhttp3:okhttp:3.2.0'
   compile 'xmlpull:xmlpull:1.1.3.1'
 }
 
--- a/android/tangram/src/com/mapzen/tangram/HttpHandler.java
+++ b/android/tangram/src/com/mapzen/tangram/HttpHandler.java
@@ -1,24 +1,26 @@
 package com.mapzen.tangram;
 
-import com.squareup.okhttp.Cache;
-import com.squareup.okhttp.Callback;
-import com.squareup.okhttp.OkHttpClient;
-import com.squareup.okhttp.Request;
+import okhttp3.Cache;
+import okhttp3.Call;
+import okhttp3.Callback;
+import okhttp3.OkHttpClient;
+import okhttp3.Request;
 
 import java.io.File;
-import java.io.IOException;
 import java.util.concurrent.TimeUnit;
 
 public class HttpHandler {
 
     private OkHttpClient okClient;
+    private OkHttpClient.Builder okClientBuilder;
     private Request.Builder okRequestBuilder;
 
     public HttpHandler() {
         okRequestBuilder = new Request.Builder();
-        okClient = new OkHttpClient();
-        okClient.setConnectTimeout(10, TimeUnit.SECONDS);
-        okClient.setReadTimeout(30, TimeUnit.SECONDS);
+        okClientBuilder = new OkHttpClient().newBuilder()
+                .connectTimeout(10, TimeUnit.SECONDS)
+                .readTimeout(30, TimeUnit.SECONDS);
+        okClient = okClientBuilder.build();
     }
 
     /**
@@ -28,7 +30,7 @@ public class HttpHandler {
      * @return true if request was successfully started
      */
     public boolean onRequest(String url, Callback cb) {
-        Request request = okRequestBuilder.tag(url).url(url).build();
+        Request request = okRequestBuilder.url(url).build();
         okClient.newCall(request).enqueue(cb);
         return true;
     }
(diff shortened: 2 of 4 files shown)


android/tangram/src/com/mapzen/tangram/HttpHandler.java
```diff
@@ -1,24 +1,39 @@
...
+    class CallEntry {
+        CallEntry(String url, Call call) {
+            this.url = url;
+            this.call = call;
+        }
+        public String url;
+        public Call call;
+    }
+
+    private ArrayList<CallEntry> calls = new ArrayList<>();
```
tallytalwar: okhttp3 dispatcher provides `queuedCalls` and `runningCalls` API, so we do not have to maintain a list on our own.

Refer: https://square.github.io/okhttp/3.x/okhttp/okhttp3/Dispatcher.html

android/tangram/build.gradle
```diff
@@ -46,7 +46,7 @@ afterEvaluate {
 }
 
 dependencies {
-  compile 'com.squareup.okhttp:okhttp:2.5.0'
+  compile 'com.squareup.okhttp3:okhttp:3.1.2'
```
tallytalwar: As off 25th Feb, there is a newer version **3.2.0** available for okhttp3 on maven. Probably we should move to that.

blair1618:
Now using version 3.2; call cancelling uses lists of running and queued calls provided by the dispatcher.

android/tangram/src/com/mapzen/tangram/HttpHandler.java
```diff
@@ -28,8 +30,9 @@ public HttpHandler() {
      * @return true if request was successfully started
      */
     public boolean onRequest(String url, Callback cb) {
-        Request request = okRequestBuilder.tag(url).url(url).build();
-        okClient.newCall(request).enqueue(cb);
+        Request request = okRequestBuilder.url(url).build();
+        Call call = okClient.newCall(request);
```
tallytalwar: nitpick. Can you collapse lines 34 and 35? (referring to the previous implementation).

I guess you had this when you were explicitly maintaining the list of calls.

android/tangram/src/com/mapzen/tangram/HttpHandler.java
```diff
@@ -49,8 +61,7 @@ public void onCancel(String url) {
      */
     public boolean setCache(File directory, long maxSize) {
         Cache okTileCache = new Cache(directory, maxSize);
-        okClient.setCache(okTileCache);
-
+        okClient = okClientBuilder.cache(okTileCache).build();
```
tallytalwar: Is there a reason you are building a new okClient here, instead of setting the cache on the already built okClient (in the HttpHandler constructor)? If this is an API change I am not able to find it.

android/tangram/src/com/mapzen/tangram/HttpHandler.java
blair1618: Resolved

android/tangram/src/com/mapzen/tangram/HttpHandler.java
blair1618: `OkHttpClient` no longer has a `setCache(Cache)` method.

android/tangram/src/com/mapzen/tangram/MapController.java
```diff
@@ -514,20 +515,18 @@ public boolean startUrlRequest(String url, final long callbackPtr) throws Except
...
-            public void onResponse(Response response) throws IOException {
+            public void onResponse(Call call, Response response) throws IOException {
                 if (!response.isSuccessful()) {
                     onUrlFailure(callbackPtr);
                     throw new IOException("Unexpected response code: " + response);
                 }
-                BufferedSource source = response.body().source();
-                byte[] bytes = source.readByteArray();
-                onUrlSuccess(bytes, callbackPtr);
+                onUrlSuccess(response.body().bytes(), callbackPtr);
```
tallytalwar: Nice! 

Don't know why I initially did not use `ResponseBody::bytes()`, instead of going to `okio.BufferedSource` and grabbing the bytes from there. Maybe this is also in their new API :D.

android/tangram/src/com/mapzen/tangram/HttpHandler.java
tallytalwar: Yup.

Referring their changelogs:

> 
> - OkHttpClient is now stateless. In the 2.x API OkHttpClient had getters and setters. Internally each request was forced to make its own complete snapshot of the OkHttpClient instance to defend against racy configuration changes. In 3.x, OkHttpClient is now stateless and has a builder. Note that this class is not strictly immutable as it has stateful members like the connection pool and cache.
> 
> - Get and Set prefixes are now avoided. With ubiquitous builders throughout OkHttp these accessor prefixes aren't necessary. Previously OkHttp used get and set prefixes sporadically which make the API inconsistent and awkward to explore.

PR #576, opened 2016-02-26
improve flinging experience on android

hjanetzek:


hjanetzek:
@blair1618 why is the start of fling delayed by one frame? 
fling -> requestRender(was missing) -> update, unset gestureOccured -> requestRender(was missing) -> update, apply fling -> requestRender

blair1618:
Ah this is starting to make sense now. I think applying fling was effectively delayed by one frame because It used to be the case that "fling" wasn't a distinct gesture with it's own handler, but rather it was inferred from the end of a pan gesture - so we could only rationally apply it when no gestures were detected in the last frame. However, now that we get explicit calls on "fling", there' should be no need to check whether a gesture occurred in the last frame before applying the fling. 

I'm kind of baffled that flinging was working correctly on iOS - I was led to believe that the problem must have been in the Android code. This change makes sense to me, I'll give it a try.

blair1618:
Oooooh this is much better!

Ok so I think [line 37](https://github.com/tangrams/tangram-es/pull/576/files#diff-2e6846abe6e81b573de7308af2838a13R37) only needs to check `isFlinging`, which means we don't need to add [line 46](https://github.com/tangrams/tangram-es/pull/576/files#diff-2e6846abe6e81b573de7308af2838a13R46).

core/src/util/inputHandler.cpp
```diff
@@ -31,21 +31,19 @@ void InputHandler::update(float _dt) {
...
-    if (!m_gestureOccured && isFlinging) {
+    if (isFlinging) {
 
         m_velocityPan -= _dt * DAMPING_PAN * m_velocityPan;
         m_view->translate(_dt * m_velocityPan.x, _dt * m_velocityPan.y);
 
         m_velocityZoom -= _dt * DAMPING_ZOOM * m_velocityZoom;
         m_view->zoom(m_velocityZoom * _dt);
-
+
```
blair1618: spaaaaace

hjanetzek:
@blair1618, sorry just tried the github editing interface..

PR #577, opened 2016-02-29
Uniform entries

karimnaaji:
Optimize string allocations for uniforms
- Statically allocate uniform strings to reduce redundant allocations
- Shader location cache: Hash on the id of the uniform (`uint16_t`) entry instead of the string

Update, (order of speedup is comparable on mobile):

master branch,
```
build/bench/bin  master ✗
> ./bench/uniform.out
Run on (8 X 1000 MHz CPU s)
2016-02-29 12:56:20
Benchmark               Time(ns)    CPU(ns) Iterations
------------------------------------------------------
BM_Tangram_SetUniform        159         61   10803970                                 
```
uniform-entries branch,
```

build/bench/bin  uniform-entries ✗
> ./bench/uniform.out 
Run on (8 X 1000 MHz CPU s)
2016-02-29 12:59:06
Benchmark               Time(ns)    CPU(ns) Iterations
------------------------------------------------------
BM_Tangram_SetUniform          8         17   36471060    
```

### DIFF
--- a/core/src/gl/shaderProgram.cpp
+++ b/core/src/gl/shaderProgram.cpp
@@ -120,21 +120,6 @@ GLint ShaderProgram::getAttribLocation(const std::string& _attribName) {
 
 }
 
-GLint ShaderProgram::getUniformLocation(const std::string& _uniformName) {
-
-    //// Get uniform location at this key, or create one valued at -2 if absent
-    //GLint& location = m_uniformMap[(intptr_t)&_uniformName].loc;
-
-    //// -2 means this is a new entry
-    //if (location == -2) {
-    //    // Get the actual location from OpenGL
-    //    location = glGetUniformLocation(m_glProgram, _uniformName.c_str());
-    //}
-
-    //return location;
-
-}
-
 GLint ShaderProgram::getUniformLocation(const UniformEntries::UniformEntry* _entry) {
 
     if (_entry == nullptr) {
--- a/core/src/gl/shaderProgram.cpp
+++ b/core/src/gl/shaderProgram.cpp
@@ -11,6 +11,42 @@
 
 namespace Tangram {
 
+namespace Uniform {
+    UniformEntries::EntryId color;
+    UniformEntries::EntryId time;
+    UniformEntries::EntryId model;
+    UniformEntries::EntryId tileOrigin;
+    UniformEntries::EntryId devicePixelRatio;
+    UniformEntries::EntryId resolution;
+    UniformEntries::EntryId mapPosition;
+    UniformEntries::EntryId normalMatrix;
+    UniformEntries::EntryId inverseNormalMatrix;
+    UniformEntries::EntryId metersPerPixel;
+    UniformEntries::EntryId view;
+    UniformEntries::EntryId proj;
+    UniformEntries::EntryId ortho;
+    UniformEntries::EntryId orthoProj;
+    UniformEntries::EntryId modelViewProj;
+    UniformEntries::EntryId tex;
+    UniformEntries::EntryId pass;
+    UniformEntries::EntryId uvScaleFactor;
+    UniformEntries::EntryId materialEmission;
+    UniformEntries::EntryId materialEmissionTexture;
+    UniformEntries::EntryId materialEmissionScale;
+    UniformEntries::EntryId materialAmbiant;
+    UniformEntries::EntryId materialAmbiantTexture;
+    UniformEntries::EntryId materialAmbiantScale;
+    UniformEntries::EntryId materialDiffuse;
+    UniformEntries::EntryId materialDiffuseTexture;
+    UniformEntries::EntryId materialDiffuseScale;
+    UniformEntries::EntryId materialShininess;
+    UniformEntries::EntryId materialSpecular;
+    UniformEntries::EntryId materialSpecularTexture;
+    UniformEntries::EntryId materialSpecularScale;
+    UniformEntries::EntryId materialNormalTexture;
+    UniformEntries::EntryId materialNormalScale;
+    UniformEntries::EntryId materialNormalAmount;
+}
 
 ShaderProgram::ShaderProgram() {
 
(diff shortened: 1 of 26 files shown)


blair1618:
Interesting... well it is true that our draw loop spends probably too much time on the process of finding and setting uniforms, but it would be nice if there were a less verbose way of handling this problem :\

core/src/gl/shaderProgram.cpp
```diff
@@ -84,15 +120,20 @@ GLint ShaderProgram::getAttribLocation(const std::string& _attribName) {
 
 }
 
-GLint ShaderProgram::getUniformLocation(const std::string& _uniformName) {
+GLint ShaderProgram::getUniformLocation(const UniformEntries::UniformEntry* _entry) {
```
blair1618: I think the input here (and for the other `ShaderProgram` methods) could be just the ID of the uniform. When the program needs to look up the location on the first call, it can get the name using the uniform namespace functions - this reduces the indirection needed for all but the first call.

core/src/gl/uniformEntries.h
```diff
@@ -0,0 +1,36 @@
...
+
+#include <cstdint>
+#include <memory>
+#include <string>
+
+namespace Tangram {
+
+namespace UniformEntries {
+
+typedef uint16_t EntryId;
```
blair1618: Why not `using`?

Also what about a signed type so we could use the sign to signify invalid IDs?

karimnaaji: This is quite equivalent to me, I'll update since we seems to prefer `using` over `typedef` in the codebase.

karimnaaji: Invalid ID is 0.

core/src/gl/uniformEntries.h
```diff
@@ -0,0 +1,36 @@
...
+
+struct UniformEntry {
```
blair1618: Seems to me like we don't need the ID/index/handle to be co-located with the name, since the name only needs to be accessed rather infrequently. We could keep a separate, global array of name strings that we index into with the IDs.

core/src/gl/shaderProgram.cpp
karimnaaji: Right, I preferred the const pointer to make it clear that this can be null, but since a uniform ID of 0 is invalid that's also a good solution, I'll update for this.

hjanetzek: Yes, this would also be my suggestion or to even cache the uniform value on the caller side. Though lights are a special (or more complicated) case where I got stuck when I tried this 'uniform caching' a while ago.

blair1618: Yeah the way that `Light`s use uniforms is pretty annoying >.< I'd like to change them

core/src/gl/uniformEntries.h
```diff
@@ -0,0 +1,36 @@
+#pragma once
+
+#include <cstdint>
+#include <memory>
+#include <string>
+
+namespace Tangram {
+
+namespace UniformEntries {
```
blair1618: Do you think we can simplify some of these names? e.g. `namespace Uniform { ...`, `using Id/Index/Handle = uint16_t; `, and we may not need `UniformEntry` at all

karimnaaji:
@blair1618 Yes I agree on that :/ I will see how feasible it is to make it less verbose for the same functionality.

core/src/gl/shaderProgram.cpp
```diff
@@ -11,6 +11,42 @@
 
 namespace Tangram {
 
+namespace Uniform {
```
blair1618: Looks a lot like a scoped enum to me! If we made this an `enum class : uint16_t` I think we could avoid duplicating this list.

core/src/gl/shaderProgram.cpp
karimnaaji: For `Light`s I think we can reliably know the number of light names at scene loading, generate those uniform names and cache then relevant uniforms when the scene just finished loading.

core/src/gl/uniformEntries.h
karimnaaji: Right, the only reason I tight them together was for debugging purpose, otherwise it's hard to tell which ID relates to a given uniform name. If this helps making it less verbose I'll update for a global static array.

core/src/gl/uniformEntries.h
karimnaaji: I'm fine to change the name for `Uniform`. See the other comment for the need of `UniformEntry`.

core/src/gl/uniformEntries.h
blair1618: Ah yeah I can see how debugging might be a pain. If we make our "built-in" uniforms an enum then those will be named in a debugger - generated uniforms would still be hard to debug.

core/src/gl/shaderProgram.cpp
hjanetzek: @karimnaaji  Sounds good!

But for the API I think it could be made a bit simpler :) Untested code below:

```
UniformLocation {
private:
   const char* name; // owned by static s_uniformStringAtoms
   int id;

public:

   UniformLocation(string name) {
           
         // important: use char* so that references stay valid when vector reallocates
         static vector<std::unique_ptr<char*>>  s_uniformStringAtoms;

         auto it = s_uniformStringAtoms.find(name);
         if (it = s_uniformStringAtoms.end()) {
            id = s_uniformStringAtoms.size();

            s_uniformStringAtoms.emplace_back(name.c_str());
            name = s_uniformStringAtoms.back().get();
            return;
         }

         id = it - s_uniformStringAtoms.begin();
         name = it->get();
  }

   UniformLocation(const UniformLocation& _other) {
       name = _other.name;
       id = _other.id;
   }


   static UniformLocation normalMatrix("u_normalMatrix");
    ...
}


void ShaderProgram::setUniformf(const UniformLocation& _loc, float _value) {}

```

hjanetzek: A better approach would be to also cache the real uniform location on the call site, this would require that UniformLocation<s> are unique for each shader. This should be easy to do except for Lights. For Lights injectOnProgram could return a struct of UniformLocations in the call light->injectOnProgram(*m_shaderProgram) for each shader program

karimnaaji:
#586

PR #579, opened 2016-02-29
fix:  lighting 'false' option

hjanetzek:
This one needs a careful look:

- the color block was not applied in vertex shader when lighting mode is ```none```
- normal blocks were applied in vertex and fragment shader for vertex lighting

### DIFF
--- a/core/resources/shaders/polygon.fs
+++ b/core/resources/shaders/polygon.fs
@@ -49,19 +49,18 @@ void main(void) {
     vec4 color = v_color;
     vec3 normal = v_normal;
 
+    // Modify material properties before lighting
     #ifdef TANGRAM_MATERIAL_NORMAL_TEXTURE
         calculateNormal(normal);
     #endif
 
-    // Modify normal before lighting
-    #pragma tangram: normal
-
-    // Modify color and material properties before lighting
-    #ifndef TANGRAM_LIGHTING_VERTEX
-        #pragma tangram: color
-    #endif
-
     #ifdef TANGRAM_LIGHTING_FRAGMENT
+        // Modify color before lighting
+        #pragma tangram: color
+
+        // Modify normal before lighting
+        #pragma tangram: normal
+
         color = calculateLighting(v_position.xyz, normal, color);
     #else
         #ifdef TANGRAM_LIGHTING_VERTEX
--- a/core/resources/shaders/polygon.vs
+++ b/core/resources/shaders/polygon.vs
@@ -87,7 +87,7 @@ void main() {
     // Set position varying to the camera-space vertex position
     v_position = u_view * position;
 
-    #ifdef TANGRAM_LIGHTING_VERTEX
+    #ifndef TANGRAM_LIGHTING_FRAGMENT
         vec4 color = v_color;
         vec3 normal = v_normal;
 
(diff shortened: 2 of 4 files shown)


bcamper:
We do apply the `normal` block when lighting is `vertex` in JS -- it seems valid to still modify a normal even if it only applies to vertex-level calculations. Obviously it only works for Lambert (N dot L) or Gouraud-style shading, but it's still useful.

https://github.com/tangrams/tangram/blob/master/src/styles/polygons/polygons_vertex.glsl#L102

hjanetzek:
It could be a bit confusing when the ```normal``` block is applied in both shaders. One time in vertex shader for use in lighting and a second time in the fragment shader. Is it likely that the same block is correct in both cases? Otherwise one would always have to differentiate with define guards in which shader the ```normal``` block is used

bcamper:
Oh, good catch, yes, I believe the `normal` block should only applied once: in the vertex shader for `vertex` lighting, and in the fragment shader for `fragment` lighting. In both cases, before the `color` block, and before lighting calculations. Make sense?

hjanetzek:
@bcamper is it intended that ```color``` is applied in for ```vertex``` lighting in vertex shader but not for ```none``` lighting?

  https://github.com/tangrams/tangram/blob/master/src/styles/polygons/polygons_fragment.glsl#L49

bcamper:
Hmm, probably not? :) This logic is quite old.

cc @blair1618 @patriciogonzalezvivo @meetar 

Does this make sense as desired behavior?

- Lighting `fragment`: `normal` and `color` applied in fragment shader
- Lighting `vertex`: `normal` and `color` applied in vertex shader
- Lighting `false`: `color` applied in vertex shader, `normal` not applied
- In all cases, `filter` applied in fragment shader for final pixel color modifications

tallytalwar:
@bcamper your bullet points make sense to me.

blair1618:
Is there harm in applying a `normal` block for `lighting: false`? The normal vector will still be accessible to users and I could imagine someone wanting to modify and use this free varying vector for lighting-like effects. Do we want to allow this?

If so, the question is then: does the `normal` block go into the vertex shader or the fragment shader? Since `color` is applied in the vertex shader for `lighting: false`, it might be natural (if somewhat arbitrary) to apply `normal` in the vertex shader as well for this case. 

If not, well, I suppose a user could create their own varying vector to mirror the normal vector and apply the normal matrix as needed, or figure out that they need to enable lighting.

blair1618:
After some further conversation, it sounds like we've found a slightly different consensus:

 - `lighting: fragment`: Blocks `normal` and `color` executed in the fragment shader
 - `lighting: vertex`: Blocks `normal` and `color` executed in the vertex shader
 - `lighting: false`: Blocks `normal` and `color` executed in the fragment shader

At first it would seem unnecessary to execute the `color` and `normal` blocks per-fragment when `lighting: false` is applied, but this is necessary to enable custom styles that create patterns or procedural textures without also forcing lighting to be applied. 

cc @bcamper @patriciogonzalezvivo

bcamper:
Yet *another* take on this here, after some more experimentation: https://github.com/tangrams/tangram/pull/271

blair1618:
The resolved changes are implemented in #647, closing this in favor of the newer PR for simplicity.

PR #580, opened 2016-02-29
fix: caching of GeometryType in StyleContext

hjanetzek:
- did not work since the switch from constants to strings


blair1618:
Looks good to me

PR #581, opened 2016-02-29
MVT source optimization

hjanetzek:
Back-porting some optimization unrelated to tile-data-sink for fairer comparison


blair1618:
Changing those hard-coded `int`s to `#define`s is much better. What's the deal with `lastWasFeature`?

hjanetzek:
> mvt: only keep first iterator of feature messages
instead of collecting *all* feature messages, just keeping the
start iterators of feature messages (which is usually one
contiguous block)

Just citing the commit message :)

blair1618:
Oh duh, I'll try reading next time haha

Looks good!

PR #586, opened 2016-03-01
[wip] Simplify uniform cache

hjanetzek:
Trying a different API to store uniform locations. The gl uniform location is cached on the caller side in a UniformLocation handle so one can avoid one lookup of some name (or id) to the actual gl handle.


karimnaaji:
@hjanetzek , this looks good so far to me, but _please_ do keep informed once comments that are being addressed in one PR are being worked on, otherwise it's a waste of time; simple coordination.
Is that taking over #577 then?

hjanetzek:
@karimnaaji I wanted to try the latter approach that I suggested in #577 this morning. Wasn't sure how it will turn out - but, yes, I should have added a notice that I'm working on it. 
The LightUniforms initialization could be still improved but overall this call-side uniform caching seems to work

karimnaaji:
Sounds good, will review once ready.

core/src/style/material.cpp
```diff
@@ -158,56 +158,56 @@ void Material::injectOnProgram(ShaderProgram& _shader ) {
 void Material::setupProgram(ShaderProgram& _shader) {
 
     if (m_bEmission) {
-        _shader.setUniformf("u_material.emission", m_emission);
+        _shader.setUniformf(m_uEmission, m_emission);
```
karimnaaji: @hjanetzek This is working properly for lights since there is a specific `LightUniform` which targets the shader it gets injected into. 

But now since `UniformLocation` now owns the location, if a material is injected to several styles (e.g. to different `ShaderProgram`) the location is now inconsistent.

Can you verify that?

hjanetzek: @karimnaaji yea right, I'll add MaterialUniforms analog to LightUniforms

karimnaaji: :+1:

karimnaaji:
Looks good to me!

PR #587, opened 2016-03-02
Add "miter_limit" draw parameter for "lines" styles

### LINKED ISSUE #500: Support 'miter_limit' parameter for lines styles
Associated feature in tangram: https://github.com/tangrams/tangram/pull/231

blair1618:
"miter_limit" and "outline:miter_limit" parameters are now parsed and applied to geometry built with `PolylineStyle`. Miter length was already being calculated in `Builders::buildPolyLine` so the changes are mostly simple. The one non-obvious change is that `buildPolyLine` can no longer take a size hint function. Now that extra triangles may be added to replace miter joins, it is no longer possible to predict the number of vertices and indices that will be used for a given line. The performance impact of this change should be negligible - the automatic growth of `std::vector` may actually be more efficient than hinting sizes. 

Resolves #500


blair1618:
@karimnaaji can you take a look? Thanks :)

karimnaaji:
Yep!

tallytalwar:
LGTM. @karimnaaji you can merge when you done reviewing this. Thanks

Issue #570, opened 2016-02-24
png24 icon transparent edges dirty


nvkelso:
When I use png8 icon with transparency have nice edges, but going to png24 ([example](https://github.com/tangrams/eraser-map/blob/gh-pages/images/poi_icons_18%402x.png)) they get dirty edges. I need to use png24 because my assets have both exterior transparency and internal transparency, to support Eraser Map app design. 

Opening the source asset up in Photoshop shows no artifacts, so I suspect this is a Tangram issue. This happens with all UI assets, and one map icon (traffic signals visible at max zoom only, look near the SF Mapzen office), which is used during routing to provide contrast with the blue route line.

![screenshot_2016-02-24-11-25-38](https://cloud.githubusercontent.com/assets/853051/13305281/094737f6-db0f-11e5-9c43-2fc93e09181c.png)

nvkelso:
(Green map pin over the light blue water edges have grey funk.)

blair1618:
I see the "dirtiness" you're talking about - it seems to be showing at edges where a true-color PNG would have partial transparency. I would guess that there's something undesirable happening in the interpolation between opaque and transparent pixels in the texture, I'll need to look a look at our shaders and blending.

hjanetzek:
The darker fringe could come from mixing in the black background pixels:
http://www.realtimerendering.com/blog/gpus-prefer-premultiplication/

hjanetzek:
Was looking if stb_image had a routing for applying premultiplication during loading. Not quite it but related: If we move the image loading to platform we could avoid back and forth premultiplication of *iphone-formatted PNGs*
https://github.com/nothings/stb/blob/master/stb_image.h#L388

blair1618:
Yeah, iOS at least has GLKit to make texture loading very convenient, including an option to premultiply alpha: https://developer.apple.com/library/ios/documentation/GLkit/Reference/GLKTextureLoader_ClassRef/index.html

matteblair:
I believe this has been resolved for a while now. If texture transparency still isn't behaving as expected then please open this or a new issue.

PR #588, opened 2016-03-02
[wip] Premultiply texture assets

hjanetzek:
#570

### DIFF
--- a/core/resources/shaders/point.fs
+++ b/core/resources/shaders/point.fs
@@ -48,16 +48,14 @@ void main(void) {
             float c2 = circle(uv, vec2(0.0), circleRadiusOut);
             color = vec4(vec3(c1) * v_color.rgb, c2 * v_alpha * v_color.a);
         #else
-            // premultiplied texture color
-            vec4 texColor = texture2D(u_tex, v_texcoords);
-            //color = vec4(texColor.rgb * v_color.rgb, texColor.a * v_color.a);
-
-            // Correct? v_color as multipler - to say how much of each texture channel is used?
-            color = texColor * v_color;
-            color *= v_alpha;
+            // assume premultiplied texture color
+            color = texture2D(u_tex, v_texcoords);
 
             // unpremultiply for blending(src_alpha,one_minus_src_alpha)
             color.rgb /= color.a;
+
+            color *= v_color;
+            color.a *= v_alpha;
         #endif
 
         #pragma tangram: color
--- a/core/resources/shaders/point.fs
+++ b/core/resources/shaders/point.fs
@@ -48,8 +48,16 @@ void main(void) {
             float c2 = circle(uv, vec2(0.0), circleRadiusOut);
             color = vec4(vec3(c1) * v_color.rgb, c2 * v_alpha * v_color.a);
         #else
+            // premultiplied texture color
             vec4 texColor = texture2D(u_tex, v_texcoords);
-            color = vec4(texColor.rgb * v_color.rgb, v_alpha * texColor.a * v_color.a);
+            //color = vec4(texColor.rgb * v_color.rgb, texColor.a * v_color.a);
+
+            // Correct? v_color as multipler - to say how much of each texture channel is used?
+            color = texColor * v_color;
+            color *= v_alpha;
+
+            // unpremultiply for blending(src_alpha,one_minus_src_alpha)
+            color.rgb /= color.a;
         #endif
 
         #pragma tangram: color
--- a/core/resources/scene.yaml
+++ b/core/resources/scene.yaml
@@ -272,16 +272,16 @@ layers:
                 draw: { icons: { sprite: restaurant } }
             cafe:
                 filter: { kind: [cafe, convenience] }
-                draw: { icons: { sprite: cafe } }
+                draw: { icons: { sprite: cafe, color: [0.5, 0.5, 1.0, 0.5] } }
             bar:
                 filter: { kind: [bar, pub] }
                 draw: { icons: { sprite: bar } }
             culture:
                 filter: { kind: [museum, library, church, place_of_worship, bank] }
-                draw: { icons: { sprite: museum } }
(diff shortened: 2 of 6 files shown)


hjanetzek:
Can we have a new label 'ready for testing'? :)

I think we should consider to use blendFunc(one, one_minus_src_alpha), now we premultiply the image on load (which can be done in advance and iOS seems to this by default for png assets) to get correctly interpolated pixels from texture2D fetches, then unpremultiply in the shader since blendFunc(src_alpha,...) does the alpha-multiplication for blending again.

blair1618:
How would you use a 'ready for testing' tag? I suppose I think of that the same as 'ready for review'.

Yeah, I think we should try a blend function that doesn't require us to undo the premultiplication in shaders. 

I'll give this a try on an iPhone and see if there is any trouble with the PNGs.

hjanetzek:
@blair1618 I mean something along the line of 'it's working, but the solution is not really optimal or ready for merging - anyone interested in this issue please try out and give feedback' :)

tallytalwar:
I guess "question" tag with an explicit mention in the PR description should suffice for such requests, instead of explicitly creating a tag (more like an rfc).

karimnaaji:
In my opinion changing for a lot of labels before merging would result in slowing down the review process. A simple comment requesting to test or a 'question' tag like here should be enough.

blair1618:
Yeah, comments can go far!

Though I did check out of curiosity and it seems like there's no limit to the length of label names :P

karimnaaji:
So we could do label comments!

core/resources/shaders/point.fs
```diff
@@ -48,8 +48,16 @@ void main(void) {
             float c2 = circle(uv, vec2(0.0), circleRadiusOut);
             color = vec4(vec3(c1) * v_color.rgb, c2 * v_alpha * v_color.a);
         #else
+            // premultiplied texture color
             vec4 texColor = texture2D(u_tex, v_texcoords);
-            color = vec4(texColor.rgb * v_color.rgb, v_alpha * texColor.a * v_color.a);
+            //color = vec4(texColor.rgb * v_color.rgb, texColor.a * v_color.a);
+
+            // Correct? v_color as multipler - to say how much of each texture channel is used?
+            color = texColor * v_color;
```
karimnaaji: I don't think this is correct, `v_color` has no concept of premultiplication so it shouldn't modify the premultiplied sampled color.
This should happen after `color.rgb /= color.a`.

core/resources/shaders/point.fs
```diff
@@ -48,8 +48,16 @@ void main(void) {
...
+            color *= v_alpha;
```
karimnaaji: Same for `v_alpha`, it shouldn't apply on the whole color, but only on `color.a` after `color.rgb /= color.a`.

What is correct to me is:

```            
color = texColor;
color.rgb /= color.a;
color *= v_color;
color.a *= v_alpha;
```

karimnaaji: @hjanetzek I just tested this it seems correct

```
draw: 
    icons:
        sprite: museum
        color: [1.0, 1.0, 1.0, 0.5]
```

Current:
screen shot 2016-03-03 at 10 11 46 am

Proposed:
screen shot 2016-03-03 at 10 12 32 am

tallytalwar: :+1:

blair1618: That sounds right for `blendFunc(SRC_ALPHA, ONE_MINUS_SRC_ALPHA)`. We could also change to `blendFunc(ONE, ONE_MINUS_SRC_ALPHA)` and return pre-multiplied color values from all of our fragment shaders, then there would be no need to divide out alpha.

karimnaaji: Polygons/polylines/text would be many more fragments than sprites though, so it's either the choice of dividing here for a small number of fragments or multiplying everywhere else for potentially a lot of fragments with a different blend mode.

blair1618: Any of the styles can sample from textures, sprites aren't the only place we would need to un-pre-multiply. 

There is a bigger problem to changing color representation though: User-defined shader code can interact with sampled color values. In the current pipeline, none of the color values available for a user to interact with are pre-multiplied, and changing that could certainly cause visual differences. We could do the necessary arithmetic on pre-defined color variables before injection points, but that won't help if a user calls a sampler in their own code. 

I do think that alpha-pre-multiplied textures are the "right" way to handle transparency, but I also want to make sure it won't disrupt existing shaders in styles or create a divergence with tangram-js.

karimnaaji: Yeah I guess what I mean is that transparency is not something that is usually used everywhere where a texture is used, mostly the case for POIs/icons. But for consistency we could change the blend mode.

I could see someone wanting to apply a Mapzen texture on all of building Manhattan, and he could get quite disappointed!
screen shot 2016-03-03 at 11 28 28 am

karimnaaji: @blair1618 what's in there this PR shouldn't be a divergence from JS, since the user can still interact with the color and this process is totally hidden for him so this seems to be the right way to handle this issue.

The material with transparency is an edge case that we could also fix in this PR, but I'm not sure how JS would behave using those.

hjanetzek:
I think we could later use an optimized shader for point.fs with blendFung(one, one_minus_src) when no custom shader blocks are defined and otherwise just add the unpremultiply step after texture2d() fetch.

karimnaaji:
@hjanetzek , can we extract https://github.com/tangrams/tangram-es/commit/a59bdd3f03cc640bc714a2896ddf9a3c729ecefb and make it a single PR, this can be useful and doesn't need to hold until the premultiplication-alpha discussion is solved.

blair1618:
https://github.com/tangrams/tangram-es/commit/ae703d29b28278cfbdb9f1d79fa34b25c23635fc is pretty orthogonal as well, and it could (should?) apply to OS X too.

Issue #539, opened 2016-02-16
Style "texcoords" parameter should determine whether texture coordinates are generated


blair1618:
Currently texture coordinates are simply added to all styles, which will not cause visual differences but is slightly inefficient and constitutes a deviation from the scene file specification.

https://github.com/tangrams/tangram-es/blob/master/core/src/scene/sceneLoader.cpp#L465

https://github.com/tangrams/tangram-docs/blob/gh-pages/pages/styles.md#texcoords

PR #589, opened 2016-03-02
Texture coordinates generation

karimnaaji:
#539

### DIFF
--- a/core/src/style/polygonStyle.cpp
+++ b/core/src/style/polygonStyle.cpp
@@ -22,20 +22,6 @@ constexpr float normal_scale = 127.0f;
 
 namespace Tangram {
 
-struct PolygonVertex {
-
-    PolygonVertex(glm::vec3 position, uint32_t order, glm::vec3 normal, glm::vec2 uv, GLuint abgr)
-        : pos(glm::i16vec4{ glm::round(position * position_scale), order }),
-          norm(normal * normal_scale),
-          texcoord(uv * texture_scale),
-          abgr(abgr) {}
-
-    glm::i16vec4 pos; // pos.w contains layer (params.order)
-    glm::i8vec3 norm;
-    uint8_t padding = 0;
-    glm::u16vec2 texcoord;
-    GLuint abgr;
-};
 
 struct PolygonVertexNoUVs {
 
@@ -50,6 +36,14 @@ struct PolygonVertexNoUVs {
     GLuint abgr;
 };
 
+struct PolygonVertex : PolygonVertexNoUVs {
+
+    PolygonVertex(glm::vec3 position, uint32_t order, glm::vec3 normal, glm::vec2 uv, GLuint abgr)
+        : PolygonVertexNoUVs(position, order, normal, uv, abgr), texcoord(uv * texture_scale) {}
+
+    glm::u16vec2 texcoord;
+};
+
 PolygonStyle::PolygonStyle(std::string _name, Blending _blendMode, GLenum _drawMode)
     : Style(_name, _blendMode, _drawMode)
 {
@@ -62,8 +56,8 @@ void PolygonStyle::constructVertexLayout() {
         m_vertexLayout = std::shared_ptr<VertexLayout>(new VertexLayout({
             {"a_position", 4, GL_SHORT, false, 0},
             {"a_normal", 4, GL_BYTE, true, 0}, // The 4th byte is for padding
-            {"a_texcoord", 2, GL_UNSIGNED_SHORT, true, 0},
             {"a_color", 4, GL_UNSIGNED_BYTE, true, 0},
+            {"a_texcoord", 2, GL_UNSIGNED_SHORT, true, 0},
         }));
 
         m_defines += "#define TANGRAM_USE_TEX_COORDS\n";
--- a/core/src/style/polygonStyle.cpp
+++ b/core/src/style/polygonStyle.cpp
@@ -46,9 +46,7 @@ struct PolygonVertex : PolygonVertexNoUVs {
 
 PolygonStyle::PolygonStyle(std::string _name, Blending _blendMode, GLenum _drawMode)
     : Style(_name, _blendMode, _drawMode)
-{
-    m_defines = "";
-}
+{}
 
 void PolygonStyle::constructVertexLayout() {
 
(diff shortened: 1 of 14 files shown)


karimnaaji:
@bcamper , trying to figure out what is the default for this, I'm gonna update the docs since it seems like we're missing some update on the default of this option.

Looking at the JS [code](https://github.com/tangrams/tangram/blob/16ea7a4cd0ec007f23987a71bc9e53a2f79e0975/src/material.js#L79) it looks like texture uvs generation default is `false` and is set to `true` when either: 
-  `texcoords` is explicitly set to `true`  on the style
- any material associated to the style with mapping `uv` is defined

Does this sound correct?

This reduces the buffer at some zoom levels by sometimes up to 2 megabytes.

screen shot 2016-03-02 at 2 14 04 pm

bcamper:
Yep that's right, for polygons and lines. For points, texcoords are always
generated.

karimnaaji:
@tallytalwar , can you review please. Thanks

tallytalwar:
Will check this today. Thanks

hjanetzek:
@karimnaaji have you considered to move the specialization to a class that holds MeshData?
Templating the builder adds an implementation which share most of the code for each vertex type.

```
struct PolygonMeshBase {
    struct {
        uint32_t order = 0;
        uint32_t color = 0xff00ffff;
    } attributes;
   PolygonVertexFn vertexFunc;
   std::vector<uint16_t> indices();
   std::unique_ptr<StyledMesh> build();
};
struct PolygonMeshSimple : PolygonMeshBase {
    static VertexLayout vertexLayout() {
        return {{{"a_position", 4, GL_SHORT, false, 0},
                 {"a_color", 4, GL_UNSIGNED_BYTE, true, 0}}};
    }
    struct Vertex {
        Vertex(glm::vec3 position, uint32_t order, GLuint abgr)
            : pos(glm::i16vec4{ glm::round(position * position_scale), order }),
              abgr(abgr) {}
        glm::i16vec4 pos;
        GLuint abgr;
    };
    MeshData<Vertex> data;

    PolygonMeshSimple() {
        vertexFunc = [this](const glm::vec3& coord, const glm::vec3& normal, const glm::vec2& uv) {
            data.vertices.push_back({ coord, attributes.order, attributes.color });
        };
    }
```

karimnaaji:
@hjanetzek , that could be solution! Trying this in a few.

tallytalwar:
@karimnaaji did you update this for review?

karimnaaji:
Not yet. Hold on this.

karimnaaji:
@tallytalwar , you can review this PR and also checkout the code for the other proposed solution in https://github.com/tangrams/tangram-es/commit/9425e5951c4e9e5776b0273ce3eb045c7dae539b (only polygon style for now)  ; it gets quite verbose unless we template the new `PolygonMeshBase`, so I would rather stick with this solution instead.

core/src/style/polygonStyle.cpp
```diff
@@ -22,33 +22,52 @@ constexpr float normal_scale = 127.0f;
...
+            {"a_normal", 4, GL_BYTE, true, 0}, // The 4th byte is for padding
+            {"a_color", 4, GL_UNSIGNED_BYTE, true, 0},
+            {"a_texcoord", 2, GL_UNSIGNED_SHORT, true, 0},
+        }));
+
+        m_defines += "#define TANGRAM_USE_TEX_COORDS\n";
+    } else {
+        m_vertexLayout = std::shared_ptr<VertexLayout>(new VertexLayout({
+            {"a_position", 4, GL_SHORT, false, 0},
+            {"a_normal", 4, GL_BYTE, true, 0},
```
tallytalwar: If we swap `a_normal` and `a_color`, we can probably get away with padding for no texcoord vertexLayout.
Or will the GPU representation still needs even bytes?

core/src/style/polygonStyle.cpp
```diff
@@ -58,16 +77,14 @@ void PolygonStyle::constructShaderProgram() {
     std::string fragShaderSrcStr = stringFromFile("shaders/polygon.fs", PathType::internal);
 
     m_shaderProgram->setSourceStrings(fragShaderSrcStr, vertShaderSrcStr);
+
+    m_shaderProgram->addSourceBlock("defines", m_defines);
```
tallytalwar: Instead of maintaining member for `defines`, we can check for `m_texCoordsGeneration` and explicitly add the define block: `#define TANGRAM_USE_TEX_COORDS\n`

core/src/style/polygonStyle.cpp
```diff
@@ -58,16 +77,14 @@ void PolygonStyle::constructShaderProgram() {
...
+template <class V>
 struct PolygonStyleBuilder : public StyleBuilder {
 
-    const PolygonStyle& m_style;
-
-    MeshData<PolygonVertex> m_meshData;
-
-    float m_tileUnitsPerMeter;
-    int m_zoom;
+public:
```
tallytalwar: Not needed, by default struct has public scope.

karimnaaji: I usually prefer this to be explicit even for structs.

core/src/style/polygonStyle.cpp
karimnaaji: :+1:

core/src/style/polygonStyle.cpp
karimnaaji: Yeah it should still be aligned on 4 bytes, especially since it's interleaved.

tallytalwar: :+1:

core/src/style/polylineStyle.cpp
```diff
@@ -112,29 +128,42 @@ struct PolylineStyleBuilder : public StyleBuilder {
 
     void addMesh(const Line& _line, const Parameters& _params);
 
-    void buildLine(const Line& _line, const Parameters::Attributes& _att,
-                   MeshData<PolylineVertex>& _mesh);
+    void buildLine(const Line& _line, const typename Parameters::Attributes& _att,
```
tallytalwar: I do not get the `typename` definition here!

tallytalwar:
I agree with @karimnaaji, that its a much cleaner/readable code by not templating PolygonMesh. 

@hjanetzek We can continue with your commit (7c4e935) as a later refactor of this work.

core/src/style/polylineStyle.cpp
tallytalwar: To answer the above: Because `Paramters::Attribute` is in a dependent scope on the template class. (refer: https://rmf.io/cxx11/dependent-names-bliss/)

PR #590, opened 2016-03-02
Use latest earcut.hpp

hjanetzek:
Quite a few fixes went into earcut since our last checkout. Unfortunately one has now to handle which vertices are actually used oneself.
 
+ fixes some wrong triangulations
+ could be faster as well

Check Papua New Guinea on low zooms as an example

### DIFF
--- a/core/src/util/builders.cpp
+++ b/core/src/util/builders.cpp
@@ -45,6 +45,7 @@ void Builders::buildPolygon(const Polygon& _polygon, float _height, PolygonBuild
         }
     }
 
+    // Run earcut, triangles are stored in _ctx.earcut.indices
     _ctx.earcut(_polygon);
 
     size_t sumPoints = 0;
@@ -52,6 +53,7 @@ void Builders::buildPolygon(const Polygon& _polygon, float _height, PolygonBuild
         sumPoints += line.size();
     }
 
+    // Mark the points that are referenced by indices as used.
     size_t sumVertices = 0;
     _ctx.used.assign(sumPoints, 0);
     for (auto i : _ctx.earcut.indices) {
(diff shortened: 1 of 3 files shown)


hjanetzek:
shot-2016-03-04_21-07-09

drawing water with 0.5 alpha and overlay blending

blair1618:
Whoa, nice example. Those water polygons must get pretty gnarly at the low zooms. 

I'll make this the next thing I review, since there seems to be a potential for nasty rendering bugs.

core/src/util/builders.cpp
```diff
@@ -59,22 +45,72 @@ void Builders::buildPolygon(const Polygon& _polygon, float _height, PolygonBuild
...
+
+    size_t sumPoints = 0;
+    for (auto& line : _polygon) {
+        sumPoints += line.size();
+    }
 
-    _ctx.numVertices += _ctx.earcut.vertices.size();
+    size_t sumVertices = 0;
+    _ctx.used.assign(sumPoints, 0);
+    for (auto i : _ctx.earcut.indices) {
```
blair1618: It seems like we should only need to iterate over the points and indices once each, possibly simultaneously. The `sizeHint` would require an initial counting pass, but we're not currently using it and we should really just remove it - it's likely to cause more allocations than necessary.

hjanetzek: Not sure if it's possible to do in less steps:
- determine which points are references by indices
- create vertices for the unique set of used points, and update the indices (shifted by the number of dropped points)
- add the updated indices

blair1618: Ah ok, I'm beginning to understand the situation. Would've been quicker if there were comments ;)

Do we really need to determine which points are referenced by indices? The only situation I can imagine that would leave points unused is if a polygon has duplicated points, which should be rare enough that it's almost certainly faster to just add them to buffers anyway and not worry about it.

core/src/util/builders.cpp
```diff
@@ -59,22 +45,72 @@ void Builders::buildPolygon(const Polygon& _polygon, float _height, PolygonBuild
...
+
+    for (size_t src = 0; src < _ctx.used.size(); src++) {
+
+        if (src - offset >= _polygon[ring].size()) {
+            offset += _polygon[ring].size();
+            ring += 1;
+        }
+        const Line& points = _polygon[ring];
+
+        if (_ctx.used[src] > 0) {
```
tallytalwar: Can you elaborate on this?

core/src/util/builders.cpp
hjanetzek: I've ported the dropDuplicateVertices method that was previously done in earcut. It's at least the start or endpoint that are duplicate in our polygons (one of them is unused depending on the winding)

blair1618: Ok I see. I'm fine with this implementation but it does need some comments describing the procedure - I for one will certainly forget how this works by the next time we need to look at it.

core/src/util/builders.cpp
tallytalwar: K, got it, but yes I agree with @blair1618, it needs some comments/documentation love.

hjanetzek: In the loop through indices above 'used' is set to 1 for any point that is used. When looping though the points only the points marked as used are added as vertex.  Then the 'used' entry is updated to store the mapping from the old to the new vertexID ```_ctx.used[src] = dst++;```

tallytalwar: Yup, figured with some dry run :D.

I think only documenting what `used` is, or maybe giving a better name, 'vertexDupTracker`?? will explain what this block is doing. Also maybe `modified` is also not an apt variable name :(.

hjanetzek:
Found another wrong triangulation both in earcut master and tangram-2 branch. Will look into it tomorrow. Tile 28/16/5

tallytalwar:
Can you revert the scene file changes?
Once done good to merge.

tallytalwar:
:+1:

PR #591, opened 2016-03-03
Fix recursive filters within a "not" filter

blair1618:
The previous code assumed that "not" could only contain a predicate.

### DIFF
--- a/core/src/scene/sceneLoader.cpp
+++ b/core/src/scene/sceneLoader.cpp
@@ -876,10 +876,7 @@ Filter SceneLoader::generateNoneFilter(Node _filter, Scene& scene) {
             filters.emplace_back(generateFilter(filt, scene));
         }
     } else if (_filter.IsMap()) { // not case
-        for (const auto& filt : _filter) {
-            std::string keyFilter = filt.first.as<std::string>();
-            filters.emplace_back(generatePredicate(_filter[keyFilter], keyFilter));
-        }
+        filters.emplace_back(generateFilter(_filter, scene));
     } else {
         LOGW("Invalid filter. 'None' expects a list or an object.");
         return Filter();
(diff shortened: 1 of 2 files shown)


hjanetzek:
Could you add a test-case for this? It's unlikely that one would use it but what if one tests ```not: function(){...}``` ? Though.. maybe it's even useful when using having function as ```achors```. Which reminds me that the set of function strings could be made unique to compile them only once.

blair1618:
Yeah, a test for this is probably a good idea. 

I believe `not: 'function() { ... }'` is a valid use of `not`, and this change enables it now.

tallytalwar:
LGTM. Once a test is added, good to merge.

core/src/scene/sceneLoader.cpp
```diff
@@ -876,10 +876,7 @@ Filter SceneLoader::generateNoneFilter(Node _filter, Scene& scene) {
...
         }
     } else if (_filter.IsMap()) { // not case
-        for (const auto& filt : _filter) {
-            std::string keyFilter = filt.first.as<std::string>();
-            filters.emplace_back(generatePredicate(_filter[keyFilter], keyFilter));
-        }
+        filters.emplace_back(generateFilter(_filter, scene));
     } else {
         LOGW("Invalid filter. 'None' expects a list or an object.");
         return Filter();
```
hjanetzek: @blair1618 what if the filter is a scalar? That case doesn't seem to  handled

blair1618: Ah yeah, whoops. That's exactly the thing I was talking about in an earlier comment haha. It should be a simple matter of replacing `_filter.IsMap()` with `_filter.IsMap() || _filter.IsScalar()`, then `generateFilter` will take care of detecting a function in a scalar. I'll push that fix to master if nobody is opposed.

hjanetzek: :+1:

PR #593, opened 2016-03-03
Add Javadoc generation

blair1618:
This was basically already present in a gradle script, mostly just needed some syntax revisions in our javadoc. 

This isn't a final revision on documentation, as the interfaces may still change before the SDK release.

### DIFF
--- a/android/tangram/gradle-mvn-push.gradle
+++ b/android/tangram/gradle-mvn-push.gradle
@@ -99,21 +99,20 @@ afterEvaluate { project ->
     sign configurations.archives
   }
 
-  //  TODO: Fix javadocs and source jars for release deployment.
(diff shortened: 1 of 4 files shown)


blair1618:
@hjanetzek take a look when you have a second. Thanks!

tallytalwar:
This looks good to me. Of course we need to scan through all documentation to finalize everything for the release.

karimnaaji:
Looks good to me too

hjanetzek:
Looks good! 

Just noticed one thing: The Renderer implementation should be an inner class so its interface is not exposed in MapController

blair1618:
@hjanetzek Good call. I'll take a look at that separately because I want to make sure it can still be usefully extended (e.g. to create a stereoscopic for a cardboard app).

PR #596, opened 2016-03-04
More timing infos in debug view

hjanetzek:


blair1618:
Cool, looks useful. 

In the near future, it would be nice to put all of our debug info code somewhere where it wouldn't clutter up the main update and render loops, but we don't need to do that in this PR.

PR #597, opened 2016-03-04
Fix point light ground attachment on tilted view

karimnaaji:
Fix point light attachment when `ground` option is on for tilted view, now relative to _eye_ position:

This branch,
relativetoview
Master,
nonrelativetoeye

### DIFF
--- a/core/src/scene/pointLight.cpp
+++ b/core/src/scene/pointLight.cpp
@@ -67,8 +67,8 @@ void PointLight::setupProgram(const View& _view, LightUniforms& _uniforms) {
         position.z = position.z - _view.getEye().z;
 
     } else if (m_origin == LightOrigin::ground) {
-        // Leave light's xy in camera space, but z needs to be moved relative to ground plane
-        position.z = position.z - _view.getPosition().z;
+        // Move light position relative to the eye position in world space
+        position -= glm::vec4(_view.getEye(), 0.0);
     }
 
     if (m_origin == LightOrigin::world || m_origin == LightOrigin::ground) {


blair1618:
Ha! This is kind of interesting :D

In master right now, a light with `origin: ground` will have it's x and y coordinates applied relative to the camera's location and it's z coordinate applied relative to the ground. So the current implementation *is* satisfying the documented behavior, see the `origin` section here: 
https://github.com/tangrams/tangram-docs/blob/gh-pages/pages/Lights-Overview.md

And the reference section here: 
https://github.com/tangrams/tangram-docs/blob/gh-pages/pages/lights.md#origin

That said, the current behavior is not very useful. I suggest that we amend the documentation to clarify that `origin: ground` will apply a light's position relative to *the point on the ground that is in the center of the current view*. This will make no semantic difference to tangram-js right now, but it would make this proposed change the correct behavior when tilt is applied. 

cc @bcamper

karimnaaji:
>will have it's x and y coordinates applied relative to the camera's location and it's z coordinate applied relative to the ground

`_view.getPosition()` is not updated, only `m_eye` gets rotated and tilted, which makes the world rotating around it, so this use the rotated/tilted eye position. Reading at the code it's the expected behavior relative to the documentation, but the value returned by `_view.getPosition()` is not what's expected in this case.

tallytalwar:
@karimnaaji view was being updated and eye was stationary before! This change (eye updates on tilting and view remains stationary) was put in place to better reflect the gesture performed also makes more sense in general.

karimnaaji:
Sure, I understand the need of it. In this case though it seems to make more sense to me to make it the light position relative to the real view positon (the eye).

tallytalwar:
I agree to this change.

core/src/view/view.h
```diff
@@ -107,6 +107,9 @@ class View {
     /* Gets the normal matrix; transforms surface normals from model space to camera space */
     const glm::mat3& getNormalMatrix() const { return m_normalMatrix; }
 
+    /* Returns the eye position in world space */
+    const glm::vec3& getEye() const { return m_eye; }
```
tallytalwar: nitpick. Can you call this just `eye` and not `getEye`?

karimnaaji: Being consistent with the rest of the view module, should we change everything in here then?

blair1618:
Yeah, bit of subtlety here: The position of the *view* is not exactly the position of the *camera*. ;)

I agree that `origin: ground` should be relative to the *view*, I just want to make the documentation reflect that - since there isn't yet a distinction there between these two things.

core/src/view/view.h
tallytalwar: I will be fine with the change.

cc. @blair1618, @hjanetzek

blair1618: I don't think there's a net gain to changing all usages of `View` getters right now. `View::pixelScale` and `View::pixelsPerMeter` already use the abbreviated form, so I have no strong opinion on `eye`/`getEye`.

blair1618:
Now that we've corrected the meaning of "correct", this is good to merge :)

Issue #578, opened 2016-02-29
Vertex lighting with world origin for isometric projection in tilted view is wrong


hjanetzek:
Flat areas change their lighting relative to tilt angle

blair1618:
Confirmed - looks like a bug. 

Seems like the factor that causes this isn't actually the isometric projection - it's the "oblique axis" value. An isometric camera with `axis: [0, 0]` results in correct lighting.

blair1618:
This chunk seems suspect to me now: https://github.com/tangrams/tangram-es/blob/master/core/src/view/view.cpp#L340-L343

I think that comment used to be true, but probably isn't in every case now.

PR #602, opened 2016-03-04
Fix lighting when orthographic projection is used

karimnaaji:
Related https://github.com/tangrams/tangram-es/issues/578

Since orthographic projection is a uniform scale transform, the normal matrix should be the identity matrix.

### DIFF
--- a/core/src/view/view.cpp
+++ b/core/src/view/view.cpp
@@ -282,9 +282,17 @@ void View::updateMatrices() {
     }
 
     if (m_type == CameraType::isometric) {
-        // Add the oblique projection scaling factors to the view matrix
-        m_view[2][0] += m_obliqueAxis.x;
-        m_view[2][1] += m_obliqueAxis.y;
+        glm::mat4 shear = m_view;
+
+        // Add the oblique projection scaling factors to the shear matrix
+        shear[2][0] += m_obliqueAxis.x;
+        shear[2][1] += m_obliqueAxis.y;
+
+        // Remove the view from the shear matrix so we don't apply it two times
+        shear *= glm::inverse(m_view);
+
+        // Inject the shear in the projection matrix
+        m_proj *= shear;
     }
 
     m_viewProj = m_proj * m_view;


tallytalwar:
Great. Had a suspicion on this when this was brought early this week.
Nice.

blair1618:
This does *look* correct, but I have questions about the math here. By the reasoning here: http://www.lighthouse3d.com/tutorials/glsl-12-tutorial/the-normal-matrix/, which seems sound to me, we should be able to calculate the normal matrix with:
```
m_normalMatrix = glm::transpose(glm::inverse(glm::mat3(m_view)));
```
The derivation makes no assumptions about the type of transformation used in the view matrix and has no dependence on the projection matrix, so we should be able to apply it to both `perspective` and `isometric` cameras. However, using this formula for the normal matrix produces the same behavior as the current code! 

@karimnaaji can you explain how you arrived at this change? Why does the projection matrix matter to the normal matrix?

karimnaaji:
It is correct for isometric projection to use the identity matrix since this kind of transformation doesn't apply on _scales_ (when later applied on the shader combined), which is mainly why a normal matrix is needed. So it looks correct, and is also mathematically correct. Does this make sense?

blair1618:
My claim is that the projection shouldn't impact the normal matrix at all. Lighting calculations are performed in view space, before any projection matrix is applied (even when fragment lighting is used, the inputs are varyings representing view space vectors). 

I'm not trying to make an academic paper out of our lighting system, but I do want to make sure that we're not just hiding another problem somewhere else ;)

karimnaaji:
Yep, you're right; this matrix doesn't modify `v_position` which is passed as is so the matrix used isn't combined, investigating more on this.

karimnaaji:
@blair1618 , Looking at the lightning calculation I think it's related to the other fixes on the light-ground/world-attachment. Eye to point should use the eye position instead of `_view.getPosition`, this https://github.com/tangrams/tangram-es/commit/057b79b918ce1379acdaf82a3420985e7c209fcc seems to give the expected behavior. Will update that branch with the changes.

core/src/view/view.cpp
```diff
@@ -281,21 +281,18 @@ void View::updateMatrices() {
             break;
     }
 
+    m_viewProj = m_proj * m_view;
+    m_invViewProj = glm::inverse(m_viewProj);
+
+    m_normalMatrix = glm::transpose(glm::inverse(glm::mat3(m_view)));
```
tallytalwar: Did we figure this out why oblique axis modification to `m_view` should not be considered when calculating the normal matrix? As per what I was seeing, normals were calculated correctly with axis contributions!

karimnaaji: Yep, normals are good with the axis contribution, something else is wrong in the pipeline.

karimnaaji:
@blair1618 , @tallytalwar , figured out that we actually need to separate the skewing transform in another matrix, please check the changes on this branch once you have time; but the contribution of the skew is now taken into account for the normal matrix, which fixes directional lights. Light position for point lights is also fixed, by separating those and using a view matrix with no skew in it to transform the point light position.

There is one commit to be dropped used for testing.

Result with skew, point light on ground & world + directional light:

screen shot 2016-03-09 at 4 00 10 pm

tallytalwar:
yup, in that case (if this is giving us the correct lighting) or skew operation was wrong before.

blair1618:
Interesting...

So this is now a somewhat different interpretation of the "oblique axis" than what we currently have in production. Currently, `axis: [0, 1]` skews extruded geometry towards the top of the screen. In this model, that axis would skew geometry North, regardless of how the map is oriented on the screen. 

This would substantially alter the appearance of styles using  an "isometric" camera (including eraser-map) in a way that is ... probably not desirable.

blair1618:
Seems like we're pretty much back to where we started here haha

We haven't yet solved the problem we set out to address with this, namely that lighting behaves in confusing ways when an isometric camera is applied (e.g. diffuse lighting changes when the camera moves). 

I think we might have been approaching this the wrong way - I think the shear represented by the camera's "axis" should rightly be applied as part of the *projection matrix*, rather than the view matrix. The trick is just figuring out how to correctly translate the axis values into a suitable transform to compose with the orthographic projection.

karimnaaji:
@blair1618 , please testing again, the issue being addressed is actually solved; I accidentally broke it while rebasing, please check out again with the test scene file.
We can revisit this by injecting the shear in the projection matrix, that would make more sense, but could get more complex too.

blair1618:
I'm still seeing the same issues :\

I do think that the projection matrix is what we need to be working with. Try plugging in:
```c++
    if (m_type == CameraType::isometric) {
        // Add the oblique projection scaling factors to the view matrix
        glm::mat4 shear = m_view;
        shear[2][0] += m_obliqueAxis.x;
        shear[2][1] += m_obliqueAxis.y;
        shear *= glm::inverse(m_view);
        m_proj *= shear;
    }
```
It's quite rough, but gives "correct" results as far as I can tell.

blair1618:
I must've been looking at an out-of-date commit, lighting looks pretty much correct now!

... but, I have another question now: When lighting calculations are performed in shaders, vertices have been transformed into "view space" using `m_viewComposed`, but lights themselves have been transformed into "view space" using `m_view`. Now, we could apply the sheared `m_viewComposed` to light positions as well, but having this additional quasi-view-matrix just seems problematic in general. 

Two alternatives:

 - Generate the normal matrix from the un-sheared view matrix and then apply the shear directly to the view matrix. This is probably the most computationally efficient approach. 

 - Make the shearing operation part of the projection matrix (as in my snippet above). This has the nice property that lighting is performed exactly the same way for both perspective and isometric cameras. 

(Hypothetically, the shear could also be a distinct transformation that's applied after the view matrix and before the projection matrix, but since we never need the intermediate result it seems most reasonable to combine it with the projection matrix in this case)

tallytalwar:
In the first approach you mentioned, normal matrix will not have any shear component to it, which will result in bad normals?

karimnaaji:
>Generate the normal matrix from the un-sheared view matrix and then apply the shear directly to the view matrix. This is probably the most computationally efficient approach.

I think this is what's happening in the branch right now, the normal matrix is computed based on the un-sheared view. 

The second approach makes sens too, though it has the cost of computing the inverse of the view it makes the steps more consistent. I'm good with either approach since they both solve the issue.

blair1618:
@tallytalwar yes, the normals won't be influenced by the shear and that's exactly how we want it. 

Consider the sides of the buildings in this screenshot, using the first-bullet-point approach, taken at 0 pitch, with `color.rgb = normal.xyz`

screen shot 2016-03-10 at 5 28 44 pm

It might seem strange that the sides of the buildings have no blue component. We can see those faces, so they must have a component facing the camera! Well, actually those faces are perpendicular to the camera direction (remember, the view is straight down), we're just taking an odd projection of it. If the normals had shear applied to them, that would imply that the geometry of the building itself is sheared in world space. 

@karimnaaji Yeah the first bullet is almost exactly the same as what you've done here, just without adding any new matrices

karimnaaji:
So I think we're good moving for the second approach which makes sense to everyone! Will update this branch in a few.

blair1618:
Okay one more question...

Just kidding :) We're good here, awesome work all around.

PR #604, opened 2016-03-05
Outline should not be drawn when no color

tallytalwar:
Before this, if no color was specified for an outline, it was rendering black (default color value in PolylineStyle
Builder is <uint32_t>(1).

However since color is a required parameter it should not draw outline, when no color is specified.
This is also the behavior with tangram-js.


tallytalwar:
We should also clarify this in the documentation, which only mention: **`outline` elements can take any `lines` style parameters"**.
https://github.com/tangrams/tangram-docs/blob/gh-pages/pages/draw.md#outline

blair1618:
Looks good. Mergin'

PR #605, opened 2016-03-05
Outline styles

### LINKED ISSUE #544: Polyline "outline" parameters should be able to override "style" of enclosing rule
https://github.com/tangrams/tangram-docs/blob/gh-pages/pages/draw.md#outline (an outline will respect *any* parameters that a line style will, including "style")

This is a bit tricky with the way we've implemented outlines in ES. I think for outlines that specify a different style than their enclosing rule, we'll have to generate a new draw rule for the outline that uses the outline's style, making sure that the outline still has the correct order offset from the original line rule.

tallytalwar:
Current approach

- Create a dummy rule for outlines which define their own drawing style
- Use this dummy rule to draw these outlines
- original draw rule will skip drawing this "outline"
- and dummy draw rule will skip drawing the "line"

Current implementation keeps the current architecture intact. 
However, things can be further improved by consolidating `polylinestyle` to only build "lines", and pass on lines and outlines drawing individually to the style builder. However this will require some architectural changes (specifically related to dependency of outline/stroke paramaters on parent line parameters, example: width), which I would like to discuss first.

Note: This is rebased off: https://github.com/tangrams/tangram-es/tree/fix-no-outline-color, request to lookat/merge #604 first.

Fixes #544

### DIFF
--- a/core/src/scene/drawRule.cpp
+++ b/core/src/scene/drawRule.cpp
@@ -48,19 +48,6 @@ DrawRule::DrawRule(const DrawRuleData& _ruleData, const SceneLayer& _layer) :
     }
 }
 
-DrawRule::DrawRule(const DrawRule& _other) : active(_other.active), name(_other.name),
-    id(_other.id), dummyOutline(_other.dummyOutline) {
-
-    for (size_t i = 0; i < StyleParamKeySize; ++i) {
-
-        if (!_other.active[i]) {
-            continue;
-        }
-
-        params[i] = _other.params[i];
-    }
-}
-
 void DrawRule::merge(const DrawRuleData& _ruleData, const SceneLayer& _layer) {
 
     evalConflict(*this, _ruleData, _layer);
@@ -122,12 +109,6 @@ size_t DrawRule::getParamSetHash() const {
         if (active[i]) { hash_combine(seed, params[i].name); }
     }
     return seed;
-}
-
-DrawRule DrawRule::outlineRule() const {
-    DrawRule outlineRule(*this);
-    outlineRule.dummyOutline = true;
-    return outlineRule;
 }
 
 void DrawRule::logGetError(StyleParamKey _expectedKey, const StyleParam& _param) const {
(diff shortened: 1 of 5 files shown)


blair1618:
Interesting, this is an approach I hadn't considered!

I do feel like this is putting off an inevitable re-write of `PolyLineStyle`, but if this quicker approach works well we may be able to live with it until we have the time for that larger change. 

Reviewing now. I'll let you resolve the merge conflict here @tallytalwar

tallytalwar:
Sure will take care of rebasing with master.

Yup, as mentioned in the description, with the current architecture we have, it was kinda convoluted to address this change by letting PolylineStyle draw only lines and then calling it twice for lines and strokes explicitly.

tallytalwar:
Rebased!!

core/src/scene/drawRule.cpp
```diff
@@ -48,6 +48,19 @@ DrawRule::DrawRule(const DrawRuleData& _ruleData, const SceneLayer& _layer) :
     }
 }
 
+DrawRule::DrawRule(const DrawRule& _other) : active(_other.active), name(_other.name),
```
blair1618: I believe we can use the default copy constructor. I see that you took care to only copy the "active" parameters, but we can safely copy all of the parameters (which are just pointers) since the "active" bits are copied too.

core/src/scene/drawRule.cpp
```diff
@@ -203,6 +222,20 @@ void DrawRuleMergeSet::apply(const Feature& _feature, const SceneLayer& _layer,
...
+
+            // build outline explicitly with outline style
+            const auto& outlineStyleName = rule.findParameter(StyleParamKey::outline_style);
+            if (outlineStyleName) {
+                auto& styleName = outlineStyleName.value.get<std::string>();
+                auto* outlineStyle = _builder.getStyleBuilder(styleName);
+                if (!outlineStyle) {
+                    LOGE("Invalid style %s", styleName.c_str());
+                } else {
+                    outlineStyle->addFeature(_feature, rule.outlineRule());
```
blair1618: I'm not sure we need to create a copy of the original rule here, could we just do something like:
```
rule.isOutlineOnly = true;
outlineStyle->addFeature(_feature, rule);
rule.isOutlineOnly = false;
```

tallytalwar: Yes thats a good suggestion. Will something like that.

tallytalwar: Works! Updated.

blair1618:
This is looking good to me now. It's kind of unfortunate to have another flag on `DrawRule` just for this case, but it gets the job done and it we can always revise later! Feel free to merge after rebasing/dropping test commits.

tallytalwar:
Yes I also do not like explicit bool living in drawrules. But it's much convinient at this point than a bigger refactor. Will definitely revisit this later.

PR #606, opened 2016-03-07
update variant submodule

hjanetzek:


hjanetzek:
Should check again with a newer version later

PR #609, opened 2016-03-08
Parse text_source fallbacks

### LINKED ISSUE #607: `text_source` param should account for language fallbacks
Language preference should be able to be defined as a serie of language fallbacks from the text style param `text_source`.

JS reference:
- https://github.com/tangrams/tangram/pull/235

Docs:
- https://github.com/tangrams/tangram-docs/blob/621fecb203c19b91514d544c58330527013cae61/pages/draw.md#text_source

karimnaaji:
Fixes #607

### DIFF
--- a/core/src/style/textStyle.cpp
+++ b/core/src/style/textStyle.cpp
@@ -112,19 +112,23 @@ bool TextStyleBuilder::checkRule(const DrawRule& _rule) const {
 
 std::string TextStyleBuilder::resolveTextSource(const std::string& textSource, const Properties& props) const {
 
-    std::vector<std::string> fallbacks;
-    std::stringstream ss(textSource);
-    std::string item;
+    std::string tmp, item;
 
-    // Parse fallbacks
-    while (std::getline(ss, item, ',')) {
-        fallbacks.push_back(item);
+    // Meaning we have a yaml sequence defining fallbacks
+    if (textSource.find(',') != std::string::npos) {
+        std::stringstream ss(textSource);
+
+        // Parse fallbacks
+        while (std::getline(ss, tmp, ',')) {
+            if (props.getAsString(tmp, item)) {
+                return item;
+            }
+        }
     }
 
-    for (const auto& fallback : fallbacks) {
-        if (props.getString(fallback, item)) {
-            return item;
-        }
+    // Fallback to default text source
+    if (props.getAsString(textSource, item)) {
+        return item;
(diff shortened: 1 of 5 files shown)


tallytalwar:
LGTM.

tallytalwar:
@karimnaaji you want to clean the scene.yaml (some commented code, etc), once done will merge.

blair1618:
Has anyone profiled this change with a label-heavy style like eraser-map or cinnabar? I'm a little concerned about the speed of stringstream operations and vector allocation.

tallytalwar:
I do not think there will be anything noticeable big, but profiling is a good call, to confirm.

karimnaaji:
@tallytalwar , I'm good with letting more test cases in the stylesheet/commented blocks that makes it easy to test.
Anyhow we should take some time writing a better default stylesheet, I will create an issue for that.

karimnaaji:
@blair1618 , good call, I will perform some profiling on that.

tallytalwar:
A quick profiling with eraser-map style sheet on master and this branch I get very similar number for `TextStyleBuilder::applyRule` on both branches, about **0.1% (15-20ms)**

karimnaaji:
@tallytalwar , eraser-map stylesheet doesn't use that much this option https://github.com/tangrams/eraser-map/search?utf8=%E2%9C%93&q=text_source. It should be profiled with eraser-map but adding `text_source: [name:en, name, area...]` to more labels.

core/src/style/textStyle.cpp
```diff
@@ -98,14 +102,36 @@ struct TextStyleBuilder : public StyleBuilder {
...
 
+std::string TextStyleBuilder::resolveTextSource(const std::string& textSource, const Properties& props) const {
+
+    std::vector<std::string> fallbacks;
+    std::stringstream ss(textSource);
+    std::string item;
+
+    // Parse fallbacks
+    while (std::getline(ss, item, ',')) {
+        fallbacks.push_back(item);
```
blair1618: Can we just check and return the property here? Do we need that `fallbacks` vector?

karimnaaji: Updated

karimnaaji:
Profiling didn't show any difference even if using `text_source` option with fallbacks for all eraser-map labels.

blair1618:
A+

Merge when ready!

PR #611, opened 2016-03-09
modelPosition() scales according to zoom used for styling

blair1618:
This ensures that values returned from `modelPosition()` are scaled appropriately for over-zoomed tiles. 

Essentially, a distance of `1.0` according to `modelPosition()` now corresponds to the size of one tile at the zoom level for which the tile was styled (not the zoom level of the original tile data, in the case of over-zoomed tiles). 

Related tangram-js PR: https://github.com/tangrams/tangram/pull/266


blair1618:
@tallytalwar review and merge when you're able, thanks!

tallytalwar:
already on it. Thanks

Varun Talwar
Mailto: varun@mapzen.com

core/resources/shaders/polyline.vs
```diff
@@ -69,7 +70,7 @@ void main() {
         vec4 extrude = UNPACK_EXTRUSION(a_extrude);
         float width = extrude.z;
         float dwdz = extrude.w;
-        float dz = u_map_position.z - abs(u_tile_origin.z);
+        float dz = u_map_position.z - u_tile_origin.w;
```
tallytalwar: shouldn't this be using style zoom?:
`float dz = u_map_position.z - u_tile_origin.z`?

tallytalwar: Ohh, 

I just saw the discussion here: https://github.com/tangrams/tangram/pull/266
We can probably address this small change in this PR only.

blair1618: It is currently correct to use the data zoom here because width is evaluated based on the data zoom. See the comments here: https://github.com/tangrams/tangram/pull/266

blair1618: Ah, you'd prefer to make that additional change in this PR?

tallytalwar: Yup, should be a small change, can be addressed in this PR only.

blair1618: Updated!

core/src/style/polylineStyle.cpp
```diff
@@ -145,15 +145,26 @@ struct PolylineStyleBuilder : public StyleBuilder {
...
 };
 
 template <class V>
-void PolylineStyleBuilder<V>::setup(const Tile& _tile) {
-    m_tileUnitsPerMeter = _tile.getInverseScale();
-    m_zoom = _tile.getID().z;
-    m_tileSize = _tile.getProjection()->TileSize();
+void PolylineStyleBuilder<V>::setup(const Tile& tile) {
+
+	auto id = tile.getID();
```
tallytalwar: Indentation.

And use reference.

core/src/style/polylineStyle.cpp
```diff
@@ -145,15 +145,26 @@ struct PolylineStyleBuilder : public StyleBuilder {
...
+    // Use the 'style zoom' to evaluate style parameters.
+    m_zoom = id.s;
+    m_tileUnitsPerMeter = tile.getInverseScale();
+    m_tileSizePixels = tile.getProjection()->TileSize();
+
+    // When a tile is overzoomed, we are actually styling the area of its
+    // 'source' tile, which will have a larger effective pixel size at the
+    // 'style' zoom level.
+    if (id.s > id.z) {
+        m_tileSizePixels *= std::exp2(id.s - id.z);
```
tallytalwar: We do not have to make this if condition here, (since style ~~id~~zoom can never be less than actual tile ~~id~~zoom).

blair1618: It shouldn't be less (save for programmer error), but it can be (and usually is) equal.

tallytalwar: And when they are equal its simply multiplying the `m_tileSizePixels` by 1 :).

core/src/style/polylineStyle.cpp
blair1618: Gah, mergetool indents with tabs! Updated.

core/src/style/polylineStyle.cpp
blair1618: Fair enough. Updated.

PR #612, opened 2016-03-09
check if GL_ARB_texture_non_power_of_two is available

hjanetzek:


core/src/gl/hardware.cpp
```diff
@@ -57,7 +60,9 @@ void loadExtensions() {
 }
 
 void loadCapabilities() {
-    glGetIntegerv(GL_MAX_TEXTURE_SIZE, &maxTextureSize);
+    int val;
+    glGetIntegerv(GL_MAX_TEXTURE_SIZE, &val);
+    maxTextureSize = val;
```
tallytalwar: why?

hjanetzek: yup. shut up the warning!

PR #613, opened 2016-03-09
Test the first matching filter only once

hjanetzek:
The second commit is a bit unrelated and could be another PR

### DIFF
--- a/core/src/scene/drawRule.cpp
+++ b/core/src/scene/drawRule.cpp
@@ -124,7 +124,6 @@ bool DrawRuleMergeSet::match(const Feature& _feature, const SceneLayer& _layer, 
     // If the first filter doesn't match, return immediately
     if (!_layer.filter().eval(_feature, _ctx)) { return false; }
 
-    // Add the first layer to the stack
     m_queuedLayers.push_back(&_layer);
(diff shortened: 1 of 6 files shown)


core/src/scene/drawRule.cpp
```diff
@@ -211,15 +210,17 @@ void DrawRuleMergeSet::apply(const Feature& _feature, const SceneLayer& _layer,
 
 void DrawRuleMergeSet::mergeRules(const SceneLayer& _layer) {
 
-    for (const auto& rule : _layer.rules()) {
+    size_t pos, end = m_matchedRules.size();
 
-        auto it = std::find_if(m_matchedRules.begin(), m_matchedRules.end(),
```
tallytalwar: What was the reason for this change?

Previously I found using `std::find_if` to be slower in comparison to doing it explicitly. Did you also discover the same?

hjanetzek: No, one needs to only compare the id's up to ```m_matchedRules.size()``` from the beginning, since the rules from _layer don't have the same id's twice.

tallytalwar:
LGTM.

PR #614, opened 2016-03-09
Tangram ES (Android) Deployment

### LINKED ISSUE #557: Android toolchain should be able to compile and publish a release AAR to maven

ecgreb:
Updates build scripts and adds necessary configuration and documentation to release Android AAR to Maven Central.

Closes #557

### DIFF
--- a/android/release-checklist.md
+++ b/android/release-checklist.md
@@ -1,0 +1,35 @@
+Tangram ES (Android) Release Checklist
+======================================
+
+Steps to build and release the Tangram ES (AAR) to Maven Central:
+
+## 1. Remove `-SNAPSHOT` tag from the version name in `gradle.properties`. Commit and push to master.
+```
+GROUP=com.mapzen.tangram
+VERSION_NAME=1.0.0
+...
+```
+
+## 2. Tag commit with release version (ex. tangram-1.0.0) and push tag to GitHub.
+```
+$ git tag tangram-1.0.0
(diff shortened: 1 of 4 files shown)


tallytalwar:
Looks good to me.

ecgreb:
More changes likely to come on this. The first commit makes it possible to upload a release AAR to Maven Central but does not put into place any process to do so.

Given the NDK configuration required for Tangram I don't feel comfortable building a release artifact locally on my laptop (which is what we do right now for other Android projects). Right now I'm looking into how we can use Travis to build and sign the release and push to Maven Central.

This will likely require changes to [script_deploy_android_snapshot.sh](https://github.com/tangrams/tangram-es/blob/master/travis/script_deploy_android_snapshot.sh) to build, sign, and upload the production artifact.

**So my question is... do we want to build additional architectures in the deploy script that will be included only for release versions? Right now several are commented out (x86_64, mips, mips64).**

blair1618:
Good question. I haven't personally encountered any android devices that use those CPU architectures. My take is that the benefit of a smaller download for the library would outweigh the benefit of reaching any of these very, very rare devices. We can always add support for those in the future if we determine that there's any demand for them.

tallytalwar:
Agreed with what @blair1618 said.

tallytalwar:
I guess this also fixes #557

ecgreb:
Yup! Updated description.

PR #615, opened 2016-03-09
Adds layer's visible parameter

### LINKED ISSUE #543: Layer "visible" property should set layer visibility
https://github.com/tangrams/tangram-es/blob/master/core/src/scene/sceneLoader.cpp#L1354

https://github.com/tangrams/tangram-docs/blob/gh-pages/pages/layers.md#visible

tallytalwar:
Fixes #543 

Added layer:visible property and respect skip drawing for layer if its marked not visible.

### DIFF
--- a/core/src/scene/sceneLayer.cpp
+++ b/core/src/scene/sceneLayer.cpp
@@ -6,11 +6,13 @@ namespace Tangram {
 
 SceneLayer::SceneLayer(std::string _name, Filter _filter,
                        std::vector<DrawRuleData> _rules,
-                       std::vector<SceneLayer> _sublayers) :
+                       std::vector<SceneLayer> _sublayers,
+                       bool _visible) :
     m_filter(std::move(_filter)),
     m_name(_name),
     m_rules(_rules),
-    m_sublayers(std::move(_sublayers)) {
+    m_sublayers(std::move(_sublayers)),
+    m_visible(_visible) {
 
     setDepth(1);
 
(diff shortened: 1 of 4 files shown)


core/src/scene/sceneLayer.cpp
```diff
@@ -6,11 +6,13 @@ namespace Tangram {
 
 SceneLayer::SceneLayer(std::string _name, Filter _filter,
                        std::vector<DrawRuleData> _rules,
-                       std::vector<SceneLayer> _sublayers) :
+                       std::vector<SceneLayer> _sublayers,
+                       bool _visible) :
```
karimnaaji: Can we default the value to `true` in the constructor instead?

karimnaaji:
Good to go once filter constructor is updated with default value or unit tests adapted!

hjanetzek:
Why do we add invisible layers at all and don't throw them out on load?

karimnaaji:
I guess this could be a dynamically changed property for debug or other purpose.

tallytalwar:
I did think about throwing them out on load,  but if we provide hooks for
dynamic changing properties then we need to have these in scene.

bcamper:
Yeah, to ensure compatibility with JS I don't think we should get too
clever with throwing them out for now.

core/src/scene/sceneLayer.cpp
tallytalwar: :+1:

PR #619, opened 2016-03-11
Fix style mixing, for merging map properties on an empty target node

### LINKED ISSUE #618: Regression on style mixing uniform shader block
While adding uniform arrays, I discovered that we might have done a regression on uniform shader mixing (probably define too).

This currently fails on master, it seems that `heightglowline` doesn't get mixed properly for the block uniforms:

```yaml
     heightglow:
         base: polygons
         shaders:
             uniforms:
                  u_test: 800.0
             blocks:
                  color: "color.rgb += vec3(position.z / u_test);"
     heightglowline:
         base: lines
         mix: heightglow
```

Compilation error on the shader while compiling `heightglowline`:
```
ERROR shaderProgram.cpp:212: Compiling shader:
ERROR: 0:451: Use of undeclared identifier 'u_test'
```

We should consolidate the test case coverage for those.

tallytalwar:
Fixes #618

### DIFF
--- a/core/src/scene/styleMixer.cpp
+++ b/core/src/scene/styleMixer.cpp
@@ -200,10 +200,10 @@ void StyleMixer::mergeFieldTakingLast(std::string key, Node target, const std::v
 
 void StyleMixer::mergeMapFieldTakingLast(std::string key, Node target, const std::vector<Node>& sources) {
 
+
     Node map = target[key];
-    if (!map || !map.IsMap()) {
-        return;
-    }
+    if (map && !map.IsMap()) { return; }
+
     for (auto it = sources.rbegin(); it != sources.rend(); ++it) {
         const auto& source = (*it)[key];
         if (!source.IsMap()) {
--- a/core/src/scene/styleMixer.cpp
+++ b/core/src/scene/styleMixer.cpp
@@ -165,7 +165,7 @@ void StyleMixer::applyShaderMixins(Node _shaders, const std::vector<Node>& _mixi
     }
 }
 
-void StyleMixer::mergeBooleanFieldAsDisjunction(std::string key, Node target, const std::vector<Node>& sources) {
+void StyleMixer::mergeBooleanFieldAsDisjunction(const std::string& key, Node target, const std::vector<Node>& sources) {
 
     auto current = target[key];
     if (current && current.as<bool>(false)) {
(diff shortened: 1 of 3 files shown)


karimnaaji:
I think it would make sense to have a test case that covers this

tallytalwar:
Yup will add one.

blair1618:
Yeahhhhh that function shouldn't return immediately if the map is null. We *do* still want to check if it is not a map, otherwise we'll crash below when we try to access elements by key.

tallytalwar:
Test added! And this test fails on master branch.

tallytalwar:
@blair1618, refer to the test case I added, we would like to merge (override the non-map target) with the mappings from the sources. I also do not see where will it crash?

core/src/scene/styleMixer.cpp
```diff
@@ -201,9 +201,6 @@ void StyleMixer::mergeFieldTakingLast(std::string key, Node target, const std::v
 void StyleMixer::mergeMapFieldTakingLast(std::string key, Node target, const std::vector<Node>& sources) {
```
karimnaaji: Just seen we were copying the keys here, we should use `const std::string&` instead. Can we update that at the same time?

blair1618:
Yeah when `map` doesn't exist we want to override it with the values from `sources`, but if `map` exists and is a scalar or a list, then `map[subkey]` will crash.

tallytalwar:
Yup, I updated my comment :D.

blair1618:
So the change should just be `if (!map || !map.IsMap()) {` -> `if (map && !map.IsMap()) {`

core/src/scene/styleMixer.cpp
blair1618: It's probably move-constructed anyway, but couldn't hurt

karimnaaji: I'm not convinced we should generally rely that much on that,
quick bench shows it's much faster to pass by const ref, even with optimization on:
https://gist.github.com/karimnaaji/d58afb771de5edfc98a1

```
~/Desktop/cpp-benchs  master ✗                                                                                                                                               80d ⚑  
> g++ const_string_vs_value_string.cpp -O3 -std=c++11

~/Desktop/cpp-benchs  master ✗                                                                                                                                               80d ⚑  
> ./a.out                                            
function_a: 0.002ms         # const ref 
function_b: 68.776ms       # copy
function_b-move: 2.86ms # move
```

blair1618: Fair enough, refs it is.

tallytalwar:
Updated.

blair1618:
LGTM

karimnaaji:
Looks good to me too, thanks for fixing this @tallytalwar

Issue #548, opened 2016-02-16
Style "uniform" declarations should support (non-vector) array types


blair1618:
https://github.com/tangrams/tangram-es/blob/master/core/src/scene/sceneLoader.cpp#L1271

https://github.com/tangrams/tangram-docs/blob/gh-pages/pages/shaders.md#uniforms

The specific behavior of the type inference from YAML is not clear from the documentation, but in the JS implementation sequences of 5 or more values and sequences whose elements are not numbers are inferred as arrays.

PR #620, opened 2016-03-11
Uniform arrays

karimnaaji:
Fixes https://github.com/tangrams/tangram-es/issues/548 

Add possibility to use uniform arrays, this part is partially documented but will complete the docs more consistently.

Comparing with the -js behavior, this adds the possibility to write the following:

```yaml
textures:
    texture0:
        url: img/poi_icons_32.png
    texture1:
        url: img/sem.jpg
styles:
    heightglow:
        base: polygons
        texcoords: true
        shaders:
            uniforms:
                u_test: [0.5, 0.5, 1.0, 0.5, 0.9]
                u_textures: [texture0, texture1]
            blocks:
                color: |
                    color.rgb *= mix(texture2D(u_textures[1], v_texcoord).rgb, vec3(u_test[0], u_test[1], u_test[4]), 0.5);
```

- Refactor uniform block scene loading
- Fix a bug in the way texture unit were used (now next available texture unit is tracked within the render states, https://github.com/tangrams/tangram-es/commit/ed29ad6bdd8df72d45f3d2a83ef2f1ccb84dbbd5 but could be a separate PR if needed)

### DIFF
--- a/core/src/gl/renderState.cpp
+++ b/core/src/gl/renderState.cpp
@@ -2,6 +2,7 @@
 
 #include "platform.h"
 #include "vertexLayout.h"
+#include "gl/hardware.h"
 
 namespace Tangram {
 
@@ -35,10 +36,6 @@ namespace RenderState {
     ClearColor clearColor;
 
     GLuint getTextureUnit(GLuint _unit) {
-        if (_unit >= TANGRAM_MAX_TEXTURE_UNIT) {
-            LOGW("trying to access unavailable texture unit");
-        }
-
         return GL_TEXTURE0 + _unit;
     }
 
@@ -85,6 +82,11 @@ namespace RenderState {
     }
 
     int nextAvailableTextureUnit() {
+        if (s_textureUnit + 1 > Hardware::maxCombinedTextureUnits) {
+            LOGE("Too many combined texture units are being used");
+            LOGE("GPU supports %d combined texture units", Hardware::maxCombinedTextureUnits);
+        }
+
         return ++s_textureUnit;
     }
 
--- a/core/src/gl/renderState.cpp
+++ b/core/src/gl/renderState.cpp
@@ -7,6 +7,7 @@ namespace Tangram {
 
  // Incremented when the GL context is invalidated
 static int s_validGeneration;
+static int s_textureUnit;
 
 namespace RenderState {
 
@@ -47,6 +48,7 @@ namespace RenderState {
     void bindTexture(GLenum _target, GLuint _textureId) { glBindTexture(_target, _textureId); }
 
     void configure() {
+        s_textureUnit = -1;
         s_validGeneration++;
         VertexLayout::clearCache();
 
@@ -63,13 +65,15 @@ namespace RenderState {
         glClearDepthf(1.0);
         glDepthRangef(0.0, 1.0);
 
+        static size_t max = std::numeric_limits<size_t>::max();
+
         clearColor.init(0.0, 0.0, 0.0, 0.0);
-        shaderProgram.init(std::numeric_limits<unsigned int>::max(), false);
-        vertexBuffer.init(std::numeric_limits<unsigned int>::max(), false);
-        indexBuffer.init(std::numeric_limits<unsigned int>::max(), false);
-        texture.init(GL_TEXTURE_2D, std::numeric_limits<unsigned int>::max(), false);
-        texture.init(GL_TEXTURE_CUBE_MAP, std::numeric_limits<unsigned int>::max(), false);
-        textureUnit.init(std::numeric_limits<unsigned int>::max(), false);
+        shaderProgram.init(max, false);
+        vertexBuffer.init(max, false);
+        indexBuffer.init(max, false);
+        texture.init(GL_TEXTURE_2D, max, false);
+        texture.init(GL_TEXTURE_CUBE_MAP, max, false);
+        textureUnit.init(max, false);
     }
 
     bool isValidGeneration(int _generation) {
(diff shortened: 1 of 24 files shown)


karimnaaji:
@tallytalwar , can you look over this when you have a moment?
Thanks :)

tallytalwar:
Sure will do. Thanks

core/src/scene/sceneLoader.cpp
```diff
@@ -965,66 +980,77 @@ void SceneLoader::parseStyleParams(Node params, Scene& scene, const std::string&
...
+                    LOGW("Can't load texture with name %s", textureName.c_str());
                 }
             }
+
+            styleUniform.value = textureArrayUniform;
         }
     } else {
         LOGW("Expected a scalar or sequence value for uniforms");
     }
-    return std::make_pair(type, std::move(uniformValues));
```
tallytalwar: Any specific reason of not returning a pair and creating an explicit struct?

tallytalwar: I am fine with this change, just curious if you noticed anything interesting.

core/src/scene/sceneLoader.cpp
```diff
@@ -965,66 +980,77 @@ void SceneLoader::parseStyleParams(Node params, Scene& scene, const std::string&
...
+            styleUniform.type = "float";
+            styleUniform.value = (float)fValue;
         } else if (getBool(value, bValue)) {
-            type = "bool";
-            uniformValues.push_back(bValue);
-
+            styleUniform.type = "bool";
+            styleUniform.value = (bool)bValue;
         } else {
             auto strVal = value.as<std::string>();
```
tallytalwar: Can we use `Scalar()` call instead of `as`?

core/src/scene/sceneLoader.cpp
karimnaaji: I think it's more clear to make the struct explicit, I found it a bit sketchy to use the pair as an implicit struct while refactoring this (mainly for readability), since you always have to look at what is `first`/`second` in the pair.

core/src/scene/sceneLoader.cpp
karimnaaji: :+1:

karimnaaji: Making this change a separate PR.

core/src/scene/sceneLoader.cpp
```diff
@@ -965,66 +980,77 @@ void SceneLoader::parseStyleParams(Node params, Scene& scene, const std::string&
...
+                    styleUniform.value = parseVec<glm::vec3>(value);
                     break;
                 case 4:
-                    uVal = parseVec<glm::vec4>(value);
+                    styleUniform.value = parseVec<glm::vec4>(value);
                     break;
                 default:
+                    UniformArray uniformArray;
+                    for (const auto& val : value) {
+                        uniformArray.push_back(val.as<double>());
```
tallytalwar: maybe it make sense to get the float value here.

core/src/gl/texture.cpp
```diff
@@ -45,6 +45,19 @@ Texture::Texture(const std::string& _file, TextureOptions _options, bool _genera
     stbi_image_free(pixels);
 }
 
+Texture::Texture(const unsigned char* data, size_t dataSize, TextureOptions options, bool generateMipmaps)
+    : Texture(0u, 0u, options, generateMipmaps) {
+    unsigned char* pixels;
+    int width, height, comp;
+
+    pixels = stbi_load_from_memory(data, dataSize, &width, &height, &comp, STBI_rgb_alpha);
```
tallytalwar: maybe you can make stbi code a separate function and call it from both the parameterized constructors you have for Texture.

core/src/gl/renderState.cpp
```diff
@@ -80,6 +81,22 @@ namespace RenderState {
         return s_validGeneration;
     }
 
+    int nextAvailableTextureUnit() {
+        if (s_textureUnit + 1 > Hardware::maxCombinedTextureUnits) {
+            LOGE("Too many combined texture units are being used");
+            LOGE("GPU supports %d combined texture units", Hardware::maxCombinedTextureUnits);
```
tallytalwar: What happens when this max is reached?

tallytalwar:
This looks good to me.

I like all the texture index tracking in render state, ideally should have had a PR for itself, but its fine now.

tallytalwar:
Once scene file is reverted, will merge this.

core/src/gl/renderState.cpp
karimnaaji: We would have those log and catch a `GL_INVALID_ENUM` error later on.

karimnaaji:
Updated

core/src/gl/renderState.cpp
tallytalwar: :+1:

core/src/gl/texture.cpp
karimnaaji: Yeah we can clean that up a bit.

hjanetzek:
How can one know whether texture0 or texture1 is texture[1] in the example? This seems to rely on the map behavior of the yaml implementation.

karimnaaji:
Yes it seems that the order is not guaranteed from the yaml spec. As far as I can tell this behave consistenly between -js and -es.

hjanetzek:
oh, just saw ```u_textures: [texture0, texture1]``` - then it's alright! :)

core/src/style/pointStyle.cpp
```diff
@@ -53,18 +54,19 @@ void PointStyle::constructShaderProgram() {
...
+        m_spriteAtlas->bind(RenderState::nextAvailableTextureUnit());
     } else if (m_texture) {
-        m_texture->update(0);
-        m_texture->bind(0);
+        m_texture->update(RenderState::nextAvailableTextureUnit());
+        m_texture->bind(RenderState::currentTextureUnit());
     }
 
     m_shaderProgram->setUniformi(m_uTex, 0);
     m_shaderProgram->setUniformMatrix4f(m_uOrtho, _view.getOrthoViewportMatrix());
```
hjanetzek: one missing ```RenderState::currentTextureUnit()``` for m_uTex

karimnaaji: Yes! Good catch.

PR #621, opened 2016-03-11
Consistent code style for shaders

karimnaaji:


blair1618:
The docs for shaders list uniforms that "are present on the vertex and fragment shaders":

 - https://github.com/tangrams/tangram-docs/blob/gh-pages/pages/Shaders-Overview.md#defines-and-uniforms
 - https://github.com/tangrams/tangram-docs/blob/gh-pages/pages/shaders.md#built-in-uniforms

`u_normalMatrix` is one of those listed. However... there are other uniforms listed there that we *do not* make available in shaders on ES :\ I'll make a separate issue to track and discuss that. This PR is good to go with the exception of `u_normalMatrix`, which we should (regrettably) leave as-is.

bcamper:
I also think it's questionable how many of these we should be publicly
exposing. Feel free to propose a trim.

karimnaaji:
Yes it seems that we're missing `u_modelView` in -es. Will revert the change on renaming `u_normalMatrix`, I didn't know this one was available for the user. It could be one to trim since the user accesses normals with this transform already applied, and now has access to `worldNormals()`.

blair1618:
After the discussion in https://github.com/tangrams/tangram-es/issues/622, seems like it should be fine now to make these edits.

Merging!

PR #624, opened 2016-03-14
Revert "Specialized mesh for vertex type with no UVs"

karimnaaji:
This reverts commit 768d2a3; which has been merged while being discussed.


tallytalwar:
Since we decided in https://github.com/tangrams/tangram-es/pull/589, that this and more refactor could be done later, merging this revert makes sense to me.

Issue #617, opened 2016-03-10
"scene.animated" should set whether the scene renders continuously


blair1618:
https://github.com/tangrams/tangram-docs/blob/gh-pages/pages/scene.md#animated

blair1618:
@bcamper does this always override the `animated` property of styles? i.e. If `scene.animated` is `false` and a style has `animated: true`, does the scene animate?

bcamper:
Yes, it's a scene-level override. If `scene.animated` is present, it's
value is used. If it's not, then the scene is animated if *one or more*
styles has `animated: true`.

tallytalwar:
Fixed by #629

PR #629, opened 2016-03-16
Animated scene property

karimnaaji:
https://github.com/tangrams/tangram-es/issues/617

### DIFF
--- a/core/src/tangram.cpp
+++ b/core/src/tangram.cpp
@@ -169,12 +169,14 @@ void update(float _dt) {
         }
     }
 
-    bool animated = false;
+    bool animated = m_scene->animated() == Scene::animate::yes;
     for (const auto& style : m_scene->styles()) {
         style->onBeginUpdate();
-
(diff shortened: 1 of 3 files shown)


core/src/tangram.cpp
```diff
@@ -196,7 +196,7 @@ void update(float _dt) {
             m_labels->update(*m_view, _dt, m_scene->styles(), tiles, cache);
         }
 
-        bool animated = false;
+        bool animated = m_scene->animated();
         for (const auto& style : m_scene->styles()) {
             if (style->isAnimated()) {
```
tallytalwar: Based on the discussion on #617, it seems if `scene.animated = false` then scene will not be animated irrespective of the value of `style.animate`.
However here, style animate will override scene animate property.

tallytalwar:
LGTM. Rebase required with master.

karimnaaji:
Rebased

PR #630, opened 2016-03-16
Cleanup debug code

karimnaaji:
Cleanup debug code from `tangram.cpp` to make it easier to step through main interface.


karimnaaji:
@tallytalwar , can you give a look to fix travis build on this, I can't figure out why linux build keeps failing and my linux setup is not yet usable.

tallytalwar:
Sure, will check it out.

tallytalwar:
Interesting that this works with clang-3.6 on linux.

tallytalwar:
Will look more closely at whats happening in this PR, which is calling g++ to bomb!

hjanetzek:
The problem is that gnu libc++ has its own debug/debug.h
How about debug/frameInfo and moving DebugFlags to tangram.h?

One thing: can we have clang-format ColumnLimit: 100?

blair1618:
LGTM

blair1618:
@hjanetzek I think the column limit in our .clang-format is 120 because it's roughly the width that can be viewed without scrolling in github. You prefer 100?

PR #634, opened 2016-03-21
scale text offset by pixelScale

hjanetzek:


hjanetzek:
This fix is necessary to show both text- and sprite-labels of features in eraser-map when pixelScale > 1. See Williamsburg Bridge for example.

blair1618:
Yeah this makes sense. 

@nvkelso this may address the regressions you were just pointing out to me.

hjanetzek:
it's indeed a regression from pre-alfons :/

hjanetzek:
Debug mode to show all labels that were added for a tile
shot-2016-03-21_22-16-02

blair1618:
Oh boy the debug drawing looks fun haha

So this is all the labels in the buffer objects?

hjanetzek:
@blair1618 that's what *was* hidden in the buffer objects. Now it's only the visible labels and the quadvertices for labels kept per tile are much smaller than the mesh buffer vertices

blair1618:
Cool. LGTM.

Issue #603, opened 2016-03-05
Repeated points in linestrings should not impact line tessellation


blair1618:
Some vector tile data can contain linestrings with consecutive identical points. This causes a normal vector calculation to fail in the polyline tessellator and results in incorrect vertex locations:

<img width="912" alt="screen shot 2016-03-04 at 6 54 16 pm" src="https://cloud.githubusercontent.com/assets/3371850/13543686/8b92bb30-e23b-11e5-9edc-ac6861ef5e56.png">

This doesn't occur for MVT tiles due to a de-duplication step during geometry decoding.

blair1618:
This seems to be a much less common occurrence in our latest set of tiles (yay), which means that we'll have to conjure up some test data to make sure we fix this (boo).

nvkelso:
Slap a /osm/v0.7/ in the source on prod and you'll have the older data.

blair1618:
Resolved by #640

PR #640, opened 2016-03-22
Fix NaN values on builder

karimnaaji:
I noticed we were sending some NaN values to OpenGL in our buffers. After some research it's unclear how this would be handled in ES 2 (mostly undefined behavior), so for sanity we should check for it. It seems that what could happen on some driver is that the value would be replaced by 0 for NaNs, which could potentially result for us in a black pixels if lightning is used.

### DIFF
--- a/core/src/util/builders.cpp
+++ b/core/src/util/builders.cpp
@@ -314,6 +314,10 @@ void buildPolyLineSegment(const Line& _line, PolyLineBuilder& _ctx) {
         coordCurr = coordNext;
         coordNext = _line[i + 1];
 
+        if (coordCurr == coordNext) {
+            continue;
+        }
+
         normPrev = normNext;
         normNext = glm::normalize(perp2d(coordCurr, coordNext));
 
--- a/core/src/util/builders.cpp
+++ b/core/src/util/builders.cpp
@@ -2,8 +2,7 @@
 
 #include "geom.h"
 #include "glm/gtx/rotate_vector.hpp"
-#include "platform.h"
-#include <memory>
+#include "glm/gtx/norm.hpp"
 
 namespace mapbox { namespace util {
 template <>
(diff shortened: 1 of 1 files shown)


core/src/util/builders.cpp
```diff
@@ -137,6 +137,12 @@ void Builders::buildPolygonExtrusion(const Polygon& _polygon, float _minHeight,
             normalVector = glm::cross(upVector, b - a);
             normalVector = glm::normalize(normalVector);
 
+            if (std::isnan(normalVector.x)
```
tallytalwar: This is good, however I think we should be catching degenerate points within a valid polygon ring before we get to this stage!

karimnaaji: Should each data sources reflect this? I'm not sure. Testing a NaN number is barely a no-op.

blair1618: I believe that only the MVT decoder removes duplicate points. Checking in the building process is probably more robust in general, e.g. client-provided data could have degenerate geometry.

blair1618:
Strictly speaking, I don't think we're passing NAN float values into GL buffers because positions and normals are all encoded as shorts before upload. Casting NAN to short results in 0 on the configurations I've been able to test, but I believe it's technically *undefined behavior*.

tallytalwar:
Sure even if we are not passing NANs to GL, short conversion of a NAN to 0 is not the right behavior. 

I am fine with this check of NAN, and ignoring rendering the geometry which resulted in a NAN.

core/src/util/builders.cpp
tallytalwar: :+1:

blair1618:
We can also get NAN normals in the line builder when two points in a row are equal, which we should cover with another commit here.

karimnaaji:
Possibly related https://github.com/tangrams/tangram-es/issues/603

blair1618:
This all seems very sensible to me. I'm still looking into rare anomalies that look potentially related, like this:
screen shot 2016-03-28 at 2 48 31 pm

PR #642, opened 2016-03-23
Revised MapData interface

### LINKED ISSUE #560: Java interface should not require use of global static methods
This is a symptom of one or more larger problems/oddities in the architecture of our Android SDK. We're currently using a combination of hand-written Java + JNI classes and SWIG-generated Java classes. Since the SWIG classes mimic the C++ code that they are based on, and our C++ interface is essentially a collection of global functions, some of our functionality can only be accessed through static methods in Java. This pattern is at best unusual and at worst dangerous (consider that the static methods can be used before the renderer is initialized), so we should revise this aspect of the interface.

blair1618:
This replaces the SWIG-generated MapData class with a new interface that works more naturally with idiomatic Java. This means:

 - Code is easily readable and much smaller
 - Code is maintainable without additional tooling
 - Interface is more precise and clear
 - Interface is safer and doesn't rely on global statics (resolves #560)

Some parts of this could potentially be improved:

 - Naming of "MapData": Users of other SDKs will likely be more familiar with this as a "layer", but we also want to make clear the connection to "data sources".
 - Efficiency of JNI data transfer: Passing features through JNI could theoretically be batched for greater speed, instead of passed with one JNI call per feature. 
 - The "addToMap", "syncWithMap", "removeFromMap" pattern: I think this will actually work pretty well, but I would like to hear feedback on it. 
 - Storing "DataSource" pointers in Java: I'd rather the library handed out integer handles for accessing data sources, then we'd be able to know whether a reference has somehow been invalidated without dereferencing a freed pointer.

TODO:

- [x] implement `Tangram::removeDataSource`
- [x] test polygon and geojson features


blair1618:
@tallytalwar @karimnaaji check this out when you get a chance, thanks!

tallytalwar:
Reviewing this now.

tallytalwar:
Can you have a PR ready on Eraser-map repo, for the corresponding changes? And then we can cut a new release whenever this gets merged! Or we can wait for eraser-map to pick these changes, till we actually have a more proper release :D.

blair1618:
Yeah, probably not a bad idea to try out these changes in eraser-map to see what it looks like on the client side. That said, I don't think I'll make a PR with those changes yet because it'll just sit there and fail builds until this interface is in a release that we can point to.

tallytalwar:
Sure since EM is dependent on an explicit release of tangram, we can do it more serially now, and will not result in breaking EM master builds.

android/tangram/jni/jniExports.cpp
```diff
@@ -143,5 +146,88 @@ extern "C" {
...
 
+        size_t n_points = jniEnv->GetArrayLength(jcoordinates) / 2;
+        size_t n_rings = (jrings == NULL) ? 0 : jniEnv->GetArrayLength(jrings);
+        size_t n_properties = (jproperties == NULL) ? 0 : jniEnv->GetArrayLength(jproperties) / 2;
+
+        Tangram::Properties properties;
+
+        for (size_t i = 0; i < n_properties; ++i) {
+            jstring jkey = (jstring) (jniEnv->GetObjectArrayElement(jproperties, 2 * i));
+            jstring jvalue = (jstring) (jniEnv->GetObjectArrayElement(jproperties, 2 * i + 1));
```
karimnaaji: Shouldn't `DeleteLocalRef` be used on `jkey` and `jvalue` here?

blair1618: Don't think so, we never allocate a local reference (i.e. `NewLocalRef`)

karimnaaji: But how `GetObjectArrayElement` is freed? It seems that it allocates memory. Referring to discussions about this on stackoverflow and other codebases this is usually what is being done (https://github.com/spurious/SDL-mirror/blob/master/src/main/android/SDL_android_main.c#L45_L52).

blair1618: Hmm ok, so local references are created for every `jobject` we access through JNI, including argument objects (this keeps those from being GC'ed while we're in a native function). This won't cause any memory leaks because local references become invalid as soon as the native function exits; those objects can then be GC'ed as usual. The only reason we'd want to `DeleteLocalRef` on one of these objects is if we expect to create a very large number of local references (e.g. in a long loop), in which case we may exceed the local reference table size of the JNI environment. I don't think it would do any harm to delete our local references in this loop, so I'll go ahead and do that.

core/src/tangram.h
```diff
@@ -89,7 +89,9 @@ uint8_t getCameraType();
 // according to the scene file using the provided data source name;
 void addDataSource(std::shared_ptr<DataSource> _source);
 
-int removeDataSource(DataSource& _source);
```
tallytalwar: Ha! This was not implemented before!

android/tangram/src/com/mapzen/tangram/geometry/Geometry.java
```diff
@@ -0,0 +1,35 @@
...
+/**
+ * Geometry is an abstract container of LngLat points and associated properties
+ */
+public abstract class Geometry {
+
+    protected double[] coordinates;
+    protected int[] rings;
+    protected String[] properties;
+
+    public double[] getCoordinateArray() {
```
tallytalwar: Can you put some comments here for documentation purposes (for javadocs)?

android/tangram/src/com/mapzen/tangram/MapData.java
```diff
@@ -1,161 +1,154 @@
...
+    /**
+     * Add the features from this {@code MapData} to a map.
+     * <p>
+     * This {@code MapData} will be associated with the given map until {@link #removeFromMap()}
+     * is called or until this method is called again.
+     * @param map The {@code MapController} managing the destination map.
+     */
+    public void addToMap(MapController map) {
+        if (map == null) {
+            throw new RuntimeException("Cannot assign a null map, use removeFromMap() instead");
```
tallytalwar: I do not get the Exception message here, specially the second part.
What has `MapController` being `null` has anything to do with `removeFromMap`?

tallytalwar: Also I do understand the association of `MapController` and `MapData` here, but I think it will be more intuitive for `MapController` to own a set of `MapData(s)` references? Your thoughts?

android/tangram/src/com/mapzen/tangram/MapData.java
```diff
@@ -1,161 +1,154 @@
...
+        syncWithMap();
     }
 
-
     /**
-     * Construct a new MapData object for adding drawable data to the map
-     * @param name Name of the data source in the scene file for styling this
-     *             object's data
+     * Remove this {@code MapData} from the map it is currently associated with. This must not be
+     * called when this {@code MapData} is not associated with a map.
```
tallytalwar: We should also add a helper method to check if a MapData is currently associated with a Map.

android/tangram/src/com/mapzen/tangram/MapData.java
```diff
@@ -1,161 +1,154 @@
...
      */
-    public MapData(String name) {
-        this(TangramJNI.new_MapData(name, ""), true);
+    public void removeFromMap() {
+        if (map == null) {
+            throw new RuntimeException("There is no associated map for this MapData");
```
tallytalwar: Maybe instead of the Exception, make this return a `boolean`, which returns true if this `MapData` object is associated with a `MapController` and false otherwise?

android/tangram/src/com/mapzen/tangram/MapData.java
```diff
@@ -1,161 +1,154 @@
...
+     * This method must be called after features are added or removed to update their
+     * appearance in the map. This method must not be called when this {@code MapData} is not
+     * associated with a map.
      */
-    public MapData clear() {
-        clearJNI();
-        return this;
+    public void syncWithMap() {
+        if (map == null) {
+            throw new RuntimeException("There is no associated map for this MapData");
```
tallytalwar: Similar to `removeFromMap` method, does it make sense to make `syncWithMap` return a `boolean` instead of the exception?

android/tangram/src/com/mapzen/tangram/geometry/Geometry.java
blair1618: Ah thanks for pointing this out, I actually think the classes in `com.mapzen.tangram.geometry` don't even need to be public, I'll change that.

android/tangram/src/com/mapzen/tangram/MapData.java
```diff
@@ -1,161 +1,154 @@
...
+        pointer = 0;
+        map = null;
     }
 
     /**
-     * Remove all data from this source
-     * @return This object, for chaining
+     * Synchronize the features in this {@code MapData} with the map.
+     * <p>
+     * This method must be called after features are added or removed to update their
```
tallytalwar: Is it a `must` call?

What if the user wants to add multiple features to this `MapData`, in which case all previous tiles and data for this datasource will be cleaned!

Also if its a must, we should handle it implicitly? Thoughts?

android/tangram/src/com/mapzen/tangram/MapData.java
```diff
@@ -1,161 +1,154 @@
...
 
-    public MapData addPolygon(Properties props, Polygon polygon) {
-        addPolyJNI(props, polygon);
+    /**
+     * Remove all features from this collection.
+     * @return This object, for chaining.
+     */
+    public MapData clear() {
+        geojson.clear();
+        features.clear();
```
tallytalwar: Example for a usecase for this be the routing icon on EM:
* get new location
* clear routeIcon MapData
* add new location Point to routeIcon MapData
* syncWithMap

android/tangram/src/com/mapzen/tangram/MapData.java
blair1618: The idea with the exception message is that someone might try to remove this data source from a map by setting its map to null, so I wanted to inform them of the right way to do so. I'll make that message more clear.

android/tangram/src/com/mapzen/tangram/geometry/Geometry.java
```diff
@@ -0,0 +1,35 @@
...
+import java.util.Map;
+
+/**
+ * Geometry is an abstract container of LngLat points and associated properties
+ */
+public abstract class Geometry {
+
+    protected double[] coordinates;
+    protected int[] rings;
+    protected String[] properties;
```
tallytalwar: Does it make sense to add a serialID to this?

Could be useful in explicitly removing a specific geometry feature from a set of features of a MapData?

android/tangram/src/com/mapzen/tangram/MapData.java
blair1618: As for the direction of association between `MapData` and `MapController`, I see trade-offs either way. What makes you say that it's more intuitive for `MapController` to own references to `MapData`?

android/tangram/src/com/mapzen/tangram/MapData.java
blair1618: That seems sensible, something like `MapController getMap()`.

android/tangram/src/com/mapzen/tangram/MapData.java
tallytalwar: I am thinking of `Map`/`MapController` having a container to `MapData(s)`. Also telling MapData to add or remove itself from a Map it kinda counter intutive. 

Maybe rename that to `Attach`/`Detach` a `MapData` to `MapController`?

android/tangram/src/com/mapzen/tangram/MapData.java
blair1618: Calling `removeFromMap()` when there is no map associated is a programmer error. The idea behind throwing an exception here is for the code to teach and enforce correct management of `MapData` instead of permitting confusing sloppiness.

android/tangram/src/com/mapzen/tangram/MapData.java
blair1618: Same reason as above, the code should enforce correct usage.

android/tangram/src/com/mapzen/tangram/MapData.java
tallytalwar: Sure, `map()`, `associatedMap()` or even `getMap()`.

android/tangram/jni/platform_android.cpp
```diff
@@ -328,5 +337,11 @@ void initGLExtensions() {
     glGenVertexArraysOESEXT = (PFNGLGENVERTEXARRAYSOESPROC) dlsym(libhandle, "glGenVertexArraysOES");
 }
 
+std::string stringFromJString(JNIEnv* jniEnv, jstring string) {
+    const char* cstring = jniEnv->GetStringUTFChars(string, NULL);
```
tallytalwar: I prefer using `GetStringUTFRegion`. Threat of not releasing the resource is not there.

Refer: https://github.com/tangrams/tangram-es/blob/jni-revisited/android/tangram/jni/platform_android.cpp#L131

android/tangram/src/com/mapzen/tangram/MapData.java
blair1618: Yeah, I've also thought about just doing the synchronization automatically whenever a feature is added. The benefit of making this an explicit call is that we can then make the feature data processing much more efficient without changing the interface. For example, if we synchronize 500 features all at once instead of one at a time, then we could process the features with geojsonvt *once* instead of *500 times*.

android/tangram/src/com/mapzen/tangram/MapData.java
blair1618: Can you explain why it's not intuitive? In the words of another, "intuition is compressed experience". What experience makes you expect that `MapController` would contain a collection of `MapData`?

android/tangram/src/com/mapzen/tangram/MapData.java
blair1618: Yes, I think? I'm a little unclear what the question is haha

tallytalwar: Was just confirming the usage :D. Thanks.

android/tangram/src/com/mapzen/tangram/MapData.java
tallytalwar: Sure, I just do not get exceptions coming from non-java background :D.

android/tangram/src/com/mapzen/tangram/MapData.java
tallytalwar: Agreed! It totally make sense to syncing explicitly.

android/tangram/src/com/mapzen/tangram/geometry/Geometry.java
blair1618: I think the better question to ask is "should we have a mechanism to remove individual features from the map?" Right now we don't, but I think we can add such a mechanism later in a natural and backwards-compatible way.

android/tangram/src/com/mapzen/tangram/MapData.java
```diff
@@ -1,161 +1,154 @@
...
+        if (map == null) {
+            throw new RuntimeException("Cannot assign a null map, use removeFromMap() instead");
+        }
+        if (this.map != null) {
+            removeFromMap();
         }
-        super.delete();
+        this.map = map;
+        this.pointer = map.nativeAddDataSource(name);
+        syncWithMap();
```
tallytalwar: We can probably avoid doing a `syncWithMap` here! Since MapData is basically empty (no features) at this point.

android/tangram/jni/platform_android.cpp
blair1618: Oh, I see! How lovely that JNI has so many ways to do the same thing :P

I guess I'm not especially worried about forgetting to release resources since this is encapsulated in a reusable function, I can change it if you have a strong preference.

android/tangram/src/com/mapzen/tangram/MapData.java
karimnaaji: But Java exception are much more consistent and reliable than c++ ones, I agree with @blair1618 on that. You can't really force the user of the API for a check with the bool return.

tallytalwar: Yes yes I also agree.  I was just expressing my usage of exceptions in a non java environment :).

android/tangram/src/com/mapzen/tangram/MapData.java
blair1618: The `MapData` could have features at this point; features can be added before being attached to a map (another reason that `syncWithMap` is explicit).

android/tangram/src/com/mapzen/tangram/geometry/Geometry.java
blair1618: Turns out we need these public to use them from the `com.mapzen.tangram` package - instead I've added comments that they are not for public use.

android/tangram/jni/platform_android.cpp
blair1618: Updated all string conversions to use the helper function, which itself uses `GetStringUTFRegion`

blair1618:
I've been reading a lot of the Google Maps API documentation today and if we're leaning towards a more similar interface (we might be soon) we could go for a lifecycle that looks more like:
```java
// Create a new data layer;
// No public constructor for MapData, you must go through a map instance.
MapData pois = map.AddDataLayer("pois");
// These can be accessed from the map object directly, 
// so we can't have an orphaned native object with no java reference to it. 
assert pois == map.dataLayers().get(0);
// Add features to the layer.
pois.addPoint(new LngLat(0, 0), properties);
// Synchronize with the map.
pois.syncWithMap();
// Remove the layer from the map,
// This removes the data layer from the map object, along with the native data source. 
pois.remove();
// The 'pois' object still exists until it's GC'ed, but now it does nothing. 
```
There are trade-offs to this approach too, just something to consider.

tallytalwar:
Yup, I was also looking at GMaps API and I agree there are trade-offs both ways. I am more inclined to mimic more of GMaps API, to have an easier learning curve for our users who are familiar with GMaps SDK.

tallytalwar:
We can also provide helper wrapper methods for LatLng to LngLat, wrt compatibility with GMaps API.

android/tangram/jni/jniExports.cpp
```diff
@@ -143,5 +146,90 @@ extern "C" {
...
+            auto key = stringFromJString(jniEnv, jkey);
+            auto value = stringFromJString(jniEnv, jvalue);
+            properties.set(key, value);
+            jniEnv->DeleteLocalRef(jkey);
+            jniEnv->DeleteLocalRef(jvalue);
+        }
+
+        auto* coordinates = jniEnv->GetDoubleArrayElements(jcoordinates, NULL);
```
tallytalwar: Though this is completely fine, we can also use the `GetDoubleArrayRegion` which is similar to the usage of `GetStringUTFRegion` :D.

blair1618:
After discussing with @tallytalwar we've agreed that there will be further refinements needed to this interface that may require more substantial change to the core library. The desired result is documented in #653. However this PR as it stands is still a major improvement to code quality and will therefore be merged shortly.

PR #645, opened 2016-03-28
Camera field of view and focal length are configurable from scene file

### LINKED ISSUE #540: Camera "fov" or "focal_length" property should set field-of-view angle for perspective cameras
https://github.com/tangrams/tangram-es/blob/master/core/src/scene/sceneLoader.cpp#L961

https://github.com/tangrams/tangram-docs/blob/gh-pages/pages/cameras.md#fov

blair1618:
Resolves #540 

This was a pretty straightforward set of changes. 

I shouldn't have needed to add code to `Stops` to complete this, but the current implementation of `Stops` is a bit silly. I'd like to revise it to be more naturally extensible, but that's for another time.

### DIFF
--- a/core/src/scene/sceneLoader.cpp
+++ b/core/src/scene/sceneLoader.cpp
@@ -722,12 +722,25 @@ void SceneLoader::loadCameras(Node _cameras, Scene& _scene) {
 
             view->setCameraType(CameraType::perspective);
 
-            if (Node fov = camera["fov"]) {
-                // TODO
-            }
-
+            // Only one of focal length and FOV is applied;
+            // according to docs, focal length takes precedence.
             if (Node focal = camera["focal_length"]) {
-                // TODO
+                if (focal.IsScalar()) {
+                    float length = focal.as<float>(view->getFocalLength());
+                    view->setFocalLength(length);
+                } else if (focal.IsSequence()) {
+                    auto stops = std::make_shared<Stops>(Stops::Numbers(focal));
+                    view->setFocalLengthStops(stops);
+                }
+            } else if (Node fov = camera["fov"]) {
+                if (fov.IsScalar()) {
+                    float degrees = fov.as<float>(view->getFieldOfView() * RAD_TO_DEG);
+                    view->setFieldOfView(degrees * DEG_TO_RAD);
+                } else if (fov.IsSequence()) {
+                    auto stops = std::make_shared<Stops>(Stops::Numbers(fov));
+                    for (auto& f : stops->frames) { f.value = f.value.get<float>() * DEG_TO_RAD; }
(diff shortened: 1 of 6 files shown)


core/src/scene/sceneLoader.cpp
```diff
@@ -722,12 +722,25 @@ void SceneLoader::loadCameras(Node _cameras, Scene& _scene) {
...
-            if (Node fov = camera["fov"]) {
-                // TODO
-            }
-
+            // Only one of focal length and FOV is applied;
+            // according to docs, focal length takes precedence.
             if (Node focal = camera["focal_length"]) {
-                // TODO
+                if (focal.IsScalar()) {
+                    float length = focal.as<float>(view->getFocalLength());
```
tallytalwar: you can use the yaml helper functions here, for consistency.

blair1618: I tried those, but found it to be less convenient than the accessors with fallbacks provided by YAML nodes.

tallytalwar: Also arguments for `as<>(const S& fallback)` are fallbacks if yaml node for which `as` is called is not valid.

In this case, since we are calling `focal.IsScalar()` on it, it will always be valid.

blair1618: It attempts to decode the node to the prescribed type and returns the fallback if either the node is invalid or the decoding fails: https://github.com/jbeder/yaml-cpp/blob/master/include/yaml-cpp/node/impl.h#L85-L99

I prefer this because it's more concise, doesn't involve exceptions, and works for any value that supports decoding.

tallytalwar: Ohh this is good. I did not look at `as_if` implementation.

core/src/scene/stops.cpp
```diff
@@ -181,6 +181,29 @@ auto Stops::Widths(const YAML::Node& _node, const MapProjection& _projection, co
...
+auto Stops::Numbers(const YAML::Node& node) -> Stops {
+    Stops stops;
+    if (!node.IsSequence()) { return stops; }
+
+    float lastKey = 0;
+
+    for (const auto frameNode : node) {
+        if (!frameNode.IsSequence() || frameNode.size() != 2) { continue; }
+
+        float key = frameNode[0].as<float>();
```
tallytalwar: Maybe use yamlhelper here :D.

blair1618: Yeah it's not terribly robust, it's just the way that every other `Stops` factory method is implemented. I intend to refactor this whole class in the not-too-distant future, so don't think of it as permanent :)

tallytalwar: Cool, in that case will merge this.

PR #648, opened 2016-03-28
Global props

### LINKED ISSUE #549: Scene file should allow 'scene properties' for run-time configuration changes
These properties would populate a `scene` object that is made available in JS contexts. The syntax for this still needs to be finalized.

tallytalwar:
Fixes #549 

- Scene global properties are parsed and applied after yaml parsing and before tangram parsing.
- This PR does not handle any dynamic updates of these properties, example setting of language preferences.

Todos:
- [x] global application to JS functions
- [x] Test/Implement Usage of global JS function inside a JS funtion
- [x] unit tests

### DIFF
--- a/core/resources/scene.yaml
+++ b/core/resources/scene.yaml
@@ -1,3 +1,6 @@
+global:
+    default_order: function() { return feature.sort_key; }
+
 scene:
     background:
         color: '#8db7d5'
@@ -300,7 +303,7 @@ layers:
         draw:
             polygons:
                 style: heightglow
-                order: function() { return feature.sort_key; }
+                order: global.default_order
                 color: [.65, .65, .63]
         # turn interactive feature selection on for buildings with names
         interactive:
--- a/core/resources/scene.yaml
+++ b/core/resources/scene.yaml
@@ -1,6 +1,16 @@
-scene:
-    background:
-        color: '#8db7d5'
+global:
+    labels: false
+    color:
+        colorA: red
+        colorC: [1, 1, 1, 1]
+        colorE: function(c) { return c; }
+        colorF: function() { return "cyan"; }
+    test:
+        background:
+            color: '#990000'
+
+scene: global.test
+
 cameras:
     iso-camera:
         # Manhattan
@@ -110,8 +120,13 @@ layers:
         draw:
             polygons:
                 order: 1
-                color: '#fffffa'
+                #color: '#fffffa'
+                color: global.color.colorA
                 interactive: true # currently ignored
+            lines:
+                order: 2
+                color: function() { return global.color.colorE("orange"); }
+                width: 5px
         green:
             filter: { kind: [park, graveyard, cemetery, forest, recreation_ground] }
             draw:
@@ -142,7 +157,7 @@ layers:
         draw:
             flatcolor:
                 order: 3
-                color: '#8db7d5'
+                color: global.color.colorF
 
     roads:
         data: { source: osm }
@@ -346,7 +361,7 @@ layers:
         draw:
             text:
                 interactive: true
-                visible: true
+                visible: global.labels
                 priority: 2
                 offset: [0, 8px]
                 transition:
@@ -365,7 +380,7 @@ layers:
                 text:
                     repeat_distance: 200px
                     repeat_group: road_labels
-                    visible: true
+                    visible: global.labels
                     #offset: [0px, 5px]
                     priority: 1
                     transition:
@@ -385,7 +400,7 @@ layers:
                     repeat_distance: 100px
                     repeat_group: roads
                     interactive: true
-                    visible: true
+                    visible: global.labels
                     priority: 3
                     transition:
                         [show, hide]:
--- a/core/resources/scene.yaml
+++ b/core/resources/scene.yaml
@@ -1,16 +1,6 @@
-global:
-    labels: false
-    color:
-        colorA: red
-        colorC: [1, 1, 1, 1]
-        colorE: function(c) { return c; }
-        colorF: function() { return "cyan"; }
-    test:
-        background:
-            color: '#990000'
-
-scene: global.test
-
+scene:
+    background:
+        color: '#8db7d5'
 cameras:
     iso-camera:
         # Manhattan
@@ -120,13 +110,8 @@ layers:
         draw:
             polygons:
                 order: 1
-                #color: '#fffffa'
-                color: global.color.colorA
+                color: '#fffffa'
                 interactive: true # currently ignored
-            lines:
-                order: 2
-                color: function() { return global.color.colorE("orange"); }
-                width: 5px
         green:
             filter: { kind: [park, graveyard, cemetery, forest, recreation_ground] }
             draw:
@@ -157,7 +142,7 @@ layers:
         draw:
             flatcolor:
                 order: 3
-                color: global.color.colorF
+                color: '#8db7d5'
 
     roads:
         data: { source: osm }
@@ -361,7 +346,7 @@ layers:
         draw:
             text:
                 interactive: true
-                visible: global.labels
+                visible: true
                 priority: 2
                 offset: [0, 8px]
                 transition:
@@ -380,7 +365,7 @@ layers:
                 text:
                     repeat_distance: 200px
                     repeat_group: road_labels
-                    visible: global.labels
+                    visible: true
                     #offset: [0px, 5px]
                     priority: 1
                     transition:
@@ -400,7 +385,7 @@ layers:
                     repeat_distance: 100px
                     repeat_group: roads
                     interactive: true
-                    visible: global.labels
+                    visible: true
                     priority: 3
                     transition:
                         [show, hide]:
--- a/core/src/scene/scene.h
+++ b/core/src/scene/scene.h
@@ -1,6 +1,7 @@
 #pragma once
 
 #include "util/color.h"
+#include "util/variant.h"
 #include <list>
 #include <memory>
 #include <string>
@@ -31,6 +32,10 @@ class Scene {
 public:
     Scene();
     ~Scene();
+
+    using ArrayValue = std::vector<double>;
+    using Value = variant<none_type, bool, double, std::string, ArrayValue>;
+    std::unordered_map<std::string, Value> m_globals;
 
     auto& view() { return m_view; }
     auto& dataSources() { return m_dataSources; };
--- a/core/src/scene/scene.h
+++ b/core/src/scene/scene.h
@@ -1,7 +1,6 @@
 #pragma once
 
 #include "util/color.h"
-#include "util/variant.h"
 #include <list>
 #include <memory>
 #include <string>
@@ -9,6 +8,7 @@
 #include <unordered_map>
 
 #include "glm/vec2.hpp"
+#include "yaml-cpp/yaml.h"
 
 namespace Tangram {
 
@@ -33,9 +33,6 @@ public:
     Scene();
     ~Scene();
 
-    using ArrayValue = std::vector<double>;
-    using Value = variant<none_type, bool, double, std::string, ArrayValue>;
-    std::unordered_map<std::string, Value> m_globals;
 
     auto& view() { return m_view; }
     auto& dataSources() { return m_dataSources; };
@@ -48,6 +45,7 @@ public:
     auto& stops() { return m_stops; }
     auto& background() { return m_background; }
     auto& fontContext() { return m_fontContext; }
+    auto& globals() { return m_globals; }
 
     const auto& dataSources() const { return m_dataSources; };
     const auto& layers() const { return m_layers; };
@@ -56,6 +54,7 @@ public:
     const auto& functions() const { return m_jsFunctions; };
     const auto& mapProjection() const { return m_mapProjection; };
     const auto& fontContext() const { return m_fontContext; }
+    const auto& globals() const { return m_globals; }
 
     const Style* findStyle(const std::string& _name) const;
     const Light* findLight(const std::string& _name) const;
@@ -88,6 +87,7 @@ private:
     std::vector<std::unique_ptr<Light>> m_lights;
     std::unordered_map<std::string, std::shared_ptr<Texture>> m_textures;
     std::unordered_map<std::string, std::shared_ptr<SpriteAtlas>> m_spriteAtlases;
+    std::unordered_map<std::string, YAML::Node> m_globals;
 
     // Container of all strings used in styling rules; these need to be
     // copied and compared frequently when applying styling, so rules use
--- a/core/src/scene/sceneLoader.cpp
+++ b/core/src/scene/sceneLoader.cpp
@@ -25,7 +25,6 @@
 #include "view/view.h"
 
 #include "csscolorparser.hpp"
-#include "yaml-cpp/yaml.h"
 
 #include <algorithm>
 #include <iterator>
@@ -74,28 +73,7 @@ void SceneLoader::applyGlobalProperties(Node& node, Scene& scene) {
             if (key.compare(0, 7, "global.") == 0) {
                 key.replace(0, 7, "");
                 key = std::regex_replace(key, std::regex("\\."), wildcard);
-                if (scene.m_globals.find(key) != scene.m_globals.end()) {
-                    auto& value = scene.m_globals[key];
-                    switch (value.which()) {
-                        case 1:
-                            node = value.get<bool>();
-                            break;
-                        case 2:
-                            node = value.get<double>();
-                            break;
-                        case 3:
-                            node = value.get<std::string>();
-                            break;
-                        case 4:
-                            for (auto& v : value.get<Scene::ArrayValue>()) {
-                                node.push_back(v);
-                            }
-                            break;
-                        case 0:
-                        default:
-                            break;
-                    }
-                }
+                node = scene.globals()[key];
             }
         }
         break;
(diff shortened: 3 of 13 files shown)


core/src/scene/scene.h
```diff
@@ -32,6 +33,10 @@ class Scene {
     Scene();
     ~Scene();
 
+    using ArrayValue = std::vector<double>;
+    using Value = variant<none_type, bool, double, std::string, ArrayValue>;
```
blair1618: This is not sufficient to encode the possible values of a global property entry. Elements of `global` can be arbitrary yaml, including sequences of strings and sequences of booleans. It seems much simpler and more natural to store the actual yaml node for each entry (or a pointer to it). 

Tangentially, I would recommend that we not introduce any more types called `Value` - even one type with a meaningless name is plenty for any given project ;)

core/src/scene/scene.h
```diff
@@ -32,6 +33,10 @@ class Scene {
...
+    std::unordered_map<std::string, Value> m_globals;
```
blair1618: I'm not clear why this is a public member. If it *needs* to be a public member, it should not be named with `m_`, otherwise it should be private like the other members.

core/src/scene/scene.h
tallytalwar: Agreed and Agreed, will update.

core/src/scene/scene.h
tallytalwar: Ohh! I meant it to be private. Forgot to put it back with accessor! Will update.

tallytalwar: In the current implementation, we actually do not need scene to store a reference of all the parsed globals. But we will need it once we add dynamic updates to scene!

core/src/scene/sceneLoader.cpp
```diff
@@ -64,6 +65,97 @@ void printFilters(const SceneLayer& layer, int indent){
     }
 };
 
+void SceneLoader::applyGlobalProperties(Node& node, Scene& scene) {
+    switch(node.Type()) {
+    case NodeType::Scalar:
+        {
+            std::string key = node.Scalar();
+            const std::string wildcard = "_$_";
```
blair1618: I think elsewhere we use `:` as a nested-value-delimiter. Think it would make sense to define a delimiter globally for this file?

tallytalwar: `_$_` is what JS code uses.

bcamper: yeah that makes sense, can change JS to use `:` as well

blair1618: Weird.. Can you link me?

That seems like a poor choice, since those characters aren't part of structural yaml syntax they seem more likely to be used in a member name than `:`.

bcamper: I am sorry

core/src/scene/sceneLoader.cpp
```diff
@@ -64,6 +65,97 @@ void printFilters(const SceneLayer& layer, int indent){
...
+            if (key.compare(0, 7, "global.") == 0) {
+                key.replace(0, 7, "");
+                key = std::regex_replace(key, std::regex("\\."), wildcard);
+                if (scene.m_globals.find(key) != scene.m_globals.end()) {
+                    auto& value = scene.m_globals[key];
+                    switch (value.which()) {
```
blair1618: This seems like a risky thing to switch on - if we add another type to the variant then all of these cases could break with no compile-time error. I think you can check the type with `is<Type>()` or something, but this becomes redundant if the map of globals just stores yaml nodes.

tallytalwar: Will get updated.. Switching the globals map to `String: YAML::Node` instead of `String: Variant_Value`.

core/src/scene/sceneLoader.cpp
```diff
@@ -64,6 +65,97 @@ void printFilters(const SceneLayer& layer, int indent){
...
+        break;
+    case NodeType::Sequence:
+        for (const auto& n : node) {
+            Node nextNode = n;
+            applyGlobalProperties(nextNode, scene);
+        }
+        break;
+    case NodeType::Map:
+        for (const auto& n : node) {
+            Node nextNode = node[n.first.Scalar()];
```
blair1618: Isn't this the same as `n.second.Scalar()`?

tallytalwar: yeah, `Node nextNode = n.second`.

core/src/scene/scene.h
tallytalwar: Fixed.

core/src/scene/scene.h
tallytalwar: Updated.

core/src/scene/sceneLoader.cpp
tallytalwar: ":" is THE delimiter now.

bcamper: Updated JS: https://github.com/tangrams/tangram/commit/03724319437093e90f07b7e0c833ab60a6c5561b

core/src/scene/scene.h
```diff
@@ -1,6 +1,7 @@
 #pragma once
 
 #include "util/color.h"
+#include "util/variant.h"
```
blair1618: Unused include?

tallytalwar: Fixed.

core/src/scene/sceneLoader.cpp
```diff
@@ -64,6 +64,56 @@ void printFilters(const SceneLayer& layer, int indent){
...
 };
 
+void SceneLoader::applyGlobalProperties(Node& node, Scene& scene) {
+    switch(node.Type()) {
+    case NodeType::Scalar:
+        {
+            std::string key = node.Scalar();
+            if (key.compare(0, 7, "global.") == 0) {
+                key.replace(0, 7, "");
+                key = std::regex_replace(key, std::regex("\\."), DELIMITER);
```
blair1618: Regex seems like a bit much for this, I think we can just use 
```std::replace(key.begin(), key.end(), '.', DELIMITER);```

File: core/src/scene/sceneLoader.cpp [hunk 8]
```diff
@@ -64,6 +64,56 @@ void printFilters(const SceneLayer& layer, int indent){
...
+                node = scene.globals()[key];
+            }
+        }
+        break;
+    case NodeType::Sequence:
+        for (const auto& n : node) {
+            Node nextNode = n;
```
blair1618: Why create another local variable here?

core/src/scene/sceneLoader.cpp
```diff
@@ -64,6 +64,56 @@ void printFilters(const SceneLayer& layer, int indent){
...
+            applyGlobalProperties(nextNode, scene);
+        }
+        break;
+    case NodeType::Map:
+        for (const auto& n : node) {
+            Node nextNode = n.second;
```
blair1618: Again, do we need another local here?

core/src/scene/sceneLoader.cpp
```diff
@@ -64,6 +64,56 @@ void printFilters(const SceneLayer& layer, int indent){
...
+    }
+}
+
+void SceneLoader::parseGlobals(const Node& node, Scene& scene, const std::string& key) {
+    switch (node.Type()) {
+    case NodeType::Scalar:
+    case NodeType::Sequence:
+        {
+            auto& g = scene.globals();
+            g[key] = YAML::Clone(node);
```
blair1618: I don't think `Clone` is necessary here, it's more efficient to assign the node directly.

core/src/scene/sceneLoader.cpp
tallytalwar: Aha, can fix that, previously was replace a "." with "_$_", hence `regex_replace` was required.

core/src/scene/sceneLoader.cpp
```diff
@@ -64,6 +64,56 @@ void printFilters(const SceneLayer& layer, int indent){
...
+            break;
+        }
+    case NodeType::Map:
+        for (const auto& g : node) {
+            std::string value = g.first.Scalar();
+            Node global = node[value];
+            std::string mapKey = (key.size() == 0) ? value : (key + DELIMITER + value);
+            parseGlobals(global, scene, mapKey);
+        }
```
blair1618: For map nodes we need to create globals for not only the leaf nodes, but for all of the mappings themselves. For example, with the global block you've added for demonstration, it's valid for the scene to reference `global.colors` but there is no entry for it in `scene.globals()`

tallytalwar: Aha, nice catch. Will fix this.

blair1618:
Global properties should be accessible in JS functions as well, just like feature properties. 

We should be able to follow a pattern similar to what's done for feature property access in `StyleContext`. There are already some members named "globals" in `StyleContext`, so we may need to do some renaming.

tallytalwar:
Ohh, yes I had a todo marked for that. Forgot about it.

Can also do a string replacement for that before JSFunction gets parsed?

blair1618:
String replacement within JS functions seems rather risky. One example: a function could declare a local variable called "global" that shadows the global "global", but a string replacer wouldn't know that.

core/src/scene/sceneLoader.cpp
tallytalwar: Need to get rid of the const! Refer the function args.

tallytalwar:
Its `global.` what we have to string replace.

core/src/scene/sceneLoader.cpp
blair1618: `for (auto n : node) { applyGlobalProperties(n, scene); }`

blair1618:
```javascript
function() {
  var global = { colors: "I like them" };
  return global.colors;
}
```

bcamper:
Agree with @blair1618, we should avoid modifying the user-defined JS function body, and expose globals via enclosing context. There is still a risk of local variable shadowing, but that's true of any scope we expose to user functions.

tallytalwar:
Sure, that was a bad suggestion. Will incorporate it in the `StyleContext` globals!

core/src/scene/sceneLoader.cpp
tallytalwar: Fixed.

core/src/scene/sceneLoader.cpp
tallytalwar: Fixed.

core/src/scene/sceneLoader.cpp
tallytalwar: Fixed.

core/src/scene/sceneLoader.cpp
tallytalwar: Fixed.

blair1618:
https://github.com/tangrams/tangram-es/pull/648/commits/c8e646bff707b7f0eff6794f9030b02aeada5cac probably shouldn't be in this PR. You can push that change directly to master and remove it here.

tallytalwar:
c8e646bff707b7f0eff6794f9030b02aeada5cac is in master.

blair1618:
So we can address members of map nodes in the `global` block with "dot notation", but is there a way to access sequence elements? Should we be able to do something like `property: global.colors[0]`?

cc @bcamper

tallytalwar:
I still need to add unit tests, but all the functionality is in place now.

@hjanetzek do you want to give this a look, since you are the master of duk API usage! I also refactored a bunch of function evaluation code for reusability.

core/src/scene/scene.h
```diff
@@ -83,6 +87,7 @@ class Scene {
     std::vector<std::unique_ptr<Light>> m_lights;
     std::unordered_map<std::string, std::shared_ptr<Texture>> m_textures;
     std::unordered_map<std::string, std::shared_ptr<SpriteAtlas>> m_spriteAtlases;
+    std::unordered_map<std::string, YAML::Node> m_globals;
```
tallytalwar: Note to self,  use `fastmap`!!

core/resources/scene.yaml
```diff
@@ -110,8 +122,13 @@ layers:
         draw:
             polygons:
                 order: 1
-                color: '#fffffa'
+                #color: '#fffffa'
+                color: global.color.colorA
                 interactive: true # currently ignored
+            lines:
+                order: 2
+                color: function() { return global.color.colorE; }
```
blair1618: Weird.. this isn't how JavaScript works. This means that the value for `color` i

tallytalwar: Hmm.. I see what you are saying there. (Sorry for the lack of JS knowledge!!).

Will check this.

bcamper: @blair1618 is right about how it should behave, however looking at the current JS flow, I realized this would not work, because the global was being replaced as a string before function compilation happens, so if `global.color.colorE` is a function, then `color: global.color.colorE` would work, but `color: function() { return global.color.colorE(); } would not (because `color: global.color.colorE` would still be a string). 

I'm updating JS behavior to enable this case, as I think it's the closest to what a typical users' expectations would be. It also enables some useful behavior, like color conversion functions, as global properties. These updates will be part of the JS 0.7 release.

tallytalwar: Ok, so whats happening right now is, when creating the global JS object for duk's context, it evaluates any JS function it finds in any of the global properties.

And when evaluating duk context for styleParams or filter it simply replaces the evaluated global value.

Even though `global.prop.propA()` is the correct way of calling a JS function, in our case should the usage be to simply put the evaluated value where ever a global property is accessed?
Also tried in JS, it doesn't seem to work there!

blair1618: I'm a little confused by the last comment, but I'll explain how I'm thinking about this now: 

When we evaluate the global properties block at scene load time, we'll create a global object called `global` in the duktape JS context. `global` will be populated with all of the members of the corresponding YAML block. While we are adding these members, we'll take any string that looks like a function and instead of giving `global` a JS string for that member, we'll give it a JS function (as evaluated by duktape). 

This supports things that can't be done with simple substitution, e.g.:
```yaml
global:
  convert_color: function(hsv) { ... }
...
  color: function() { return global.convert_color(feature.color); }
```

tallytalwar: Fixed. Updated tests also.

core/src/scene/styleContext.cpp
```diff
@@ -82,6 +82,77 @@ StyleContext::~StyleContext() {
...
+                    duk_push_string(m_ctx, "");
+                    if (duk_pcompile(m_ctx, DUK_COMPILE_FUNCTION) == 0) { // compile function
+                        duk_put_prop(m_ctx, -3); // put {key: function()}
+                    } else {
+                        LOGW("Compile failed: %s\n%s\n---",
+                             duk_safe_to_string(m_ctx, -1),
+                             nodeValue.c_str());
+                        duk_pop(m_ctx); //pop error
+                        duk_pop(m_ctx); //pop key
+                    }
```
blair1618: Thank you for the comments, I can totally grok this!

Here's a small dilemma: For strings that "look like" a JS function, we try to compile them and if that doesn't work we don't add a value for that key at all. But what if I'm a very crafty user and I want a global property that is the string "function"? Right now we would try to compile it, fail, and then not create that property at all. 

On the one hand, we could trust our users and infer that if the string doesn't compile as a function, it should just be a string property. 

On the other hand, it might be more useful to provide stronger feedback if something that "looks like" a function doesn't compile. 

cc @bcamper What does JS do in a case like this?

tallytalwar: One small solution to this will be match against "function(". 

In that case, we can also do the same when parsing yaml for "function" in style params and filters.

bcamper: For strings that "look like" a JS function, Tangram JS tries to compile it, and if the compilation fails, it just reverts to retaining the original string value. This is done in a `try/catch` block.

tallytalwar: We can surely do what you mentioned @bcamper.

core/src/scene/styleContext.cpp
```diff
@@ -330,37 +424,18 @@ bool StyleContext::parseStyleResult(StyleParamKey _key, StyleParam::Value& _val)
     } else if (duk_is_null_or_undefined(m_ctx, -1)) {
         // Ignore setting value
         LOGD("duk evaluates JS method to null or undefined.\n");
+    } else if (duk_is_function(m_ctx, -1)) {
+        if (duk_pcall(m_ctx, 0) != 0) {
+            LOGE("EvalFilterFn: %s", duk_safe_to_string(m_ctx, -1));
+            duk_pop(m_ctx);
+        } else {
+            parseStyleResult(_key, _val);
+        }
```
blair1618: This case probably shouldn't be allowed. This means that if a parameter is a function that returns *another* function, that returned function will be evaluated for another result. I don't believe this is how tangram-js would handle this situation and I don't see a reason for us to add it.

tallytalwar: Ohh yeah, I forgot to take that out. Was trying out stuff.

core/src/scene/styleContext.cpp
tallytalwar: Fixed

blair1618:
This looks good to me now. Nice work! 

This is good to merge after dropping the test scene changes. You could leave in some useful globals just so we test them regularly, like:
```
global:
  default_order: 'function() { return feature.sort_key; }'
```

Issue #598, opened 2016-03-04
Light position param in meters/pixels parsing


karimnaaji:
tangram-es should be able to parse light position params in `meters`/`pixels`:

Something like,
```yaml
lights:
    light:
        position: [0, 0m, 100px]
```
gives currently some parsing errors.

https://github.com/tangrams/tangram-docs/blob/gh-pages/pages/lights.md#point-light-properties

blair1618:
Resolved by #651

PR #651, opened 2016-03-30
Light position params

karimnaaji:
#598

### DIFF
--- a/core/src/scene/light.h
+++ b/core/src/scene/light.h
@@ -27,6 +27,15 @@ enum class LightOrigin {
     world
 };
 
+static std::string lightOriginString(LightOrigin origin) {
+    switch(origin) {
+        case LightOrigin::camera: return "camera";
+        case LightOrigin::ground: return "ground";
+        case LightOrigin::world: return "world";
+        default: return "undefined";
+    }
+}
+
 struct LightUniforms {
     LightUniforms(ShaderProgram& _shader, const std::string& _name)
         : shader(_shader),
@@ -67,6 +76,9 @@ public:
 
     /*  Set the origin relative to which this light will be positioned */
     void setOrigin(LightOrigin origin);
+
+    /*  Get the origin relative to which this light is positioned */
+    LightOrigin getOrigin() const { return m_origin; }
 
     /*  Get the instances light name defined on the shader */
     virtual std::string getInstanceName();
--- a/core/src/scene/light.h
+++ b/core/src/scene/light.h
@@ -27,7 +27,7 @@ enum class LightOrigin {
     world
 };
 
-static std::string lightOriginString(LightOrigin origin) {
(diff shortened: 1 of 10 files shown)


core/src/scene/light.h
```diff
@@ -27,6 +27,15 @@ enum class LightOrigin {
     world
 };
 
+static std::string lightOriginString(LightOrigin origin) {
```
tallytalwar: You can make this a member function of the `LightOrigin` enum class. Something like `toString`.

blair1618: Alas, I really wish c++ typed enums allowed this - Java enums do and it's very very convenient.

blair1618:
Ok, we do need to parse these units correctly, this approach should work.. however I do feel that our treatment of value-unit parameters is in need of some love. We do unit parsing in too many ways in too many places (where n > 1 is too many). There are some details that I'll comment on in-line, but this approach will do for now with the understanding that we can (should) revisit units more holistically in the near future.

core/src/scene/styleParam.h
```diff
@@ -60,6 +60,28 @@ constexpr size_t StyleParamKeySize = static_cast<size_t>(StyleParamKey::NUM_ELEM
...
+        case Unit::meter: return "meter";
+        case Unit::seconds: return "seconds";
+        default: return "undefined";
+    }
+}
+
+template <typename T, int size>
+struct UnitVec {
+    T value = T(0.0);
+    Unit units[size];
```
blair1618: I wonder if we could do `Unit units[sizeof(T)/sizeof(T[0])];`

karimnaaji:
Yep that's what I noticed too; those unit conversion really need a refactoring after we all get that working.

core/src/scene/styleParam.h
karimnaaji: Will give that a try.

blair1618:
*poke* @karimnaaji this still has conflicts

karimnaaji:
Yes, rebasing this in a few.

core/src/scene/styleParam.h
karimnaaji: Unfortunately, this does not work.. :/

blair1618: *facepalm* my bad, it makes no sense to index into a type name haha. `sizeof(value)/sizeof(value[0])` should work though. Then we can remove the `size` template parameter.

karimnaaji: Even in that case it wound't be a `constexpr` right?

blair1618: It actually is! At least in C++14. The values are all definitely known at compile-time, I think arithmetic might not have been allowed until C++14.

blair1618:
Lots of warnings about `unitString` as an unused function. It's a natural thing to have, but maybe we can comment it for now :\

PR #652, opened 2016-03-30
Make Android map initialization asynchronous

### LINKED ISSUE #558: Android SDK should allow asynchronous initialization
Currently the only way to initialize the renderer on Android is with a synchronous call to the `MapController` constructor, which can take a long time (on the order of a second) for large scenes on low-power devices. A client application could currently make initialization asynchronous by calling the constructor on a background thread, but it would be more convenient to library users if we provided a built-in way to initialize the renderer asynchronously and run a callback on completion.

blair1618:
Resolves #558 

With this change, the primary means of initializing a tangram map on android is with the method `getMapAsync()` on `MapView`. This is very similar to the way a Google Map is initialized, the main difference being that our initializer takes a string argument to specify the scene file that should be loaded. There are a few advantages to this:

 - Loading large scene files will no longer block the UI thread
 - Client apps no longer need to provide an `Activity` or `Context`, (we get a `Context` from the `MapView` itself)
 - Similarity to a widely-adopted API should make it easier for new users to understand. 

This PR also _heavily_ refactors the `MainActivity` in the Android demo. It was rather messy and I had to do a bit of cleaning for the new async initializer anyway.

### DIFF
--- a/android/tangram/src/com/mapzen/tangram/MapView.java
+++ b/android/tangram/src/com/mapzen/tangram/MapView.java
@@ -51,6 +51,10 @@ public class MapView extends FrameLayout {
 
         final Context context = getContext();
 
+        if (getMapTask != null) {
+            getMapTask.cancel(true);
+        }
+
         getMapTask = new AsyncTask<Void, Void, MapController>() {
 
             @Override
--- a/android/tangram/src/com/mapzen/tangram/MapView.java
+++ b/android/tangram/src/com/mapzen/tangram/MapView.java
@@ -28,19 +28,26 @@ public class MapView extends FrameLayout {
 
     }
 
+    /**
+     * Interface for receiving a {@code MapController} once it is ready to be used
+     */
     public interface OnMapReadyCallback {
 
+        /**
+         * Called when the map is ready to be used
+         * @param mapController A non-null {@code MapController} instance for this {@code MapView}
+         */
         void onMapReady(MapController mapController);
 
     }
 
     /**
-     * Construct a MapController asynchronously; may only be called from the UI thread
-     * @param listener The object to receive the resulting MapController in a callback;
+     * Construct a {@code MapController} asynchronously; may only be called from the UI thread
+     * @param callback The object to receive the resulting MapController in a callback;
      * the callback will be made on the UI thread
      * @param sceneFilePath Location of the YAML scene file within the asset bundle
      */
-    public void getMapAsync(final OnMapReadyCallback listener, final String sceneFilePath) {
+    public void getMapAsync(final OnMapReadyCallback callback, final String sceneFilePath) {
 
         final Context context = getContext();
 
@@ -55,7 +62,7 @@ public class MapView extends FrameLayout {
             protected void onPostExecute(MapController map) {
                 map.setView(glSurfaceView);
                 addView(glSurfaceView);
-                listener.onMapReady(map);
+                callback.onMapReady(map);
             }
 
         }.execute();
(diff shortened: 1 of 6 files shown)


blair1618:
@ecgreb How's this look to you?

ecgreb:
What is the longest amount of time it can take to initialize the MapController?
What happens if there is a configuration change during this time?

blair1618:
Good question! There's no well-defined time limit right now. In practice, loading a large scene file like the one in eraser-map can take as long as 2 seconds on older phones. By "configuration change" do you mean something like changing device orientation? I am not totally sure, but I can try it out.

ecgreb:
Yeah exactly. What if the user opens the app then immediately rotates the device?

Given the `MapView` will be destroyed and recreated I'm not sure what the result will be when the `AsyncTask` completes (or the new one is started).

blair1618:
The current result is that `OnMapReady` is called twice, with two different `MapController` instances. This actually doesn't cause any problems in our demo app, but I can imagine how it could be a pain for other situations. 

We could hold a reference to the `AsyncTask` used in `getMapAsync` and cancel it if `getMapAsync` is called again before the task completes.

blair1618:
Wait, no actually we can't do that because the `MapView` will be a new instance as well. Hmm. Is there something we can query on a `View` to determine whether it has been "destroyed"?

tallytalwar:
@blair1618 can we maintain a state internally in our core which increments a counter if config changes in between and asyncMap has not finished.

blair1618:
@tallytalwar The core library has even less knowledge about application lifecycle than the view does, I don't think that will help. 

The Google Maps MapView requires that activity lifecycle listeners are forwarded from the Activity or Fragment that owns it: https://developers.google.com/android/reference/com/google/android/gms/maps/MapView

That's a possible solution. If our `MapView` received the `OnDestroy()` call from the application, we could cancel any pending tasks. This would result in `MapView` requiring more code to integrate than it currently does, but this may not be a real "cost" if it's already required for other libraries.

ecgreb:
Good catch on forwarding the lifecycle events. I guess that is something else the `MapFragment` does for you automatically.

This does seem like a clean way to solve the problem and like you said it is in line with what other libraries already require.

blair1618:
Added lifecycle listeners to `MapView`, this solves the issue of repeated calls to `getMapAsync`.

android/tangram/src/com/mapzen/tangram/MapView.java
```diff
@@ -22,13 +28,93 @@ public MapView(Context context, AttributeSet attrs) {
...
+        getMapTask = new AsyncTask<Void, Void, MapController>() {
+
+            @Override
+            protected MapController doInBackground(Void... params) {
+                return new MapController(context, sceneFilePath);
+            }
+
+            @Override
+            protected void onPostExecute(MapController map) {
+                map.setView(glSurfaceView);
```
ecgreb: Should we add some checks here to ensure `map`, `glSurfaceView`, and `callback` are not null? I know they shouldn't be but crazier thinks have been known to happen (especially during device rotation).

blair1618: Hm. If any of those are somehow null at this point, there's been some kind of error by the client application in managing the application lifecycle. The correct response to that is to throw an exception, which is what will happen in the methods here. 

We could create our own, slightly more informative exception here to denote that an invalid state has been reached.

ecgreb: I guess I just don't have total confidence that `onPostExecute(...)` won't fire even if the task has been properly canceled in `onDestroy()`. Kind of an edge case but its been known to happen.

blair1618: Yeah I do want to be robust to unexpected-but-correct usage, I just want to make sure we also make it easy to diagnose incorrect usage. 

I will continue to expect that `AsyncTask` behaves as documented unless we find a demonstrated case to the contrary, but I can also add additional safeguards here. I'll post again after trying some changes.

blair1618:
Now that I've got our travis android config happy again, I feel like this is pretty good to merge. It's not perfect yet but I want to keep these changes for now. 

The most immediate change I can foresee would be to replace the string argument to `getMapAsync` with a configuration object, `SceneConfiguration` or something. That would let us build a scene configuration with things like an API key, language settings, anything we know *a priori* and then get a fully-configured map object when the async call completes.

ecgreb:
:shipit:

Issue #592, opened 2016-03-03
Depth test function should use GL_LESS


blair1618:
Testing with `GL_LESS` instead of `GL_LEQUAL` should result in many fewer fragments being processed with no change in visual quality. 

The only cases where this could change visual results would be overlapping geometries set to the same `order`, but this case is effectively a user error and those geometries will have visual artifacts from z-fighting anyway.

blair1618:
Hey @karimnaaji can you try this out? We'll want to survey the results with a few stylesheets to make sure there aren't any unforeseen consequences.

karimnaaji:
Yep

blair1618:
Resolved by #656

PR #656, opened 2016-03-31
Use GL_LESS for depth function

karimnaaji:
https://github.com/tangrams/tangram-es/issues/592

See https://github.com/tangrams/tangram/pull/283 for details

### DIFF
--- a/core/src/gl/renderState.cpp
+++ b/core/src/gl/renderState.cpp
@@ -58,7 +58,7 @@ namespace RenderState {
         depthWrite.init(GL_TRUE);
 
         glDisable(GL_STENCIL_TEST);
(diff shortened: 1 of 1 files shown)


tallytalwar:
@karimnaaji did you finish testing this with a bunch of our styles? And what's the word?

karimnaaji:
@tallytalwar yes. Word is mostly what's in https://github.com/tangrams/tangram/pull/283. There's not any differences on render time and visually, and the renderer doesn't need to use `GL_LEQUAL` for what we do; it would have been useful for multi-pass alpha blended rendering, but the blending modes we have that require alpha blending don't rely on z-test.

PR #657, opened 2016-04-01
Add vanishing_point perspective camera property

### LINKED ISSUE #541: Camera "vanishing_point" property should set the vanishing point for perspective cameras
It's in the docs, but I'm honestly kind of unsure what effect this is supposed to have on cameras with arbitrary orientation.

https://github.com/tangrams/tangram-es/blob/master/core/src/scene/sceneLoader.cpp#L969

https://github.com/tangrams/tangram-docs/blob/gh-pages/pages/cameras.md#vanishing_point

blair1618:
Resolves #541 

`vanishing_point` now has the same behavior as it does in tangram-js for top-down views. There is a known issue that for very large `vanishing_point` vectors, a top-down view may show reduced-level-of-detail tiles. This is not the intended LOD behavior, but I think it's an acceptably small degradation of quality, given the rarity of its occurrence.

### DIFF
--- a/core/src/scene/sceneLoader.cpp
+++ b/core/src/scene/sceneLoader.cpp
@@ -744,7 +744,12 @@ void SceneLoader::loadCameras(Node _cameras, Scene& _scene) {
             }
 
             if (Node vanishing = camera["vanishing_point"]) {
-                // TODO
+                if (vanishing.IsSequence() && vanishing.size() >= 2) {
+                    // Values are pixels, unit strings are ignored.
+                    float x = std::stof(vanishing[0].Scalar());
+                    float y = std::stof(vanishing[1].Scalar());
+                    view->setVanishingPoint(x, y);
+                }
             }
         } else if (type == "isometric") {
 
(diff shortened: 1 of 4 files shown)


core/src/scene/sceneLoader.cpp
```diff
@@ -744,7 +744,12 @@ void SceneLoader::loadCameras(Node _cameras, Scene& _scene) {
             }
 
             if (Node vanishing = camera["vanishing_point"]) {
-                // TODO
+                if (vanishing.IsSequence() && vanishing.size() >= 2) {
```
tallytalwar: Should we be checking for strictness of `size == 2`?

blair1618: Probably not super important, but I'm guessing that JS doesn't check the array length is == 2.

bcamper: Right, I think it's fine as long as XY coords are available, we're lenient
for cases like this elsewhere.

PR #659, opened 2016-04-01
Polyline join

### LINKED ISSUE #550: Polygons drawn with line styles should not have caps at first vertex

tallytalwar:
Fixes #550

Todo:
- [x] test more to resolve some edge cases.

### DIFF
--- a/core/src/style/polylineStyle.cpp
+++ b/core/src/style/polylineStyle.cpp
@@ -111,6 +111,7 @@ public:
         } fill, stroke;
 
         bool keepTileEdges = false;
+        bool closedPolygon = false;
         bool outlineOn = false;
         bool lineOn = true;
     };
@@ -339,6 +340,8 @@ void PolylineStyleBuilder<V>::addFeature(const Feature& _feat, const DrawRule& _
             addMesh(line, params);
         }
     } else {
+        params.closedPolygon = true;
+
         for (auto& polygon : _feat.polygons) {
             for (const auto& line : polygon) {
                 addMesh(line, params);
@@ -377,6 +380,7 @@ void PolylineStyleBuilder<V>::addMesh(const Line& _line, const Parameters& _para
     m_builder.join = _params.fill.join;
     m_builder.miterLimit = _params.fill.miterLimit;
     m_builder.keepTileEdges = _params.keepTileEdges;
+    m_builder.closedPolygon = _params.closedPolygon;
 
     if (_params.lineOn) { buildLine(_line, _params.fill, m_meshData[0]); }
 
--- a/core/src/util/builders.cpp
+++ b/core/src/util/builders.cpp
@@ -288,12 +288,17 @@ bool isOutsideTile(const glm::vec3& _a, const glm::vec3& _b) {
     return false;
 }
 
-void buildPolyLineSegment(const Line& _line, PolyLineBuilder& _ctx) {
+void buildPolyLineSegment(const Line& _line, PolyLineBuilder& _ctx,
+                          size_t startIndex, size_t endIndex, bool endCap = true) {
 
-    int lineSize = (int)_line.size();
+    size_t origLineSize = _line.size();
+    int lineSize = (int)((endIndex > startIndex) ?
+                   (endIndex - startIndex) :
+                   (origLineSize - startIndex + endIndex));
     if (lineSize < 2) { return; }
 
-    glm::vec3 coordPrev(_line[0]), coordCurr(_line[0]), coordNext(_line[1]);
+    glm::vec3 coordCurr(_line[startIndex]);
+    glm::vec3 coordNext(_line[(startIndex + 1) % origLineSize]);
     glm::vec2 normPrev, normNext, miterVec;
 
     int cornersOnCap = (int)_ctx.cap;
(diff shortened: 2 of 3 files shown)


tallytalwar:
Looks mostly good. Have to test some more to resolve some edge cases with tile edges!

core/src/style/polylineStyle.cpp
```diff
@@ -339,6 +340,8 @@ void PolylineStyleBuilder<V>::addFeature(const Feature& _feat, const DrawRule& _
             addMesh(line, params);
         }
     } else {
+        params.closedPolygon = true;
```
blair1618: Is this necessary? I think it is implicit in the input data whether a linestring is closed (i.e. if the first point is the same as the last point, it is closed).

core/src/util/builders.cpp
```diff
@@ -288,103 +288,123 @@ bool isOutsideTile(const glm::vec3& _a, const glm::vec3& _b) {
     return false;
 }
 
-void buildPolyLineSegment(const Line& _line, PolyLineBuilder& _ctx) {
+glm::vec2 processPoint(const Point& coordCurr, const Point& coordNext, const glm::vec2& normPrev,
```
blair1618: It's not at all clear what this function does or what it returns. Naming is hard, so keeping things inline where they are used can be a good strategy to minimize the amount of naming that one has to do :)

core/src/util/builders.cpp
```diff
@@ -288,103 +288,123 @@ bool isOutsideTile(const glm::vec3& _a, const glm::vec3& _b) {
...
+                       int trianglesOnJoin, int uv_v, PolyLineBuilder& _ctx, bool first = true) {
```
blair1618: `uv_v` is almost certainly supposed to be a `float` here.

blair1618:
I'm unclear why `buildPolyLineSegment` has different arguments now - it should not need to behave dramatically differently. The one new case it needs to handle is a closed loop, which it can infer and act on by simply checking whether the first and last input points are the same. Logic for tile edge cases is handled by `buildPolyLine`.

core/src/style/polylineStyle.cpp
blair1618: Resolved IRL: A "closed polygon" flag is used in the JS line building code and is set according to the calling function, not the data itself. So if a "line" feature happened to have the same first and last points, JS would still build two caps (as opposed to inferring the line to be closed). Therefore we *should* use a flag like this to match the behavior in JS ... for now.

core/src/util/builders.cpp
tallytalwar: Will either try to rename in properly or stick it back inline!

core/src/util/builders.cpp
tallytalwar: yup, my bad (all the calls have floats though)! Will fix.

tallytalwar:
Updated, with cleaner logic.

This also cleans out the previous `buildPolyLine` to not instantiate temporary Line vectors. Instead uses indices (wrapped) in the original Line Geometry.

core/src/util/builders.cpp
```diff
@@ -288,12 +288,20 @@ bool isOutsideTile(const glm::vec3& _a, const glm::vec3& _b) {
...
+void buildPolyLineSegment(const Line& _line, PolyLineBuilder& _ctx,
+                          size_t startIndex, size_t endIndex, bool endCap = true) {
 
-    int lineSize = (int)_line.size();
+    size_t origLineSize = _line.size();
+
+    // endIndex/startIndex could be wrapped values, calculate lineSize accordingly
+    int lineSize = (int)((endIndex > startIndex) ?
+                   (endIndex - startIndex) :
+                   (origLineSize - startIndex + endIndex));
```
blair1618: Ahh I see what's going on here, wrapping indices can make the arithmetic tricky. Do you think it would be cleaner to use a while loop over the intermediate points instead?

tallytalwar: While loop can be used but still to calculate `v` value, tricky arithmetic needs to be done. 
I do not see much difference between using `while` vs `for` here.

P.S. BTW its not that tricky .. just a modulo operator :D

core/src/util/builders.cpp
```diff
@@ -384,32 +396,50 @@ void buildPolyLineSegment(const Line& _line, PolyLineBuilder& _ctx) {
...
             }
         }
 
-        Line line = Line(&_line[cut], &_line[_line.size()]);
-        buildPolyLineSegment(line, _ctx);
+        if (_ctx.closedPolygon) {
+            if (cut == 0) {
+                // no tile edge cuts!
+                // loop and close the polygon with no endcaps
+                buildPolyLineSegment(_line, _ctx, 0, lineSize+2, false);
```
blair1618: Looking at this `lineSize+2`: so we build and draw the first line segment twice? I guess it's not a *huge* waste, but maybe we should leave a note here to remind us for future refactoring.

tallytalwar: Discussed in IRL: This extra line segment will not be of huge waste, and GL pipeline will be culling this anyhow!

blair1618:
Looks good to me. It might be nice to squash these commits so that the messages have a bit more context.

tallytalwar:
Updated comments and squashed some irrelevant commits.

PR #662, opened 2016-04-04
Fixes #661

### LINKED ISSUE #661: Bad Polygon extrusion geometry
screen shot 2016-04-04 at 5 42 52 pm

Looks like the culprit here is: https://github.com/tangrams/tangram-es/blob/master/core/src/util/builders.cpp#L139-L143

Whats happening here is part of the vertices are added to the buffers, I think we should throw out the entire extruded set of vertices for that polygon when either one of the extrusion results in a nan normal.

tallytalwar:
- Removes unused sizeHint
- Decrease vertex count in context, when vertices do not end up getting added because of nan normals!

### DIFF
--- a/core/src/util/builders.cpp
+++ b/core/src/util/builders.cpp
@@ -64,7 +64,6 @@ void Builders::buildPolygon(const Polygon& _polygon, float _height, PolygonBuild
 
     uint16_t vertexDataOffset = _ctx.numVertices;
     _ctx.numVertices += sumVertices;
-    _ctx.sizeHint(_ctx.numVertices);
 
     size_t ring = 0;
     size_t offset = 0;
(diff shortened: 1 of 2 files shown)


core/src/util/builders.cpp
```diff
@@ -117,12 +116,10 @@ void Builders::buildPolygonExtrusion(const Polygon& _polygon, float _minHeight,
         size_t lineSize = line.size();
         sumIndices += lineSize * 6;
         sumVertices += (lineSize - 1) * 4;
-
-        _ctx.numVertices = sumVertices;
     }
```
blair1618: I would suggest that instead of adjusting a pre-computed vertex count, we just don't pre-compute that number in the first place. It used to make sense to predict the vertex count when we were trying to hint buffer sizes, but now there's no reason for it. We can simply count vertices as they're added and set the context vertex count to that number after we're done.

tallytalwar: Sure, I wanted to get the fix in asap before I left for the day!

Will update this PR to remove any size pre-computation.

tallytalwar: We actually do need `numVertices` in `Builder` context. Reason being, we will have to count the vertices everytime, to make sure the vertex offsets are correct.

Here `numVertices` is not used for `std::vector::reserve` but to determine next vertexOffsets for indices.

blair1618:
Let's just merge this as-is and I'll make my nit-picky changes later haha

Issue #663, opened 2016-04-04
Data source configuration should include setting URL parameters


blair1618:
```yaml
sources:
   osm:
      type: TopoJSON
      url: http://vector.mapzen.com/osm/all/{z}/{x}/{y}.topojson
      url_params:
         api_key: vector-tiles-HqUVidw
```

Tangram-JS implementation: https://github.com/tangrams/tangram/pull/204

PR #664, opened 2016-04-06
Data source configurations can specify url parameters

blair1618:
Resolves https://github.com/tangrams/tangram-es/issues/663

### DIFF
--- a/core/resources/scene.yaml
+++ b/core/resources/scene.yaml
@@ -64,6 +64,8 @@ sources:
         type: MVT
         url:  https://vector.mapzen.com/osm/all/{z}/{x}/{y}.mvt
         max_zoom: 18
+        url_params:
+            api_key: vector-tiles-tyHL4AY
 
 layers:
     touch:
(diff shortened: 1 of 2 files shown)


bcamper:
URL params need to be inserted *after* the base URL, but *before* the `#` (if one is already present in the URL), and take into account possibly existing parameters:

Examples where `url_params: { a: 1, b: 2 }`:

- `http://baseurl.com/` => `http://baseurl.com/?a=1&b=2`
- `http://baseurl.com/?c=3&d=4` => `http://baseurl.com/?c=3&d=4&a=1&b=2`
- `http://baseurl.com/#lat/lng/z` => `http://baseurl.com/?a=1&b=2#lat/lng/z`
- `http://baseurl.com/?c=3&d=4#lat/lng/z` => `http://baseurl.com/?c=3&d=4&a=1&b=2#lat/lng/z`

See https://github.com/tangrams/tangram/pull/204/files#diff-be73193a7de1c86c1903f5cf92321d2aR84.

blair1618:
Huh, well I hadn't foreseen that someone would use a hash in a data source URL but I guess it's not impossible. I'll update. 

Just to check: trailing ampersands and duplicated parameters are generally permissible, yes?

bcamper:
Using a hash in a URL data source template is unlikely but valid, so I think it should not break.

Trailing ampersands shouldn't be a problem as far as I know, and duplicated parameters I explicitly punted on for the JS implementation because it's not rational to put in both the template and the `url_params`, and can be avoided by the scene file author.

blair1618:
Updated! I was tempted to make a new utility file for URL functions, but I'm trying to be diligent about not abstracting anything until I use it in at least two places.

PR #666, opened 2016-04-08
Refactor Android feature pick callbacks

blair1618:
- Rename `FeatureTouchListener` and `onTouch` to `FeaturePickListener` and `onFeaturePick`
- Simplify callback cycle; if a listener is provided, the native function makes the callback directly


tallytalwar:
LGTM. Merging.

PR #667, opened 2016-04-08
Consistently prefix native methods with 'native'

blair1618:
This is a pretty typical practice for code that uses JNI. For us it means that 1) we can immediately tell when JNI is and is not being invoked in our Java classes and 2) we'll have less trouble with method name collisions (e.g. `setMapPosition` could just be called `setPosition` now).


tallytalwar:
LGTM.

Merging #666, caused this to have conflicts. Once resolved, will merge.

PR #670, opened 2016-04-08
Run-time scene properties update

karimnaaji:
Rebase of #658

### DIFF
--- a/core/src/scene/scene.cpp
+++ b/core/src/scene/scene.cpp
@@ -92,7 +92,7 @@ void Scene::setComponent(std::string componentPath, std::string value) {
         {"lights", StyleComponent::lights},
         {"textures", StyleComponent::textures},
         {"styles", StyleComponent::styles},
-        {"source", StyleComponent::sources},
+        {"sources", StyleComponent::sources},
         {"layers", StyleComponent::layers}
     };
 
--- a/core/src/scene/scene.cpp
+++ b/core/src/scene/scene.cpp
@@ -26,7 +26,7 @@ Scene::Scene(std::string path) : id(s_serial++), m_path(path) {
     m_mapProjection.reset(new MercatorProjection());
 }
 
-Scene::Scene(std::string path, std::map<std::string, UserDefinedSceneValue> userDefined) :
+Scene::Scene(std::string path, std::vector<UserDefinedSceneValuePair> userDefined) :
     Scene(path)
 {
     m_userDefinedValues = userDefined;
@@ -80,9 +80,9 @@ bool Scene::texture(const std::string& textureName, std::shared_ptr<Texture>& te
     return true;
 }
 
-void Scene::setComponent(std::string componentPath, std::string value) {
+void Scene::queueComponentUpdate(std::string componentPath, std::string value) {
     std::vector<std::string> splitPath = splitString(componentPath, COMPONENT_PATH_DELIMITER);
-    m_userDefinedValues[componentPath] = { splitPath, value };
+    m_userDefinedValues.push_back({ componentPath, { splitPath, value }});
 }
 
 }
--- a/core/src/scene/scene.cpp
+++ b/core/src/scene/scene.cpp
@@ -21,15 +21,15 @@ namespace Tangram {
 
 static std::atomic<int32_t> s_serial;
 
-Scene::Scene(std::string path) : id(s_serial++), m_path(path) {
+Scene::Scene(std::string path, std::string scene) : id(s_serial++), m_scene(scene), m_path(path) {
     m_view = std::make_shared<View>();
     // For now we only have one projection..
     // TODO how to share projection with view?
     m_mapProjection.reset(new MercatorProjection());
 }
 
-Scene::Scene(std::string path, std::vector<UserDefinedSceneValue> userDefined) :
-    Scene(path)
+Scene::Scene(std::string path, std::vector<UserDefinedSceneValue> userDefined, std::string scene) :
+    Scene(path, scene)
 {
     m_userDefinedValues = userDefined;
 }
--- a/core/src/scene/scene.cpp
+++ b/core/src/scene/scene.cpp
@@ -28,10 +28,10 @@ Scene::Scene(std::string scene) : id(s_serial++), m_scene(scene) {
     m_mapProjection.reset(new MercatorProjection());
 }
 
-Scene::Scene(std::vector<UserDefinedSceneValue> userDefined, std::string scene) :
+Scene::Scene(std::vector<UpdateValue> updates, std::string scene) :
     Scene(scene)
 {
-    m_userDefinedValues = userDefined;
+    m_updates = updates;
 }
 
 Scene::~Scene() {}
@@ -84,7 +84,7 @@ bool Scene::texture(const std::string& textureName, std::shared_ptr<Texture>& te
 
 void Scene::queueComponentUpdate(std::string componentPath, std::string value) {
     std::vector<std::string> splitPath = splitString(componentPath, COMPONENT_PATH_DELIMITER);
-    m_userDefinedValues.push_back({ splitPath, value });
+    m_updates.push_back({ splitPath, value });
 }
 
 }
--- a/core/src/scene/scene.cpp
+++ b/core/src/scene/scene.cpp
@@ -26,7 +26,7 @@ Scene::Scene(std::string path) : id(s_serial++), m_path(path) {
     m_mapProjection.reset(new MercatorProjection());
 }
 
-Scene::Scene(std::string path, std::unordered_map<StyleComponent, StyleComponents> userDefined) :
+Scene::Scene(std::string path, std::map<StyleComponent, StyleComponents> userDefined) :
     Scene(path)
 {
     m_userDefinedValues = userDefined;
--- a/core/src/scene/scene.cpp
+++ b/core/src/scene/scene.cpp
@@ -21,15 +21,15 @@ namespace Tangram {
 
 static std::atomic<int32_t> s_serial;
 
-Scene::Scene(std::string path, std::string scene) : id(s_serial++), m_scene(scene), m_path(path) {
+Scene::Scene(std::string scene) : id(s_serial++), m_scene(scene) {
     m_view = std::make_shared<View>();
     // For now we only have one projection..
     // TODO how to share projection with view?
     m_mapProjection.reset(new MercatorProjection());
 }
 
-Scene::Scene(std::string path, std::vector<UserDefinedSceneValue> userDefined, std::string scene) :
-    Scene(path, scene)
+Scene::Scene(std::vector<UserDefinedSceneValue> userDefined, std::string scene) :
+    Scene(scene)
 {
     m_userDefinedValues = userDefined;
 }
--- a/core/src/scene/scene.cpp
+++ b/core/src/scene/scene.cpp
@@ -15,6 +15,8 @@
 #include <atomic>
 #include <algorithm>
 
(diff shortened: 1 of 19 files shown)


blair1618:
The implementation here seems to only allow updating properties that are already defined in the scene file. To most nearly match tangram-js and to get the most utility from this feature, we should also be able to create properties that are not already defined in the scene file. For example, with our current demo scene I should be able to call:
```c++
Tangram::setSceneComponent("scene.animated", "true");
Tangram::applySceneUpdates();
```
Even though `scene` does not have a member called `animated` before this call. 

Naturally, this raises the question of what to do about intermediate properties along a path are not defined, e.g.
```c++
Tangram::setSceneComponent("lights.new_light.type", "point");
Tangram::applySceneUpdates(); // Does this create 'lights.new_light'??
```
In the tangram-js api, because you are just manipulating JS objects, you would have to first create `lights.new_light` before accessing its properties. We could place a similar restriction here and only allow creating properties whose path is fully defined.

I think it would also be possible to unambiguously create any needed intermediate properties when defining a new property, since any property accessed with dot notation must be considered an element of a mapping. 

If the first approach (the JS-like approach) is easier to implement, I would recommend that we start there and see whether we find it to be limiting.

core/src/scene/scene.cpp
```diff
@@ -73,4 +80,9 @@ bool Scene::texture(const std::string& textureName, std::shared_ptr<Texture>& te
     return true;
 }
 
+void Scene::setComponent(std::string componentPath, std::string value) {
+    std::vector<std::string> splitPath = splitString(componentPath, COMPONENT_PATH_DELIMITER);
+    m_userDefinedValues[componentPath] = { splitPath, value };
```
blair1618: Component updates should be stored in a structure that preserves the order of their creation, otherwise we'll end up with update sequences that don't have well-defined outcomes, e.g.
```c++
Tangram::setSceneComponent("scene.animated", "true");
Tangram::setSceneComponent("scene", "null");
Tangram::applySceneUpdates(); // Does this succeed? Depends on your map hash function!
```

core/src/tangram.h
```diff
@@ -139,6 +139,10 @@ void loadScene(const char* _scenePath, bool _setPositionFromScene = false);
 
 void runOnMainLoop(std::function<void()> _task);
 
+void setSceneComponent(const char* componentName, const char* value);
```
blair1618: Can we make it more explicit that updates are deferred? Something like `queueSceneUpdate` sounds more honest about what's going on here.

karimnaaji: Yes, will update.

core/src/scene/scene.cpp
karimnaaji: 👍

core/src/util/util.h
```diff
@@ -0,0 +1,29 @@
...
+inline bool tryFind(M& map, const std::string& key, T& out) {
+    auto it = map.find(key);
+
+    if (it != map.end()) {
+        out = it->second;
+        return true;
+    }
+
+    return false;
+}
```
blair1618: trailing newlines for the trailing newline deity

karimnaaji:
@blair1618 , comments have been addressed, you can refer to unit tests.

blair1618:
Sweet. Reviewing!

core/src/util/util.h
tallytalwar: I think you can configure your editor to take care of this.
I have added this to my vimrc.

blair1618:
It feels weird to me to call these scene updates "user-defined values" because the original scene file is *also* entirely user-defined. They're really just updates or edits.

core/src/scene/scene.h
```diff
@@ -28,12 +28,21 @@ struct Stops;
...
  */
 
+#define COMPONENT_PATH_DELIMITER '.'
+
+struct UserDefinedSceneValue {
+    std::vector<std::string> splitPath;
+    std::string value;
+};
+
+using UserDefinedSceneValuePair = std::pair<std::string, UserDefinedSceneValue>;
```
blair1618: Do we need to keep around both the original path and the split path?

core/src/scene/sceneLoader.cpp
```diff
@@ -42,16 +42,56 @@ const std::string DELIMITER = ":";
 // TODO: make this configurable: 16MB default in-memory DataSource cache:
 constexpr size_t CACHE_SIZE = 16 * (1024 * 1024);
 
-bool SceneLoader::loadScene(const std::string& _sceneString, Scene& _scene) {
+bool SceneLoader::loadScene(const std::string& _sceneString, Scene& _scene, Node& root) {
```
blair1618: I'm unclear why this method now takes a `Node&` argument, there doesn't seem to be anything functionally different going on here.

core/src/scene/scene.h
```diff
@@ -28,12 +28,21 @@ struct Stops;
  * Scene is a singleton containing the styles, lighting, and interactions defining a map scene
  */
 
+#define COMPONENT_PATH_DELIMITER '.'
```
blair1618: Putting `#define` directives in headers is somewhat dangerous because they have global scope wherever they end up being included. It doesn't look like this is used in the header, so can we put it in the implementation file instead?

core/src/tangram.cpp
```diff
@@ -500,7 +502,20 @@ const std::vector<TouchItem>& pickFeaturesAt(float _x, float _y) {
...
+void queueSceneUpdate(const char* componentName, const char* value) {
+
+    return m_scene->queueComponentUpdate(componentName, value);
+}
+
+void applySceneUpdates() {
 
+    if (m_scene->userDefines().size() > 0) {
+        // reload the entire scene for now
+        loadScene(m_scene->path().c_str());
```
blair1618: So it looks like applying scene updates will load the scene file from the file system and parse the entire thing again, which can take multiple seconds and is not really necessary to this task. 

Instead we should probably be keeping a reference to the in-memory representation of the YAML document and simply applying the scene updates to that object.

core/src/tangram.cpp
```diff
@@ -102,11 +102,13 @@ void loadScene(const char* _scenePath, bool _setPositionFromScene) {
 
     bool setPositionFromCurrentView = bool(m_scene);
 
-    auto scene = std::make_shared<Scene>();
+    auto scene = std::make_shared<Scene>(m_scene->path(), m_scene->userDefines());
```
blair1618: To behave consistently, we'd need to set the "path" of the new `Scene` object to `_scenePath` here, otherwise the scene will become very confused about which file it is supposed to use if you load another scene file after initialization. 

More generally though, it doesn't seem super useful for the scene to store this path in the first place.

tests/unit/sceneUpdateTests.cpp
```diff
@@ -0,0 +1,73 @@
...
+    scene.queueComponentUpdate("layers.poi_icons.draw.icons.interactive", "false");
+    scene.queueComponentUpdate("styles.heightglow.shaders.uniforms.u_time_expand", "5.0");
+    scene.queueComponentUpdate("cameras.iso-camera.active", "true");
+    scene.queueComponentUpdate("cameras.iso-camera.type", "perspective");
+    scene.queueComponentUpdate("global.default_order", "function() { return 0.0; }");
+    scene.queueComponentUpdate("global.non_existing_property0", "true");
+    scene.queueComponentUpdate("global.non_existing_property1.non_existing_property_deep", "true");
+
+    // Tangram apply scene updates, reload the scene
+    REQUIRE(SceneLoader::loadScene(sceneString, scene, root));
```
blair1618: Ok, I think I can understand the reasoning behind it but I think it could be confusing that `loadScene` always applies queued updates. By putting `applySceneUpdates` in the public API we have given users a way to explicitly tell us to apply queued scene updates, but here we just go right ahead and do it without the user ever telling us to.

tests/unit/sceneUpdateTests.cpp
```diff
@@ -0,0 +1,73 @@
...
+    scene.clearUserDefines();
```
blair1618: I think it could also be confusing that updates are not "consumed" when they are applied - it feels more natural to me to empty the update queue when the user says to apply them.

tests/unit/sceneUpdateTests.cpp
```diff
@@ -0,0 +1,73 @@
...
+    auto sceneString = stringFromFile(setResourceRoot(path).c_str(), PathType::resource);
+
+    REQUIRE(!sceneString.empty());
+
+    Node root;
+    REQUIRE(SceneLoader::loadScene(sceneString, scene, root));
+
+    // Update
+    scene.queueComponentUpdate("lights.light1.ambient", "0.9");
+    scene.queueComponentUpdate("lights.light1.ambient", "0.0");
```
blair1618: This test is necessary, but not sufficient to determine whether sequence of entry is preserved (for instance, this would still pass if updates were placed in a `map` by their component path, like before). We should test two components with different paths, where one depends on the other being updated first.

tests/unit/sceneUpdateTests.cpp
```diff
@@ -0,0 +1,73 @@
...
+#include "scene/sceneLoader.h"
+#include "style/style.h"
+#include "scene/scene.h"
+#include "platform.h"
+#include "tangram.h"
+
+using YAML::Node;
+using namespace Tangram;
+
+const char* path = "scene.yaml";
```
blair1618: Relying on the name and contents of our testing scene file is probably not a good idea. If we go and refactor or rename `scene.yaml` then these tests will all fail even though the code they're meant to test is unchanged. I'd recommend just using some cutsom YAML, it doesn't even have to be a complete scene specification.

core/src/scene/sceneLoader.cpp
```diff
@@ -42,16 +42,56 @@ const std::string DELIMITER = ":";
...
+    auto& userDefines = scene.userDefines();
+
+    for (const auto& define : userDefines) {
+        auto& userDefine = define.second;
+        auto& path = userDefine.splitPath;
+
+        std::vector<Node> stack;
+        stack.push_back(root);
+        for (auto& str : path) {
+            if (stack.back()[str]) {
```
blair1618: This bit is elegant, very nice!

core/src/scene/scene.h
karimnaaji: No actually this can go now, it was only used for user error messages but we can reconstruct it from the split path.

core/src/scene/sceneLoader.cpp
karimnaaji: It's used only to enable unit testing, since it needs access to the root.

core/src/scene/scene.h
karimnaaji: Yes we can move that directive.

tests/unit/sceneUpdateTests.cpp
karimnaaji: That's what already happens through the Tangram API, but in the unit test we directly manipulate the scene object so this call is needed, here it basically mimics what `Tangram::applySceneUpdates();` does in a shorter way. 
We can go through Tangram API for unit tests, but this involves having all of the other components (View/TileManager...), so I think we're good on that.

tests/unit/sceneUpdateTests.cpp
karimnaaji: Yes agreed on that, this has been updated.

tests/unit/sceneUpdateTests.cpp
blair1618: Oh you're right, I see it now. :+1:

karimnaaji:
Comments addressed.

PR #675, opened 2016-04-12
Yet Another Refactor of MapData usage

blair1618:
- `MapData` can only be instantiated through a `MapController`, so now they are always associated with a non-null map
 - You no longer need to call `syncWithMap` after modifying a `MapData`, synchronization is automatic
 - To prevent native data sources becoming unreachable from Java code, `MapController` keeps a static collection of all `MapData` instances

### DIFF
--- a/android/tangram/src/com/mapzen/tangram/MapController.java
+++ b/android/tangram/src/com/mapzen/tangram/MapController.java
@@ -12,6 +12,7 @@ import com.squareup.okhttp.Request;
 import com.squareup.okhttp.Response;
 
 import java.io.IOException;
+import java.util.HashMap;
 import java.util.Map;
 
 import javax.microedition.khronos.egl.EGLConfig;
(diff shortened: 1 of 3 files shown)


blair1618:
@hjanetzek and @tallytalwar take a look when you get a chance - the part that feels most questionable to me is the static collection of `MapData` instances, but I think it's the only way to make sure that no data source pointers become unreachable from Java code.

hjanetzek:
LGTM

android/tangram/src/com/mapzen/tangram/MapController.java
```diff
@@ -312,6 +313,37 @@ public LngLat coordinatesAtScreenPosition(double screenX, double screenY) {
...
+        mapData = new MapData(name, pointer, this);
+        clientDataSources.put(name, mapData);
+        return mapData;
+    }
+
+    /**
+     * For package-internal use only; remove a {@code MapData} from this map
+     * @param mapData The {@code MapData} to remove
+     */
+    void removeDataLayer(MapData mapData) {
```
tallytalwar: As a user, will one not expect to call `removeDataLayer` from `MapController` object, like is done for `addDataLayer`?Instead of making a `MapData` object remove/disassociate itself from the `MapController` object?

blair1618: Indeed, that would've been my guess as well. I made `remove()` a method on `MapData` instead to more closely resemble the gmaps API (https://developers.google.com/maps/documentation/android-api/groundoverlay#remove_an_overlay).

tallytalwar: In that case, can we also allow users to remove a `MapData` from `MapController` and not having `* For package-internal use only;` in the code comments/documentation? Will be sometime more intuitive to offer along with mimicking gmaps API.

blair1618: The method isn't public, so it won't be visible in documentation. Having two ways to do one thing is generally more confusing than having one way.

tallytalwar: Fine, but I also know that we decided on not mimicking gmaps 100%, and do things which are more correct for our usage.
Anyways, I am going to merge this right now. Will be good to take @ecgreb and @sarahlensing opinion on this/these APIs as well.

ecgreb: I agree either one seems fine (but not both) so perhaps using the convention Android developers have been trained to expect and are using in existing mapping apps should be the tie-breaker as long as it does not go against the architecture of Tangram ES.

PR #679, opened 2016-04-13
Clean up camera-related code in Android SDK

blair1618:
Consider this a prelude to more substantial changes that will bring the "camera" interface closer to that of gmaps. Notable changes here:

 - `setMapThing` -> `setThing` in `MapController`
 - Eased setters in `MapController` now take duration values in milliseconds instead of seconds
 - Positions are always represented as a `LngLat` in `MapController` (a method that takes two `double` arguments makes it too easy to put them in the wrong order)
 - Removed the methods to get/set camera type (this can now be done with scene component updates)

### DIFF
--- a/android/tangram/src/com/mapzen/tangram/MapController.java
+++ b/android/tangram/src/com/mapzen/tangram/MapController.java
@@ -31,12 +31,6 @@ public class MapController implements Renderer {
 
     protected static EaseType DEFAULT_EASE_TYPE = EaseType.CUBIC;
 
-    public enum CameraType {
-        PERSPECTIVE,
-        ISOMETRIC,
-        FLAT,
-    }
-
     public enum DebugFlag {
         FREEZE_TILES,
         PROXY_COLORS,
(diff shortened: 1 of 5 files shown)


android/tangram/src/com/mapzen/tangram/MapController.java
```diff
@@ -125,54 +121,44 @@ public void setHttpHandler(HttpHandler handler) {
...
-     * @param lat Degrees latitude of the position to set
-     * @param duration Time in seconds to ease to the given position
+     * @param position LngLat of the position to set
+     * @param duration Time in milliseconds to ease to the given position
      * @param ease Type of easing to use
      */
-    public void setMapPosition(double lng, double lat, float duration, EaseType ease) {
-        nativeSetPositionEased(lng, lat, duration, ease.ordinal());
+    public void setPosition(LngLat position, int duration, EaseType ease) {
+        float seconds = duration / 1000.f;
```
tallytalwar: So internally its still in seconds. Just curious if the switch is to have compat with gmaps?

tallytalwar:
LGTM.

android/tangram/src/com/mapzen/tangram/MapController.java
blair1618: Exactly. Other map APIs (and animations in Android in general) tend to specify times in milliseconds.

PR #685, opened 2016-04-16
Raster source

tallytalwar:
Refer Tangram PR for conceptual details: https://github.com/tangrams/tangram/pull/282

Some of the technical details in this PR specific to implementation on tangram-es

- Each `TileTask` can have a set of extra dependent tasks (`m_rasterTasks`) responsible to "Download" the raster textures from appropriate DataSources. The `tileWorker`s continue to build the tile geometry and once done will wait on these dependent raster tasks to have valid data. Once done the downloaded Textures are stored with their corresponding "overzoomed" tileIDs in the built `Tile`.
Note: All these download tasks are restricted within the bounds of the `MAX_DOWNLOADS`!

- RasterSource is the central storage for all the raster textures belonging to itself. `Tile`s will share references to these `textures` stored in the RasterSource. Texture in the datasource will only be deleted when:
  - all the `tile`s referring the `texture` are popped from the `TileCache`, that is when there is only one reference to the `Texture` which is the one stored in the `RasterSource` itself.
  - Note: Maybe Texture deletion logic could be more aggressive and take DataSource RawDataCache into account. Currently this is not being done.

- `Rasters` structure
  - This stores a `shared` reference of the raster texture and the corresponding **overzoomed** `TileID`. This `TileID` is used to determine the texture `offset` and `position` when the raster tile is overzoomed.
  - Tile stores a vector of these `Rasters`, which is then used to bind and upload the raster textures at draw time.

- Raster sampler shader injection
  - Any style which defines a `rasters` (values: `color`/`normal`/`custom`), will have shader injections to define texture/raster samplers. The number of samplers injected will be dependent on the number of rasters textures a tile has.

Note: We now have a `default` **1x1 black texture**. If for any reason (no valid texture data from a raster datasource, or bad texture image, etc) texture creation fails, a default 1x1 black pixel texture is created to avoid having random texture unit state set at draw time. This now mimics the behavior we have on JS side.

### DIFF
--- a/bench/src/tileLoading.cpp
+++ b/bench/src/tileLoading.cpp
@@ -54,7 +54,7 @@ struct TestContext {
         styleContext.initFunctions(*scene);
         styleContext.setKeywordZoom(0);
 
-        source = scene->dataSources()[0];
+        source = scene->dataSources().begin()->second;
         tileBuilder.setScene(scene);
     }
 
@@ -76,7 +76,7 @@ struct TestContext {
 
     void parseTile() {
         Tile tile({0,0,10,10,0}, s_projection);
-        source = scene->dataSources()[0];
+        source = scene->dataSources().begin()->second;
         auto task = source->createTask(tile.getID());
         auto& t = dynamic_cast<DownloadTileTask&>(*task);
         t.rawTileData = std::make_shared<std::vector<char>>(rawTileData);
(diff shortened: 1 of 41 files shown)


tallytalwar:
We noticed a bug in queueing loading tasks in tileManager. Will be pushing a fix for this most likely tonight. Hopefully :D.

tallytalwar:
Closing in favor of a better implemented tileManager behavior.
https://github.com/tangrams/tangram-es/pull/700

Issue #683, opened 2016-04-15
Tangram should provide a mechanism to notify clients when a view is "complete"


blair1618:
The motivation for this is mostly to support capturing screenshots of the current map view after the view is finished loading and rendering. The interface could be something very simple, like:

C++
```c++
void setViewCompleteCallback(std::function<void()> _onViewComplete);
```

Java
```java
public void setOnViewCompleteListener(OnViewCompleteListener listener);
```
The callback should be executed after all visible tiles have been loaded and made renderable. Some potential edge cases and how I feel they should be handled:

 - If there are any eases in progress, the view should not be considered "complete".
 - If labels collisions or transitions are still in progress, the view should not be considered "complete". 
 - Scene animation should be disregarded when determining whether the view is "complete".

hjanetzek:
I came across a problem with using this callback to take a screenshot: one does not know whether the view is already complete when doing the screenshot render-request. So one would need to call onViewComplete after each render() with a complete view. I would suggest to have a

```OnRenderDoneCallback(bool complete, bool animating)```

with 'complete' meaning all tiles loaded and no label animations

blair1618:
Resolved by https://github.com/tangrams/tangram-es/commit/90d259cee29c436540277ec9177f71221fbb1629. `Tangram::update(dt)` now returns a boolean representing whether the view is "complete".

PR #686, opened 2016-04-18
[wip] View complete callback

hjanetzek:
- [ ] TODO: cleanups + docs

### DIFF
--- a/android/demo/src/com/mapzen/tangram/android/MainActivity.java
+++ b/android/demo/src/com/mapzen/tangram/android/MainActivity.java
@@ -2,6 +2,7 @@ package com.mapzen.tangram.android;
 
 import android.app.Activity;
 import android.os.Bundle;
+import android.util.Log;
 import android.view.Window;
 import android.widget.Toast;
 
@@ -87,9 +88,7 @@ public class MainActivity extends Activity implements OnMapReadyCallback, TapRes
                 public void onViewComplete() {
                     runOnUiThread(new Runnable() {
                             public void run() {
-                                Toast.makeText(getApplicationContext(),
-                                               "View Complete",
-                                               Toast.LENGTH_SHORT).show();
+                                Log.d("Tangram", "View complete");
                             }
                         });
                 }});
--- a/android/demo/src/com/mapzen/tangram/android/MainActivity.java
+++ b/android/demo/src/com/mapzen/tangram/android/MainActivity.java
@@ -9,6 +9,7 @@ import com.mapzen.tangram.HttpHandler;
 import com.mapzen.tangram.LngLat;
 import com.mapzen.tangram.MapController;
 import com.mapzen.tangram.MapController.FeaturePickListener;
+import com.mapzen.tangram.MapController.ViewCompleteListener;
 import com.mapzen.tangram.MapData;
 import com.mapzen.tangram.MapView;
 import com.mapzen.tangram.MapView.OnMapReadyCallback;
@@ -82,6 +83,16 @@ public class MainActivity extends Activity implements OnMapReadyCallback, TapRes
         map.setLongPressResponder(this);
         map.setFeaturePickListener(this);
 
+        map.setViewCompleteListener(new ViewCompleteListener() {
+                public void onViewComplete() {
+                    runOnUiThread(new Runnable() {
+                            public void run() {
+                                Toast.makeText(getApplicationContext(),
+                                               "View Complete",
+                                               Toast.LENGTH_SHORT).show();
+                            }
+                        });
+                }});
         markers = map.addDataLayer("touch");
     }
 
--- a/android/tangram/src/com/mapzen/tangram/MapController.java
+++ b/android/tangram/src/com/mapzen/tangram/MapController.java
@@ -63,6 +63,14 @@ public class MapController implements Renderer {
         void onFeaturePick(Map<String, String> properties, float positionX, float positionY);
     }
 
+    public interface ViewCompleteListener {
+        /**
+         * Called on the render-thread at the end of whenever the view is fully loaded and
+         * no ease- or label-animation is running.
+         */
+        void onViewComplete();
+    }
+
     /**
      * Construct a MapController using a custom scene file
      * @param context Context in which the map will function; the asset bundle for this activity
@@ -513,6 +521,12 @@ public class MapController implements Renderer {
     }
 
     /**
+     */
+    public void setViewCompleteListener(ViewCompleteListener listener) {
+        viewCompleteListener = listener;
+    }
+
+    /**
      * Enqueue a Runnable to be executed synchronously on the rendering thread
      * @param r Runnable to run
      */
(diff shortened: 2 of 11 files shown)


hjanetzek:
Implementing https://github.com/tangrams/tangram-es/issues/683

@blair1618 How about using simple notification events on the C++ side and implement the callback/listener style just for MapController? I've added this interface to platform.h
```
enum MapEvent {
    view_completed,
    ...
}
```
```
handleMapEvent(MapEvent){}
```

blair1618:
Hmm that could work. So then each platform implementation would have code to resolve the events? Let's try it and see how we like it. 

Not a fan of scoped enums I take it? ;) 
I tend to use `enum class` unless there's a compelling reason to use the older style.

hjanetzek:
@blair1618  it's actually implemented as enum class. the example above was just a draft

blair1618:
Oh my bad. :+1:

core/src/tangram.cpp
```diff
@@ -44,6 +44,11 @@ std::unique_ptr<Labels> m_labels;
 std::unique_ptr<Skybox> m_skybox;
 std::unique_ptr<InputHandler> m_inputHandler;
 
+bool m_viewComplete = false;
+std::function<void()> m_viewCompleteCallback = [](){
```
blair1618: Is this just for testing?

hjanetzek: Yes, but I'm currently thinking about switching to the setViewCompletedCallback API that you suggeted

core/src/tangram.cpp
```diff
@@ -241,6 +247,23 @@ void render() {
...
+                   m_tileManager->hasTileSetChanged() ||
+                   m_tileManager->hasLoadingTiles() ||
+                   m_labels->needUpdate());
+    if (ready) {
+        for (auto& ease : m_eases) {
+            if (!ease.finished()) { ready = false; }
+        }
+    }
+
+    m_renderDoneCallback(ready);
```
blair1618: Why not something like
```
if (ready) {
    m_renderDoneCallback();
} else if(!isContinuousRendering()) {
    requestRender();
}
```
?

hjanetzek: It's just the name that implies for me that this callback should be called at the end of each frame, but yes, for this
ViewCompleteCallback would probably be more appropriate (though it sounds more like it is only triggered when the state changes to complete, not for each frame that is complete)

I think it would be nicer to query the frame ready state in onDraw() instead
```
      if (viewCompleteListener != null) {
           if (nativeFrameRenderedWithAllTilesLoadedAndNoAnimationRunning()) {
               viewCompleteListener.onViewComplete();
           }
      }
```

blair1618: Yeah, the original goal here was a "view complete" event/callback. You're right that it should really only be called when "complete" changes from false to true, and should probably be called something more like "view complete" rather than "render done".

hjanetzek: @blair1618 How about this last commit? Just return true from render() whenever the view is fully loaded, keeping the listener thing completely in MapController

blair1618: Oh that makes sense! How simple :)

Would it be more natural to return the boolean from `update()`? `update()` is where tangram actually figures out whether the view is complete.

hjanetzek: yea. moving this to update()

android/demo/src/com/mapzen/tangram/android/MainActivity.java
```diff
@@ -170,5 +213,52 @@ public void onFeaturePick(Map<String, String> properties, float positionX, float
                 "Selected: " + name + " at: " + positionX + ", " + positionY,
                 Toast.LENGTH_SHORT).show();
     }
-}
 
+    private void saveScreenshot(Bitmap bitmap) {
+
+        ByteArrayOutputStream bos = new ByteArrayOutputStream();
+        bitmap.compress(Bitmap.CompressFormat.PNG, 90, bos);
```
blair1618: I think you can pass a `FileOutputStream` directly to `Bitmap$compress` and skip the byte array altogether.

android/tangram/src/com/mapzen/tangram/MapController.java
```diff
@@ -51,6 +54,52 @@
...
+    public Bitmap capture() {
+        // FIXME Check if on render thread!
+        int w = mapView.getWidth();
+        int h = mapView.getHeight();
+
+        int b[] = new int[(int) (w * h)];
+        int bt[] = new int[(int) (w * h)];
+        IntBuffer buffer = IntBuffer.wrap(b);
+        buffer.position(0);
+        GLES20.glReadPixels(0, 0, w, h, GLES20.GL_RGBA, GLES20.GL_UNSIGNED_BYTE, buffer);
```
karimnaaji: What is the reason of not reading the raw pixels in a native call? It would be the same for other platforms (and could still use platform image types like `Bitmap` and `CImage`.

hjanetzek: It seemed to be more work to write jni wrapper for glReadPixels without much gain. When we add a frame capture api for other platforms this can be revisited.

android/demo/src/com/mapzen/tangram/android/MainActivity.java
hjanetzek: yep

android/tangram/src/com/mapzen/tangram/MapController.java
karimnaaji: Though, one use case for it is to run it along with osmesa; so it would make sense to wrap it in a native call.

hjanetzek: I think we should do this when we decide on a common implemention for a screenshot API for all platforms

blair1618:
Works like a charm!

There's an issue with Android permissions at API level 23 and up (we would need to ask the user for WRITE_STORAGE permission at run-time) but that's really outside the domain of the library. 

I think we can remove the demo activity code for saving a screenshot before merging this. Taking screenshots is useful for testing, but we can already do that with the built-in Android screen capture function. 

Think you can clean up or squash some of these commits?

hjanetzek:
Write permissions are only needed for the Demo's manifest. Why is this a problem? I though one could at least use a Share Intent to push screenshots to slack but this also requires storing as file or FileProvider from v4 support package..

blair1618:
Starting with Android 6.0, certain "dangerous" permissions (like WRITE_STORAGE) can't just be declared in an app's manifest, they must be explicitly requested from the user at run-time. This means saving a screenshot to disk is probably more work than we really want in our demo. Sharing to Slack could be cool, but that might have the same problem. 

https://developer.android.com/training/permissions/requesting.html

blair1618:
LGTM. If nobody has objections, I'll merge this shortly.

hjanetzek:
should the capture callback be called on the UI thread?

blair1618:
Hm, maybe so. The GLSurfaceView render loop **is** on a separate thread from the UI thread, but I'm not sure that one is clearly a better context for this callback.

blair1618:
Actually, let's just merge this as-is. If we decide later that it would be better to execute the capture callback on a different thread, we can change that without changing the interface.

PR #687, opened 2016-04-18
Scene update tweaks

blair1618:
Two major updates here:

 - Instead of retaining a string containing the yaml scene configuration, the scene object retains the parsed yaml scene representation, so that we don't have to re-parse the unchanged text. 
 - Data sources added dynamically were dropped whenever a scene was loaded (and, by extension, whenever a scene component was updated). These dynamic "client" sources are now stored separately in the scene object in order to preserve them across scene loads. 

Aside from that there's a good bit of refactoring for clarity and succinctness.

### DIFF
--- a/core/src/scene/scene.cpp
+++ b/core/src/scene/scene.cpp
@@ -21,16 +21,14 @@ namespace Tangram {
 
 static std::atomic<int32_t> s_serial;
 
-Scene::Scene(std::string scene) : id(s_serial++), m_scene(scene) {
+Scene::Scene() : id(s_serial++) {
     m_view = std::make_shared<View>();
     // For now we only have one projection..
     // TODO how to share projection with view?
     m_mapProjection.reset(new MercatorProjection());
 }
 
-Scene::Scene(std::vector<UpdateValue> updates, std::string scene) :
-    Scene(scene)
-{
+Scene::Scene(std::vector<Update> updates) : Scene() {
     m_updates = updates;
 }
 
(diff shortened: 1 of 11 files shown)


hjanetzek:
LGTM.

At some point we should look into having a more efficient representation though: keeping the bubble-wrap style around as yaml object takes about 8MB

blair1618:
Cool. Yeah I totally agree that a more compact representation would be better, but right now I think this is a good trade-off of memory usage for speed.

karimnaaji:
Looks good to me, merging.

core/src/scene/scene.cpp
```diff
@@ -82,9 +83,25 @@ bool Scene::texture(const std::string& textureName, std::shared_ptr<Texture>& te
...
+}
+
+void Scene::addClientDataSource(std::shared_ptr<DataSource> _source) {
+    m_clientDataSources.push_back(_source);
+}
+
+void Scene::removeClientDataSource(DataSource& _source) {
+    auto it = std::remove_if(m_clientDataSources.begin(), m_clientDataSources.end(),
+        [&](auto& s) { return s.get() == &_source; });
+    m_clientDataSources.erase(it, m_clientDataSources.end());
```
tallytalwar: I do not get this second `erase` here.

First `remove_if` will remove all the clients datasources and return `past the end iterator` for the set of elements remaining in the container.
And the second statement is erasing everything from the `past the end iterator` (which is/should be `m_clientDataSources.end()`) to `m_clientDataSources.end()`. Not sure if the second erase is needed here.

Can you explain if I am misreading something here?

Thanks

blair1618: Yeah this is just how the "remove" algorithm works in the STL, "removing" means moving all the designated elements to the end of the collection and returning an iterator to the first element to be removed (http://en.cppreference.com/w/cpp/algorithm/remove). This trips me up all the time.

tallytalwar: Ohh I should have read the whole thing :D.
Yes, that is confusing.

PR #688, opened 2016-04-19
Enable subclassing MapController and MapView

blair1618:
`MapController` and `MapView` subclasses should implement `MapController$getInstance` and `MapView$getMapInstance`.

I had to suppress a warning in our `AsyncTask` because Android thinks that every method on a subclass of `View` can only be run on the UI thread, even though `getMapInstance` is perfectly safe to call from a worker thread.


ecgreb:
LGTM

tallytalwar:
👍

PR #690, opened 2016-04-20
fix: skip empty map fields

hjanetzek:


core/src/scene/styleMixer.cpp
```diff
@@ -205,9 +205,10 @@ void StyleMixer::mergeMapFieldTakingLast(const std::string& key, Node target, co
 
     for (auto it = sources.rbegin(); it != sources.rend(); ++it) {
         const auto& source = (*it)[key];
-        if (!source.IsMap()) {
+        if (!source || !source.IsMap()) {
```
blair1618: Huh. This really shouldn't change any behavior, an undefined node should return `false` for `IsMap()`. This might be a bug in yaml-cpp.

hjanetzek: Yeah. 

The raster-source branched emitted for this parsing error. The fix that I had there was ```if (!(*it) || !it->IsMap()) { continue; }``` While writing the unit test the above logic seemed to include(i.e. also handle) that case.

blair1618:
LGTM

PR #693, opened 2016-04-21
Small revisions to positioning method names and javadoc

blair1618:


android/tangram/src/com/mapzen/tangram/MapController.java
```diff
@@ -133,21 +133,21 @@ public void setPosition(LngLat position) {
...
      */
-    public void setPosition(LngLat position, int duration) {
-        setPosition(position, duration, DEFAULT_EASE_TYPE);
+    public void setPositionEased(LngLat position, int duration) {
+        setPositionEased(position, duration, DEFAULT_EASE_TYPE);
     }
 
     /**
-     * Set the geographic position of the center of the map view
+     * Set the geographic position of the center of the map view with easing
```
tallytalwar: Doc should reflect the setting of EaseType in the description? Can you update please?

blair1618: You mean the description should say "cubic easing"? This method doesn't care what easing type is used, if someone does care then they should use the other method. Plus now the docs will remain correct if we change the default easing.

tallytalwar: I just mean to say the description should point the difference between the usage of:

`public void setPositionEased(LngLat position, int duration)` and
`public void setPositionEased(LngLat position, int duration, EaseType ease)`

blair1618: Oh gotcha, will edit that!

blair1618: Updated

tallytalwar:
Looks Good.Merging when travis goes green.

PR #694, opened 2016-04-21
Polyline texture coordinates grow linearly with path distance

### LINKED ISSUE #678: Texture coordinates for polylines should be "linear" in tile space
See https://github.com/tangrams/tangram/pull/275

blair1618:
Resolves #678 

Before:
screen shot 2016-04-21 at 1 05 39 am

After:
screen shot 2016-04-21 at 1 01 59 am

There's currently no special logic for the texture coordinates of caps or joins. I haven't noticed significant visual impact from these, but they can be addressed incrementally as we see fit.

### DIFF
--- a/core/resources/scene.yaml
+++ b/core/resources/scene.yaml
@@ -61,6 +61,12 @@ styles:
     icons:
         base: points
         texture: pois
+    stripeline:
+        base: lines
+        texcoords: true
+        shaders:
+            blocks:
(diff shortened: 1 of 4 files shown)


tallytalwar:
LGTM. Will merge.

bcamper:
@patriciogonzalezvivo what (if anything) are you doing for caps/joins in JS?

PR #696, opened 2016-04-22
Use constant to scale linear y uv

karimnaaji:
Using the interpolated width creates a "sliding" effect for the uvs, we should use a constant instead (to match js behavior). Following https://github.com/tangrams/tangram-es/pull/694.


blair1618:
LGTM

PR #700, opened 2016-04-23
Raster source

### LINKED ISSUE #672: Raster tile source and style
Refer JS implementation and notes: https://github.com/tangrams/tangram/pull/282

Some Major Tasks to be done:
- [x] Raster DataSource
  - [x] texture store
  - [x] mesh store
  - [x] raster datasource tiles and tile manager interactions
  - [x] proxy tiles handling?
- [x] base raster style
- [x] raster tile sampler
- [x] mix raster datasource samplers with other vector tile data.

tallytalwar:
Fixes #672 

Refer Tangram PR for conceptual details: https://github.com/tangrams/tangram/pull/282

Some of the technical details in this PR specific to implementation on tangram-es

- Each `TileTask` can have a set of extra dependent tasks (`m_subTasks`) responsible to "Download" the raster textures from appropriate DataSources. The `tileWorker`s continue to build the tile geometry and once done, the built tile is stored in the parent tileTask. A dependent raster task has a valid raster even even the download resulted in bad/no data. Once done the downloaded Textures are stored with their corresponding "overzoomed" tileIDs (just x/y/z) in the built `Tile`.
- Once all the dependent rasters are ready, the stored task tile is put to draw and moved to the ready tiles set in the tileManager.
- `RasterTileTask` is a specialized class of `DownloadTileTask`.
Note: All these download tasks are restricted within the bounds of the `MAX_DOWNLOADS`!

  - SubTasks (raster tasks) is loaded with a callback which makes sure a `renderRequest` is called for the tileManager to pick the updates, and fill the respective tiles with the "Ready" subtask's rasters.

- `RasterSource` is the central storage for all the raster textures belonging to itself. `Tile`s will share references to these `textures` stored in the `RasterSource`. `Texture` in the datasource will only be deleted when:
  - all the `tile`s referring the `texture` are popped from the `TileCache`, that is when there is only one reference to the `Texture` which is the one stored in the `RasterSource` itself.
  - Textures are keyed on TileID's x/y/z and ignores TileID.s and TileID.w (similar to how RawDataCache is cached).
  - Note: Maybe Texture deletion logic could be more aggressive and take DataSource RawDataCache into account. Currently this is not being done.

- `Rasters` structure
  - This stores a `shared` reference of the raster texture and the corresponding **overzoomed** `TileID`. This `TileID` is used to determine the texture `offset` and `position` when the raster tile is overzoomed.
  - Tile stores a vector of these `Rasters`, which is then used to bind and upload the raster textures at draw time.

- Raster sampler shader injection
  - Any style which defines a `rasters` (values: `color`/`normal`/`custom`), will have shader injections to define texture/raster samplers. The number of samplers injected will be dependent on the number of rasters textures a tile has.

Note: We now have a `default` **1x1 black texture**. If for any reason (no valid texture data from a raster datasource, or bad texture image, etc) texture creation fails, a default 1x1 black pixel texture is created to avoid having random texture unit state set at draw time. This now mimics the behavior we have on JS side.

Generic Note: The commit history here has a couple of implementation for TileManagement. I will try my best to squash some of these commits, but appologies if the history looks dirty. It will be better not to look at the commit history while reviewing the PR and to instead look at the actual diff with master.

### DIFF
--- a/core/src/gl/shaderProgram.cpp
+++ b/core/src/gl/shaderProgram.cpp
@@ -431,12 +431,30 @@ void ShaderProgram::setUniformMatrix4f(const UniformLocation& _loc, const glm::m
     }
 }
 
-void ShaderProgram::setUniformf(const UniformLocation& _loc, const UniformArray& _value) {
+void ShaderProgram::setUniformf(const UniformLocation& _loc, const UniformArray1f& _value) {
     if (!use()) { return; }
     GLint location = getUniformLocation(_loc);
     if (location >= 0) {
         bool cached = getFromCache(location, _value);
         if (!cached) { glUniform1fv(location, _value.size(), _value.data()); }
+    }
+}
+
+void ShaderProgram::setUniformf(const UniformLocation& _loc, const UniformArray2f& _value) {
+    if (!use()) { return; }
+    GLint location = getUniformLocation(_loc);
+    if (location >= 0) {
+        bool cached = getFromCache(location, _value);
+        if (!cached) { glUniform2fv(location, _value.size(), (float*)_value.data()); }
+    }
+}
+
+void ShaderProgram::setUniformf(const UniformLocation& _loc, const UniformArray3f& _value) {
+    if (!use()) { return; }
+    GLint location = getUniformLocation(_loc);
+    if (location >= 0) {
+        bool cached = getFromCache(location, _value);
+        if (!cached) { glUniform3fv(location, _value.size(), (float*)_value.data()); }
     }
 }
 
--- a/core/src/gl/shaderProgram.cpp
+++ b/core/src/gl/shaderProgram.cpp
@@ -297,7 +297,14 @@ std::string ShaderProgram::applySourceBlocks(const std::string& source, bool fra
     //     }
     // }
 
-    return sourceOut.str();
+    // Certain graphics drivers have issues with shaders having line continuation backslashes "\".
+    // Example raster.glsl was having issues on s6 and note2 because of the "\"s in the glsl file.
+    // This also makes sure if any "\"s are present in the shaders coming from style sheet will be
+    // taken care of.
+    auto str = sourceOut.str();
+    std::regex backslashMatch("\\\\\\s*\\n");
+
+    return std::regex_replace(str, backslashMatch, " ");
 }
 
 void ShaderProgram::checkValidity() {
--- a/core/src/tile/tileManager.cpp
+++ b/core/src/tile/tileManager.cpp
@@ -220,7 +220,7 @@ void TileManager::updateTileSet(TileSet& _tileSet, const ViewState& _view,
                     }
                     m_tilesInProgress++;
 
-                } else if (!bool(entry.task) ||
+                } else if (!bool(entry.task) || !entry.m_mainTaskLoaded ||
                            (entry.isCanceled() &&
                             (entry.task->sourceGeneration() < generation))) {
                     // Start loading when no task is set or the task stems from an
@@ -407,6 +407,7 @@ void TileManager::loadTiles() {
             entry.task = task;
             loadRasterTasks(tileSet.source->rasterSources(), entry.task, tileId);
             if (tileSet.source->loadTileData(std::move(task), m_dataCallback)) {
+                entry.m_mainTaskLoaded = true;
                 m_loadPending++;
             } else {
                 // Set canceled state, so that tile will not be tried
--- a/core/src/tile/tileManager.cpp
+++ b/core/src/tile/tileManager.cpp
@@ -84,7 +84,7 @@ bool TileManager::removeDataSource(DataSource& dataSource) {
                 tile.second.cancelTask();
             }
             // Remove the textures for this data source
-            it->source->clearTextures();
+            it->source->clearRasters();
             // Remove the tile set associated with this data source
             it = m_tileSets.erase(it);
             removed = true;
@@ -472,8 +472,8 @@ void TileManager::removeTile(TileSet& _tileSet, std::map<TileID, TileEntry>::ite
 
     // Remove tile from set
     _tileIt = _tileSet.tiles.erase(_tileIt);
-    //Remove textures from this DS
-    _tileSet.source->clearTexture(id);
+    // Remove textures from this DS
+    _tileSet.source->clearRaster(id);
 }
 
 bool TileManager::updateProxyTile(TileSet& _tileSet, TileEntry& _tile,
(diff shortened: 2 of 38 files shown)


core/src/tile/tileWorker.cpp
```diff
@@ -144,6 +185,10 @@ void TileWorker::setScene(std::shared_ptr<Scene>& _scene) {
     }
 }
 
+void TileWorker::notifyAll() {
+    m_condition.notify_all();
+}
+
```
hjanetzek: For which case is it necessary to wake up the workers other than to enqueue tasks?

tallytalwar: Since tasks are requeued within the worker if a dependent task is not ready, we need to inform/notify the worker to run again on the "conditions" if one of the dependent tasks finishes.

so notifyAll is called from rasterCallback!

Also, its possible workers are done and waiting for a notification, when a task was cancelled, so all workers are notified on a task cancel from the main thread. (This might have some redundant notifications when a task was cancelled from the main thread and workers are not waiting on anything but are in middle of building a tile)

hjanetzek: As long as !rastersReady in TileWorker the task is re-enqueued, so it should loop until all rasters are available. 
Commenting out line 189 seems to make no difference here - maybe this was needed earlier to work around a different problem?

tallytalwar: I thought that the worker will wait on the conditional variable until
1. condition is modified and satisfied
2. "and" a notification is sent.
http://en.cppreference.com/w/cpp/thread/condition_variable

Because in this case, `m_queue` will not be empty and if it is not required of the conditional variable to be notified to check for wait condition, then I agree we do not need to send the notification. But according to the description of conditional variable, it seems a notification has to be sent explicitly. Correct me here if I am misinterpreting conditional variable usage.

hjanetzek: It will block on ```wait``` *once the condition is not met*, but since the re-enqueue causes either another worker thread to wake up (with notify_one) or lets the current worker continue to run in the next ```while(true)``` iteration since ```!m_queue.empty()```

karimnaaji: We needed this notification at some point where we were making two conditions to stop the worker thread (the initial one that is still present here; enqueue, not running) and when raster tasks were cancelled are ready. 
It is probably ok to remove those notifications now.

core/src/gl/shaderProgram.cpp
```diff
@@ -292,7 +292,14 @@ std::string ShaderProgram::applySourceBlocks(const std::string& source, bool fra
     //     }
     // }
 
-    return sourceOut.str();
+    // Certain graphics drivers have issues with shaders having line continuation backslashes "\".
+    // Example raster.glsl was having issues on s6 and note2 because of the "\"s in the glsl file.
+    // This also makes sure if any "\"s are present in the shaders coming from style sheet will be
+    // taken care of.
+    auto str = sourceOut.str();
+    std::regex backslashMatch("\\\\ *?\\n");
```
bcamper: I'm confused about the ` *?` part of your regex -- removing spaces is usually done by matching `\s`, and the `?` should be unnecessary after the `*`. Shouldn't `\\s*` (double-escaped backslash) work?

tallytalwar: I think I put in `?` for optional space. And I can definitely use what you suggested, will update.

Thanks

core/src/tile/tileWorker.cpp
```diff
@@ -112,23 +112,63 @@ void TileWorker::run(Worker* instance) {
...
+                    rastersReady = false;
+                    break;
+                }
+            }
+        }
+
+        if (!rastersReady) {
+            // enqueue this task again
+            enqueue(std::move(task));
+            continue;
```
hjanetzek: I think there needs to be a better way than to constantly loop (enque - dequeue) in the TileWorker until rasters are available.

tallytalwar: Other thing I had before was to add another conditional wait on the threads but that means other tiles which could be build were just sitting until these dependent rasters are fetched.

tallytalwar: I removed this behavior here: https://github.com/tangrams/tangram-es/pull/700/commits/55effcdd88b39921472927c7d9d948d1f95e0a78

hjanetzek: One option seems to be to renqueue the dependent-tileTask once all the rasters are loaded: Instead of having a fixed ```m_rasterCallback``` one could pass a lambda as callback that binds the dependent task

tallytalwar: That is one approach, but will get more complicated with managing download tasks if there are a number of dependent tasks! Will have to think more about it.

core/src/tile/tileManager.cpp
```diff
@@ -212,18 +233,20 @@ void TileManager::updateTileSet(TileSet& _tileSet, const ViewState& _view,
...
                 }
             } else {
-                if (entry.isLoading()) {
+
+                if (entry.isLoading() && entry.rastersPending() == 0) {
                     if (newTiles) {
                         // check again for proxies
                         updateProxyTiles(_tileSet, visTileId, entry);
                     }
                     m_tilesInProgress++;
```
hjanetzek: ```entry.rastersPending() == 0``` should not be necessary -- if a proxy tile became available for use it does not matter if the actual tile has pending rasters

core/src/tile/tileTask.cpp
```diff
@@ -9,4 +9,10 @@ TileTask::TileTask(TileID& _tileId, std::shared_ptr<DataSource> _source) :
     m_sourceGeneration(_source->generation()),
     m_priority(0) {}
 
+void TileTask::cancel() {
+    for (auto& raster : m_rasterTasks) {
+        raster->cancel();
+    }
+    m_canceled = true;
```
hjanetzek: ```m_rasterTasks``` can be modified concurrently (though with an almost minimal chance) when TileTask::cancel() is called on the TileWorker while m_rasterTasks are added or canceld in TileManager update

core/src/tile/tileWorker.h
```diff
@@ -28,14 +28,6 @@ class TileWorker : public TileTaskQueue {
 
     bool isRunning() const { return m_running; }
 
-    virtual bool checkProcessedTiles() override {
```
tallytalwar: Nice, this was kinda unnecessary in master also.

tallytalwar:
Merging this.

There is a separate issue on handling depth offsets for proxy tiles, which is already present in master. That can be independently worked on.

PR #705, opened 2016-04-27
Descriptive mesh drawing failure errors

karimnaaji:
When the mesh is not able to draw, our logs give no information about the style/shader it is coming from. This mesh will also keep drawing and trying to send uniform to uncompiled shader which leads to GL errors.
- Add a `NOTIFY` macro to send messages once to the log (can be renamed to `LOGN` for consistency with others) 
- Uniforms of uncompiled shader are not anymore asked for their location
- If a shader fails to compile, the style to which this mesh is associated will be notified on the log

### DIFF
--- a/core/src/gl/shaderProgram.cpp
+++ b/core/src/gl/shaderProgram.cpp
@@ -319,7 +319,7 @@ std::string ShaderProgram::getExtensionDeclaration(const std::string& _extension
 }
 
 void ShaderProgram::setUniformi(const UniformLocation& _loc, int _value) {
-    use();
+    if (!use()) { return; }
     GLint location = getUniformLocation(_loc);
     if (location >= 0) {
         bool cached = getFromCache(location, _value);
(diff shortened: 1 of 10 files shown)


core/src/gl/shaderProgram.cpp
```diff
@@ -97,18 +98,21 @@ GLint ShaderProgram::getUniformLocation(const UniformLocation& _uniform) {
 }
 
 bool ShaderProgram::use() {
+    bool valid = true;
 
     checkValidity();
 
     if (m_needsBuild) {
-        build();
+        valid |= build();
```
tallytalwar: not sure if this or'ing is required?

karimnaaji: Actually it's not since the shader program id is tested on line `109`, will remove.

core/src/platform.h
```diff
@@ -100,8 +103,19 @@ void initGLExtensions();
 #endif
 
 #if LOG_LEVEL >= 0
+#define NOTIFY(fmt, ...) do {                                                       \
```
tallytalwar: this is good.

tallytalwar:
Looks good to me. Once travis is ok with it, can be merged.

Issue #706, opened 2016-04-27
Android: setContinuousRendering may be called before map init async task is finished


tallytalwar:
If the scene file has any of the styles marked with `animate: true`, `Tangram::setScene()` calls `setContinuousRendering`, which for android will try to set `RenderMode` of the `MapView` which has not been initialized yet. Hence causing a `NullPointerException` on `android.opengl.GLSurfaceView.setRenderMode(int)`.

Log:
```
D/Tangram (17069): TANGRAM tangram.cpp:94: finish initialize
E/AndroidRuntime(17069): FATAL EXCEPTION: AsyncTask #1
E/AndroidRuntime(17069): Process: com.mapzen.tangram.android, PID: 17069
E/AndroidRuntime(17069): java.lang.RuntimeException: An error occured while executing doInBackground()
E/AndroidRuntime(17069): 	at android.os.AsyncTask$3.done(AsyncTask.java:300)
E/AndroidRuntime(17069): 	at java.util.concurrent.FutureTask.finishCompletion(FutureTask.java:355)
E/AndroidRuntime(17069): 	at java.util.concurrent.FutureTask.setException(FutureTask.java:222)
E/AndroidRuntime(17069): 	at java.util.concurrent.FutureTask.run(FutureTask.java:242)
E/AndroidRuntime(17069): 	at android.os.AsyncTask$SerialExecutor$1.run(AsyncTask.java:231)
E/AndroidRuntime(17069): 	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1112)
E/AndroidRuntime(17069): 	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:587)
E/AndroidRuntime(17069): 	at java.lang.Thread.run(Thread.java:818)
E/AndroidRuntime(17069): Caused by: java.lang.NullPointerException: Attempt to invoke virtual method 'void android.opengl.GLSurfaceView.setRenderMode(int)' on a null object reference
E/AndroidRuntime(17069): 	at com.mapzen.tangram.MapController.setRenderMode(MapController.java:410)
E/AndroidRuntime(17069): 	at com.mapzen.tangram.MapController.nativeInit(Native Method)
E/AndroidRuntime(17069): 	at com.mapzen.tangram.MapController.<init>(MapController.java:142)
E/AndroidRuntime(17069): 	at com.mapzen.tangram.MapController.getInstance(MapController.java:147)
[...]

PR #707, opened 2016-04-27
Fixes #706

tallytalwar:
- Saves the continuous rendering state if one is called before mapView gets initialized.
- On the mapView initialize callback sets the appropriate rendering mode if required.

PR #708, opened 2016-04-27
Fix android setContinuousRendering on init

hjanetzek:
Fixes https://github.com/tangrams/tangram-es/issues/706

### DIFF
--- a/android/tangram/src/com/mapzen/tangram/MapController.java
+++ b/android/tangram/src/com/mapzen/tangram/MapController.java
@@ -119,8 +119,11 @@ public class MapController implements Renderer {
 
     /**
      * Construct a MapController using a custom scene file
-     * @param context Context in which the map will function; the asset bundle for this activity
-     * must contain all the local files that the map will need
+     * @param view GLSurfaceView for the map display; input events from this
+     * view will be handled by the MapController's TouchInput gesture detector.
+     * It also provides the Context in which the map will function; the asset
+     * bundle for this activity must contain all the local files that the map
+     * will need.
      * @param sceneFilePath Location of the YAML scene file within the assets directory
      */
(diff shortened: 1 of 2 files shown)


android/tangram/src/com/mapzen/tangram/MapController.java
```diff
@@ -123,35 +123,13 @@ private Bitmap capture() {
...
-        return new MapController(context, sceneFilePath);
-    }
-
-    /**
-     * Set the view in which the map will be drawn
-     * @param view GLSurfaceView where the map will be displayed; input events from this view will
-     * be handled by the resulting MapController
-     */
-    void setView(GLSurfaceView view) {
         touchInput = new TouchInput(view.getContext());
```
tallytalwar: I do not think we should allow to have touch handling if the view is not set. This change will allow for one.

tallytalwar:
I think for now ... https://github.com/tangrams/tangram-es/pull/707 is has the minimal change, which will work.
A more API changing change for this will mean redoing a lot of API demo work.
We can talk more on this later, when we are done with SDC!

android/tangram/src/com/mapzen/tangram/MapController.java
hjanetzek: I thought MapView is our external API now (setView is package visible only) and GLSurfaceView is  added to the Fragment in postExecute (again) so it cant get input before complete initialiazation

hjanetzek:
ok

tallytalwar:
This looks good to me.
@blair1618 it will be good for you to have a look at this one.

blair1618:
Yeah this looks good, just update the Javadoc for `MapController()`. I would also prefer the new method be called `init()` to correspond to `nativeInit()` but that's a smaller detail.

PR #709, opened 2016-04-28
Fix tile manager crash on app exit

karimnaaji:
Storing the workers as reference within the Tile Manager is not entirely safe when the workers get deleted in the following scenario:
- Data source runs a network request
- App exits
- Tile worker gets dereferenced from main thread and deallocated
- Data source callback gets invoked, and tries to enqueue on null reference worker

A solution to fix this is to store the tile worker as a pointer, and test its validity before enqueue the worker task when the callback gets invoked.

Backtrace of crash (not main thread):
```
(lldb) bt
* thread #2: tid = 0x3d65a4, 0x00000001002a8ad8 tangram`Tangram::TileManager::TileManager(this=0x0000700000199808, task=std::__1::shared_ptr<Tangram::TileTask>::element_type @ 0x00006000000eb618 strong=1 weak=1)::$_0::operator()(std::__1::shared_ptr<Tangram::TileTask>&&) const + 72 at tileManager.cpp:23, queue = 'NSOperationQueue 0x60000003c1c0 :: NSOperation 0x6100000438a0 (QOS: LEGACY)', stop reason = EXC_BAD_ACCESS (code=EXC_I386_GPFLT)
  * frame #0: 0x00000001002a8ad8 tangram`Tangram::TileManager::TileManager(this=0x0000700000199808, task=std::__1::shared_ptr<Tangram::TileTask>::element_type @ 0x00006000000eb618 strong=1 weak=1)::$_0::operator()(std::__1::shared_ptr<Tangram::TileTask>&&) const + 72 at tileManager.cpp:23
[...]

PR #711, opened 2016-04-29
osx: cancel download tasks when exiting mainloop

hjanetzek:
This is analog to how it works for glfw/linux. I'm not sure if one has to check the cancel state in the completionHandler.

### DIFF
--- a/osx/src/main.cpp
+++ b/osx/src/main.cpp
@@ -381,6 +381,8 @@ int main(int argc, char* argv[]) {
 
     }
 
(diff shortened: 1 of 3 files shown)


tallytalwar:
This looks good to me.

Irrespective of this, we should any how have the fixes in https://github.com/tangrams/tangram-es/pull/709, which handle destruction in a more graceful way.

tallytalwar:
Merging this.

PR #712, opened 2016-04-29
Gl check errors

karimnaaji:
When a gl error happens we have no information on the line related to an error, we just know that an error happened during the last frame without any information on the related GL call that has produced this error. This PR adds a macro to check for GL errors after gl function calls when `TANGRAM_DEBUG` is defined. 

- Don't check for gl errors unless `TANGRAM_DEBUG` is defined
- Consistently give line number / file / gl function call that relates to a gl error

### DIFF
--- a/core/src/gl/error.cpp
+++ b/core/src/gl/error.cpp
@@ -7,7 +7,8 @@ std::unordered_map<GLenum, std::string> Error::s_GlErrorCodesToStrings = {
         {GL_INVALID_ENUM, "GL_INVALID_ENUM"},
         {GL_INVALID_VALUE, "GL_INVALID_VALUE"},
         {GL_INVALID_OPERATION, "GL_INVALID_OPERATION"},
-        {GL_OUT_OF_MEMORY, "GL_OUT_OF_MEMORY"}
+        {GL_OUT_OF_MEMORY, "GL_OUT_OF_MEMORY"},
+        {GL_INVALID_FRAMEBUFFER_OPERATION, "GL_INVALID_FRAMEBUFFER_OPERATION"}
     };
 
 bool Error::hadGlError(const std::string& _locationTag) {
@@ -29,10 +30,14 @@ bool Error::hadGlError(const std::string& _locationTag) {
 void Error::glError(const char* stmt, const char* fname, int line) {
     GLenum err = glGetError();
 
-    auto it = s_GlErrorCodesToStrings.find(err);
+    while (err != GL_NO_ERROR) {
+        auto it = s_GlErrorCodesToStrings.find(err);
 
-    if (it != s_GlErrorCodesToStrings.end() && err != GL_NO_ERROR) {
-        LOGE("OpenGL error %s, at %s:%i - for %s\n", it->second.c_str(), fname, line, stmt);
+        if (it != s_GlErrorCodesToStrings.end()) {
+            LOGE("OpenGL error %s, at %s:%i - for %s\n", it->second.c_str(), fname, line, stmt);
+        }
+
+        err = glGetError();
     }
 }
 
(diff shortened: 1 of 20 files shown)


core/src/gl/error.cpp
```diff
@@ -26,4 +26,14 @@ bool Error::hadGlError(const std::string& _locationTag) {
     return false;
 }
 
+void Error::glError(const char* stmt, const char* fname, int line) {
+    GLenum err = glGetError();
```
blair1618: `glGetError` should always be called in a loop because multiple error flags can be set at the same time (even by one function) and they're returned in an arbitrary order (http://docs.gl/es2/glGetError).

core/src/gl/error.h
```diff
@@ -18,10 +18,18 @@ class Error {
...
+    static void glError(const char* stmt, const char* fname, int line);
+
 private:
 
     static std::unordered_map<GLenum, std::string> s_GlErrorCodesToStrings;
 
 };
 
+#ifdef TANGRAM_DEBUG
+#define GL_CHECK(STMT) do { STMT; Tangram::Error::glError(#STMT, __FILE__, __LINE__); } while (0)
```
blair1618: mapbox-gl-native has a pretty clever approach here to allow this macro to return the same value as the function it wraps: https://github.com/mapbox/mapbox-gl-native/blob/master/include/mbgl/gl/gl.hpp#L71

This may be *too* clever. I'm not sure how much the lambda calls would impact performance, but it does simplify usage of the macro. 

At a minimum, we should be able to simply write `GL_CHECK()` if we wrap `STMT` in another pair of parentheses, I think.

karimnaaji: We don't have to wrap `STMT` in a set of parenthesis to be able to write `GL_CHECK()`, the macro would be an extra `;` in that case. 
I'm not sure we want something that clever, there's a pretty light set of gl functions we use that return a value, we already have all of the information we need to debug with this.

blair1618: Yeah I'm cool with that.

karimnaaji:
@blair1618 : do we want to enable this by default on debug builds?

blair1618:
I think we do want this enabled by default on debug builds, unless it's disastrously slow. 

This is looking pretty good! I'll give it a try locally and if nobody has concerns then I'm in favor of merging.

blair1618:
Is it just me, or does this branch now draw nothing except the gl clear color?

karimnaaji:
No there's something going on here. Will fix this.

core/src/gl/shaderProgram.cpp
```diff
@@ -92,7 +93,8 @@ GLint ShaderProgram::getUniformLocation(const UniformLocation& _uniform) {
     }
 
     _uniform.generation = m_generation;
-    _uniform.location = glGetUniformLocation(m_glProgram, _uniform.name.c_str());
+    _uniform.location = glGetUniformLocation(0, _uniform.name.c_str());
```
karimnaaji: @blair1618 this line broke it, it'll be good after fixing this.

blair1618: Hoooooo that's a likely culprit haha

blair1618:
LGTM now. 

I just pushed one small fix and a rebase onto master. The call to `glClear` in the main render loop was missing a `GL_CHECK()`, which demonstrates an important consequence of this error-handling approach: If the GL call that produces an error is missing this macro, the error will be incorrectly reported as originating from the next occurrence of the macro. For example, I put an invalid argument into the `glClear` call when it didn't have the `GL_CHECK` macro and the error was reported as coming from `glBindBuffer` in `RenderState`. I don't think this is a fatal flaw, but it's definitely something we need to be diligent about.

PR #713, opened 2016-05-02
drop labels for which no glyphs are available

hjanetzek:


core/src/text/fontContext.cpp
```diff
@@ -185,6 +185,11 @@ bool FontContext::layoutText(TextStyle::Parameters& _params, const std::string&
                      (metrics.aabb.y + height * 0.5) * TextVertex::position_scale);
 
     auto it = _quads.begin() + quadsStart;
+    if (it == _quads.end()) {
```
tallytalwar: Why not just check for size of `_quads` at the beginning before doing any textWrapping stuff?

hjanetzek: The _quads are set during textWrap/drawShapeRange. 

Actually the line.shapes().size() test at the beginning already checks for no-glyphs-at-all. So this new test only catches lines for which only whitespaces were found (and this really happend when no fallbacks were loaded, making the map unusable slow for Cairo). 

Should we rather drop labels if they have any missing glyphs?

tallytalwar: What will happen currently if a label is missing a glyph, will it show a default quad or an empty space?

If so, then yes it makes sense to drop the label if its missing any glyph required by it to be rendered completely.

hjanetzek: Missing glyphs will just be ignored - i.e. not added to the LineLayout and not adding 'empty space' to it

PR #715, opened 2016-05-02
[rfc] raster source

hjanetzek:
PR is just for discussing the idea!

### DIFF
--- a/core/src/tile/tileManager.cpp
+++ b/core/src/tile/tileManager.cpp
@@ -27,13 +27,13 @@ TileManager::TileManager(TileTaskQueue& _tileWorker) : m_workers(_tileWorker) {
         }
     }};
 
-    m_rasterCallback = TileTaskCb{[this](std::shared_ptr<TileTask>&& rasterTask) {
-        // If a rasterTask ever reach this point its hasRaster must have been set!
-        // assert(rasterTask->hasRaster());
-        // if (rasterTask->hasRaster()) {
-        //     m_workers.notifyAll();
-        // }
-    }};
+    // m_rasterCallback = TileTaskCb{[this](std::shared_ptr<TileTask>&& rasterTask) {
+    //     // If a rasterTask ever reach this point its hasRaster must have been set!
+    //     // assert(rasterTask->hasRaster());
+    //     // if (rasterTask->hasRaster()) {
+    //     //     m_workers.notifyAll();
+    //     // }
+    // }};
 }
 
 TileManager::~TileManager() {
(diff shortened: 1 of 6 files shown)


core/src/tile/tileManager.cpp
```diff
@@ -415,7 +415,13 @@ void TileManager::loadTiles() {
         auto tileIt = tileSet.tiles.find(tileId);
         auto& entry = tileIt->second;
 
-        if (entry.task && entry.rastersPending() > 0 && !entry.isCanceled()) {
+        // FIXME: When can this happen?
+        if (entry.isCanceled()) {
```
tallytalwar: I think this can happen because of this: https://github.com/tangrams/tangram-es/pull/715/commits/76c5e007878847342214ff429988fc62de483529#diff-88eb1c828b234b81cac47e5e9c23ab2bR242

core/src/tile/tileManager.cpp
```diff
@@ -376,26 +376,40 @@ void TileManager::enqueueTask(TileSet& _tileSet, const TileID& _tileID,
...
+
+        } else if (m_loadPending < MAX_DOWNLOADS) {
+            rasterTasks.push_back(rasterTask);
+            // Load tile if it is not in raw cache.
+
+            if (rasterSource->loadTileData(std::move(rasterTask),
+                                           TileTaskCb{[this, _tileTask](auto rasterTask) {
+                                                   auto task = _tileTask;
+                                                   if (task->hasTile()) {
+                                                       m_workers.enqueue(std::move(task));
```
tallytalwar: This makes sense to me.

Instead of enqueuing in the tileWorker you make it to the rasterTask callback's responsibility to enqueue it.

But what if we have more than 1 dependent rasters, that will make the same tile task enqueued more than once!

tallytalwar: Maybe before we enqueue the task again, we can check if all the dependent tasks have their rasters, in that case it will only enqueue the task again when all the rasters have been fetched!

hjanetzek: umm, I think there is some room to simplify this :) 

raster tasks can run on the worker completely independent of the geom task. when any task finishes it requests a redraw. then the geom-tile can pick their raster textures in the next TileManager update.. tomorrow more

tallytalwar: Brainstorming. ... while walking now

Dependent rasters need not be sent to workers at all.

Whenever geom task is done the tile is added to this task.
Tilemanager on its next update can add rafter textures to this built tile
in the task if it's corresponding dependent raster tasks are done
(hasRaster) and then move this tile to entry and delete the main geom task.

If in the next tilemanager update, rasters for this task are not done then
the task  continues to hold the built tile (task not deleted) and is
checked in next tilemanager update.

tallytalwar: This simplifies the dependent task callback to be just a bool (true).. to
make sure tilemanager update is run whenever a dependent download finishes.

tallytalwar: https://github.com/tangrams/tangram-es/compare/raster-source-rasterCallbacks...varun/raster-task-rfc

^^ Implements what I was describing in the above comments. A quick run looks good on a tethered mobile network.

hjanetzek: @tallytalwar yeah this is very similar to what I had in mind. However the decoding of images should still be done out of the main thread. It may be possible to do this as part of the download task with a RasterTileTask and do the decoding in RasterSource::onTileLoaded

tallytalwar: Yup RasterSource::raster can be called from RasterSource::onTileLoaded to put the rasters in m_textures.

A call later from tilemanager will simply fetch the respective raster from m_textures.

tallytalwar: @hjanetzek this should take care of it: https://github.com/tangrams/tangram-es/commit/8a3f49dbf46bf9e0466ecc013583d56bc213b9a0

I can cherry pick these last 2 commits to the raster-source branch, if we both are happy with this. I am :D

PR #716, opened 2016-05-03
osx: safely cancel download tasks on shutdown

hjanetzek:


osx/src/platform_osx.mm
```diff
@@ -176,6 +189,10 @@ void cancelUrlRequest(const std::string& _url) {
 
 void finishUrlRequests() {
 
+    std::lock_guard<std::mutex> lock(s_urlRequestsMutex);
```
tallytalwar: we can probably wrap this in its own scope.

hjanetzek: yeah, I think so as well

PR #717, opened 2016-05-04
[rfc] raster source

hjanetzek:
- specialize loadTileData for RasterSource

TODO 
- check again task cancellation logic

Nice-to-have
- raster textures could be stored with map\<tileID, weak_ptr\> and custom deleter so that they get automatically removed when the last reference is dropped.

Issues:
- possible duplicate raster downloads at a time for tiles referencing raster with zoom \> maxZoom of RasterSource, or when more than one source references a RasterSource

- orphaned textures: dependent textures may remain in RasterSource::m_textures when the parent task was canceled before RasterSource::onTileLoaded succeeds.

- rasters from rawCache are again decoded on the main-thread. This might not be a big issue though

### DIFF
--- a/scenes/raster-double.yaml
+++ b/scenes/raster-double.yaml
@@ -1,0 +1,35 @@
+sources:
+    terrain-normals:
+        type: Raster
+        url: https://terrain-preview.mapzen.com/normal/{z}/{x}/{y}.png
+    mapzen-osm:
+        type: TopoJSON
+        url: https://vector.mapzen.com/osm/all/{z}/{x}/{y}.topojson
+        rasters: [basemap, terrain-normals]
+    basemap:
+        type: Raster
+        #url: http://a.tile.stamen.com/terrain-background/{z}/{x}/{y}.jpg
+        url: http://a.tile.openstreetmap.org/{z}/{x}/{y}.png
+        #rasters: [terrain-normals]
+
+styles:
+    normals:
+        base: polygons
+        raster: custom
+        shaders:
+            blocks:
+                color: |
+                    color = sampleRaster(0); // color from first raster (basemap)
+                normal: |
+                    normal = normalize(sampleRaster(1).xyz * 2. - 1.); // normal from second raster (normal tiles)
+
+lights:
+    light2: { type: directional, direction: [1, 1, -.9], diffuse: 0.8, ambient: 0.3 }
+
+layers:
+    earth:
+        data: { source: mapzen-osm }
+        draw:
+            normals:
+                color: white
+                order: 0
--- a/scenes/raster-simple.yaml
+++ b/scenes/raster-simple.yaml
@@ -1,0 +1,13 @@
+sources:
+    stamen-terrain:
+        type: Raster
+        #url: http://a.tile.stamen.com/terrain-background/{z}/{x}/{y}.jpg
+        url: http://a.tile.openstreetmap.org/{z}/{x}/{y}.png
+
+layers:
+    terrain:
+        data: { source: stamen-terrain }
+        draw:
+            raster:
+                color: [0.5, 0.5, 0.5]
+                order: 0 # draw on bottom
--- a/scenes/raster-terrain.yaml
+++ b/scenes/raster-terrain.yaml
@@ -1,0 +1,39 @@
+sources:
+    terrain-normals:
+        type: Raster
+        url: https://terrain-preview.mapzen.com/normal/{z}/{x}/{y}.png
+    mapzen-osm:
+        type: TopoJSON
+        url: https://vector.mapzen.com/osm/all/{z}/{x}/{y}.topojson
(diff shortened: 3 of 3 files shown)


tallytalwar:
I agree on the nice to have. I think I did add a todo for the same.. in my current implementation of counting references and only deleting if its the last reference (ugly :|)

core/src/data/rasterSource.cpp
```diff
@@ -37,37 +37,37 @@ std::shared_ptr<TileData> RasterSource::parse(const TileTask& _task, const MapPr
 
 }
 
+std::shared_ptr<TileTask> RasterSource::createTask(TileID _tileId) {
```
tallytalwar: `RasterSource` does not need to have its own implementation for `createTask`. It is same as the uber `DataSource::createTask`.

hjanetzek: It could first check if the texture is already available - Didn't came to add this yet

tallytalwar:
Also not sure if we need `hasRaster` and `rasterReady` now. Previously that was needed for tileWorker.

We can do with TileTask::hasData, as is done for all other tasks.

hjanetzek:
Right, that was also on my TODO to check again

tallytalwar:
Some comments on the issues mentioned above:

1. possible duplicate raster downloads at a time for tiles referencing raster with zoom > maxZoom of RasterSource, or when more than one source references a RasterSource**
  - RawCache/RasterSource::m_textures should be able to return the decoded texture immediately in such situations, right?

2. orphaned textures: dependent textures may remain in RasterSource::m_textures when the parent task was canceled before RasterSource::onTileLoaded succeeds.
  - Maybe we can store a reference of the parent in the dependent task, and only put the textures in, when the parent task is not cancelled. But this might still require some synchronization. Another way we can handle this is to do a clearRaster when a tileTask is cancelled. (I think currently this only happens when a tile is thrown out of tilecache).

3. rasters from rawCache are again decoded on the main-thread. This might not be a big issue though
  - This should only when RasterSource::m_textures does not have a reference of the decoded texture. Very unlikely though, isn't it?

hjanetzek:
1. Two downloads for the same raster can be initiated during one TileManager update. But it's no big deal.
2. This sounds complicated. Maybe it's better to just have a LRU cache for textures (also instead of the rawcache)
3. yep

One more issue: RasterSource may not be disposed on the main-thread when a task keeps a reference to it. So it must be ensured to dispose the texture cache safely

core/src/data/rasterSource.cpp
```diff
@@ -10,10 +10,86 @@
...
+        }
+    }
+
+    void onProcess(TileBuilder& _tileBuilder) override {
+
+        auto source = reinterpret_cast<RasterSource*>(m_source.get());
+
+        if (!m_texture) {
+            // Decode texture data
+            m_texture = source->createTexture(*rawTileData);
```
tallytalwar: can we put in an assert here that m_texture was never set before!

But, I do not think there is no code path currently which will make onProcess be called twice for the same RasterTask!

core/src/tile/tileManager.cpp
```diff
@@ -18,7 +18,10 @@ TileManager::TileManager(TileTaskQueue& _tileWorker) : m_workers(_tileWorker) {
 
     m_tileCache = std::unique_ptr<TileCache>(new TileCache(DEFAULT_CACHE_SIZE));
 
+    // Callback to pass task from Download-Thread to Worker-Queue
     m_dataCallback = TileTaskCb{[this](std::shared_ptr<TileTask>&& task) {
+        if (task->isReady()) { return; }
```
tallytalwar: there should be a `requestRender` if the task is ready.

core/src/tile/tileManager.cpp
```diff
@@ -352,31 +351,36 @@ void TileManager::enqueueTask(TileSet& _tileSet, const TileID& _tileID,
...
+                                   [&](auto& task, auto id){ return task->subTaskId() < id; });
+
+        if (it != subTasks.end() && (*it)->subTaskId() == subTaskId) { continue; }
+
         TileID subTileID = tileID;
         if (subTileID.z > subSource->maxZoom()) {
             subTileID = subTileID.withMaxSourceZoom(subSource->maxZoom());
         }
-        auto subTask = subSource->createTask(subTileID, true);
+        auto subTask = subSource->createTask(subTileID, index+1);
```
tallytalwar: `index+1` == `subTaskId`.

hjanetzek: yeah.. I think it's better now :)

tallytalwar: Yup... the the change in https://github.com/tangrams/tangram-es/pull/717/commits/08d627357e333889b11b7a903a0581c8aab29476

PR #720, opened 2016-05-05
Scene Import

### LINKED ISSUE #714: Scene imports - Ability to "import" and modify another scene file inside tangram scene yaml
Refer JS PR and conversation in the PR for all conceptual details: https://github.com/tangrams/tangram/pull/278

tallytalwar:
For conceptual description refer JS PR: https://github.com/tangrams/tangram/pull/278

Fixes #714

TODO:
- [x] Maintain relative paths on imported assets, example normalize texture paths with respect to the import scene file, etc
-  Network request for imported scenes and handle downloading of all the referenced assets in the scene. (to be done in a separate PR, as this will require more work on the networking side of things) Refer: https://github.com/tangrams/tangram-es/pull/776
- [x] Unit tests

### DIFF
--- a/core/src/scene/importer.cpp
+++ b/core/src/scene/importer.cpp
@@ -57,11 +57,12 @@ void Importer::normalizeSceneImports(Node& root, const std::string& parentPath) 
 void Importer::setNormalizedTexture(Node& texture, const std::vector<std::string>& names,
         const std::string& parentPath) {
 
-    for (auto& name : names) {
+    for (size_t index = 0; index < names.size(); index++) {
+        auto& name = names[index];
         if (m_textureNames.find(name) == m_textureNames.end()) {
             m_textureNames.insert(name);
             if (names.size() > 1) {
-                texture.push_back(normalizePath(name, parentPath));
+                texture[index] = normalizePath(name, parentPath);
             } else {
                 texture = normalizePath(name, parentPath);
             }
@@ -131,7 +132,7 @@ void Importer::normalizeSceneTextures(Node& root, const std::string& parentPath)
 bool Importer::loadScene(const std::string& path) {
 
     auto scenePath = path;
-    auto sceneString = stringFromFile(setResourceRoot(scenePath.c_str()).c_str(), PathType::resource);
+    auto sceneString = stringFromFile(scenePath.c_str(), PathType::resource);
     auto sceneName = getFilename(scenePath);
 
     if (m_scenes.find(sceneName) != m_scenes.end()) { return true; }
--- a/core/src/scene/importer.cpp
+++ b/core/src/scene/importer.cpp
@@ -1,0 +1,124 @@
+#include "importer.h"
+#include "platform.h"
+#include "util/topologicalSort.h"
+#include "yaml-cpp/yaml.h"
+
+#include <regex>
+
+using YAML::Node;
+using YAML::NodeType;
+
+namespace Tangram {
+
+Node Importer::applySceneImports(const std::string& scenePath) {
+
+    if (loadScene(scenePath)) {
+        return importScenes(scenePath);
+    }
+
+    return Node();
+}
+
+bool Importer::loadScene(const std::string& scenePath) {
+
+    auto sceneString = stringFromFile(setResourceRoot(scenePath.c_str()).c_str(), PathType::resource);
+    auto sceneName = scenePath;
+
+    try {
+        auto root = YAML::Load(sceneString);
+        auto imports = getScenesToImport(root);
+        std::regex r("[^//]+$");
+        std::smatch match;
+        if (std::regex_search(sceneName, match, r)) { sceneName = match[0]; }
+        for (const auto& import : imports) {
+            // TODO: What happens when parsing fails for an import
+            // Relative path of imports with respect to the original
+            auto importPath = std::regex_replace(scenePath, r, import);
+            loadScene(importPath);
+        }
+        m_scenes[sceneName] = root;
+    }
+    catch (YAML::ParserException e) {
+        LOGE("Parsing scene config '%s'", e.what());
+        return false;
+    }
+    return true;
+}
+
+std::vector<std::string> Importer::getScenesToImport(const Node& scene) {
+
+    std::vector<std::string> sceneNames;
+
+    if (const Node& import = scene["import"]) {
+        if (import.IsScalar()) {
+            sceneNames.push_back(import.Scalar());
+        }
+        else if (import.IsSequence()) {
+            for (const auto& imp : import) {
+                if (imp.IsScalar()) { sceneNames.push_back(imp.Scalar()); }
+            }
+        }
+    }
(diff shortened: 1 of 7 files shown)


tallytalwar:
@karimnaaji you want to have a look at this?

The only thing left for the basic import (non network) is to add unit tests.

karimnaaji:
Sure, I will start to look at this today.

core/src/scene/importer.cpp
```diff
@@ -0,0 +1,251 @@
...
+}
+
+void Importer::setNormalizedTexture(Node& texture, const std::vector<std::string>& names,
+        const std::string& parentPath) {
+
+    for (auto& name : names) {
+        if (m_textureNames.find(name) == m_textureNames.end()) {
+            m_textureNames.insert(name);
+            if (names.size() > 1) {
+                texture.push_back(normalizePath(name, parentPath));
```
karimnaaji: It should not `push_back` on the container here, we should access `texture[current_index]` and update it instead.

core/src/scene/importer.cpp
```diff
@@ -0,0 +1,251 @@
...
+                }
+            }
+        }
+    }
+}
+
+bool Importer::loadScene(const std::string& path) {
+
+    auto scenePath = path;
+    auto sceneString = stringFromFile(setResourceRoot(scenePath.c_str()).c_str(), PathType::resource);
```
karimnaaji: `s_resourceRoot` is static within the platform so once modified we would stick with that modified resource root for any other loaded resource. I guess that `setResourceRoot` should be called again when going out of the recursion, otherwise it would consider the root used for the last imported scene (which may be a different/deeper path).

karimnaaji:
@tallytalwar , this looks good to me! Last thing would be unit tests for some edge cases and clarifying the comments above.

core/src/scene/importer.cpp
tallytalwar: Yes thats a good catch. Instead of calling it again, I can just call it once from `tangram.cpp` and pass just the resultant `scenePath` to the `Importer::loadScene`.

core/src/scene/importer.cpp
tallytalwar: Yup, this is a bug, would have added more textures to the YAML::Node. Will fix this.

tallytalwar: Fixed by 6563ea1

core/src/scene/importer.cpp
tallytalwar: Fixed by 6563ea1

core/src/scene/importer.cpp
```diff
@@ -0,0 +1,252 @@
...
+        }
+    }
+}
+
+void Importer::setNormalizedTexture(Node& texture, const std::vector<std::string>& names,
+        const std::string& parentPath) {
+
+    for (size_t index = 0; index < names.size(); index++) {
+        auto& name = names[index];
+        if (m_textureNames.find(name) == m_textureNames.end()) {
```
karimnaaji: I don't think that we need that test. If we have two scenes that reference the same texture, this will fail to normalize the texture path where one imports the other, we will fail in normalizing the path and the second path will try to load a resource that is not relevant to its own resource root.
In general I think there's also an issue with the texture management within the scene, where they should be stored by URI at some point. I will create an issue and look into fixing that texture management.

karimnaaji:
https://github.com/tangrams/tangram-es/commit/e7c3b696cb6d9da4ba19d2f45f1794a6163aa690 looks good!

core/src/scene/importer.cpp
```diff
@@ -0,0 +1,257 @@
...
+    }
+}
+
+bool Importer::loadScene(const std::string& path) {
+
+    auto scenePath = path;
+    auto sceneString = stringFromFile(scenePath.c_str(), PathType::resource);
+    auto sceneName = getFilename(scenePath);
+
+    if (m_scenes.find(sceneName) != m_scenes.end()) { return true; }
```
blair1618: Differentiating scenes by their file name alone seems inadequate. If a scene imports both `folder_a/scene.yaml` and `folder_b/scene.yaml` then by this logic, I think the second will be skipped. 

Also, if this is meant to be an early exit from the function, it should happen before the expensive call to `stringFromFile()`.

tallytalwar: Sure both are legitimate concerns, will fix these.

core/src/scene/importer.cpp
```diff
@@ -0,0 +1,257 @@
...
+std::vector<std::string> Importer::getImportOrder() {
+
+    std::vector<std::pair<std::string, std::string>> dependencies;
+
+    for (const auto& scene : m_scenes) {
+        const auto& name = scene.first.k;
+        const auto& sceneRoot = scene.second;
+        for (const auto& import : getScenesToImport(sceneRoot)) {
+            auto importName = getFilename(import);
+            // TODO: fixme do not allow cycles
```
blair1618: The current implementation seems to do Very Bad Things when there is an import cycle. There's clearly not one 'correct' way to resolve a cycle, but we definitely can't crash. 

When `topologicalSort` encounters a cycle, it treats it as an error in the inputs and returns an empty list. You can check for this result and respond appropriately - for instance, using only the root scene and logging an error.

tallytalwar: Yeah ... I did notice that and put in a TODO to think more on the solution.
We should not probably load/try fix the cycle right? Just do not load the scene if a cycle is found or maybe just load the default scene with no imports when a cycle is found.,

core/src/scene/importer.cpp
```diff
@@ -0,0 +1,257 @@
...
+
+    if (Node styles = root["styles"]) {
+
+        for (auto styleNode : styles) {
+
+            auto style = styleNode.second;
+            //style->texture
+            if (Node texture = style["texture"]) {
+                if (texture.IsScalar()) {
+                    setNormalizedTexture(texture, {texture.Scalar()}, parentPath);
```
blair1618: Ok, this is gonna be somewhat tricky... the values being replaced from here down in this function, none of them are strictly URLs. They can also refer to named textures declared in the scene file.

tallytalwar: Good you pointed this out. I missed it before.Will handle this in following commits. Thanks

core/src/scene/importer.cpp
tallytalwar: Fixed by 5d98937

core/src/scene/importer.cpp
tallytalwar: Fixed by 5d98937

core/src/scene/importer.cpp
tallytalwar: Fixed by 359eed1ca3a2af641b06e1114ca3371224e95990

tallytalwar:
Closing in favor of #776.

PR #721, opened 2016-05-05
Unified style

karimnaaji:
See https://github.com/tangrams/tangram/pull/306 for more details about the unified style.

Tangram-js uses a _unified_ point style since the architecture uses the same shaders for both text and sprites rendering (due to the use of canvas). Here, the implementation is different since we need to use different draw call since the rendering techniques for text and points are different (signed distance field text rendering shaders vs texture atlas sprites shaders):

- Point style now _redirects_ the draw rule to a text style when the draw rule is considered as _unified_
- A style param is flagged as unified when related to a text within a point style
- At draw rule merging: any merge where an ubber style param is unified would flag the draw rule as unified
- The draw rule are now keyed differently for text and icons so there is no conflicts when accessing the property at feature building time
- Line/polygon geometry for text behaves differently when unification happens (see https://github.com/tangrams/tangram-es/pull/721/commits/a19ddc3b7e5d87ce5a7cae0ff27b2277035ce6b0)

This new unification involves two draw calls per frame when unified (one for the text, and one for the sprites associated).


external/alfons
```diff
@@ -1 +1 @@
-Subproject commit afb61f7143d6d0a78032123b61ade297a90691c0
+Subproject commit cfdab1782c3294c54c0732b9c4868024af374e3d
```
hjanetzek: reverting alfons submodule again

hjanetzek:
- For what is it necessary to change unique_ptr<Label> to shared_ptr?
  The reference Label::m_parent could be a raw pointer since both Text- and SpriteLabels are owned    by IconMesh and will be destructed with it at the same time

- Line geometries for icons need to be handled differently in TextStyleBuilder (or could be ignored instead)

karimnaaji:
- You're right, we can safely use the raw pointer from the unique_ptr. I'll make that change.
- We could ignore the line otherwise it would behave incorrectly, checking what's the behavior in -js first.

core/src/scene/sceneLoader.cpp
```diff
@@ -626,15 +626,13 @@ void SceneLoader::loadStyleProps(Style& style, Node styleNode, Scene& scene) {
...
-
         if (auto pointStyle = dynamic_cast<PointStyle*>(&style)) {
-
             const std::string& textureName = textureNode.Scalar();
             auto atlases = scene.spriteAtlases();
             auto atlasIt = atlases.find(textureName);
             if (atlasIt != atlases.end()) {
                 pointStyle->setSpriteAtlas(atlasIt->second);
-            } else {
+            }  else {
```
tallytalwar: extra space!

core/src/scene/styleParam.cpp
```diff
@@ -43,14 +34,48 @@ const std::map<std::string, StyleParamKey> s_StyleParamMap = {
     {"outline:width", StyleParamKey::outline_width},
     {"outline:style", StyleParamKey::outline_style},
     {"priority", StyleParamKey::priority},
-    {"repeat_distance", StyleParamKey::repeat_distance},
-    {"repeat_group", StyleParamKey::repeat_group},
     {"size", StyleParamKey::size},
     {"sprite", StyleParamKey::sprite},
     {"sprite_default", StyleParamKey::sprite_default},
     {"style", StyleParamKey::style},
+    {"text:anchor", StyleParamKey::text_anchor},
```
tallytalwar: I know we came up with this together, but we should refactor this later to avoid adding so many more parameters (if we can).

tallytalwar:
This mostly looks good. WIll be finalizing the review to check some latest code and changes and then we are good with this.

Tested and compared with JS, we get the same behavior, which is nice :D.

core/src/scene/drawRule.h
```diff
@@ -70,6 +70,8 @@ struct DrawRule {
     // 480 (on 32bit arch) or 980 byte for params array.
     std::bitset<StyleParamKeySize> active = { 0 };
 
+    bool unified = false;
+
```
hjanetzek: Why is it necessary to have the unified flag as part of the drawrule? Can't one check in ```PointStyleBuilder::addFeature``` if the drawrule containts text elements?

core/src/scene/sceneLoader.cpp
```diff
@@ -1160,14 +1158,15 @@ Filter SceneLoader::generateNoneFilter(Node _filter, Scene& scene) {
 }
 
 void SceneLoader::parseStyleParams(Node params, Scene& scene, const std::string& prefix,
-                                   std::vector<StyleParam>& out) {
+                                   std::vector<StyleParam>& out, bool unified) {
 
     for (const auto& prop : params) {
+        unified |= prefix == "text";
         std::string key;
```
hjanetzek: What happens when an icon has an empty ```text``` block? The unified property would need to be propagated up. Or one adds some default text property so that it can be identified as icon with text

karimnaaji: What do you mean by that? Can you provide a yaml example to show what you're thinking of.

hjanetzek: A reduced case: This would not create any DrawRule, so one couldn't store the unified property

    points:
        text:

karimnaaji: Ok I see, can this be used though? Maybe if you want to use default params of the text?

tallytalwar: Yeah ... maybe will use the default params for text builder.
Can you also try how this behaves on JS?

hjanetzek: It seems to be allowed since all text params have defaults - even though it's unlikely to be used this way.

karimnaaji: JS doesn't support empty text draw rule.

tallytalwar: @bcamper Should we have support for empty `text` property for an icon style, in which case it will use the default text properties specified in the style?

Issue #643, opened 2016-03-24
IOS font not appearing


haidershahzad:
Hello, Im trying to display vector data from url  which is loading properly like buildings and poi but the labels are not appearing below i have pasted the console log please guide what is the issue, im using xcode 7.2.1. and cmake version 3.4.0

TANGRAM tileManager.cpp:67: add source local
TANGRAM tangram.cpp:99: finish initialize
TANGRAM tangram.cpp:577: setup GL
TANGRAM hardware.cpp:56: Driver supports map buffer: 1
TANGRAM hardware.cpp:57: Driver supports vaos: 1
TANGRAM hardware.cpp:71: Hardware max texture size 4096
TANGRAM hardware.cpp:72: Hardware max combined texture units 8
TANGRAM tangram.cpp:132: resize: 1242 x 2208
TANGRAM tileWorker.cpp:70: Passed new TileBuilder to TileWorker
Failed to read file at path: /var/mobile/Containers/Bundle/Application/40249C13-C398-4624-9D27-D6A170598A5E/tangram.app/fonts/default-400normal.ttf
Failed to read file at path: /var/mobile/Containers/Bundle/Application/40249C13-C398-4624-9D27-D6A170598A5E/tangram.app/fonts/default-400normal.ttf
Failed to read file at path: 
ERROR fontContext.cpp:263: Could not load font file default_400_normal
TANGRAM tileWorker.cpp:70: Passed new TileBuilder to TileWorker
Missing font: error: 1 NotoSans-Regular.ttf
Missing font: error: 1 NotoNaskh-Regular.ttf
Missing font: error: 1 NotoSansHebrew-Regular.ttf
Missing font: error: 1 DroidSansJapanese.ttf
Missing font: error: 1 DroidSansFallback.ttf
Missing font: error: 1 NotoSans-Regular.ttf
Missing font: error: 1 NotoNaskh-Regular.ttf
Missing font: error: 1 NotoSansHebrew-Regular.ttf

tallytalwar:
Hi @haidershahzad 

Yes, I am able to reproduce this. We will look into this.

Thanks

haidershahzad:
What will be the eta of documention ios.

tallytalwar:
Hey Haider.

Currently we have not spent much time exposing a lot of functionality for iOS SDK. Our focus has been for Android, but soon we will be shifting gears and working on iOS SDK and documentation.

Tangram-es is a work in progress project, and we appreciate your desire of using it.

PR #649, opened 2016-03-28
Revise FontContext font loading

karimnaaji:
- Remove platform dependent code
- Let platform layer resolve font path and font binary loading
- Fix #643

PR #732, opened 2016-05-13
Ios fallback fonts

karimnaaji:
Fixes https://github.com/tangrams/tangram-es/issues/643
Related https://github.com/tangrams/tangram-es/pull/649

### DIFF
--- a/core/src/text/fontContext.cpp
+++ b/core/src/text/fontContext.cpp
@@ -268,6 +268,7 @@ auto FontContext::getFont(const std::string& _family, const std::string& _style,
     }
 
     font->addFace(m_alfons.addFontFace(alfons::InputSource(reinterpret_cast<char*>(data), dataSize), fontSize));
+    free(data);
 
     // add fallbacks from default font
     font->addFaces(*m_font[sizeIndex]);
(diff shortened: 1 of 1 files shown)


karimnaaji:
Merging this as is to fix issues we're having on iOS right now. Though, this would need to be cleaned up and to remove platform-dependent code/conditionals from core.

core/src/text/fontContext.cpp
```diff
@@ -57,6 +59,36 @@ FontContext::FontContext() :
             m_font[i]->addFace(m_alfons.addFontFace(alfons::InputSource(fontPath), size));
         }
     }
+#elif defined(PLATFORM_IOS)
```
tallytalwar: We definitely need a refactor of this once we have more knowledge of ios font loading, at least to get rid of the `PLATFORM_IOS` specific handling here.

MajidBashir:
well related to fonts, sometimes application is crashing over fonts

> NOTIFY textStyleBuilder.cpp:300: stroke_width too large: 12.094118 / 2.015686


how can i fix this issue

blair1618:
@MajidBashir this crash doesn't seem to be related to fonts - that "NOTIFY" message is only a warning. If you are still having this problem could you please open a new issue?

PR #733, opened 2016-05-13
[rfc] Moved bundled fonts out of core/resources

tallytalwar:
- Android does not need the tangram bundled fonts so do not include these with libtangram (saved
~5MB of space with the aar, from 11MB to 4.7 MB).
- Refactor other platforms to include the bundled fonts which now live under
{Uber_Path}/resources/fonts along with some test scenes under {Uber_Path}/resources/scenes

TODO:
- [ ] Remove some imgs from libtangram bundle
- [ ] Test on linux


tallytalwar:
It made sense to put the bundled fonts under the root project path, instead of `core`, we can move it accordingly.

Also we need to work around something which is currently not possible to do with gradle for android ... If we could exclude certain assets dirs from `android/tangram/build.gradle`, we would have not required to do this. (Refer: https://github.com/tangrams/tangram-es/blob/master/android/tangram/build.gradle#L35 and http://stackoverflow.com/questions/25910206/gradle-exclude-file-from-android-assets-folder).

blair1618:
Yay for putting the AAR on a diet! 

Strictly speaking, the only "resources" that need to be bundled with the library are the shader source files (and even those could be put into code as part of the build process). The other files are associated with the default test scene, which should be part of our demo applications but not the library itself. 

Perhaps it makes the most sense to move `core/resources/scene.yaml`, `core/resources/img/`, and `core/resources/fonts/` to `scenes/`. Demo applications will then bundle these test scene resources (and Android can exclude specific assets like fonts in the application build process).

hjanetzek:
Moving scene resources to ```scenes/``` sounds good. Each platform could have their own ```copy_resources``` task to explicitly add what is needed

blair1618:
Implemented with #752

PR #736, opened 2016-05-17
icon_text StyleParam

hjanetzek:
use icon_text StyleParam to identify icons with text


blair1618:
This is definitely a good move, but why is the parameter called `icon_text`? The actual syntax uses a parameter named `text`, so why don't we use that? Or `point_text`, to use the names of the actual styles.

karimnaaji:
We can definitely rename that in #721.

Issue #729, opened 2016-05-12
Destroy MapController instance on MapView destroy


tallytalwar:
Currently we only clear the async tasks on MapView::onDestroy.
But if one creates a new activity, it grabs the old/previously initialized MapController instance.

Maybe it makes sense to destroy the mapController instance on a MapView::onDestroy.

ecgreb:
Upon further thought there are times we do indeed want `MapController` to persist when `MapView` and its parent `Activity` are destroyed. For example on device rotation. Maybe instead of automatically destroying the `MapController` in `MapView#onDestroy()` we should add a public method like `MapView#resetMapController()`?

Also as @sarahlensing pointed out currently [`MapView#getMapAsyc(OnMapReadyCallback, String)`](https://github.com/tangrams/tangram-es/blob/master/android/tangram/src/com/mapzen/tangram/MapView.java#L54) ignores the `sceneFilePath` param if there is a `MapController` already in existence. It should probably at least call [`MapController#loadSceneFile(String)`](https://github.com/tangrams/tangram-es/blob/master/android/tangram/src/com/mapzen/tangram/MapController.java#L182) if the scene files are not the same.

blair1618:
It sounds like the practical problem here is that re-initializing a tangram instance with a different scene file doesn't actually load the new scene. This wasn't an intentional design choice and can be easily remedied. I'll submit a PR today.

PR #737, opened 2016-05-17
Compare new and old scene paths on initialization

blair1618:
Previously, if `Tangram::initialize` or `MapView$getMapAsync` were called repeatedly with different scene file paths, the new scene would not be loaded. This change compares the new and old scene file paths in `Tangram::initialize` and loads the new scene if the paths differ. 

Resolves https://github.com/tangrams/tangram-es/issues/729

### DIFF
--- a/core/src/scene/scene.cpp
+++ b/core/src/scene/scene.cpp
@@ -22,14 +22,14 @@ namespace Tangram {
 
 static std::atomic<int32_t> s_serial;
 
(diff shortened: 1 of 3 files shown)


tallytalwar:
Code changes look good to me. Will give these a shot on a device in a test application I have and merge.

ecgreb:
👍

PR #739, opened 2016-05-17
Fix over-zoomed line precision

blair1618:
See https://github.com/tangrams/tangram/issues/267 and https://github.com/tangrams/tangram/pull/310

When a tile is over-zoomed many times, the width attributes of lines (encoded as 16-bit integers) can become small enough, relative to their "source" tile, that they are rounded to zero. This causes lines to disappear on tiles that are sufficiently over-zoomed, particularly when the width is very small to begin with. 

This change moves a re-scaling operation from the polyline building process to the vertex shader, making width precision is undiminished with repeated over-zooming.

### DIFF
--- a/core/resources/shaders/polyline.vs
+++ b/core/resources/shaders/polyline.vs
@@ -96,7 +96,8 @@ void main() {
         // Interpolate between zoom levels
         width += dwdz * clamp(dz, 0.0, 1.0);
         // Scale pixel dimensions to be consistent in screen space
-        width *= exp2(-dz);
+        // and adjust scale for overzooming.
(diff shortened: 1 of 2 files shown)


hjanetzek:
LGTM. 

We could also increase the extrusion_scale factor from 2^12 to 2^15 since an extrusion width of 1 is a line twice as wide as a tile, which would still be a generous upper limit.

blair1618:
Yeah true, I first tried adjusting `extrusion_scale` to gain some precision. It helps, but if you over-zoom far enough you'll eventually reach the same problem. In truth, the position attribute exhibits this problem too, but the results are far less visually noticeable.

tallytalwar:
This looks good. Going to merge.

Issue #409, opened 2015-11-30
Desktop client can crash on unexpected HTTP response


blair1618:
A URL request can sometimes return data in an unexpected form, e.g. when a malformed MVT tile request returns with a status 429 the response body is a JSON object of metadata (instead of a protocol buffer). 

For URL requests processed by OkHTTP on Android, these qualify as errors and so the response body is never sent to Tangram. 

However NSURLSession (and probably libcurl as well) does not count these status codes as errors, it simply passes any response body along to Tangram. For MVT tiles, this causes the protobuff parser to segfault. 

Ideally, the protobuff parser would be robust enough to at least not segfault when data is malformed. But a simpler solution to this particular issue is to check the status codes of HTTP responses received through NSURLSession and libcurl and skip any responses that are not "successful".

blair1618:
Relevant docs for [NSURLSession](https://developer.apple.com/library/ios/documentation/Foundation/Reference/NSURLSession_class/index.html#//apple_ref/occ/instm/NSURLSession/dataTaskWithURL:completionHandler:) and [NSHTTPURLResponse](https://developer.apple.com/library/ios/documentation/Cocoa/Reference/Foundation/Classes/NSHTTPURLResponse_Class/index.html#//apple_ref/occ/instp/NSHTTPURLResponse/statusCode)

blair1618:
Docs for OkHTTP [Response](http://square.github.io/okhttp/2.x/okhttp/com/squareup/okhttp/Response.html) (see `isSuccessful()`)

PR #742, opened 2016-05-19
Add checks to HTTP status codes for NSURLSession requests

blair1618:
Status codes [200, 300) are accepted and their body data are used.

Resolves https://github.com/tangrams/tangram-es/issues/409.

### DIFF
--- a/ios/src/platform_ios.mm
+++ b/ios/src/platform_ios.mm
@@ -144,6 +144,8 @@ bool startUrlRequest(const std::string& _url, UrlCallback _callback) {
 
         int statusCode = [httpResponse statusCode];
 
+        std::vector<char> rawDataVec;
+
         if (error != nil) {
 
             LOGE("Response \"%s\" with error \"%s\".", response, [error.localizedDescription UTF8String]);
@@ -154,11 +156,11 @@ bool startUrlRequest(const std::string& _url, UrlCallback _callback) {
                 statusCode,
                 [[NSHTTPURLResponse localizedStringForStatusCode: statusCode] UTF8String],
                 [response.URL.absoluteString UTF8String]);
+            _callback(std::move(rawDataVec));
 
         } else {
 
             int dataLength = [data length];
-            std::vector<char> rawDataVec;
             rawDataVec.resize(dataLength);
             memcpy(rawDataVec.data(), (char *)[data bytes], dataLength);
             _callback(std::move(rawDataVec));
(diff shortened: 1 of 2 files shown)


tallytalwar:
This looks good.
We might have to do the same for libcurl for linux and rpi platforms.

blair1618:
I thought so too, but the linux platform code does a similar check already https://github.com/tangrams/tangram-es/blob/master/linux/src/urlWorker.cpp#L51

tallytalwar:
Sure. thanks for pointing to the code. Can we use the same status code range in linux then? Currently it seems it considers downloaded data valid only when status code is `200`.

blair1618:
After some IRL discussion with @bcamper it's starting to sound like we should "accept" the response body for all status codes and perform any content checks at the next step when we're trying to use the data. Essentially, this is because it is a common practice to provide a meaningful response body in the event of errors (e.g. a 404 on an image resource can result in an "error image"). At the platform networking layer, we'll just issue a warning for status codes that are likely errors. 

I'll comment again here after adjusting this behavior.

bcamper:
Ideally we should probably try to process the response if it has the expected content type for the resource (e.g. images for raster tiles, protobuf for MVT tiles, etc.). This does involve some guesswork though. And throw a warning on any resources that can't be processed successfully. Of course this is problematic if protobuf decoding is segfaulting on invalid data. Stability should take priority here.

tallytalwar:
So this means error check needs to be done on parsing end?

blair1618:
After a bit more poking around, I'm now feeling more like we should *not* be processing response bodies with "unsuccessful" status codes.

 - The [example code](https://github.com/square/okhttp/blob/master/samples/guide/src/main/java/okhttp3/recipes/AsynchronousGet.java#L39-L48) from OkHTTP checks status codes using the [`isSuccessful()`](http://square.github.io/okhttp/3.x/okhttp/okhttp3/Response.html#isSuccessful--) method to decide whether to read a response body, and our Android SDK has behaved this way since the initial release. 
 - The [HTTP request handling in mapbox gl](https://github.com/mapbox/mapbox-gl-native/blob/master/platform/default/http_file_source.cpp#L333-L390) goes further and only uses response body data when the status code is 200. 

It makes sense for a browser to attempt to process response bodies for "failure" status codes, but since tangram-es is not a browser (yet) I think we should process only "successful" responses until both:

1. we find a use case requiring the engine to process response bodies with "failure" status codes, and
2. we make certain that all code paths that receive HTTP responses are robust to malformed data.

tallytalwar:
Looks good. Merging.

PR #745, opened 2016-05-20
Link with static version of libc++ on Android

blair1618:
Reduces combined library size by 1.3MB and reduces AAR size by 0.5MB


tallytalwar:
LGTM.

Merging and saving 1.3MB of space :D

PR #749, opened 2016-05-24
[wip] Tile label placement

hjanetzek:
Determine initial label placement during tile generation.

shot-2016-06-03_15-32-00
Positions tested during initial placements (discarded labels in purple)

shot-2016-06-03_15-31-50
All labels in final tiles 

TODO:
- [ ] Improve line-label candidate positions

### DIFF
--- a/core/src/labels/label.cpp
+++ b/core/src/labels/label.cpp
@@ -167,7 +167,8 @@ bool Label::canOcclude() {
                         State::skip_transition |
                         State::fading_in |
                         State::sleep |
-                        State::out_of_screen);
+                        State::out_of_screen |
+                        State::dead);
 
     return (occludeFlags & m_state) && !(m_type == Type::debug);
 }
@@ -190,6 +191,8 @@ glm::vec2 Label::center() const {
 }
 
 void Label::enterState(const State& _state, float _alpha) {
+    if (m_state == State::dead) { return; }
+
     m_state = _state;
     setAlpha(_alpha);
 }
@@ -207,42 +210,37 @@ void Label::setAlpha(float _alpha) {
 }
 
 void Label::resetState() {
+    if (m_state == State::dead) { return; }
+
     m_occludedLastFrame = false;
     m_occluded = false;
     m_updateMeshVisibility = true;
     m_dirty = true;
     m_proxy = false;
-    //enterState(State::wait_occ, 0.0);
+    enterState(State::wait_occ, 0.0);
 }
 
 bool Label::update(const glm::mat4& _mvp, const glm::vec2& _screenSize, float _zoomFract, bool _allLabels) {
-
-    if (!_allLabels && (m_state == State::dead || (m_parent && m_parent->state() == State::dead))) {
-        return false;
-    }
 
     m_occludedLastFrame = m_occluded;
     if (m_state != State::fading_out) {
         m_occluded = false;
     }
 
-    bool ruleSatisfied = updateScreenTransform(_mvp, _screenSize, _allLabels);
+    if (m_state == State::dead) {
+        if (!_allLabels) {
+            return false;
+        } else {
+            m_occluded = true;
+        }
+    }
+
+    bool ruleSatisfied = updateScreenTransform(_mvp, _screenSize, !_allLabels);
 
     // one of the label rules has not been satisfied
     if (!ruleSatisfied) {
-        if (m_state == State::wait_occ) {
-            // go to dead state, this breaks determinism, but reduce potential
-            // label set since a lot of discarded labels are discared for line
-            // exceed (lots of tiny small lines on a curve for example), which
-            // won't have their rule satisfied
-            enterState(State::dead, 0.0);
-            pushTransform();
-        } else {
-            enterState(State::sleep, 0.0);
-            pushTransform();
-        }
+        enterState(State::sleep, 0.0);
         return false;
-
     }
 
     // update the view-space bouding box
@@ -306,11 +304,11 @@ bool Label::evalState(const glm::vec2& _screenSize, float _dt) {
             break;
         case State::wait_occ:
             if (m_occluded) {
-                if (m_parent) {
-                    enterState(State::sleep, 0.0);
-                } else {
-                    enterState(State::dead, 0.0);
-                }
+                // if (m_parent) {
+                enterState(State::sleep, 0.0);
+                // } else {
+                //     enterState(State::dead, 0.0);
+                // }
             } else {
                 m_fade = FadeEffect(true, m_options.showTransition.ease,
                                     m_options.showTransition.time);
@@ -320,8 +318,8 @@ bool Label::evalState(const glm::vec2& _screenSize, float _dt) {
             break;
         case State::skip_transition:
             if (m_occluded) {
-                enterState(State::dead, 0.0);
-
+                enterState(State::sleep, 0.0);
+                // enterState(State::dead, 0.0);
             } else {
                 enterState(State::visible, 1.0);
             }
--- a/core/src/labels/label.cpp
+++ b/core/src/labels/label.cpp
@@ -212,12 +212,12 @@ void Label::resetState() {
     m_updateMeshVisibility = true;
     m_dirty = true;
     m_proxy = false;
-    enterState(State::wait_occ, 0.0);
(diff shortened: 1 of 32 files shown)


karimnaaji:
Starting to look over the details of this, looks really good so far!

core/src/labels/labels.cpp
```diff
@@ -169,187 +174,135 @@ void Labels::skipTransitions(const std::vector<std::unique_ptr<Style>>& _styles,
...
+    if (_a.priority != _b.priority) {
+        return _a.priority < _b.priority;
+    }
+
+    auto l1 = _a.label;
+    auto l2 = _b.label;
+
+    // Note: This causes non-deterministic placement, i.e. depending on
+    // navigation history.
+    if (l1->occludedLastFrame() != l2->occludedLastFrame()) {
```
karimnaaji: Is this needed as a sorting rule? So we base the sorting only on things not related to label state.

core/src/labels/labels.cpp
```diff
@@ -169,187 +174,135 @@ void Labels::skipTransitions(const std::vector<std::unique_ptr<Style>>& _styles,
...
+        return l2->occludedLastFrame();
+    }
+
+    // if (l1->options().repeatGroup != l2->options().repeatGroup) {
+    //     return l1->options().repeatGroup < l2->options().repeatGroup;
+    // }
+
+    if (l1->type() == Label::Type::line && l2->type() == Label::Type::line) {
+        // Prefer the label with longer line segment as it has a chance
+        return glm::length(l1->transform().modelPosition1 - l1->transform().modelPosition2) >
```
karimnaaji: We can use `glm::length2`.

core/src/labels/labelCollider.cpp
```diff
@@ -0,0 +1,221 @@
...
+    mvp[3][0] = -1;
+    mvp[3][1] = 1;
+
+    for (auto* label : m_labels) {
+        label->update(mvp, m_screenSize, 1, true);
+
+         m_aabbs.push_back(label->aabb());
+    }
+
+    m_isect2d.resize({m_screenSize.x / 128, m_screenSize.y / 128}, m_screenSize);
```
karimnaaji: Have you noticed any difference using `128`? If not, we could probably use the same grid size as the one in `labels.cpp`. I also noticed we're not using the pixel density to scale those grid, we should probably scale them.

core/src/labels/labelCollider.cpp
```diff
@@ -0,0 +1,221 @@
...
+                      return l1->options().priority > l2->options().priority;
+                  }
+
+                  if (l1->type() == Label::Type::line &&
+                      l2->type() == Label::Type::line) {
+                      // Prefer the label with longer line segment as it has a chance
+                      // to be shown earlier (also on the lower zoom-level)
+                      // TODO compare fraction segment_length/label_width
+
+                      return glm::length(l1->transform().modelPosition1 - l1->transform().modelPosition2) >
```
karimnaaji: Could also use `glm::length2`.

core/src/labels/labels.cpp
```diff
@@ -169,187 +174,135 @@ void Labels::skipTransitions(const std::vector<std::unique_ptr<Style>>& _styles,
...
+    if (l1->hash() != l2->hash()) {
+        return l1->hash() < l2->hash();
+    }
+
+    return l1 < l2;
+}
 
-    struct GroupElement {
-        Label* label;
+void Labels::sortLabels() {
```
karimnaaji: We can inline that one liner if not used elsewhere.

core/src/labels/labels.cpp
hjanetzek: It's not 'needed' to do this. It's a matter of preference of less flickering vs having always the same set of labels shown for a viewport - This is the behavior we had before so I kept it and added the Note.

core/src/labels/labels.cpp
hjanetzek: yup

core/src/labels/labelCollider.cpp
hjanetzek: Hard to say what a good value is for the grid :)
With the default pixel scale a tile is projected to 512x512, so this would make a 4x4 grid which seem reasonable.

core/src/labels/labels.cpp
hjanetzek: yes, it can be inlined again. I've split it up to see the time spent for the different phases more easily in perf

karimnaaji: Oh right that's an issue if we want to see this in perf tools.

core/src/labels/labels.cpp
```diff
@@ -169,187 +174,135 @@ void Labels::skipTransitions(const std::vector<std::unique_ptr<Style>>& _styles,
     }
 }
 
-void Labels::checkRepeatGroups(std::vector<Label*>& _visibleSet) const {
+bool Labels::labelComparator(const LabelEntry& _a, const LabelEntry& _b) {
```
karimnaaji: As the discussion we had in the meeting, we could use the depth (ground distance from camera, or perspective factor already available when projecting to screen) to sort  labels within the same priority level. 
Probably to consider out of this PR.

core/src/labels/spriteLabel.cpp
```diff
@@ -36,59 +35,54 @@ void SpriteLabel::applyAnchor(const glm::vec2& _dimension, const glm::vec2& _ori
...
 void SpriteLabel::updateBBoxes(float _zoomFract) {
     glm::vec2 halfSize = m_dim * 0.5f;
     glm::vec2 sp = m_transform.state.screenPos;
     glm::vec2 dim = m_dim + glm::vec2(m_extrudeScale * 2.f * _zoomFract);
-    m_obb = OBB(sp.x + halfSize.x, sp.y - halfSize.y, m_transform.state.rotation, dim.x, dim.y);
-    m_aabb = m_obb.getExtent();
-}
 
-void SpriteLabel::align(glm::vec2& _screenPosition, const glm::vec2& _ap1, const glm::vec2& _ap2) {
+    if (m_occludedLastFrame) { dim += 2; }
```
karimnaaji: Is that for debug?

hjanetzek: This should reduce flickering. The idea was to increase the OBB of occluded labels so that they can only become active when there is some buffer to other labels and thereby are less likely to get occluded again immediately by minimal map movements. 
https://github.com/tangrams/tangram-es/pull/749/commits/ef535cb73d3ed08d51a7ba31e17549e8c2ca211c

karimnaaji: I see, it's a good idea but I think this value should also be scaled by the pixel scale.

blair1618:
Seems like a good time to consider merging this - it's lingered for a long time now.

karimnaaji:
@blair1618 yes we talked about merging this yesterday and I did a last review, it seems ready to go to me if @hjanetzek doesn't have any blockers on it.

hjanetzek:
yeah, ready to merge. 

♫just wait-ing for travis♫

blair1618:
LGTM :+1:

core/src/labels/label.cpp
```diff
@@ -21,107 +23,80 @@ Label::Label(Label::Transform _transform, glm::vec2 _size, Type _type, Options _
...
 
-            rot = angleBetweenPoints(p1, p2) + M_PI_2;
+            // default heuristic : allow label to be 30% wider than segment
+            float minLength = m_dim.x * 0.7;
 
-            if (rot > M_PI_2 || rot < -M_PI_2) { // un-readable labels
-                rot += M_PI;
-            } else {
-                std::swap(p1, p2);
+            if (_testVisibility && length < minLength) {
```
tallytalwar: No need to test `_testVisibility` here, already done above.

hjanetzek: The name _testVisibility may not be the best choice. It is used to return false and return early when the label is not renderable. In this case when the line segment is too short for the label.

tallytalwar: Ohh I just mean, to do `if (length < minLength)`. `_testVisibility` must be `true` here before of `if (_testVisibility && clipped)` check before.

PR #750, opened 2016-05-24
use shared fontContext for text- and icon-styles

hjanetzek:


tallytalwar:
LGTM.

Will cherry-pick this in url-scene-import work. Once all 0.3.2 patches are done, we can merge this. That said, this change can probably still go in 0.3.2 patched release. So whatever the group decides.

blair1618:
Makes sense, and seems to work fine in the demos. Merging!

PR #752, opened 2016-05-24
Unbundle default scene resources

blair1618:
Resources required for the default demo scene are now located in `scenes/` and the build steps for demo applications now include a step to copy those scene resources as needed. The only files still required to be packaged with the library are GLSL source files, which remain in `core/resources/shaders/`. 

This also adds a CI step to update the CMake version on Travis for OS X builds; the pre-installed version seems to have a bug affecting the `copy_directory` command. 

See also https://github.com/tangrams/tangram-es/pull/733


blair1618:
Ran into some trouble here with the resource copy tasks. The command run by the macro `add_resources` (i.e. `cmake -E copy_directory`) would intermittently fail in the unit tests build target without apparent explanation. https://github.com/tangrams/tangram-es/pull/752/commits/c149f32d04f54a6a69397859806e4b45b0d0e801 fixed this by ensuring that these copy commands are only run once for all of the unit test executables (specifically, the commands are run as part of the _last_ unit test target). 

This can be made somewhat more elegant once shader source files no longer need to be copied as output (as in #753).

tallytalwar:
This looks good to me. Hitting shaders next in #753 will further reduce our lib size by a very tiny fraction though. (But diet is diet :D).

blair1618:
I was thinking of merging the shader binary bundling first so that this change would be simpler, but I suppose there's little difference in the end - I'll modify these tasks again after that.

PR #753, opened 2016-05-24
Shaders included within static/shared lib

karimnaaji:
Going along with https://github.com/tangrams/tangram-es/pull/752; we can now make the tangram library having no direct dependency on resources. 

- generate the shaders as header file to be part of the build process, 
- included them as regular cpp headers within the source code

That way, the library distribution would only need a static/shared binary file of the tangram library and a minimal set of headers.


blair1618:
This doesn't seem to re-build the shader data headers when the GLSL source is modified - is this intended? I think it would be rather error-prone to manually request those headers to be rebuilt.

karimnaaji:
I haven't tried this yet, but yes it should be working this way. There might be a way to tell cmake to look over those headers before running the custom command. I'll into the cmake doc for custom targets.

blair1618:
Working now! Thanks

PR #756, opened 2016-05-26
[rfc] Common platform code

karimnaaji:
Reorganize platform modules to be able to share common code between several platforms.


blair1618:
I like this idea. In fact, I want to go a step further! 

Right now we're connecting platform-specific functions to the library at compile-time, but it could be much more flexible to connect those at run-time, using code that doesn't have to be compiled with tangram. We could do this by defining the "platform" as a class that extends a `Platform` class and implements a set of functions. Then a tangram instance can be provided a `Platform` pointer at initialization, e.g. `auto map = new tangram::map(new linuxPlatform());`

That may be getting a little ahead of ourselves though. This PR seems like a solid incremental improvement.

PR #759, opened 2016-05-30
fix: use the actual size of shader data, data is not null-terminated

hjanetzek:
- fixes out of bounds reads of embedded shader data
- use SHADER_SOURCE macro for safe use of 'data' with respective 'size'

### DIFF
--- a/core/src/gl/primitives.cpp
+++ b/core/src/gl/primitives.cpp
@@ -29,10 +29,8 @@ void init() {
     if (!s_initialized) {
         s_shader = std::unique_ptr<ShaderProgram>(new ShaderProgram());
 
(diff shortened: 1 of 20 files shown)


hjanetzek:
Merging as this fix is necessary for running on linux. Implementation details are still open for discussion

blair1618:
Ah sorry, I hadn't tested the new encoded shaders with linux. Odd that it works on Android, but it could be a driver-specific issue.

hjanetzek:
Probably all char arrays were stored with zero'ed padding bytes at the end. Also worked with gcc debug builds - so I guess optimized builds just packs constant arrays tight

PR #762, opened 2016-05-31
Update alfons submodule

karimnaaji:
We were dealing with some issues building tangram-es on @dmvaldman machine. It turns out that freetype was dynamically linking with either libpng, zlib, bzip2 or harfbuzz whenever installed on the machine. Since we have a custom build for harfbuzz and we're not using those other libraries in freetype we can safely remove those dependencies in the build process.


dmvaldman:
This fixed my build. Thanks!

PR #764, opened 2016-06-01
Add a check in CMake to inform a user that submodules are missing

blair1618:
A very frequent problem encountered by new users is that they will clone the repository and try to build it, only to receive error messages that some directories don't contain a `CMakeLists.txt`. This just means that submodules haven't been initialized correctly. This change adds a macro to stop the build and print an informational message when it detects that a certain directory is empty or doesn't exist. This macro is applied for all submodule uses in the build process.

### DIFF
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -33,6 +33,11 @@ file(GLOB_RECURSE FOUND_HEADERS "${CMAKE_CURRENT_SOURCE_DIR}/${SOURCE_DIR}/*.h")
 set(INCLUDE_DIRS "")
 list(APPEND INCLUDE_DIRS "${CMAKE_CURRENT_SOURCE_DIR}/${SOURCE_DIR}")
 
+check_submodule(${INCLUDE_DIR}/earcut.hpp)
+check_submodule(${INCLUDE_DIR}/glm)
+check_submodule(${INCLUDE_DIR}/isect2d)
+check_submodule(${INCLUDE_DIR}/variant)
+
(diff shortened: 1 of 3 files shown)


blair1618:
This change won't be able to determine whether a user has initialized only the direct submodules of tangram-es but not the recursive submodules. This can happen if submodules are initialized without the `--recursive` flag. I can't think of any way to check this except for directly checking the existence of all the recursed submodules.

hjanetzek:
Looks good -- but a bit over-engineered compared to https://github.com/tangrams/tangram-es/commit/cd2ad11a866e79a0eb56806e11790fad543f424d :)

hjanetzek:
ok, if the check would be for ```external/alfons/deps/harfbuzz-icu-freetype/CMakeLists.txt``` it should also catch missing ```--recursive``` in 99.9% of the cases.

blair1618:
Oh I hadn't even seen that commit you made for this a couple days ago - you're right, that's much simpler and should catch almost all cases.

karimnaaji:
cd2ad11 is checking only one submodule though. This was adding a check for all the submodule we're using, which is still useful.

blair1618:
It's *possibly* useful, but I've never seen an instance where someone had the recursive alfons submodules initialized but not other submodules - it's usually all-or-nothing. If this problem persists then we can revisit these changes but I'd prefer keeping the CMake more concise.

karimnaaji:
Oh right, I haven't noticed this one was checking for a one more depth submodule 👍

PR #772, opened 2016-06-03
Dash array patterns

karimnaaji:
Refer to https://github.com/tangrams/tangram/pull/312 for more details.

screen shot 2016-06-03 at 5 53 09 pm


core/src/style/polylineStyle.cpp
```diff
@@ -129,10 +171,10 @@ struct PolylineStyleBuilder : public StyleBuilder {
         : StyleBuilder(_style), m_style(_style),
           m_meshData(2) {}
 
-    void addMesh(const Line& _line, const Parameters& _params);
+    void addMesh(const Line& _line, const Parameters& _params, float _overzoom2);
```
hjanetzek: Why have _overzoom as parameter when it's accessible as member?

core/src/util/builders.cpp
```diff
@@ -384,7 +384,7 @@ void buildPolyLineSegment(const Line& _line, PolyLineBuilder& _ctx,
         }
     }
 
-    distance += glm::distance(coordCurr, coordNext);
+    distance += glm::distance(coordCurr, coordNext) * _overzoom2;
 
     // Process last point in line with a cap
     addPolyLineVertex(coordNext, normNext, {1.f, distance}, _ctx); // right corner
```
hjanetzek: Could overzoom instead be applied in the vertex callback?

core/src/style/polylineStyle.cpp
karimnaaji: Leftover, will remove.

karimnaaji:
Based on the -js implementation, I found the discard method quite aggressive in some cases; for texture patterns on regular pixel density with no aa we can have big loss of texture infos.

We could default the blending to use a blending mode with transparency when a texture is used instead of alpha testing discard. (with a priority on user defined style blend mode that could lead in that case to alpha discard).

alha-discard:


defaulting-blend:


cc @bcamper

bcamper:
I think it's a good point -- but can you send me the texture you're using, because the ones I tried didn't seem nearly that bad? That said, I'm hesitant to have conditional defaults based on something this subtle (dash pattern vs. explicit texture), plus I feel like use of direct texture is not likely to be a very common use case, based on the tests I've done (in general the line width needs to be pretty thick to be able to get most textures to look OK). I'm OK waiting to see how popular this is, and adjusting defaults or adding more support later if needed.

karimnaaji:
Yes I agree that this is a default on a pretty specific case, we can wait to see how this is used. 
I guess that on -js canvas aa might help on that end. 

The texture is [here](https://github.com/tangrams/tangram-es/blob/2c06acb80f6e651eb2246296b32a2f6aa4411f95/scenes/img/arrow-3.png).

blair1618:
The alpha threshold used here is also pretty high. You might experiment with smaller values like 0.25 or 0.1 and see what results that produces.

PR #776, opened 2016-06-07
Url scene import

### LINKED ISSUE #714: Scene imports - Ability to "import" and modify another scene file inside tangram scene yaml

### LINKED ISSUE #723: run applySceneUpdates on bg thread
I'm trying to migrate EM to use 0.2.2 and need to apply scene updates to change the camera when the user starts navigation mode (https://github.com/mapzen/eraser-map/blob/tangram-2.2/app/src/main/kotlin/com/mapzen/erasermap/controller/MainActivity.kt#L1129). When `applySceneUpdates` is called, we are also changing the tilt/zoom, etc. Currently, applying the scene updates breaks the animation of these properties.

tallytalwar:
Fixes #714 
Fixes #723 

Is everything async yet!
Task of making scene loading asynchronously, resulted in making a bunch of other things run asyncly.

Some tasks handles in this PR:
- Scene Imports (Refer [import](https://github.com/tangrams/tangram-docs/blob/c506f7c65e676844b5418e5404c8efe9cb65157d/pages/import.md) documentation.)
- Async scene loading
- Async scene updating
- Import/Load network resources (scene files, textures, non-tiled datasource)
- Removed static `s_resourceRoot` for each platform, its now a part of each Scene been loaded. This is also required for multiple Tangram instancing.
- retains previous synchronous `sceneLoad` functionality (specifically for android). To be used with android clients. Basically it gets tricky for android platform to pass callback from android through tangram pipeline, while making sure callback executed on the initial UI thread it originated from. Which is handled gracefully by the `AsyncTask` implementation. All other platforms can use `loadSceneAsync` functionality and can also pass in callback to perform tasks post loading of the scene/map.
  - NOTE: For android, we can still use `sceneLoadAsync` (and get rid of synchronous `sceneLoad` call all together), a `callback`, to be executed post scene loading could be posted back from JNI to be run on the UI thread using `runOnUiThread` functionality on android. This can take care of running `addView(glSurfaceView)` on the UI Thread. 

    This being said, it might require some complicated JNI handshakes, and could be done later as a separate PR.

TODO:
- [x] Tangram "View Complete" logic should take async scene loading/updating into account. Basically if any scene load/update is pending it should return `false`.

Note: 
- A subsequent PR can help reduce some code here by using a generic URI handler.
- A subsequent PR can handle remote loading of font files, which in itself could be a big task. Plus we currently do not have a use for the same. (cc. @hjanetzek)

### DIFF
--- a/android/tangram/jni/platform_android.cpp
+++ b/android/tangram/jni/platform_android.cpp
@@ -26,6 +26,8 @@
 #include <fstream>
 #include <algorithm>
 
+#include <regex>
+
 /* Followed the following document for JavaVM tips when used with native threads
  * http://android.wooyd.org/JNIExample/#NWD1sCYeT-I
  * http://developer.android.com/training/articles/perf-jni.html and
@@ -176,10 +178,19 @@ bool isContinuousRendering() {
 
 std::string setResourceRoot(const char* _path, std::string& _sceneResourceRoot) {
 
-    _sceneResourceRoot = std::string(dirname(_path));
+    std::regex r("^(http|https):/");
+    std::smatch match;
+    std::string path(_path);
+
+    if (std::regex_search(path, match, r)) {
+        _sceneResourceRoot = "";
+        return path;
+    }
+
+    _sceneResourceRoot = std::string(dirname(path));
 
     // TODO: InternalResource boolean should also be on Scene instead of being static
-    s_useInternalResources = (*_path != '/');
+    s_useInternalResources = (*path != '/');
 
     // For unclear reasons, the AAssetManager will fail to open a file at
     // path "filepath" if the path is instead given as "./filepath", so in
@@ -191,7 +202,7 @@ std::string setResourceRoot(const char* _path, std::string& _sceneResourceRoot) 
         _sceneResourceRoot += '/';
     }
 
-    return std::string(basename(_path));
+    return std::string(basename(path));
 
 }
 
--- a/android/tangram/jni/platform_android.cpp
+++ b/android/tangram/jni/platform_android.cpp
@@ -53,7 +53,6 @@ static AAssetManager* assetManager = nullptr;
 
 static bool s_isContinuousRendering = false;
 static bool s_useInternalResources = true;
-static std::string s_resourceRoot;
 
 PFNGLBINDVERTEXARRAYOESPROC glBindVertexArrayOESEXT = 0;
 PFNGLDELETEVERTEXARRAYSOESPROC glDeleteVertexArraysOESEXT = 0;
@@ -175,20 +174,21 @@ bool isContinuousRendering() {
 
 }
 
-std::string setResourceRoot(const char* _path) {
+std::string setResourceRoot(const char* _path, std::string& _sceneResourceRoot) {
 
-    s_resourceRoot = std::string(dirname(_path));
+    _sceneResourceRoot = std::string(dirname(_path));
 
+    // TODO: InternalResource boolean should also be on Scene instead of being static
     s_useInternalResources = (*_path != '/');
 
     // For unclear reasons, the AAssetManager will fail to open a file at
     // path "filepath" if the path is instead given as "./filepath", so in
     // cases where dirname returns "." we simply use an empty string. For
     // all other cases, we add a "/" for appending relative paths.
-    if (!s_resourceRoot.empty() && s_resourceRoot.front() == '.') {
-        s_resourceRoot = "";
+    if (!_sceneResourceRoot.empty() && _sceneResourceRoot.front() == '.') {
+        _sceneResourceRoot = "";
     } else {
-        s_resourceRoot += '/';
+        _sceneResourceRoot += '/';
     }
 
     return std::string(basename(_path));
@@ -245,10 +245,10 @@ unsigned char* bytesFromFileSystem(const char* _path, unsigned int* _size) {
 
 }
 
-std::string stringFromFile(const char* _path, PathType _type) {
+std::string stringFromFile(const char* _path, PathType _type, const char* _resourceRoot) {
 
     unsigned int length = 0;
-    unsigned char* bytes = bytesFromFile(_path, _type, &length);
+    unsigned char* bytes = bytesFromFile(_path, _type, &length, _resourceRoot);
     std::string out(reinterpret_cast<char*>(bytes), length);
     free(bytes);
 
@@ -256,9 +256,14 @@ std::string stringFromFile(const char* _path, PathType _type) {
 
 }
 
-unsigned char* bytesFromFile(const char* _path, PathType _type, unsigned int* _size) {
+unsigned char* bytesFromFile(const char* _path, PathType _type, unsigned int* _size, const char* _resourceRoot) {
 
-    std::string resourcePath = s_resourceRoot + _path;
+    std::string resourcePath;
+    if (_resourceRoot) {
+        resourcePath = std::string(_resourceRoot) + _path;
+    } else {
+        resourcePath = _path;
+    }
 
     switch (_type) {
     case PathType::absolute:
--- a/android/tangram/jni/platform_android.cpp
+++ b/android/tangram/jni/platform_android.cpp
@@ -187,10 +187,10 @@ std::string setResourceRoot(const char* _path, std::string& _sceneResourceRoot) 
         return path;
     }
 
-    _sceneResourceRoot = std::string(dirname(path));
+    _sceneResourceRoot = std::string(dirname(_path));
 
     // TODO: InternalResource boolean should also be on Scene instead of being static
-    s_useInternalResources = (*path != '/');
+    s_useInternalResources = (*_path != '/');
 
     // For unclear reasons, the AAssetManager will fail to open a file at
     // path "filepath" if the path is instead given as "./filepath", so in
@@ -202,7 +202,7 @@ std::string setResourceRoot(const char* _path, std::string& _sceneResourceRoot) 
         _sceneResourceRoot += '/';
     }
 
-    return std::string(basename(path));
+    return std::string(basename(_path));
 
 }
 
--- a/android/tangram/jni/platform_android.cpp
+++ b/android/tangram/jni/platform_android.cpp
@@ -229,10 +229,10 @@ unsigned char* bytesFromFileSystem(const char* _path, size_t& _size) {
 
 }
 
-std::string stringFromFile(const char* _path, PathType _type) {
+std::string stringFromFile(const char* _path) {
 
     unsigned int length = 0;
-    unsigned char* bytes = bytesFromFile(_path, length, _type);
+    unsigned char* bytes = bytesFromFile(_path, length);
     std::string out(reinterpret_cast<char*>(bytes), length);
     free(bytes);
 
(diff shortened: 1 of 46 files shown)


tallytalwar:
@hjanetzek can you start looking at this? I have all the code in here, will be doing some minor code cleaning.

core/src/tangram.cpp
```diff
@@ -119,33 +138,105 @@ void setScene(std::shared_ptr<Scene>& _scene) {
...
+
+    m_sceneLoading = true;
+
+    std::thread t([&]() {
+            while(m_newScene) {
+                std::string scenePath;
+                {
+                    std::lock_guard<std::mutex> lock(m_newScenePathMutex);
+                    scenePath = setResourceRoot(m_newScenePath.c_str());
+                    m_newScene = false;
```
hjanetzek: This makes a concurrent modification of s_resourceRoot. 

I guess during async loading all resources should be requested with absolute/internal URIs. I.e. relative paths need to be translated in SceneLoader to absolute paths.

tallytalwar: Oops I missed this one! Thanks for the catch.

I agree on getting rid of `s_resourceRoot` to enable multiple Tangram instance support. This should be given more thought, and guess should/will be handled in a subsequent PR. For the purposes of this PR, I will put a lock in `setResourceRoot`.

core/src/tangram.cpp
```diff
@@ -119,33 +138,105 @@ void setScene(std::shared_ptr<Scene>& _scene) {
...
+                }
+
+                // Copy old scene
+                std::shared_ptr<Scene> scene;
+                {
+                    std::lock_guard<std::mutex> lock(m_sceneMutex);
+                    scene = std::make_shared<Scene>(*m_scene);
+                }
+
```
hjanetzek: Should it really copy the old scene when loading a new one? This also copies the m_updates of the old scene. I think the passing over of View and ClientDataSources should be separate from the Scene's copy constructor.

tallytalwar: @blair1618 do you remember why we needed to copy the view from previous scene for a new scene load (or for that matter to use the old scene to create a new one)? I think it was related retaining certain view parameters (context dimensions, etc).

blair1618: The view was copied when loading a new scene so that if loading the new scene failed, the view could be returned to its previous, valid configuration.

android/tangram/jni/platform_android.cpp
```diff
@@ -245,20 +256,25 @@ unsigned char* bytesFromFileSystem(const char* _path, unsigned int* _size) {
...
+    unsigned char* bytes = bytesFromFile(_path, _type, &length, _resourceRoot);
     std::string out(reinterpret_cast<char*>(bytes), length);
     free(bytes);
 
     return out;
 
 }
 
-unsigned char* bytesFromFile(const char* _path, PathType _type, unsigned int* _size) {
+unsigned char* bytesFromFile(const char* _path, PathType _type, unsigned int* _size, const char* _resourceRoot) {
```
karimnaaji: Why changing this to give the resource root?

tallytalwar: resourceRoot is no longer a static on platform, but is held by the Scene instance.
To resolve the path to a resource, we need to provide the resourceRoot from scene/fontContext instance to `bytesFromFile` or `stringFromFile`.

karimnaaji: It was in the scene at some point https://github.com/tangrams/tangram-es/pull/776/commits/bb0a0570e0d12b7f16a781158bcd4477570a49c9, how was that an issue?

hjanetzek: Would it be possible to load all scene resources *through* Scene and thereby keep resourceRoot internal to Scene?
Then we could remove all the PathType::resource special handling.

karimnaaji: @hjanetzek that's exactly what we were speaking about IRL. @tallytalwar will actually remove some part of the code that doesn't need to know about the resource root (e.g. texture for example).

tallytalwar: I agree on it.
To start with I removed I removed a Texture constructor which was kinda redundant as a Texture should only care about the blob of texture data and not the file/resource storing that data.

tallytalwar: Discussing more on this:

Currently file loading is required in 3 different places!
- sceneLoader: to load textures from resource Paths (Scene can handle this).
- fontContext: to load atlas (a const ref to resource root from Scene's resourceRoot member can be set to handle this internally in fontContext).
- DataSource: (could be internal/absolute or relative to resource path). Scene can not handle this. One proposal could be make DataSource aware of the same `resourceRoot` which Scene has and handle it internally in DataSource. But I am not sure about it.

core/src/scene/sceneLoader.cpp
```diff
@@ -430,9 +440,18 @@ MaterialTexture SceneLoader::loadMaterialTexture(Node matCompNode, Scene& scene,
...
+        matTex.tex = scene.textures()[name];
+    }
 
-    if (!matTex.tex) { matTex.tex = std::make_shared<Texture>(name); }
+    if (!matTex.tex) {
+        // Load inline material  textures
+        if (!loadTexture(name, scene)) {
+            LOGW("Not able to load material texture: %s", name.c_str());
+            return MaterialTexture();
+        };
```
karimnaaji: extra `;`

core/src/scene/sceneLoader.cpp
```diff
@@ -1278,11 +1335,14 @@ bool SceneLoader::parseStyleUniforms(const Node& value, Scene& scene, StyleUnifo
         } else {
             const std::string& strVal = value.Scalar();
             styleUniform.type = "sampler2D";
-            std::shared_ptr<Texture> texture = scene.getTexture(strVal);
 
-            if (!texture && !loadTexture(strVal, scene)) {
-                LOGW("Can't load texture with name %s", strVal.c_str());
-                return false;
+            if (scene) {
```
karimnaaji: In which cases the scene would be `nullptr`?

tallytalwar: https://github.com/tangrams/tangram-es/blob/74a467a932b8e6dc0af6960ae667e8fb5f795922/core/src/scene/importer.cpp#L217

Need to reuse it to determine type of uniforms to normalize texture paths!

karimnaaji: Oh yes I remember now.

core/src/scene/spriteAtlas.cpp
```diff
@@ -24,6 +24,14 @@ bool SpriteAtlas::getSpriteNode(const std::string& _name, SpriteNode& _node) con
     return true;
 }
 
+void SpriteAtlas::updateSpriteNodes(std::shared_ptr<Texture> _texture) {
```
karimnaaji: Can you add some comment that this is needed because the texture atlas is created before knowing anything about the texture it is associated with (width/height) when fetched over network.

core/src/tangram.cpp
```diff
@@ -58,6 +69,58 @@ static std::bitset<8> g_flags = 0;
 static bool g_cacheGlState = false;
 
 
+class AsyncWorker {
```
karimnaaji: We can move that to `util/`

core/src/tangram.h
```diff
@@ -154,6 +156,7 @@ bool getDebugFlag(DebugFlags _flag);
 void toggleDebugFlag(DebugFlags _flag);
 
 void runOnMainLoop(std::function<void()> _task);
+void runAsyncTask(std::function<void()> _task);
```
karimnaaji: Could probably add some comments on how this differentiate from `runOnMainLoop`.

tallytalwar: Sure, it is implied though.

karimnaaji: True, it's probably more `runOnMainLoop` that can be confusing if you don't know what is a tangram main loop.

tallytalwar: True, it will also have to be re-factored when we have multiple instancing, specifically around which instance's update/render needs to be performed. But thats a job of another PR :D.

Will document it here.

bench/src/tileLoading.cpp
```diff
@@ -48,12 +49,12 @@ struct TestContext {
             LOGE("Parsing scene config '%s'", e.what());
             return;
         }
-        scene = std::make_shared<Scene>();
         SceneLoader::applyConfig(sceneNode, *scene);
 
         styleContext.initFunctions(*scene);
         styleContext.setKeywordZoom(0);
 
+        LOG("Here");
```
blair1618: Remove

core/src/scene/scene.h
```diff
@@ -2,8 +2,11 @@
 
 #include "util/color.h"
 #include "util/fastmap.h"
+#include "view/view.h"
```
karimnaaji: Can remove the extra forward declaration of `View`.

core/src/scene/sceneLoader.h
```diff
@@ -2,13 +2,14 @@
 
 #include "gl/uniform.h"
 #include "scene/scene.h"
-
 #include <string>
 #include <vector>
 #include <memory>
 #include <tuple>
+#include <unordered_map>
```
karimnaaji: Extra include. (along with `unordered_set`).

tallytalwar: Was included for a previous implementation I had. Will remove. Thanks

osx/src/main.cpp
```diff
@@ -278,6 +278,7 @@ void init_main_window(bool recreate) {
 
     // Setup tangram
     Tangram::initialize(sceneFile.c_str());
```
karimnaaji: Discussed IRL with @tallytalwar : we can have `Tangram::initialize` taking no scene path now, since it's not loading the scene anymore. In that case it would be a dummy initialization for the main objects accessing that does not access  any resources.

tallytalwar: I am leaving this as it is right now. Even though `_scenePath` has no purpose in `Tangram::initialize` now and is redundant. But this will save some extra commits, since this code is being re-factored already to handle multiple tangram instances.

core/src/scene/importer.cpp
```diff
@@ -0,0 +1,325 @@
...
+    std::regex r("^(http|https):/");
+    std::smatch match;
+    if (path[0] == '/' || std::regex_search(path, match, r)) {
+        //absolute path or network url , return as it is
+        return path;
+    }
+
+    r = std::regex("[^//]+$");
+    return std::regex_replace(parentPath, r, path);
+}
```
hjanetzek: Could you add a comment what this replacement is supposed to do or an example case?

blair1618:
LGTM. Will merge once the Travis build passes.

core/src/scene/scene.h
```diff
@@ -2,8 +2,11 @@
...
+
 #include <list>
 #include <memory>
+#include <mutex>
```
karimnaaji: Is this include needed anymore?

tallytalwar: Na, I do not think so. Will take it out and rebase with last commit.

PR #778, opened 2016-06-08
Synchronize data source methods in MapController

blair1618:
May prevent memory leaks or corruption if sources are manipulated while scene is being modified.


tallytalwar:
LGTM. Merging.

PR #779, opened 2016-06-08
Delete local jstring reference

tallytalwar:
- seems to fix crash on Marshmallow having the exception of "local reference table overflow"

Fixes https://github.com/mapzen/android/issues/110


blair1618:
LGTM

PR #780, opened 2016-06-08
Delete local refs to 'jstring's in JNI functions

blair1618:
This fixes the JNI local reference table overflow described in https://github.com/mapzen/android/issues/110

There may be other places in our JNI code that temporarily leak local references, but these seem to be the only problematic cases.


blair1618:
Duplicate of #779

Issue #777, opened 2016-06-07
Render state: Texture unit binding should invalidate texture binding


karimnaaji:
@bcamper found out a bug in the tracking of texture unit/id binding within tangram-js that could possibly happen in -es.
We currently don't use multiple texture unit on default shaders but this could happen on user defined uniforms textures. A fix would be, every time a new texture unit is bound the current state tracking of the bound texture should be invalidated.

bcamper:
FYI, fixed in JS here: https://github.com/tangrams/tangram/commit/0cee05f76de2d0c630d209c7ea1d02068f9d46e4

tallytalwar:
Closed #777 via #784

PR #784, opened 2016-06-10
Invalidate render state tracking of current texture binding

karimnaaji:
#777

### DIFF
--- a/core/src/gl/renderState.cpp
+++ b/core/src/gl/renderState.cpp
@@ -39,10 +39,25 @@ namespace RenderState {
         return GL_TEXTURE0 + _unit;
     }
 
-    void bindVertexBuffer(GLuint _id) { GL_CHECK(glBindBuffer(GL_ARRAY_BUFFER, _id)); }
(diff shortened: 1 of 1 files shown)


blair1618:
Ah now I understand the problem this aims to fix :)

This looks correct. We could be slightly more efficient if we really wanted to by caching texture bindings as pairs of (unit, texture).

karimnaaji:
It's ok to store them separately, storing them as a pair `<unit,texture>` would make it discard transition `<0, 1>` to `<0, 2>` where only the texture binding should be updated.

blair1618:
What I mean is that if our state goes from (unit: 0, texture: 1) to (unit: 1, texture: 2), texture 1 is still bound to unit 0, so if we make 0 our active unit again we don't have to re-bind texture 1.

blair1618:
But this LGTM as-is and we can consider further refinement in the future as-needed.

karimnaaji:
Yes we can see the trace and see what's going on but I wonder if it's likely to happen in our context; styles most likely use next texture unit when previous one was not available for them since we always use the units starting from 0, but something to consider if we share gl-texture object between styles.

PR #785, opened 2016-06-10
Invalidate frame render state

karimnaaji:
Add an API call to be able to invalidate render state on a frame basis. By default invalidation is on. If that proves to be an inefficient default, we can change it to false (as discussed with @blair1618 and @tallytalwar) 

A quick stress test shows render state changes can have a big impact: 

**render state on**
screen shot 2016-06-10 at 11 44 48 am
**render state off**
screen shot 2016-06-10 at 11 44 17 am

Even if this invalidation would involve small redundant call between frames, we should test this carefully on device.

### DIFF
--- a/android/tangram/jni/jniExports.cpp
+++ b/android/tangram/jni/jniExports.cpp
@@ -232,8 +232,8 @@ extern "C" {
         Tangram::setDebugFlag(static_cast<Tangram::DebugFlags>(flag), on);
     }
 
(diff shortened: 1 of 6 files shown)


blair1618:
I've done some profiling of this now on both desktop and mobile devices, the difference in performance is much smaller than that presented in the screenshots above (I believe these were taken using a different implementation of this feature). On an Intel laptop GPU the difference was imperceptible. On an iPhone there was a small performance penalty measurable in the profiler and a GL trace showed roughly twice as many redundant state changes with invalidation turned on, but overall the performance wasn't visibly different.

karimnaaji:
Good to hear! The screenshot were not taken using this specifically, they were taken not using render any render state caching at all.

PR #786, opened 2016-06-10
Set client data sources to always be geometry-generating sources

blair1618:
Fixes an issue where client data sources were incorrectly removed when the scene is updated or reloaded.


tallytalwar:
LGTM.

PR #790, opened 2016-06-13
Debug toggle for frame graph

karimnaaji:
Add another debug toggle to represent the frame infos we are capturing (render + cpu time).

screen shot 2016-06-13 at 12 58 43 pm

### DIFF
--- a/core/src/debug/frameInfo.cpp
+++ b/core/src/debug/frameInfo.cpp
@@ -28,7 +28,7 @@ static clock_t s_startFrameTime = 0,
 
 void FrameInfo::beginUpdate() {
 
-    if (getDebugFlag(DebugFlags::tangram_infos) || getDebugFlag(DebugFlags::stats)) {
+    if (getDebugFlag(DebugFlags::tangram_infos) || getDebugFlag(DebugFlags::tangram_stats)) {
         s_startUpdateTime = clock();
     }
 
@@ -36,7 +36,7 @@ void FrameInfo::beginUpdate() {
 
 void FrameInfo::endUpdate() {
 
-    if (getDebugFlag(DebugFlags::tangram_infos) || getDebugFlag(DebugFlags::stats)) {
+    if (getDebugFlag(DebugFlags::tangram_infos) || getDebugFlag(DebugFlags::tangram_stats)) {
         s_endUpdateTime = clock();
         s_lastUpdateTime = TIME_TO_MS(s_startUpdateTime, s_endUpdateTime);
     }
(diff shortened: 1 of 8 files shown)


blair1618:
HOT. 

Maybe I'm getting greedy here but can we also draw a horizontal line at 16.6ms? :)

karimnaaji:
Yeah that'd be useful!

core/src/debug/frameInfo.cpp
```diff
@@ -41,23 +45,37 @@ void FrameInfo::endUpdate() {
...
-void FrameInfo::draw(const View& _view, TileManager& _tileManager) {
 
-    if (getDebugFlag(DebugFlags::tangram_infos)) {
+void FrameInfo::draw(const View& _view, TileManager& _tileManager, float _pixelsPerPoint) {
+
+    if (getDebugFlag(DebugFlags::tangram_infos) || getDebugFlag(DebugFlags::stats)) {
         static int cpt = 0;
 
+        static std::deque<float> cputime;
+        static std::deque<float> rendertime;
```
blair1618: Nitpick: `std::list` will do the job here.

core/src/debug/frameInfo.cpp
```diff
@@ -87,25 +105,48 @@ void FrameInfo::draw(const View& _view, TileManager& _tileManager) {
...
+            for (float t : cputime) {
+                i += 4 * _pixelsPerPoint;
+                Primitives::drawLine(glm::vec2(i, 0), glm::vec2(i, t * scale));
+            }
+
+            i = 2 * _pixelsPerPoint;
+            Primitives::setColor(0x0000ff);
+            for (float t : rendertime) {
+                i += 4 * _pixelsPerPoint;
+                Primitives::drawLine(glm::vec2(i, 0), glm::vec2(i, t * scale));
```
blair1618: To properly visualize the frame budget, the CPU and GPU times should probably be stacked vertically. I think that should be pretty easy?

core/src/tangram.cpp
```diff
@@ -273,7 +276,7 @@ void render() {
 
     m_labels->drawDebug(*m_view);
 
-    FrameInfo::draw(*m_view, *m_tileManager);
+    FrameInfo::draw(*m_view, *m_tileManager, m_pixelsPerPoint);
```
blair1618: Why not use `m_view->pixelScale()` in this function?

core/src/tangram.h
```diff
@@ -130,7 +130,8 @@ enum DebugFlags {
     tile_infos,         // Debug tile infos
     labels,             // Debug label bounding boxes
     tangram_infos,      // Various text tangram debug info printed on the screen
-    all_labels          // Draw all labels
+    all_labels,         // Draw all labels
+    stats               // Frame states
```
blair1618: Typo: "Frame stats". I also find it nice to leave a trailing comma on the last entry in an enum.

core/src/debug/frameInfo.cpp
karimnaaji: Actually what we show currently for a frame is not really GPU time because:
- cputime: time taken in `tangram.cpp:render()`
- rendertime: time taken in `tangram.cpp:render()` + `glFinish()` so it encompasses all gl commands dispatched from from current frame.

We should stack them like proposed, but also have in `tangram.cpp:update()` in it so it's the whole tangram frame and the 16.6ms would make sense as a notice of 60fps.

I will update for this.

core/src/debug/frameInfo.cpp
karimnaaji: Will keep `std::deque` because for array access (https://github.com/tangrams/tangram-es/pull/790/commits/483679b7c78a43e3137f99023359c9b593d57884).

karimnaaji:
Updated.

blair1618:
LGTM.

PR #791, opened 2016-06-14
Support the following style params

tallytalwar:
- text:order (not used in ES!!, but documentation mentions this to be a required property for all style
types)
- text:visible
- outline:visible


tallytalwar:
@blair1618 do you want to have a quick look at this. Can get away with the release after this gets merged.

tallytalwar:
I am unsure of `text:order` parameter. But the documentation says all styles need to have `order` property.

blair1618:
These parameters will be parsed now, but they aren't used in any style-building process so this doesn't actually change any behavior (though it does silence some warnings, I suppose). If we are adding keys for these parameters, we should also add their respective behaviors.

karimnaaji:
In here `text:order` would indeed be parsed, but would be a no-op on ES because of how label works, we can probably be more specific and output a warning.

tallytalwar:
Aiiiiii... I am too eager to push the release out that I missed a part of
the work here :(. Will update.

tallytalwar:
Updated, along with a warning for usage of `text:order`.

core/src/scene/styleParam.cpp
```diff
@@ -221,12 +222,17 @@ StyleParam::Value StyleParam::parseString(StyleParamKey key, const std::string&
...
+    case StyleParamKey::text_visible:
+    case StyleParamKey::outline_visible:
     case StyleParamKey::collide:
     case StyleParamKey::text_collide:
         if (_value == "true") { return true; }
         if (_value == "false") { return false; }
         LOGW("Bool value required for capitalized/visible. Using Default.");
         break;
+    case StyleParamKey::text_order:
+        LOGW("text:order style paramater is not used in ES.");
```
blair1618: Spelling/style: please revise to "text:order parameter is ignored." or omit.

core/src/scene/drawRule.cpp
```diff
@@ -174,7 +174,9 @@ void DrawRuleMergeSet::apply(const Feature& _feature, const SceneLayer& _layer,
         }
 
         bool visible;
-        if (rule.get(StyleParamKey::visible, visible) && !visible) {
+        if ((rule.get(StyleParamKey::visible, visible) && !visible) ||
+            (rule.get(StyleParamKey::outline_visible, visible) && !visible) ||
+            (rule.get(StyleParamKey::text_visible, visible) && !visible)) {
```
blair1618: This doesn't produce the correct behavior for these parameters. Under this logic, if a `lines` style has an `outline` node with `visible: false`, this will cause both the line and the outline to __not__ draw.

tallytalwar:
Updated.

tallytalwar:
Marking this for review, since we have come to a conclusion with respect to `visible` params' usage.

tallytalwar:
Updated and rebased with master.

blair1618:
LGTM. Merging!

PR #793, opened 2016-06-15
Re-add direct setter and getter for camera type

blair1618:
We had this feature before, and removed it in favor of using a "scene update" to change the camera type. While technically sufficient and correct, the "scene update" approach is much more cumbersome than a direct setter! Of course we can't add direct setters for every part of the scene, but since we provide direct control over every other aspect of the view (zoom, position, tilt, etc.) it seems reasonable that the camera type should be just as easy to access and modify.

### DIFF
--- a/android/tangram/jni/jniExports.cpp
+++ b/android/tangram/jni/jniExports.cpp
@@ -95,6 +95,14 @@ extern "C" {
         Tangram::setPixelScale(scale);
     }
 
+    JNIEXPORT void JNICALL Java_com_mapzen_tangram_MapController_nativeSetCameraType(JNIEnv* jniEnv, jobject obj, jint type) {
+        Tangram::setCameraType(type);
+    }
+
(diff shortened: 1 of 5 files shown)


tallytalwar:
I agree. 

The only concern I have is out of all the view calls we allow (zoom, position, tilt, cameraType change) this is the only one dependent on the scene file. If the scene file does not specify these camera types what should be the behavior?

blair1618:
The `camera` element in the scene file has parameters for `position` and `zoom` as well, so this doesn't seem that different. Can you clarify what you mean?

tallytalwar:
`perspective` and `isometric` camera types use different configurations specified in the scene file.
If one does not provide these configuration in the scene file and change the camera type, it will result in bad camera/renders. 

Example if one only has `isometric` camera set in the scene file with apt parameter ('axis') set.. and then camera type is changed from `isometric` to `perspective` (with no information of perspective camera parameters (`fov`, `focal_length`, etc) available).

blair1618:
All of these parameters have documented default values, so I don't think
one could produce an "invalid" configuration.

tallytalwar:
Note: Update documentation when this gets released with our next release.

PR #796, opened 2016-06-15
Make version name and version code available from BuildConfig

### LINKED ISSUE #795: Provide method to query version name
Tangram should provide a public method to query the version name included in `gradle.properties`. This way client apps can determine which version of the rendering engine is currently in use.

https://github.com/tangrams/tangram-es/blob/master/android/tangram/gradle.properties#L2

blair1618:
The Gradle plugin generates a class called BuildConfig at compile time, we can use it to access values determined at build time like the version name and version code:

```java
// Print the version of tangram in use.
Log.d(com.mapzen.tangram.BuildConfig.VERSION_NAME);
```

The version name is probably more useful than the version code, but version code seems to be used sometimes as well, so I [stole the script used in Eraser Map](https://github.com/mapzen/eraser-map/blob/master/versioning.gradle) to generate version codes automatically :)

Closes #795

### DIFF
--- a/android/tangram/build.gradle
+++ b/android/tangram/build.gradle
@@ -25,6 +25,7 @@ android {
   defaultConfig {
     minSdkVersion 15
     targetSdkVersion 23
+    versionName VERSION_NAME
   }
 
   sourceSets.main {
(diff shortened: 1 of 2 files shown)


nvkelso:
Is there a JS equivalent for this already?

tallytalwar:
This is good to go. Should also resolve #795!

PR #797, opened 2016-06-16
Adding console arguments to RPI tangram

patriciogonzalezvivo:
Hi there! this is the first attempt in a line of pulls to use tangram ES in a RaspberrPi to get thumbnails of tangram .yaml scenes. 

This pull essentially adds arguments to tangram so can load custom `scene` files, place (`lat`, `lon`, `zoom`) with specify `tilt` and `rotation` at a given `width` and `height`

### DIFF
--- a/rpi/src/context.cpp
+++ b/rpi/src/context.cpp
@@ -261,6 +261,10 @@ void initGL(int argc, char **argv){
         EGL_GREEN_SIZE, 8,
         EGL_BLUE_SIZE, 8,
         EGL_ALPHA_SIZE, 8,
+        EGL_SAMPLE_BUFFERS, 1,
+        EGL_SAMPLES, 4,
+        EGL_SURFACE_TYPE, EGL_WINDOW_BIT,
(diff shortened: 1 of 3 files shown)


patriciogonzalezvivo:
After this is merge I'll speak to @blair1618 about the possibility of using https://github.com/nothings/stb/blob/master/stb_image_write.h to save the pixels on the frame buffer to an image (and add the arguments to make that possible)

blair1618:
Looks like you committed old versions of the submodules for `alfons` and `geojson-vt-cpp`, let's remove those changes from the PR.

patriciogonzalezvivo:
While you are there, Could you add this lines https://github.com/tangrams/paparazzi/blob/master/src/context.cpp#L46-L49

blair1618:
Yep, added!

Can you give the code a try now and see if the continuous rendering is working as expected?

patriciogonzalezvivo:
Working smoothly!

patriciogonzalezvivo:
Hi! can we merge this to master? only affect rpi code

PR #803, opened 2016-06-23
Texture: fix destruction and uninitialized member access

hjanetzek:
if m_validData was accidental 0 font atlas were constantly re-uploaded

### DIFF
--- a/core/src/gl/mesh.cpp
+++ b/core/src/gl/mesh.cpp
@@ -37,17 +37,21 @@ MeshBase::~MeshBase() {
     // Deleting a index/array buffer being used ends up setting up the current vertex/index buffer to 0
     // after the driver finishes using it, force the render state to be 0 for vertex/index buffer
 
-    if (m_glVertexBuffer) {
-        if (RenderState::vertexBuffer.compare(m_glVertexBuffer)) {
-            RenderState::vertexBuffer.init(0, false);
+    if (RenderState::isValidGeneration(m_generation)) {
+        if (m_glVertexBuffer) {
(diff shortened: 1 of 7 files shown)


karimnaaji:
LGTM

tallytalwar:
This is good. And some nice catches.

core/src/gl/mesh.cpp
```diff
@@ -37,17 +37,21 @@ MeshBase::~MeshBase() {
     // Deleting a index/array buffer being used ends up setting up the current vertex/index buffer to 0
     // after the driver finishes using it, force the render state to be 0 for vertex/index buffer
 
-    if (m_glVertexBuffer) {
-        if (RenderState::vertexBuffer.compare(m_glVertexBuffer)) {
-            RenderState::vertexBuffer.init(0, false);
+    if (RenderState::isValidGeneration(m_generation)) {
```
karimnaaji: Do you think we should also defer those in main thread loop?

tallytalwar: isnt Mesh destructor already invoked from mainthread? (tileWorker does that?)

hjanetzek: Yes, only textures were a special case for the caches in RasterSource so far. At some point we could have a GLResource class that holds the gl handle and moves things for destruction to the mainloop unless it's called on the mainloop already.

tallytalwar: Making more work for @blair1618 to rebase/redo!
Refer: https://github.com/tangrams/tangram-es/commit/ba0dcf9bb4f94a383d61aff036cf98107a3308cf

PR #805, opened 2016-06-24
[rfc] simplify resources root handling

hjanetzek:
- either a path is relative, then bytesFromFile uses bundled resource path
- or a path is absolute then bytesFromFile just loads that file
- relative path replacements are all done in Scene Importer.

no more headaches, hopefully :)

PathType could also be removed, unless I'm missing something

### DIFF
--- a/android/tangram/jni/platform_android.cpp
+++ b/android/tangram/jni/platform_android.cpp
@@ -229,10 +229,10 @@ unsigned char* bytesFromFileSystem(const char* _path, size_t& _size) {
 
 }
 
-std::string stringFromFile(const char* _path, PathType _type) {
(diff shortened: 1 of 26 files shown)


blair1618:
I believe you're right, now that we don't need to access any files bundled with the library itself (what used to be 'internal' files) there's no need for PathType.

karimnaaji:
This seems like a nice change. No more headaches!

tallytalwar:
This is good. Will clean this a bit and merge/squash to `url-scene-import` branch, to which this is based off.

tallytalwar:
Cleaned this a bit and pushed these commits to `url-scene-import` PR.

PR #809, opened 2016-07-01
Convert latitude/longitude to screen coordinates

karimnaaji:


core/src/view/view.cpp
```diff
@@ -372,6 +372,20 @@ void View::updateMatrices() {
 
 }
 
+glm::vec2 View::latLonToScreen(double lat, double lon) {
```
bcamper: Can we not mix order or latitude and longitude? :) e.g. `View::latLonToScreen` followed by `m_projection->LonLatToMeters`. My preference is either something agnostic, or if we have to choose, longitude first (X first for consistency with graphics).

tallytalwar:
This looks good to me. Merging.

PR #810, opened 2016-07-05
Refactor C++ interface into Map class

blair1618:
A few things remain to finish this implementation:
- ~~Make async task dispatch safe for multiple instances~~
- Make debug flags trigger tile rebuilding when necessary (not strictly necessary?)
- ~~Create and access a `RenderState` instance for each GL thread~~ (separate PR)

Items to consider:

1. There are other possible approaches to managing multiple Tangram instances. An alternative would be to require an object handle as an argument to all of the main interface functions, e.g.

  ```
  auto mapHandle = Tangram::init();
  Tangram::loadSceneAsync(mapHandle, sceneFilePath);
  ```

  where `mapHandle` is an integer or a void pointer. I spent a little time trying this and it seems very possible, but often more verbose. The advantages of this would be that the internal operation of Tangram could be totally hidden from the calling context and it would then be much easier to write a pure C interface to the library. 

2. This current implementation has the drawback of exposing the types and declarations for all the members in `Tangram::Map` to the calling application. This could be addressed by wrapping the actual map data structure in another class such that the outer class only needs to hold an opaque pointer to the implementation class. 

3. The main interface in this implementation is now the `Tangram::Map` class, but it resides in files called "tangram.h" and "tangram.cpp". For consistency with the rest of our source files, we may want to rename these to "map.h" and "map.cpp".

### DIFF
--- a/android/tangram/jni/jniExports.cpp
+++ b/android/tangram/jni/jniExports.cpp
@@ -6,138 +6,203 @@
 
 extern "C" {
 
-    JNIEXPORT void JNICALL Java_com_mapzen_tangram_MapController_nativeSetPosition(JNIEnv* jniEnv, jobject obj, jdouble lon, jdouble lat) {
-        Tangram::setPosition(lon, lat);
+    JNIEXPORT void JNICALL Java_com_mapzen_tangram_MapController_nativeSetPosition(JNIEnv* jniEnv, jobject obj, jlong mapPtr, jdouble lon, jdouble lat) {
+        assert(mapPtr > 0);
+        auto map = reinterpret_cast<Tangram::Map*>(mapPtr);
+        map->setPosition(lon, lat);
     }
 
-    JNIEXPORT void JNICALL Java_com_mapzen_tangram_MapController_nativeSetPositionEased(JNIEnv* jniEnv, jobject obj, jdouble lon, jdouble lat, jfloat duration, jint ease) {
-        Tangram::setPosition(lon, lat, duration, static_cast<Tangram::EaseType>(ease));
+    JNIEXPORT void JNICALL Java_com_mapzen_tangram_MapController_nativeSetPositionEased(JNIEnv* jniEnv, jobject obj, jlong mapPtr, jdouble lon, jdouble lat, jfloat duration, jint ease) {
+        assert(mapPtr > 0);
+        auto map = reinterpret_cast<Tangram::Map*>(mapPtr);
+        map->setPositionEased(lon, lat, duration, static_cast<Tangram::EaseType>(ease));
     }
 
-    JNIEXPORT void JNICALL Java_com_mapzen_tangram_MapController_nativeGetPosition(JNIEnv* jniEnv, jobject obj, jdoubleArray lonLat) {
+    JNIEXPORT void JNICALL Java_com_mapzen_tangram_MapController_nativeGetPosition(JNIEnv* jniEnv, jobject obj, jlong mapPtr, jdoubleArray lonLat) {
+        assert(mapPtr > 0);
+        auto map = reinterpret_cast<Tangram::Map*>(mapPtr);
         jdouble* arr = jniEnv->GetDoubleArrayElements(lonLat, NULL);
-        Tangram::getPosition(arr[0], arr[1]);
+        map->getPosition(arr[0], arr[1]);
         jniEnv->ReleaseDoubleArrayElements(lonLat, arr, 0);
     }
 
-    JNIEXPORT void JNICALL Java_com_mapzen_tangram_MapController_nativeSetZoom(JNIEnv* jniEnv, jobject obj, jfloat zoom) {
-        Tangram::setZoom(zoom);
+    JNIEXPORT void JNICALL Java_com_mapzen_tangram_MapController_nativeSetZoom(JNIEnv* jniEnv, jobject obj, jlong mapPtr, jfloat zoom) {
+        assert(mapPtr > 0);
+        auto map = reinterpret_cast<Tangram::Map*>(mapPtr);
+        map->setZoom(zoom);
     }
 
-    JNIEXPORT void JNICALL Java_com_mapzen_tangram_MapController_nativeSetZoomEased(JNIEnv* jniEnv, jobject obj, jfloat zoom, jfloat duration, jint ease) {
-        Tangram::setZoom(zoom, duration, static_cast<Tangram::EaseType>(ease));
+    JNIEXPORT void JNICALL Java_com_mapzen_tangram_MapController_nativeSetZoomEased(JNIEnv* jniEnv, jobject obj, jlong mapPtr, jfloat zoom, jfloat duration, jint ease) {
+        assert(mapPtr > 0);
+        auto map = reinterpret_cast<Tangram::Map*>(mapPtr);
+        map->setZoomEased(zoom, duration, static_cast<Tangram::EaseType>(ease));
     }
 
-    JNIEXPORT jfloat JNICALL Java_com_mapzen_tangram_MapController_nativeGetZoom(JNIEnv* jniEnv, jobject obj) {
-        return Tangram::getZoom();
+    JNIEXPORT jfloat JNICALL Java_com_mapzen_tangram_MapController_nativeGetZoom(JNIEnv* jniEnv, jobject obj, jlong mapPtr) {
+        assert(mapPtr > 0);
+        auto map = reinterpret_cast<Tangram::Map*>(mapPtr);
+        return map->getZoom();
     }
 
-    JNIEXPORT void JNICALL Java_com_mapzen_tangram_MapController_nativeSetRotation(JNIEnv* jniEnv, jobject obj, jfloat radians) {
-        Tangram::setRotation(radians);
+    JNIEXPORT void JNICALL Java_com_mapzen_tangram_MapController_nativeSetRotation(JNIEnv* jniEnv, jobject obj, jlong mapPtr, jfloat radians) {
+        assert(mapPtr > 0);
+        auto map = reinterpret_cast<Tangram::Map*>(mapPtr);
+        map->setRotation(radians);
     }
 
-    JNIEXPORT void JNICALL Java_com_mapzen_tangram_MapController_nativeSetRotationEased(JNIEnv* jniEnv, jobject obj, jfloat radians, jfloat duration, jint ease) {
-        Tangram::setRotation(radians, duration, static_cast<Tangram::EaseType>(ease));
+    JNIEXPORT void JNICALL Java_com_mapzen_tangram_MapController_nativeSetRotationEased(JNIEnv* jniEnv, jobject obj, jlong mapPtr, jfloat radians, jfloat duration, jint ease) {
+        assert(mapPtr > 0);
+        auto map = reinterpret_cast<Tangram::Map*>(mapPtr);
+        map->setRotationEased(radians, duration, static_cast<Tangram::EaseType>(ease));
     }
 
-    JNIEXPORT jfloat JNICALL Java_com_mapzen_tangram_MapController_nativeGetRotation(JNIEnv* jniEnv, jobject obj) {
-        return Tangram::getRotation();
+    JNIEXPORT jfloat JNICALL Java_com_mapzen_tangram_MapController_nativeGetRotation(JNIEnv* jniEnv, jobject obj, jlong mapPtr) {
+        assert(mapPtr > 0);
+        auto map = reinterpret_cast<Tangram::Map*>(mapPtr);
+        return map->getRotation();
     }
 
-    JNIEXPORT void JNICALL Java_com_mapzen_tangram_MapController_nativeSetTilt(JNIEnv* jniEnv, jobject obj, jfloat radians) {
-        Tangram::setTilt(radians);
+    JNIEXPORT void JNICALL Java_com_mapzen_tangram_MapController_nativeSetTilt(JNIEnv* jniEnv, jobject obj, jlong mapPtr, jfloat radians) {
+        assert(mapPtr > 0);
+        auto map = reinterpret_cast<Tangram::Map*>(mapPtr);
+        map->setTilt(radians);
     }
 
-    JNIEXPORT void JNICALL Java_com_mapzen_tangram_MapController_nativeSetTiltEased(JNIEnv* jniEnv, jobject obj, jfloat radians, jfloat duration, jint ease) {
-        Tangram::setTilt(radians, duration, static_cast<Tangram::EaseType>(ease));
+    JNIEXPORT void JNICALL Java_com_mapzen_tangram_MapController_nativeSetTiltEased(JNIEnv* jniEnv, jobject obj,  jlong mapPtr, jfloat radians, jfloat duration, jint ease) {
+        assert(mapPtr > 0);
+        auto map = reinterpret_cast<Tangram::Map*>(mapPtr);
+        map->setTiltEased(radians, duration, static_cast<Tangram::EaseType>(ease));
     }
 
-    JNIEXPORT jfloat JNICALL Java_com_mapzen_tangram_MapController_nativeGetTilt(JNIEnv* jniEnv, jobject obj) {
-        return Tangram::getTilt();
+    JNIEXPORT jfloat JNICALL Java_com_mapzen_tangram_MapController_nativeGetTilt(JNIEnv* jniEnv, jobject obj, jlong mapPtr) {
+        assert(mapPtr > 0);
+        auto map = reinterpret_cast<Tangram::Map*>(mapPtr);
+        return map->getTilt();
     }
 
-    JNIEXPORT jboolean JNICALL Java_com_mapzen_tangram_MapController_nativeScreenPositionToLngLat(JNIEnv* jniEnv, jobject obj, jdoubleArray coordinates) {
+    JNIEXPORT jboolean JNICALL Java_com_mapzen_tangram_MapController_nativeScreenPositionToLngLat(JNIEnv* jniEnv, jobject obj, jlong mapPtr, jdoubleArray coordinates) {
+        assert(mapPtr > 0);
+        auto map = reinterpret_cast<Tangram::Map*>(mapPtr);
         jdouble* arr = jniEnv->GetDoubleArrayElements(coordinates, NULL);
-        bool ret = Tangram::screenPositionToLngLat(arr[0], arr[1], &arr[0], &arr[1]);
+        bool ret = map->screenPositionToLngLat(arr[0], arr[1], &arr[0], &arr[1]);
         jniEnv->ReleaseDoubleArrayElements(coordinates, arr, 0);
         return ret;
     }
 
-    JNIEXPORT jboolean JNICALL Java_com_mapzen_tangram_MapController_nativeLngLatToScreenPosition(JNIEnv* jniEnv, jobject obj, jdoubleArray coordinates) {
+    JNIEXPORT jboolean JNICALL Java_com_mapzen_tangram_MapController_nativeLngLatToScreenPosition(JNIEnv* jniEnv, jobject obj, jlong mapPtr, jdoubleArray coordinates) {
+        assert(mapPtr > 0);
+        auto map = reinterpret_cast<Tangram::Map*>(mapPtr);
         jdouble* arr = jniEnv->GetDoubleArrayElements(coordinates, NULL);
-        bool ret = Tangram::lngLatToScreenPosition(arr[0], arr[1], &arr[0], &arr[1]);
+        bool ret = map->lngLatToScreenPosition(arr[0], arr[1], &arr[0], &arr[1]);
         jniEnv->ReleaseDoubleArrayElements(coordinates, arr, 0);
         return ret;
     }
 
-    JNIEXPORT void JNICALL Java_com_mapzen_tangram_MapController_nativeInit(JNIEnv* jniEnv, jobject obj, jobject tangramInstance, jobject assetManager, jstring stylePath) {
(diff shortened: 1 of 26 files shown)


blair1618:
I'm realizing it's going to be another big heap of work to get multiple instances working in a truly safe way on all of our platforms. 2 issues so far:

 - `RenderState` is accessed from a global scope right now, but we actually need a separate `RenderState` for each GL thread. There happens to be a perfect solution for this in the C++11 [thread_local](http://en.cppreference.com/w/cpp/language/storage_duration) storage specifier, but this feature isn't supported in AppleClang until the upcoming Xcode 8 (even though upstream Clang has supported it since 3.5). An alternative approach is to create and switch between multiple instances in our own code - I've implemented most of that solution and it's a huge pain and probably slower, but should technically work. 

 - On Android we render with a `GlSurfaceView` which has a separate thread used just for GL rendering. Our JNI code right now uses global variables to track which thread the JNI environment should attach to, and that will simply be broken if we have multiple `GlSurfaceViews` with multiple rendering threads. 

It may be sensible to merge the changes for the `Map` class refactoring and then incrementally fix support for rendering with multiple instances.

karimnaaji:
@blair1618 : we can probably easily convert `RenderState` to a class/struct and instantiate one `RenderState` instance for each of the GL threads instead of keeping it global and adding thread safety within it.
It could be a separate work/PR only for that and will probably make things simpler on that end.

blair1618:
Yeah, the `thread_local` specifier is the most elegant solution for `RenderState` but it's not the only way. I've tried just using an `unordered_map` of `thread_id`s to `RenderState` instances here: https://github.com/tangrams/tangram-es/commit/128d1c196a223e6d122e560101e5a7b7736f5dd1

karimnaaji:
Yes `thread_local` seems the most appropriate. In https://github.com/tangrams/tangram-es/commit/128d1c196a223e6d122e560101e5a7b7736f5dd1 the render state access might be a bottleneck, we could probably pass it around to our gl objects (by accessing the thread id once at the beginning of the render loop).

blair1618:
I'm worried about the speed of `RenderState::get()` lookups too, I plan to benchmark it.

ios/src/ViewController.mm
```diff
@@ -201,16 +200,18 @@ - (void)setupGL
 {
     [EAGLContext setCurrentContext:self.context];
 
-    Tangram::initialize("scene.yaml");
-    Tangram::loadSceneAsync("scene.yaml");
-    Tangram::setupGL();
+    if (!self.map) {
+        self.map = new Tangram::Map("scene.yaml");
```
karimnaaji: There's a missing `delete` in `dealloc`

karimnaaji: It seems to be the same for other platforms layers.

blair1618: Yeah, not too worried about it because we would only ever delete it when the process is going to exit anyway - but it's probably better practice to explicitly manage it.

karimnaaji: Yes that's true, I just noticed that we would retain all pointers held by the `Map` class, but it only really depends on how the platform manages process leaks.

blair1618: Updated. Glad I tried manual deletion - it exposed a pretty subtle bug https://github.com/tangrams/tangram-es/pull/810/commits/dacf44e8bdd0d9146363431a2730305ef8d59cbf

lygstate:
Looks good for me.

blair1618:
To clean up the `Map` class declaration, we can use a private implementation class like I've done here: https://github.com/tangrams/tangram-es/commit/6be9f398605f468adfc89b0dc0c90c9ad0861672

This could either go into this PR or a quick follow-up, but it's a closely-related change that (I think) should be considered necessary for maintainability. 

Also, I can't decide what to name the inner class! Considered "Base", "Bridge", "Detail", "Impl", "Inner", ...

core/src/gl/texture.cpp
```diff
@@ -189,6 +191,8 @@ void Texture::generate(GLuint _textureUnit) {
     GL_CHECK(glTexParameteri(m_target, GL_TEXTURE_WRAP_T, m_options.wrapping.wrapt));
 
     m_generation = RenderState::generation();
+
+    m_mainThreadJobQueue.makeCurrentThreadTarget();
```
tallytalwar: Still looking at code, so I might be missing something, but should this thread be made current gl one before making calls to gl functions? Can you clarify please.

blair1618: `makeCurrentThreadTarget()` sets the current thread as the thread that any queued jobs will be run on later. It doesn't have anything directly to do with GL.

tallytalwar: Ahh! I should have looked the implementation of that. Name looked familar to the evas call to set current gl thread :P. Thanks

tallytalwar:
So I get a crash on android by performing a gesture while the scene is being loading. Can reproduce this with the default scene file (though harder to hit because its faster to load) as well as with a remote scene file. You can try reoriented the device to trigger a context loss just as you do a gesture motion.

Last few lines on the logs show some more information. Looks like MapController is null
```
7-18 17:25:45.579 12923-13804/com.mapzen.tangram.android D/Tangram: TANGRAM tileWorker.cpp:67: Passed new TileBuilder to TileWorker
07-18 17:25:45.669 12923-13803/com.mapzen.tangram.android D/Tangram: Failed to open asset at path: fonts/sans-serif-400normal.ttf
07-18 17:25:45.679 12923-13803/com.mapzen.tangram.android I/Alfons: LOADED Font: Roboto Regular size: 52
07-18 17:25:46.159 12923-12923/com.mapzen.tangram.android D/Activity: performCreate Call secproduct feature valuefalse
07-18 17:25:46.159 12923-12923/com.mapzen.tangram.android D/Activity: performCreate Call debug elastic valuetrue
07-18 17:25:46.169 12923-12923/com.mapzen.tangram.android E/ViewRootImpl: sendUserActionEvent() mView == null
07-18 17:25:46.179 12923-13387/com.mapzen.tangram.android D/Tangram: TANGRAM tangram.cpp:49: initialize
07-18 17:25:46.189 12923-13387/com.mapzen.tangram.android D/Tangram: TANGRAM tangram.cpp:72: finish initialize
07-18 17:25:46.189 12923-13387/com.mapzen.tangram.android D/Tangram: TANGRAM tangram.cpp:143: Loading scene file: scene.yaml
07-18 17:25:46.259 12923-12923/com.mapzen.tangram.android I/Timeline: Timeline: Activity_idle id: android.os.BinderProxy@20b96fd3 time:50440299
07-18 17:25:46.289 12923-13387/com.mapzen.tangram.android D/Tangram: WARNING sceneLoader.cpp:1524: Can't find data source touch for layer touch
07-18 17:25:46.359 12923-12923/com.mapzen.tangram.android D/Activity: performCreate Call secproduct feature valuefalse
07-18 17:25:46.359 12923-12923/com.mapzen.tangram.android D/Activity: performCreate Call debug elastic valuetrue
07-18 17:25:46.369 12923-13059/com.mapzen.tangram.android D/Tangram: TANGRAM tangram.cpp:49: initialize
07-18 17:25:46.369 12923-13059/com.mapzen.tangram.android D/Tangram: TANGRAM tangram.cpp:72: finish initialize
07-18 17:25:46.379 12923-12923/com.mapzen.tangram.android D/Activity: performCreate Call secproduct feature valuefalse
07-18 17:25:46.379 12923-12923/com.mapzen.tangram.android D/Activity: performCreate Call debug elastic valuetrue
07-18 17:25:46.389 12923-12923/com.mapzen.tangram.android D/AndroidRuntime: Shutting down VM
07-18 17:25:46.389 12923-12923/com.mapzen.tangram.android E/AndroidRuntime: FATAL EXCEPTION: main
                                                                            Process: com.mapzen.tangram.android, PID: 12923
                                                                            java.lang.NullPointerException: Attempt to invoke virtual method 'void com.mapzen.tangram.MapController.dispose()' on a null object reference
                                                                                at com.mapzen.tangram.MapView$1.onCancelled(MapView.java:80)
                                                                                at com.mapzen.tangram.MapView$1.onCancelled(MapView.java:62)
                                                                                at android.os.AsyncTask.finish(AsyncTask.java:630)
                                                                                at android.os.AsyncTask.access$600(AsyncTask.java:177)
                                                                                at android.os.AsyncTask$InternalHandler.handleMessage(AsyncTask.java:645)
                                                                                at android.os.Handler.dispatchMessage(Handler.java:102)
                                                                                at android.os.Looper.loop(Looper.java:145)
                                                                                at android.app.ActivityThread.main(ActivityThread.java:5835)
                                                                                at java.lang.reflect.Method.invoke(Native Method)
                                                                                at java.lang.reflect.Method.invoke(Method.java:372)
                                                                                at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:1399)
                                                                                at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1194)
07-18 17:25:46.399 12923-13263/com.mapzen.tangram.android D/Tangram: TANGRAM tangram.cpp:49: initialize
07-18 17:25:46.399 12923-13263/com.mapzen.tangram.android D/Tangram: TANGRAM tangram.cpp:72: finish initialize
07-18 17:25:46.399 12923-13263/com.mapzen.tangram.android D/Tangram: TANGRAM tangram.cpp:143: Loading scene file: scene.yaml
07-18 17:25:46.479 12923-13263/com.mapzen.tangram.android D/Tangram: WARNING sceneLoader.cpp:1524: Can't find data source touch for layer touch
07-18 17:25:47.989 12923-12923/com.mapzen.tangram.android I/Process: Sending signal. PID: 12923 SIG: 9
```

core/src/tangram.cpp
```diff
@@ -98,11 +66,27 @@ void initialize(const char* _scenePath) {
...
+    // Explicitly destroy all member objects so that we have a chance
+    // to run any resulting jobs sent to the JobQueue.
+    m_asyncWorker.reset();
+    m_inputHandler.reset();
+    m_labels.reset();
+    m_view.reset();
+    m_nextScene.reset();
+    m_tileManager.reset();
+    m_tileWorker.reset();
+    JobQueue::runJobsForCurrentThread();
```
tallytalwar: There might be an issue here where async scene load puts in a task on the JobQueue, to do some operation on the newly loaded (async) scene. Will test this further and update, if this actually results in bad behavior.

blair1618:
The lifecycle management of `MapController` from `MapView` has been updated - this should correct the crashes that occurred during rapid Activity re-creation.

core/src/tangram.cpp
tallytalwar: K, tested this, because of [line 218](https://github.com/tangrams/tangram-es/blob/826fa42f774343ee1a86eb1c460f713957943c3b/core/src/tangram.cpp#L218) and the `if block` at [line 225](https://github.com/tangrams/tangram-es/blob/826fa42f774343ee1a86eb1c460f713957943c3b/core/src/tangram.cpp#L225), handles this issue.

`m_nextScene` is copied as `scene` down the async Task and hence is still valid when this Task is run from the destructor. But since `m_nextScene` is reset, this task simply returns. And same applied to tasks created from `Map::applySceneUpdates`.

tallytalwar: Ha! looks like I was handling this already from async scene loading code on an app exit! which is same as Map object Destruction

core/src/tangram.h
```diff
@@ -2,144 +2,213 @@
...
+class TileWorker;
+class View;
+
+struct TouchItem {
+    std::shared_ptr<Properties> properties;
+    float position[2];
+    float distance;
+};
+
+struct SceneUpdate {
```
tallytalwar: Can you elaborate on why to move `SceneUpdate` outside `Scene`?

blair1618: So that we don't need to `#include "scene.h"`

The (not-too-distant) goal here is for library users to only need `tangram.h` (or `map.h`).

karimnaaji: Yes this one could probably be forward declared.

blair1618: Well... not in this implementation it can't. The `vector<SceneUpdate>` declared below needs to know the size of a `SceneUpdate` (just one reason why it'll be nice to make the `Map` implementation private).

core/src/util/jobQueue.cpp
```diff
@@ -0,0 +1,40 @@
+#include "jobQueue.h"
+#include <mutex>
+
+namespace Tangram {
+
+std::unordered_map<JobQueue::ThreadId, std::vector<JobQueue::Job>> JobQueue::jobsForThreads;
+std::mutex jobMutex;
```
tallytalwar: can also make this mutex a static member of `JobQueue`.

blair1618: Why?

core/src/tangram.h
tallytalwar: Ahh! +1 for private `Map` implementation.

core/src/util/jobQueue.cpp
tallytalwar: Static member vs global variable!
Purpose is to make read/writes to `jobsForThreads` synchronous. Both will achieve the same thing.

core/src/tile/tileWorker.cpp
```diff
@@ -29,17 +31,16 @@ TileWorker::~TileWorker(){
...
-        Tangram::runOnMainLoop(disposer);
+        // Create an std::function that will run on the main thread
+        // and dispose the TileBuilder, including its Scene reference
+        // with OpenGL resources, on the main thread. This ensures
+        // that no GL functions are called on the worker thread.
+        auto builderPtr = _builder.release();
+        m_mainThreadJobQueue.add([=]() {
+            delete builderPtr;
+        });
     }
```
hjanetzek: This is nice, makes the intention way more clear :)

core/src/tile/tileWorker.cpp
```diff
@@ -16,6 +16,8 @@ namespace Tangram {
 TileWorker::TileWorker(int _num_worker) {
     m_running = true;
 
+    m_mainThreadJobQueue.makeCurrentThreadTarget();
+
```
hjanetzek: I think it would be sensible to set thread_id in the default constructor

core/src/tangram.h
karimnaaji: Oh I see, this would be fixed later with the private imp.

core/src/util/jobQueue.cpp
blair1618: I'm not sure I follow. This declaration creates static storage duration with no external visibility, which seems to be what we need here.

core/src/tile/tileWorker.cpp
karimnaaji: I think it's nice to have it explicit so there's no mis-usage behavior on that. Since the default constructor could be called on other object creation holding jobQueues, it seems harder to follow.

core/src/util/jobQueue.cpp
tallytalwar: Though both are fine here.
But, in a general practice, global scoped mutex are used to sync global scoped variables.
Its a better practice to have a static mutex class member to sync a static class member variable.

android/tangram/jni/jniExports.cpp
```diff
@@ -148,37 +212,45 @@ extern "C" {
...
+    JNIEXPORT void JNICALL Java_com_mapzen_tangram_MapController_nativeClearDataSource(JNIEnv* jniEnv, jobject obj, jlong mapPtr, jlong sourcePtr) {
+        assert(mapPtr > 0);
+        auto map = reinterpret_cast<Tangram::Map*>(mapPtr);
+        assert(sourcePtr > 0);
+        auto source = reinterpret_cast<Tangram::DataSource*>(sourcePtr);
+        map->clearDataSource(*source, true, true);
     }
 
-    JNIEXPORT void JNICALL Java_com_mapzen_tangram_MapController_nativeAddFeature(JNIEnv* jniEnv, jobject obj, jlong pointer,
+    JNIEXPORT void JNICALL Java_com_mapzen_tangram_MapController_nativeAddFeature(JNIEnv* jniEnv, jobject obj, jlong mapPtr, jlong sourcePtr,
```
tallytalwar: I guess this should not require `mapPtr`.

android/tangram/src/com/mapzen/tangram/MapController.java
```diff
@@ -656,7 +664,27 @@ public void applySceneUpdates() {
...
+    void clearDataSource(long sourcePtr) {
+        nativeClearDataSource(mapPointer, sourcePtr);
+    }
+
+    void addFeature(long sourcePtr, double[] coordinates, int[] rings, String[] properties) {
+        nativeAddFeature(mapPointer, sourcePtr, coordinates, rings, properties);
+    }
+
+    void addGeoJson(long sourcePtr, String geoJson) {
+        nativeAddGeoJson(mapPointer, sourcePtr, geoJson);
```
tallytalwar: this should not require `mapPointer`. Also the corresponding jni call does not match in the `jniExports.cpp`.

android/tangram/jni/jniExports.cpp
```diff
@@ -6,138 +6,202 @@
...
         return ret;
     }
 
-    JNIEXPORT void JNICALL Java_com_mapzen_tangram_MapController_nativeInit(JNIEnv* jniEnv, jobject obj, jobject tangramInstance, jobject assetManager, jstring stylePath) {
+    JNIEXPORT jlong JNICALL Java_com_mapzen_tangram_MapController_nativeInit(JNIEnv* jniEnv, jobject obj, jobject tangramInstance, jobject assetManager) {
         setupJniEnv(jniEnv, tangramInstance, assetManager);
-        const char* cStylePath = jniEnv->GetStringUTFChars(stylePath, NULL);
-        Tangram::initialize(cStylePath);
-        jniEnv->ReleaseStringUTFChars(stylePath, cStylePath);
+        assert(mapPtr > 0);
```
tallytalwar: No `mapPtr` to check here!

tallytalwar:
@blair1618 Do you plan to merge `private-map-impl` with this or you will create a subsequent PR following this?

With respect to this PR, once the last few comments and issues are addressed, its good to be merged from my side.

core/src/tile/tileWorker.cpp
blair1618: Yeah same, I'd rather have exactly one way to set the target thread and not set it implicitly. The default constructor for `std::thread::id` creates an ID that represents no thread, which I think is a sensible initial condition.

core/src/util/jobQueue.cpp
blair1618: Oh you make a fair point - I thought that neither of those were class members. I've edited this such that both variables are now non-members.

android/tangram/jni/jniExports.cpp
blair1618: Technically no, it doesn't ~ but ~ I'd like to pass the map pointer into all of the native methods so that we can always `assert` that there is a valid map object that we're working with.

android/tangram/src/com/mapzen/tangram/MapController.java
blair1618: Yup, the native declaration didn't match. Fixed!

android/tangram/jni/jniExports.cpp
blair1618: Oops! Fixed.

android/tangram/jni/jniExports.cpp
tallytalwar: Ok.
But `sourcePtr` will be invalid if `mapPtr` is invalid. Since `Map` holds `TileManager` which holds the `ClientSources` in `m_tileSets`. So its a redundant check to add `mapPtr > 0` here!

android/tangram/src/com/mapzen/tangram/MapController.java
```diff
@@ -179,11 +176,22 @@ void init() {
...
+        mapPointer = nativeInit(this, assetManager);
+    }
+
+    void dispose() {
+        // Disposing native resources involves GL calls, so we need to run on the GL thread.
+        queueEvent(new Runnable() {
+            @Override
+            public void run() {
+                nativeDispose(mapPointer);
+                mapPointer = 0;
```
tallytalwar: Should we not be clearing `clientDataSources` for this Map instance, at this point?

android/tangram/jni/jniExports.cpp
blair1618: `MapController$dispose()` doesn't destroy or invalidate any `MapData` objects or the pointers that they hold, so I'm trying to safeguard against calling a function to update a data source (with a pointer > 0) while the map object itself has been disposed.

tallytalwar: Hmm.. MapController::dispose() results in deletion of native map instance, which resets its `m_tileManager` which means native pointers to its client datasources will be lost!

So there will be inconsistency between pointer value stored for MapData on this MapController.

blair1618:
@tallytalwar The private Map implementation is one more commit on top of these with about 200 LOC changed - I can submit it in a PR immediately following this one.

android/tangram/src/com/mapzen/tangram/MapController.java
blair1618: Hmm that's an interesting point. We could clear the `HashMap` of data sources, but the client application could still be holding references and so they could still be used. I think for the sake of producing useful errors in `MapController$AddDataLayer()` it probably does make sense to clear the client data sources here.

android/tangram/jni/jniExports.cpp
blair1618: Yep! The values in `MapData` will effectively be dangling pointers, but `mapPointer` is reliably set to zero, so we can check that instead.

android/tangram/src/com/mapzen/tangram/MapController.java
blair1618: Updated, `clientDataSources` is cleared here now.

blair1618:
Anything else to discuss? If not I'll merge this tonight so we can move ahead with subsequent improvements.

tallytalwar:
Go for it. I am good with it now :D.

PR #812, opened 2016-07-06
View not complete if any dependent resources are being fetched

tallytalwar:


blair1618:
It seems like this should be covered by checking whether `m_nextScene` is null in the main update loop, which is already done. Does this mean that resources may not be all retrieved when the new scene is set? How/why does that happen?

tallytalwar:
When a texture resource is a network resource, a dummy 1x1 black texture is created and scene is loaded.

Once the requested texture is fetched, its callback replaces the texture data appropriately in the scene.

`m_nextScene` only take care of actual scene loading, whereas this check, checks if any of the scene's resources are in process of being downloading.

blair1618:
Hmmmmm so that means we're rendering these "dummy" textures while we wait for the network requests? It seems somewhat broken to be displaying an incorrect texture, even temporarily, if the user has specified everything in the scene correctly. 

I think the "right" behavior here is to not set the new scene until resource requests have resolved (either in success or failure).

tallytalwar:
According to me, rest of the scene describes information for tile building, whereas textures are not needed for tile building, hence it will be wasteful for scene loading thread to wait for these textures to be downloaded and not start tile building process.

Also, it has been done to be consistent with JS. JS also has a similar behavior. cc @bcamper to confirm.

nvkelso:
Can we please use white dummy textures, instead? Black is great for debug but looks egregious in production settings.

blair1618:
@nvkelso noted: https://github.com/tangrams/tangram-es/issues/813

That's a fair point about waiting to start tile requests and geometry building. It does still seem wrong to have visible dummy textures, but maybe a better solution is to only delay drawing the new tiles until the resource requests have resolved.

bcamper:
Actually it turns out that JS does currently wait for all textures to load before displaying anything. This may have been a side effect of other resource refactoring rather than intentional :)  And it does have the downside of delaying the whole scene if only some resources are slow to retrieve. But the overall UX may be better.

I'm not sure about what happens if you swap in a new texture at run-time (after initial scene load), I would need to check that out. In any case, there are 1x1 black dummy textures for any cases that "fell through" this logic.

+1 to white dummy textures, makes sense.

tallytalwar:
Ok, I agree we can probably postpone new scene rendering until scene resources are loaded, I created an issue to do that work (https://github.com/tangrams/tangram-es/issues/814). I would stick with this approach for now though, as https://github.com/tangrams/tangram-es/issues/814 will require in a larger change and time.

core/src/tangram.cpp
```diff
@@ -329,7 +329,8 @@ bool update(float _dt) {
     bool tilesLoading = m_tileManager->hasLoadingTiles();
     bool labelsNeedUpdate = m_labels->needUpdate();
 
-    if (viewChanged || tilesChanged || tilesLoading || labelsNeedUpdate || bool(m_nextScene)) {
+    if (viewChanged || tilesChanged || tilesLoading || labelsNeedUpdate ||
+            m_scene->m_resourceLoad > 0 || bool(m_nextScene)) {
```
tallytalwar: Sure will do.

core/src/scene/scene.h
```diff
@@ -107,6 +108,8 @@ class Scene {
 
     std::shared_ptr<Texture> getTexture(const std::string& name) const;
 
+    std::atomic_ushort m_resourceLoad;
```
blair1618: I think any operations on word-sized variables are guaranteed to be atomic without any extra guards (on ARM at least, need to check for x86).

blair1618: Ok nevermind, atomic makes sense here (load and store are safe but increment can be a problem).

core/src/tangram.cpp
tallytalwar: Updated!

blair1618:
LGTM

PR #820, opened 2016-07-11
Snapshot API call

karimnaaji:
Read from framebuffer directly in core library and add an API call for it (used in tizen related work).

### DIFF
--- a/android/tangram/jni/jniExports.cpp
+++ b/android/tangram/jni/jniExports.cpp
@@ -244,6 +244,17 @@ extern "C" {
         Tangram::useCachedGlState(use);
     }
 
+    JNIEXPORT void JNICALL Java_com_mapzen_tangram_MapController_nativeCaptureSnapshot(JNIEnv* jniEnv, jobject obj, jintArray buffer) {
+        jint* arr = jniEnv->GetIntArrayElements(buffer, NULL);
+        std::vector<unsigned int> snapshot = Tangram::captureSnapshot();
+
+        for (int i = 0; i < snapshot.size(); ++i) {
+            arr[i] = snapshot[i];
+        }
+
+        jniEnv->ReleaseIntArrayElements(buffer, arr, 0);
+    }
+
     JNIEXPORT void JNICALL Java_com_mapzen_tangram_MapController_nativeQueueSceneUpdate(JNIEnv* jnienv, jobject obj, jstring path, jstring value) {
         const char* cPath = jnienv->GetStringUTFChars(path, NULL);
         const char* cValue = jnienv->GetStringUTFChars(value, NULL);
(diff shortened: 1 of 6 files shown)


tallytalwar:
Looks good. Merging.

core/src/tangram.h
```diff
@@ -166,6 +166,9 @@ void runOnMainLoop(std::function<void()> _task);
 // Run this task asynchronously to Tangram's main update loop.
 void runAsyncTask(std::function<void()> _task);
 
+// Capture a snapshot of the current frame
+std::vector<unsigned int> captureSnapshot();
```
blair1618: We can do better with this interface. The big issues are 1) the width, height, and pixel size of the captured image are not obvious and 2) `std::vector` is a pretty restrictive container, you can't get the data out without copying. 

1) is easily resolved by taking in-out integer arguments to populate with the image width, height, and pixel size.

2) is normally solved by taking a pointer argument to a buffer, but the trouble is then ensuring a large enough buffer.

android/tangram/jni/jniExports.cpp
```diff
@@ -244,6 +244,17 @@ extern "C" {
...
 
+    JNIEXPORT void JNICALL Java_com_mapzen_tangram_MapController_nativeCaptureSnapshot(JNIEnv* jniEnv, jobject obj, jintArray buffer) {
+        jint* arr = jniEnv->GetIntArrayElements(buffer, NULL);
+        std::vector<unsigned int> snapshot = Tangram::captureSnapshot();
+
+        for (int i = 0; i < snapshot.size(); ++i) {
+            arr[i] = snapshot[i];
+        }
+
+        jniEnv->ReleaseIntArrayElements(buffer, arr, 0);
```
blair1618: `SetIntArrayRegion()` should be able to handle this more succinctly and efficiently.

core/src/tangram.h
karimnaaji: Yes I agree, I'm in favor for changing that for just an allocated pointer.
Something like this would make sense to me:
```
int width = Tangram::getViewportWidth();
int height = Tangram::getViewportHeight();
// [...] alloc on snapshot by width/height
Tangram::captureSnapshot(snapshot, &scale); // write on allocated unsigned int*
```

The Java API/iOS can ensure the large enough buffer since we store can store the width and height in `MapController`/`ViewController`.

android/tangram/jni/jniExports.cpp
karimnaaji: Yes if we give a pointer to a buffer this would be more succint.

core/src/tangram.h
blair1618: I like it. For C++ interface users: Good luck ¯\\\_(ツ)\_/¯

PR #822, opened 2016-07-12
More consistent interface for converting between screen and geographic coordinates

blair1618:
1. Uses consistent types and naming for the relevant function calls. It should be obvious that these functions are the inverse of one another. 

2. `lngLatToScreenPosition` returns `true` if the point _is_ visible (the opposite of the old behavior). This is a more subjective point but to me it seems much more intuitive to associate a "positive" result with something being visible, e.g. the following use case seems reasonable:
  ```c++
  if (lngLatToScreenPosition(lng, lat, &x, &y)) {
    // draw a marker on screen at (x, y)
  }
  ```
3. `screenPositionToLngLat` returns `true` if the screen position corresponds to a geographic location (i.e. if the ray cast has a positive intersection). This is meaningful information for tilted views.

### DIFF
--- a/android/tangram/src/com/mapzen/tangram/MapController.java
+++ b/android/tangram/src/com/mapzen/tangram/MapController.java
@@ -2,6 +2,7 @@ package com.mapzen.tangram;
 
 import android.content.res.AssetManager;
 import android.graphics.Bitmap;
+import android.graphics.PointF;
 import android.opengl.GLES20;
 import android.opengl.GLSurfaceView;
 import android.opengl.GLSurfaceView.Renderer;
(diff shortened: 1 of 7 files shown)


core/src/tangram.h
```diff
@@ -88,12 +88,11 @@ float getTilt();
 
 // Transform coordinates in screen space (x right, y down) into their longitude
 // and latitude in the map view
```
tallytalwar: Can you update the comments here to reflect (`lng, lat`) pair has the appropriate value. Currently if one just reads the comments it seems `x` and `y` will be transformed/updated to have `lng`/`lat` respectively.

blair1618: Updated!

blair1618:
Updated with improved comments and one more commit.

tallytalwar:
Looks good to me. Merging.

blair1618:
One question first!

blair1618:
Ok, last issue resolved, should be good to go!

android/tangram/src/com/mapzen/tangram/MapController.java
```diff
@@ -379,12 +379,15 @@ public CameraType getCameraType() {
     /**
      * Find the geographic coordinates corresponding to the given position on screen
      * @param screenPosition Position in pixels from the top-left corner of the map area
-     * @return LngLat corresponding to the given point
+     * @return LngLat corresponding to the given point, or null if the screen position
+     * does not intersect a geographic location.
```
tallytalwar: We can mention "or null if the screen position does not intersect a geographic location, which can happen at higher tilt angles".

blair1618: Updated.

PR #824, opened 2016-07-12
Follow up of #820

karimnaaji:


core/src/tangram.h
```diff
@@ -166,8 +166,21 @@ void runOnMainLoop(std::function<void()> _task);
 // Run this task asynchronously to Tangram's main update loop.
 void runAsyncTask(std::function<void()> _task);
 
-// Capture a snapshot of the current frame
-std::vector<unsigned int> captureSnapshot();
+// Gets the viewport height in logical pixels
```
blair1618: I think this is actually hardware pixels, i.e. the dimensions of the framebuffer (framebuffer size is what matters for the dimensions of a snapshot, so I think this is the right value to return).

karimnaaji: Comment is wrong it's the physical size.

blair1618:
Beautiful :)

Merging.

PR #828, opened 2016-07-13
Update variant submodule

hjanetzek:


hjanetzek:
@blair1618 could you check what's up with travis osx/ios builds?

blair1618:
@hjanetzek It took far too many tries, but I finally got the Travis OS X environment to behave itself!

PR #840, opened 2016-07-16
Fix: make sure glMapBuffer returns valid data!

tallytalwar:
Fixes crashes on tizen emulator where mapBuffer are supported but glMapBuffers returns an invalid (nullptr) data.

Fixes one of the issues listed [here](https://github.com/mapzen/tizen-maps-plugin-mapzen/issues/22).

### DIFF
--- a/core/src/gl/mesh.cpp
+++ b/core/src/gl/mesh.cpp
@@ -102,16 +102,17 @@ void MeshBase::subDataUpload(GLbyte* _data) {
     // invalidate/orphane the data store on the driver
     GL_CHECK(glBufferData(GL_ARRAY_BUFFER, vertexBytes, NULL, m_hint));
 
+    GLvoid* dataStore = nullptr;
     if (Hardware::supportsMapBuffer) {
-        GLvoid* dataStore = glMapBuffer(GL_ARRAY_BUFFER, GL_WRITE_ONLY);
+        dataStore = glMapBuffer(GL_ARRAY_BUFFER, GL_WRITE_ONLY);
         GL_CHECK();
+        GL_CHECK(glUnmapBuffer(GL_ARRAY_BUFFER));
+    }
 
+    if (dataStore) {
         // write memory client side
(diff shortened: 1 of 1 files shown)


tallytalwar:
I verified this in my emulator setup with Bubble-wrap and pois rendering. @blair1618 can you also verify this on your end and merge this fix.

karimnaaji:
Isn't that a symptom of another issue? If `NULL` is returned, a gl error should also be triggered. Did you experience that on a `DEBUG` build? Because if `NULL` is returned and no error is returned it could mean that something else is wrong. We haven't done any code to load the gl extensions in tizen (loading the gl es so file and mapping the functions to it, as done for Android), so that might be part of the problem.

karimnaaji:
Some reference code here: https://developer.tizen.org/ko/development/api-tutorials/native-application/graphics/opengl-es?langredirect=1#ext

tallytalwar:
K, this line is the reason we are getting a nullptr with `glMapBuffer`. https://github.com/tangrams/tangram-es/compare/tizen-library#diff-c177d5f2bdeb429fe98622ab9353065fR251.

Its implemented to return nullptr in tizen gl headers for us.
It should look like this though:
```
void __glMapBuffer(GLenum target, GLenum access) {
    __evas_gl_glapi->__glMapBuffer(target, access);
}

GLboolean __glUnmapBuffer(GLenum target) {
    __evas_gl_glapi->__glUnmapBuffer(target);
}
```

Will try this and see what happens.

tallytalwar:
Ok, d436097 commit, fixes the issue on tizen.

`glMapBuffer` was hard coded to return `nullptr` and `glUnmapBuffer` was hard coded to return false for platform tizen gl implementation. I have no clue why was this done in the beginning (might have been for debugging something I suppose).

But the above mentioned commit fixes the issue for tizen and also makes the `glMapBuffer` call work in the core tangram code.

I am closing this PR as this was not the cause of the issue.

PR #842, opened 2016-07-19
Automatic label anchor (rebased)

karimnaaji:
Rebase of #806 

- [x] Update unit tests
- [x] Update syntax:
    - single value: `anchor: top`, 
    - multiple values: anchor: `[right, bottom, left, top]`
- [x] Set defaults according to JS
- [ ] parse YAML list

Optimizations
- [ ] ~~Don't use multiple sets of quads (for each anchor)~~, follow up work going in #862 
- [x] Reference three alignments instead of one per anchor
- [x] LabelOptions::m_fallbackAnchors could be stored more efficiently as 9*4 bytes fixed array.  

Bugs
~~- [ ] label offset does not work (check bubble-wrap with debug_all_labels)~~

Notes:
Label now stores a set of Anchors which are defined by the anchor-style parameter 
https://github.com/tangrams/tangram-es/pull/842/files#diff-7d439434d5d36de1c37a6cb2f01b81aaR37.

When a label got occluded (or couldn't be placed when its tile becomes active) the label state changes from ```visible/(none) -> fallback``` and the next anchor position is tried on the next frame until all positions have been tried or the label could be placed. 
https://github.com/tangrams/tangram-es/pull/842/files#diff-d66207e00a5f8416ddf6f8aea50052dcR284

A TextLabel now stores quads for left/right/center alignments which are chosen to match the currently active Anchor.  https://github.com/tangrams/tangram-es/pull/842/files#diff-1fd6c9d719003eb34d010a7063a941e7R38, https://github.com/tangrams/tangram-es/pull/842/files#diff-4cb2204fedb2ba4ea43314c75ab033f0R204

### DIFF
--- a/android/tangram/src/com/mapzen/tangram/MapController.java
+++ b/android/tangram/src/com/mapzen/tangram/MapController.java
@@ -60,7 +60,7 @@ public class MapController implements Renderer {
         TILE_INFOS,
         LABELS,
         TANGRAM_INFOS,
-        ALL_LABELS,
+        DRAW_ALL_LABELS,
         TANGRAM_STATS,
     }
 
--- a/core/src/labels/fadeEffect.h
+++ b/core/src/labels/fadeEffect.h
@@ -35,6 +35,13 @@ public:
         return st;
     }
 
+    void reset(bool _in, Interpolation _interpolation, float _duration) {
+        m_in = _in;
+        m_interpolation = _interpolation;
+        m_duration = _duration;
+        m_step = 0.f;
+    }
+
     bool isFinished() {
         return m_step > m_duration;
     }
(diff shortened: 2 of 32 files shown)


karimnaaji:
Both @hjanetzek and me are happy with the state of this branch, @tallytalwar can you give a quick review on this.

tallytalwar:
Will get to reviewing this soon.
Do you want to do a cleanup of the git history here? I see some specific "fix", "squash" commits here.

tallytalwar:
There are a bunch of conceptual changes here, can we document these in the PR description please. Will help reviewing this and also for debugging in future.

hjanetzek:
@tallytalwar added some notes to the PR description.

hjanetzek:
@karimnaaji thinking about the fallback logic again: why not try the fallbacks immediately instead of waiting for the next frame? I'll see if this works

tallytalwar:
Good to go.

hjanetzek:
Continuing at https://github.com/tangrams/tangram-es/pull/893

PR #844, opened 2016-07-19
Make Tangram::Map implementation private

blair1618:
Instead of holding all of our subsystem objects directly in the `Map` class, `Map` now holds a `unique_ptr` to an object of the private class `Map::Impl` and all of the subsystem objects are members of `Map::Impl`.

This significantly reduces the declarations needed in `tangram.h` and in the future will allow us to change the implementation of `Map` without changing this header.

### DIFF
--- a/core/src/tangram.h
+++ b/core/src/tangram.h
@@ -1,24 +1,14 @@
 #pragma once
 
 #include "data/properties.h"
-#include "util/ease.h"
-#include <array>
+#include <functional>
 #include <memory>
-#include <mutex>
-#include <queue>
 #include <string>
 #include <vector>
 
 namespace Tangram {
 
-class AsyncWorker;
 class DataSource;
-class InputHandler;
-class Labels;
-class Scene;
-class TileManager;
-class TileWorker;
-class View;
 
 struct TouchItem {
     std::shared_ptr<Properties> properties;
@@ -29,6 +19,13 @@ struct TouchItem {
 struct SceneUpdate {
     std::string keys;
     std::string value;
+};
+
+enum class EaseType : char {
+    linear = 0,
+    cubic,
+    quint,
+    sine,
 };
 
 class Map {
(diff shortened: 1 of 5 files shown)


karimnaaji:
Could we clear tiles for all map instance when triggering a debug flag that requests tile reloading?

blair1618:
Maybe! We'd have to register all map instances to a global variable somewhere, which seemed undesirable. 

I had a weird idea on fixing this: `getDebugFlag()` should remain a global because it means we can use these test conditions from anywhere in the code, and that's really useful. But `setDebugFlag()` could be a non-static function in `Map` so that we have access to the tile sets whenever we set a flag that needs to refresh them. It's weirdly asymmetrical, but kinda fits our needs.

tallytalwar:
Reviewing this now.

core/src/tangram.h
```diff
@@ -184,29 +181,8 @@ class Map {
...
-    std::shared_ptr<Scene> m_nextScene;
-    std::shared_ptr<View> m_view;
-    std::unique_ptr<Labels> m_labels;
-    std::unique_ptr<InputHandler> m_inputHandler;
-    std::unique_ptr<AsyncWorker> m_asyncWorker;
-    std::vector<SceneUpdate> m_sceneUpdates;
-    std::array<Ease, 4> m_eases;
-    bool m_cacheGlState;
+    class Impl;
+    std::unique_ptr<Impl> impl;
```
tallytalwar: Just a note here that `Map::Impl` class has its move constructor suppressed, unless explicitly defined/defaulted. This might have an affect if the user of the `Map` class tries to `move` an instance.
Reference: http://en.cppreference.com/w/cpp/language/pimpl

tallytalwar: On similar note, should we restrict copying of a Map object by explicitly deleting copy constructor? RFC!

blair1618: `Map` is implicitly non-copyable - the compiler can't generate a copy constructor because of the `unique_ptr`

Supporting move semantics for `Map` would be trivial, I don't know if there's a serious need for it though.

tallytalwar: Yeah I also do not see a serious need or use case for a move for `Map`. Wanted to bring this out, just in case, someone has a use case in mind.

tallytalwar:
This looks good to me. If no one has further comments will merge (squash, to save one merge commit) this.

blair1618:
Thanks! Yeah, squash makes sense to me.

PR #845, opened 2016-07-20
Get system font file on tizen

tallytalwar:
`systemFontPath` returns a closest matching font to the specified `family`, `style` and `weight`. This uses the `FontConfig` API (https://www.freedesktop.org/software/fontconfig/fontconfig-devel/).

Todo:
- [x] Need some special code to handle fallback fonts.

### DIFF
--- a/tizen/src/platform_tizen.cpp
+++ b/tizen/src/platform_tizen.cpp
@@ -6,6 +6,7 @@
 #include <list>
 #include <memory>
 #include <algorithm>
+#include <vector>
 
 //#include "urlWorker.h"
 #include "platform_tizen.h"
@@ -26,6 +27,9 @@
 
 static bool s_isContinuousRendering = false;
 static std::function<void()> s_renderCallbackFunction = nullptr;
+
+static std::vector<std::string> s_fallbackFonts;
+static FcConfig* s_fcConfig = nullptr;
 
 #if USE_ECORE_CON
 #include <Eina.h>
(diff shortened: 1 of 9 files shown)


hjanetzek:
One way to get fallback fonts could be to query fontconfig for 'Regular' style fonts and then pick the first font for each FC_LANG

tallytalwar:
Yeah thats a good suggestion, will try that.

We should also add this code for linux platforms.

tizen/src/platform_tizen.cpp
```diff
@@ -225,6 +227,46 @@ std::string systemFontFallbackPath(int _importance, int _weightHint) {
     return "";
 }
 
+// No system fonts implementation (yet!)
```
blair1618: Not anymore :)

tizen/src/platform_tizen.cpp
```diff
@@ -225,6 +227,46 @@ std::string systemFontFallbackPath(int _importance, int _weightHint) {
...
+
+    std::string fontFile;
+    FcValue fcFamily, fcFace, fcWeight;
+
+    fcFamily.type = fcFace.type = fcWeight.type = FcType::FcTypeString;
+    fcFamily.u.s = (const FcChar8*)(_name.c_str());
+    fcWeight.u.s = (const FcChar8*)(_weight.c_str());
+    fcFace.u.s = (const FcChar8*)(_face.c_str());
+
+    FcConfig* config = FcInitLoadConfigAndFonts();
```
blair1618: Seems like we should probably do this once instead of every function call.

tallytalwar: Yeah I am taking care of this in subsequent commits which also implement fallback fonts support.

tallytalwar:
@hjanetzek Updated with fallback implemetation and some fixups.

tizen/src/platform_tizen.cpp
```diff
@@ -220,9 +226,103 @@ std::string stringFromFile(const char* _path) {
     return out;
 }
 
-// No system fonts fallback implementation (yet!)
+void initPlatformFontSetup() {
```
karimnaaji: Do you think we could make that a lazy initialization on tizen? It seems pretty inconvenient to add this to all other platforms.

tallytalwar: Hmm.. we can probably do this in `FontContext` constructor! Which means for every scene load/update this will be called.

karimnaaji: A simple local static `isInitialized` within `getFallback` should do the job. And on the first call to it branch to everything within `initPlatformFontSetup`.

tallytalwar: Updated.

tizen/src/platform_tizen.cpp
```diff
@@ -220,9 +226,110 @@ std::string stringFromFile(const char* _path) {
...
+
+    static bool s_platformFontsInit = false;
+    if (s_platformFontsInit) { return; }
+
+    s_fcConfig = FcInitLoadConfigAndFonts();
+
+    std::string style = "Regular";
+
+    FcStrSet* fcLangs = FcGetLangs();
+    FcStrList* fcLangList = FcStrListCreate(fcLangs);
```
karimnaaji: Should `FcStrListDone` be called at the end?

karimnaaji: http://refspecs.linuxfoundation.org/fontconfig-2.6.0/r4065.html

tallytalwar: Yup, testing that right now. Nice catch.

tallytalwar: Which does not crash unlike `FcValueDestroy`.

tizen/src/platform_tizen.cpp
```diff
@@ -220,9 +226,110 @@ std::string stringFromFile(const char* _path) {
...
+    FcResult res;
+    FcPattern* font = FcFontMatch(s_fcConfig, pattern, &res);
+    if (font) {
+        FcChar8* file = nullptr;
+        FcChar8* fontFamily = nullptr;
+        if (FcPatternGetString(font, FC_FILE, 0, &file) == FcResultMatch &&
+            FcPatternGetString(font, FC_FAMILY, 0, &fontFamily) == FcResultMatch) {
+            // We do not want the "best" match, but an "exact" or at least the same "family" match
+            // We have fallbacks to cover rest here.
+            if (std::string((char*)fontFamily) == _name) {
```
karimnaaji: We could save a string copy `strcmp((char*)fontFamily, _name.c_str());`.

tizen/src/platform_tizen.cpp
tallytalwar: Fixed here:  bd2231c

tizen/src/platform_tizen.cpp
tallytalwar: Updated.

karimnaaji:
LGTM, @hjanetzek can review and merge. 

`FontContext` would need a major cleanup from platform dependent code.

tizen/src/platform_tizen.cpp
```diff
@@ -242,7 +242,7 @@ void initPlatformFontSetup() {
         FcValue fcStyleValue, fcLangValue;
 
         fcStyleValue.type = fcLangValue.type = FcType::FcTypeString;
-        fcStyleValue.u.s = (const FcChar8*)(style.c_str());
+        fcStyleValue.u.s = reinterpret_cast<const FcChar8*>(style.c_str());
```
tallytalwar: :+1

Issue #821, opened 2016-07-11
Perspective camera 'vanishing_point' behaves differently from tangram-js


blair1618:
The apparent center of the view is different in tangram-es and tangram-js when a non-zero vanishing point vector is applied.

PR #846, opened 2016-07-20
Add interface to offset the map position within the viewport

### LINKED ISSUE #841: Enable offsets to map position within viewport
This is a common need in navigation applications. When displaying a navigation UI, it is often useful to track the user's current location and place the user's current location in the lower portion of the display in order to show more area in the direction of travel. 

Currently in tangram-es there is no simple mechanism to do this - the location given to `setPosition()` will always be in the center of the display. The only work-around is to call `setPosition()` with a location that is ahead of the user's location in the direction of travel. 

A possible interface for this could be:

``` c++
// Set the displacement of the map's position from the center of the display.
// The location set as the map position will appear at the screen position 
// (w/2 + xPixels, h/2 + yPixels) where 'w' and 'h' are the viewport width and height.
void setViewOffset(int xPixels, int yPixels);
```

matteblair:
This adds a parameters to the projection matrix that effectively shift the viewport by a given offset in screen space. This works great for vertical offsets, but feels a little weird for horizontal offsets since the camera perspective is no longer centered on the middle of the screen.

Resolves #841

### DIFF
--- a/android/tangram/jni/jniExports.cpp
+++ b/android/tangram/jni/jniExports.cpp
@@ -161,6 +161,20 @@ extern "C" {
         return map->getCameraType();
     }
 
+    JNIEXPORT void JNICALL Java_com_mapzen_tangram_MapController_nativeSetViewOffset(JNIEnv* jniEnv, jobject obj, jlong mapPtr, jint xPixels, jint yPixels) {
+        assert(mapPtr > 0);
+        auto map = reinterpret_cast<Tangram::Map*>(mapPtr);
+        map->setViewOffset(xPixels, yPixels);
+    }
+
+    JNIEXPORT void JNICALL Java_com_mapzen_tangram_MapController_nativeGetViewOffset(JNIEnv* jniEnv, jobject obj, jlong mapPtr, jintArray offset) {
+        assert(mapPtr > 0);
+        auto map = reinterpret_cast<Tangram::Map*>(mapPtr);
+        jint* arr = jniEnv->GetIntArrayElements(offset, NULL);
+        map->getViewOffset(&arr[0], &arr[1]);
+        jniEnv->ReleaseIntArrayElements(offset, arr, JNI_ABORT);
+    }
+
     JNIEXPORT void JNICALL Java_com_mapzen_tangram_MapController_nativeHandleTapGesture(JNIEnv* jniEnv, jobject obj, jlong mapPtr, jfloat posX, jfloat posY) {
         assert(mapPtr > 0);
         auto map = reinterpret_cast<Tangram::Map*>(mapPtr);
--- a/android/tangram/src/com/mapzen/tangram/MapController.java
+++ b/android/tangram/src/com/mapzen/tangram/MapController.java
@@ -2,6 +2,7 @@ package com.mapzen.tangram;
 
 import android.content.res.AssetManager;
 import android.graphics.Bitmap;
+import android.graphics.Point;
 import android.graphics.PointF;
 import android.opengl.GLES20;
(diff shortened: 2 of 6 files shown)


tallytalwar:
Reviewing this now.

tallytalwar:
Ok, seems to mess-up orthographic ("isometric") projection.
Will look at this more tomorrow.

tallytalwar:
Updated this to include a fix for isometric projection.

blair1618:
Gonna rebase this and see if I can fix the build error - hang tight.

blair1618:
Ok so this is weird - the transformation we're applying for this view offset is essentially the same transformation we use for 'vanishing_point':

https://github.com/tangrams/tangram-es/blob/8c0bbcf15374d78bfc6e16ffaa64a118d03080d2/core/src/view/view.cpp#L349-L350
```cpp
m_proj[2][0] -= m_vanishingPoint.x / getWidth();
m_proj[2][1] -= m_vanishingPoint.y / getHeight();
```

https://github.com/tangrams/tangram-es/blob/8c0bbcf15374d78bfc6e16ffaa64a118d03080d2/core/src/view/view.cpp#L377-L378
```cpp
m_proj[2][0] = (-2.f * m_offset.x) / m_vpWidth; // (right + left) / (right - left);
m_proj[2][1] = (-2.f * m_offset.y) / m_vpHeight; // (top + bottom) / (top - bottom);
```

This explains why the view starts to look 'skewed' with some offset values - the effective 'vanishing point' is shifting. This also starts to explain https://github.com/tangrams/tangram-es/issues/821. I'm thinking that one or both of these transformations needs to be adjusted.

blair1618:
More research! I noticed that there was a factor of 2 different between the two vanishing-point-like transforms we were applying, so I investigated by setting `vanishing_point: [600, 0]` and turning on the tile grid lines:

vanishing-point-300-0

The vanishing point is displaced by half the expected distance!

I checked in tangram-js as well and it turns out the same thing happens there too: https://github.com/tangrams/tangram/issues/372

This suggests to me that _both_ of these transforms should be multiplying the pixel values by 2 before they are applied to the matrix. This was the logical conclusion of creating the 'offset' for this PR - the values in the matrix are "(right + left) / (right - left)" and "(top + bottom) / (top - bottom)" which end up being proportional to 2 * the x offset and 2 * the y offset, respectively. 

The problem with our 'vanishing point' implementation (which leads to https://github.com/tangrams/tangram-es/issues/821) is that we don't negate the translation created by the 'vanishing point' change, as the tangram-js camera does: https://github.com/tangrams/tangram/blob/392e44cce9b137dbb2903209ade3e81b4d8cee81/src/camera.js#L186-L187

Work on this continues...

tallytalwar:
Nice this solves the mystery for me for the matrix update for the vanishing point. Both `vanishing point` and `principle point offset` are the same thing.

tallytalwar:
This does not fix https://github.com/tangrams/tangram-es/issues/821 at the moment, but on the other hand fixes the halfing of vanishing point affect and adds the offsetting functionality for `perspective` and `orthographic` projections.

I can spend some time to get #821 with this PR, but since its a separate issue, we can address that in another PR and merge this one in, thoughts?

blair1618:
This PR has left me a bit baffled and I'm not sure what to do with it yet. I'd prefer to understand and correct the erroneous screen-center-displacement that `vanishing_point` produces before adding this new change on top of that.

This change also doesn't do exactly what we would want for the view offset, since we've discovered that it also changes the vanishing point. I think this could benefit from a bit more tinkering and thinking.

PR #847, opened 2016-07-21
added beginChangeBlock() and endChangeBlock() to MapData.

redfish64:
I'm converting my android app, Tiny Travel Tracker, to mapzen. It contains an database on the phone of gps points collected continuously tracking the users movements for later retrieval, which I'd like to add as a tile source (my idea is it would just loop back to the android app to query its database). Since my app can handle hundreds of thousands of points (personally I have over 500,000 stored tracking me over the last 6 years),

I've tried using addFeature() to display a subset of points (of what is on the screen currently) but it creates a lot of flickering as the points are deleted and redrawn.

I was able to change the source to fix my problem by adding two methods to MapData, beginChangeBlock(), which prevents feature changes from immediately being reflected in the rendered view, and endChangeBlock(), which commits the features after the beginChangeBlock() all at once. This successfully prevents flickering in my app and fixes my problem.

I hope this change is useful.

### DIFF
--- a/android/tangram/jni/jniExports.cpp
+++ b/android/tangram/jni/jniExports.cpp
@@ -241,8 +241,24 @@ extern "C" {
         assert(mapPtr > 0);
         auto map = reinterpret_cast<Tangram::Map*>(mapPtr);
         assert(sourcePtr > 0);
-        auto source = reinterpret_cast<Tangram::DataSource*>(sourcePtr);
-        map->clearDataSource(*source, true, true);
+        auto source = reinterpret_cast<Tangram::ClientGeoJsonSource*>(sourcePtr);
+        source->clearData();
+    }
+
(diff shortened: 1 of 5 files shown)


blair1618:
That's a lot of points - this interface makes a lot of sense for your needs.

Funnily enough, the `MapData` class used to work this way (https://github.com/tangrams/tangram-es/commit/163bbcb07f1044abcfa453c0a9ec29b573510767)! Features could be added in Java and then synchronized to the map in batches so that large updates wouldn't choke the renderer. Then it turned out that most of the time `MapData` was being used to draw one or two features at a time and the extra call to synchronize the data was troublesome for those cases, so we changed it to just synchronize on every new feature. 

We're now planning to replace the common usage of `MapData` for individual markers and annotations with a new, more focused interface, which means that `MapData` could once again become more dedicated to large sets of features. At that point we'll add an interface like the one you have here. We're going to hold off on adding this change to `master` until then.

PR #863, opened 2016-07-22
Update versioning.gradle to allow an empty patch

tallytalwar:
0.x will work, and one will not have to put in an extra "0" for the patch.


tallytalwar:
@blair1618 give it a look. Once this is in, will start 0.4 versioning process.

blair1618:
LGTM

PR #868, opened 2016-07-25
Use gcc-5 for Travis Linux builds

blair1618:
GCC 5 is the first version to fully support the C++14 standard: https://gcc.gnu.org/projects/cxx-status.html

The ubuntu-toolchain-r-test PPA provides version 5.4.0-3 for Ubuntu 12.04: https://launchpad.net/~ubuntu-toolchain-r/+archive/ubuntu/test

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,11 +6,11 @@ matrix:
           env: PLATFORM=osx
           compiler: clang
         - os: linux
-          env: PLATFORM=linux CXX=g++-4.9 CC=gcc-4.9
+          env: PLATFORM=linux CXX=g++-5 CC=gcc-5
           addons:
             apt:
               sources: [ 'kubuntu-backports', 'ubuntu-toolchain-r-test', 'george-edison55-precise-backports' ]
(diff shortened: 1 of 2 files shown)


tallytalwar:
Can you try cherry-picking this commit on renderstate branch to confirm gcc 5.4 fixes that issue for us on travis?

blair1618:
`renderstate-non-static` works locally with GCC 5.3 on Trusty

tallytalwar:
We should also test on rpi and update the readme for that also.

blair1618:
Yeah the Pi is an interesting question. As far as I can tell, the most recent version that can be installed from a package is 4.9.2 (any newer and you have to compile gcc yourself).

blair1618:
Verified that Clang 3.5 can successfully build tangram-es on the raspberry pi - I updated the README to make it the recommended compiler.

blair1618:
All good here?

tallytalwar:
Yes yes, will merge. Was in middle of tizen gerrit stuff when you updated this before, then forgot about this! (Gerrit can make you forget stuff).

PR #872, opened 2016-07-27
fix label flickering/duplicate tiles with different source-zoom

hjanetzek:
When tiles are mapped to the maximum source zoom, tiles with the same coordinates (but different style-zoomlevels) can appear in the the visible-set.

This change puts only the tile with the higher style-zoomlevel to the visible-set.


tallytalwar:
Going to merge this, and as discussed in the meeting we "might" refactor the logic of the view sending the visible tile set taking style zooms into account. (It looks kinda not possible because different datasources have different max zooms, which means tilemanager should be the place this check should be happening)

Issue #839, opened 2016-07-15
Add support for scene `fonts` block


bcamper:
See https://github.com/tangrams/tangram/pull/326 for details.
Also relevant: #838.

Issue #814, opened 2016-07-06
Postpone scene rendering until all resources are ready


tallytalwar:
Currently we start rendering a scene if texture resources are not ready and are in process of being downloaded. And the current state of the scene is to substitute dummy textures, which get replaced once actual resources are fetched.

Instead of the current approach of rendering dummy resources, we would like to postpone the scene rendering until all scene's resources are ready. (This should also include a tile's raster texture resource, in that case).

bcamper:
To clarify, in JS we do continue *building* a tile that has dependent raster sources, we just can't *display* it until the textures have loaded. We don't need to wait around for textures to load to do other feature/style parsing work though.

tallytalwar:
Yeah, we do the same on ES, this issue is just postponing the actual _drawing_ of the tiles. _Tile building_ will still take place.

PR #874, opened 2016-07-27
Support for remote fonts

karimnaaji:
https://github.com/tangrams/tangram-es/issues/839

### DIFF
--- a/core/src/tangram.cpp
+++ b/core/src/tangram.cpp
@@ -313,7 +313,7 @@ bool Map::update(float _dt) {
     bool tilesChanged = impl->tileManager.hasTileSetChanged();
     bool tilesLoading = impl->tileManager.hasLoadingTiles();
     bool labelsNeedUpdate = impl->labels.needUpdate();
-    bool resourceLoading = (impl->scene->m_resourceLoad > 0);
+    bool resourceLoading = (impl->scene->resourceLoad > 0);
     bool nextScene = bool(impl->nextScene);
 
     if (viewChanged || tilesChanged || tilesLoading || labelsNeedUpdate || resourceLoading || nextScene) {
--- a/core/src/tangram.cpp
+++ b/core/src/tangram.cpp
@@ -257,6 +257,11 @@ void Map::resize(int _newWidth, int _newHeight) {
 
 bool Map::update(float _dt) {
 
+    // Wait until font resources are fully loaded
+    if (impl->scene->fontContext()->resourceLoad > 0) {
+        return false;
+    }
+
     FrameInfo::beginUpdate();
 
     JobQueue::runJobsForCurrentThread();
--- a/core/src/text/fontContext.cpp
+++ b/core/src/text/fontContext.cpp
@@ -35,6 +35,8 @@ FontContext::FontContext() :
     m_atlas(*this, GlyphTexture::size, m_sdfRadius),
     m_batch(m_atlas, m_scratch) {
 
+    resourceLoad = 0;
+
 // TODO: make this platform independent
 #if defined(PLATFORM_ANDROID)
     auto fontPath = systemFontPath("sans-serif", "400", "normal");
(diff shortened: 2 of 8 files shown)


core/src/tangram.cpp
```diff
@@ -257,6 +257,11 @@ void Map::resize(int _newWidth, int _newHeight) {
 
 bool Map::update(float _dt) {
 
+    // Wait until font resources are fully loaded
+    if (impl->scene->fontContext()->resourceLoad > 0) {
```
tallytalwar: We can do something similar for any uri resource from the scene.
"Wait until texture resources are loaded".

tallytalwar: Maybe make sense to do this in render` call for texture resources, since these are not required for "tile building" which happens in the update loop.

core/src/text/fontContext.cpp
```diff
@@ -242,6 +248,73 @@ bool FontContext::layoutText(TextStyle::Parameters& _params, const std::string&
...
+    std::smatch match;
+
+    if (std::regex_search(_ft.uri, match, regex)) {
+        // Load remote
+        resourceLoad++;
+        startUrlRequest(_ft.uri, [&, _ft](std::vector<char>&& rawData) {
+            if (rawData.size() == 0) {
+                LOGE("Bad URL request for font %s at URL %s", _ft.alias.c_str(), _ft.uri.c_str());
+            } else {
+                std::lock_guard<std::mutex> lock(m_mutex);
```
tallytalwar: nitpick, lock_guard can start from just before you start adding faces to the font. So just before the for loop.

core/src/tangram.cpp
karimnaaji: Yes there's a slight difference between remote texture resource and remote font resource:
- remote texture resource doesn't lock tile building and will render the black texture until being fetched
- remote font resource would lock tile building and thus rendering

I guess there's also some upcoming work to go in https://github.com/tangrams/tangram-es/issues/814 for that concern.

tallytalwar: Yup, my comments were in reference to #814.

core/src/text/fontContext.cpp
karimnaaji: There's no runtime cost in `reinterpret_cast`, except if we want to save the `int`. I'd rather leave it at the beginning of the block for easier reading. otherwise I'll be tempted to write another `{`,`}` to guard.

core/src/text/fontContext.cpp
```diff
@@ -254,64 +327,48 @@ void FontContext::ScratchBuffer::drawGlyph(const alfons::Rect& q, const alfons::
              {glm::vec2{q.x2, q.y2} * TextVertex::position_scale, {g.u2, g.v2}}}});
 }
 
-auto FontContext::getFont(const std::string& _family, const std::string& _style,
-                          const std::string& _weight, float _size) -> std::shared_ptr<alfons::Font> {
+std::shared_ptr<alfons::Font> FontContext::getFont(const std::string& _family, const std::string& _style,
+                                                   const std::string& _weight, float _size) {
```
tallytalwar: nitpick. Whitespace :D.

core/src/text/fontContext.cpp
```diff
@@ -254,64 +327,48 @@ void FontContext::ScratchBuffer::drawGlyph(const alfons::Rect& q, const alfons::
...
-    auto font = m_alfons.getFont(fontName, fontSize);
+    auto font =  m_alfons.getFont(FontDescription::Alias(_family, _style, _weight), fontSize);
     if (font->hasFaces()) { return font; }
 
     size_t dataSize = 0;
     unsigned char* data = nullptr;
 
     // Assuming bundled ttf file follows this convention
-    auto bundledFontPath = "fonts/" + _family + "-" + _weight + _style + ".ttf";
+    std::string bundleFontPath = BUNDLE_FONT_PATH + FontDescription::BundleAlias(_family, _weight, _style);
```
tallytalwar: :+1:

tallytalwar:
This looks good to me.

PR #876, opened 2016-07-27
Additional cleanup and safety checks for Android

blair1618:
- Checks the validity of every value used as a native pointer and throws Exceptions for bad values
 - Calls remove() on every MapData associated with a MapController in dispose()

### DIFF
--- a/android/tangram/src/com/mapzen/tangram/MapController.java
+++ b/android/tangram/src/com/mapzen/tangram/MapController.java
@@ -16,6 +16,7 @@ import com.squareup.okhttp.Response;
 import java.io.IOException;
 import java.nio.IntBuffer;
 import java.util.HashMap;
+import java.util.Iterator;
 import java.util.Map;
 
 import javax.microedition.khronos.egl.EGLConfig;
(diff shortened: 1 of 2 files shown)


android/tangram/src/com/mapzen/tangram/MapController.java
```diff
@@ -184,6 +184,9 @@ void dispose() {
         queueEvent(new Runnable() {
             @Override
             public void run() {
+                for (Map.Entry<String, MapData> entry : clientDataSources.entrySet()) {
```
tallytalwar: Can you also do a `checkPointer(mapPointer)` when map is init. Line: 179.

P.S. Github did not allow me to comment of that specific line!!

android/tangram/src/com/mapzen/tangram/MapData.java
```diff
@@ -49,8 +49,9 @@ public String name() {
     }
 
     /**
-     * Remove this {@code MapData} from the map it is currently associated with. After
-     * {@code remove} is called, this object can no longer apply changes to the map.
+     * Remove this {@code MapData} from the map it is currently associated with. Using this object
+     * after {@code remove} is called will cause a crash. {@code remove} is called on every
```
tallytalwar: will cause an `exception` instead of a `crash`.

tallytalwar:
Good to go, once couple of commits are addressed.

blair1618:
Thanks, updated!

PR #879, opened 2016-07-28
Flip y coordinates of texture data from stb_image

### LINKED ISSUE #869: Raster normals with spheremap produces reversed output from Tangram JS
This is demonstrated in the walkabout style: https://tangrams.github.io/magazine/?scene=https://tangrams.github.io/walkabout-style/walkabout-style.yaml#13/37.7639/-122.4323

screen shot 2016-07-26 at 9 50 54 am

In walkabout this effect is implemented with a custom shader: https://github.com/tangrams/walkabout-style/blob/gh-pages/walkabout-style.yaml#L709

Tangram-es injects normals in the same way as JS: https://github.com/tangrams/tangram-es/blob/5d80c7e586dc7c76142547de73dd1a264f518569/core/shaders/polygon.fs#L66

Tangram-es also appears to calculate normals from textures in the same way as JS too: https://github.com/tangrams/tangram/blob/b4a9ec50700c880df272c983acd2ab6e93bc0ef6/src/gl/shaders/material.glsl#L122

blair1618:
Resolves #869 

This could use a careful eye - texture y-coordinates get confusing very quickly.

### DIFF
--- a/core/src/scene/sceneLoader.cpp
+++ b/core/src/scene/sceneLoader.cpp
@@ -544,7 +544,7 @@ std::shared_ptr<Texture> SceneLoader::fetchTexture(const std::string& name, cons
                 std::lock_guard<std::mutex> lock(m_textureMutex);
 				auto texture = scene->getTexture(name);
                 if (texture) {
-                    if (!texture->loadImageFromMemory(ptr, dataSize, false)) {
+                    if (!texture->loadImageFromMemory(ptr, dataSize)) {
                         LOGE("Invalid texture data '%s'", url.c_str());
                     }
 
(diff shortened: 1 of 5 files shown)


karimnaaji:
Reviewing this.

core/src/scene/sceneLoader.cpp
```diff
@@ -544,7 +544,7 @@ std::shared_ptr<Texture> SceneLoader::fetchTexture(const std::string& name, cons
                 std::lock_guard<std::mutex> lock(m_textureMutex);
 				auto texture = scene->getTexture(name);
                 if (texture) {
-                    if (!texture->loadImageFromMemory(ptr, dataSize, false)) {
+                    if (!texture->loadImageFromMemory(ptr, dataSize, true)) {
```
karimnaaji: We can remove this param and set it to `true` by default within texture loading. It's a global state within the `stb` lib, and it can be called in worker threads that changing this option while loading another texture. Or make it thread safe, but first option sounds better.

blair1618: Good call - definitely not threadsafe: https://github.com/nothings/stb/blob/master/stb_image.h#L931-L936

I can't think of a reason to load images in the opposite orientation, so I agree on getting rid of this parameter. I'll update.

tallytalwar:
This looks good to me.

PR #881, opened 2016-07-28
Coverity issues

hjanetzek:


core/include/earcut.hpp
```diff
@@ -1 +1 @@
-Subproject commit 1384c97a79c322e1aad37fd317bf87be5dbb24fa
+Subproject commit 195b2fdae9a354c0f86592f320df4988c896f822
```
blair1618: This update looks fine to me. Have you checked whether these issues are fixed in the original mapbox repo?

hjanetzek: This code deviated a bit from mapbox/master - so I can't tell. The nullptr deref is also probably only shown as defect since coverity didn't do the math for not hitting sum != 0

external/alfons
```diff
@@ -1 +1 @@
-Subproject commit bbb81cc67f646e4c97a861103a55275c4458e7ed
+Subproject commit 118e3cfaa7293e02ccf4eb73cb8064674b125cb2
```
blair1618: Looks good, but maybe it's time to call this branch something besides 'wip' :P

blair1618:
LGTM. Will merge if nobody has more to say on this.

tallytalwar:
Looks good to me too.

PR #887, opened 2016-07-29
Remove global state from GL management

blair1618:
or, *Towards Enabling Multiple Concurrent Tangram Map Instances*.

Major changes:

 - `RenderState` is now a class instead of a namespace of globally-scoped variables. 
 - `RenderState` now uses a collection of small, anonymous structs to track GL state. The class templates used before became difficult to maintain once members could no longer be accessed globally. 
 - A `RenderState&` is passed into all functions that need to modify an aspect of the tracked state. 
 - `JobQueue` is simpler now and uses no global state. 
 - `VertexLayout` now uses no global state, instead using a member of `RenderState` to track attribute bindings. 
 - Added a disposal mechanism that ensures GL resources are only disposed during the `Map::render()` call. 
 - The `QuadIndices`, `ShaderProgram`, and `Vao` classes were simplified or refactored to accommodate other changes.


blair1618:
I am looking into a bug that prevents textures from being correctly re-generated after GL context loss on Android. Feel free to look for clues while reviewing :)

core/src/gl/mesh.cpp
```diff
@@ -143,51 +130,65 @@ void MeshBase::upload() {
...
+            if (m_glVertexBuffer) {
+                rs.vertexBufferUnset(m_glVertexBuffer);
+                GL_CHECK(glDeleteBuffers(1, &m_glVertexBuffer));
+            }
+            if (m_glIndexBuffer) {
+                rs.indexBufferUnset(m_glIndexBuffer);
+                GL_CHECK(glDeleteBuffers(1, &m_glIndexBuffer));
+            }
+            m_vaos.dispose();
+        }
```
hjanetzek: This is a copy of m_vaos before it was initialized. Since this keeps a reference to RenderState in disposer one could also keep a reference to it in Mesh and create the disposer in Mesh's destructor.

hjanetzek:
ok, the last commits are very WIP -- still valgrinding

core/src/gl/mesh.cpp
blair1618: Technically yes, GL objects could just keep a reference to the RenderState - though I was trying to avoid to prevent accidentally using the RenderState when we don't have a valid context.

hjanetzek: I see. The Disposer is back!

blair1618:
Those last commits look good to me. The iterator fix was a good catch - I got greedy with caching `end()`. Refactoring `Disposer` to be called in destructors is probably a good move for readability - I would've preferred to keep the function call more descriptive but this is fine. And cleaning up `disposeBuilder()` is a nice perk here that didn't occur to me!

There's still an issue with textures (and maybe meshes?) after context loss on Android - I'll take another look at that today.

core/src/gl/dynamicQuadMesh.cpp
```diff
@@ -1,41 +1,28 @@
 #include "gl/dynamicQuadMesh.h"
+#include "gl/error.h"
 #include "gl/renderState.h"
 #include "gl/shaderProgram.h"
 
 namespace Tangram {
 
 GLuint QuadIndices::quadIndexBuffer = 0;
 int QuadIndices::quadGeneration = -1;
-std::atomic<int> QuadIndices::meshCounter(0);
```
tallytalwar: What was the purpose of `meshCounter` here?
I see it was used with every construction of `DyanmicQuadMesh`.

This could be the issue resulting in quad mesh issues with android gl context losses!

core/src/gl/dynamicQuadMesh.cpp
```diff
@@ -1,41 +1,28 @@
...
-void QuadIndices::unref() {
-    meshCounter--;
-
-    if (quadIndexBuffer != 0 && (!RenderState::isValidGeneration(quadGeneration) ||
-                                 meshCounter <= 0)) {
-
-        if (RenderState::indexBuffer.compare(quadIndexBuffer)) {
-            RenderState::indexBuffer.init(0, false);
-        }
+    if (quadIndexBuffer != 0 && !rs.isValidGeneration(quadGeneration)) {
```
tallytalwar: Ahh.. it might be too late to unset the quadIndexBuffer here, maybe it make sense to put this in the constructor of `DyanmicQuadMesh`.

tallytalwar: By "late" I mean, currently quadMesh data will be reset after the first `DyanmicQuadMesh::draw`.

tallytalwar: Ignore these, I see whats happening here.

core/src/tangram.cpp
```diff
@@ -94,6 +95,7 @@ Map::~Map() {
     // The unique_ptr to Impl will be automatically destroyed when Map is destroyed.
     TextDisplay::Instance().deinit();
     Primitives::deinit();
+    QuadIndices::reset();
```
tallytalwar: Maybe not the right place to put this.
(Maybe RenderState destructor is the right place for this reset)
Or reintroduce `meshCounter` thingy which is being done on master.

blair1618:
@tallytalwar I just overwrote your last commit, since we figured out a new solution IRL

The global quad index buffer was still using global state and when we start using multiple GL contexts that's gonna break - because the GL handle for the global index buffer generated in one context will almost certainly not be valid in another context! This was indirectly the cause of the weirdness with text and icons. Texture objects were actually fine, it's just the quad indices that were bogus.

I _think_ all the outstanding issues with this PR are resolved now.

tallytalwar:
Yup, will test this once more and good to go.

P.S. Wait for patched release to the sent for android first, and then merge this.

hjanetzek:
Looks good to go in!

PR #892, opened 2016-07-31
tizen library

tallytalwar:
First steps to bring tizen-library work into our master branch

Leaving alone `tangram-library` branch, since we might need that for pre-multiple instances work, until we finalize tizen plugin work for that.

- [x] Some cleanup required of the generic gl header. We still might have to stick around with the redefinition of gl methods to use `evas_gl_api`.


tallytalwar:
Thoughts on merging this branch to master.
Someone should give this a look and merge to master. Its getting really complicated to get tizen feature work done on tangram without creating multiple branches and keeping a track of it. Also causes confusion amongst others who are building tangram code for some tizen work.

Currently I have created an `old-rb-tizen-library` branch which reflects the one currently on gerrit top. There should not be any other "tizen-library" branch sitting around in `remote origin`.

core/src/gl.h
```diff
@@ -426,212 +424,4 @@ extern "C" {
...
-    #define glUnmapBuffer __glUnmapBufferOES
-
-    #define glFinish __glFinish
-
-    // VAO
-    #define glBindVertexArray __glBindVertexArray
-    #define glDeleteVertexArrays __glDeleteVertexArrays
-    #define glGenVertexArrays __glGenVertexArrays
-
-#endif
```
tallytalwar: This is great.

blair1618:
Ah yes, this is a much more sensible way to redefine the GL functions. +1

blair1618:
This branch has been sitting around for long enough and been tested enough through the tizen maps plugin that we should consider merging it.

The only serious concern I still have is the GL abstraction layer; it seems regrettably verbose and error-prone, especially if we expand the set of GL symbols that we use (note that we will certainly need to do this in the near future, e.g. for framebuffers).

tallytalwar:
Agreed, also as previously stated :D.

I would wait on getting the curl stuff in with this PR.

tallytalwar:
Will spend some time today to prune some of the gl headers here.

tallytalwar:
Rebased this with master. Tested this build with tizen emulator and device.
@hjanetzek can you also confirm this? Once done we can merge this.

PR #893, opened 2016-08-02
Dynamic anchors

hjanetzek:
Continuing https://github.com/tangrams/tangram-es/pull/842 

Since the changes of tile-label-placement branch introduced incremental collision testing it's possible to try fallback anchors directly instead of going through the fallbacks during multiple frames.

TODO
- [x] When the parent label has multiple anchors the child label must be updated() after the parent has been placed
- [x] Bug in alignFromParent where the parent's offset is constantly added to childs m_offset
- [x] Should offset for child labels be ignored since it conflicts the offset from the parent?

BUGS
- [x] ```icon:text:transition [show,hide] time:``` form is not parsed correctly (old bug)

### DIFF
--- a/core/src/scene/sceneLoader.cpp
+++ b/core/src/scene/sceneLoader.cpp
@@ -560,7 +560,7 @@ std::shared_ptr<Texture> SceneLoader::fetchTexture(const std::string& name, cons
                 auto ptr = (unsigned char*)(rawData.data());
                 size_t dataSize = rawData.size();
                 std::lock_guard<std::mutex> lock(m_textureMutex);
-				auto texture = scene->getTexture(name);
+                auto texture = scene->getTexture(name);
                 if (texture) {
                     if (!texture->loadImageFromMemory(ptr, dataSize)) {
                         LOGE("Invalid texture data '%s'", url.c_str());
@@ -804,7 +804,7 @@ void SceneLoader::loadStyleProps(Style& style, Node styleNode, const std::shared
             const std::string& textureName = textureNode.Scalar();
             auto& atlases = scene->spriteAtlases();
             auto atlasIt = atlases.find(textureName);
-			auto styleTexture = scene->getTexture(textureName);
+            auto styleTexture = scene->getTexture(textureName);
             if (atlasIt != atlases.end()) {
                 pointStyle->setSpriteAtlas(atlasIt->second);
             } else if (styleTexture){
@@ -1347,8 +1347,8 @@ void SceneLoader::parseStyleParams(Node params, const std::shared_ptr<Scene>& sc
         } else {
             key = prop.first.Scalar();
         }
-        if (key == "transition") {
-            parseTransition(prop.second, scene, out);
+        if (key == "transition" || key == "text:transition") {
+            parseTransition(prop.second, scene, key, out);
             continue;
         }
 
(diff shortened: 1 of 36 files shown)


core/src/style/pointStyleBuilder.cpp
```diff
@@ -278,7 +278,7 @@ void PointStyleBuilder::addFeature(const Feature& _feat, const DrawRule& _rule)
 
         size_t textStart = textLabels.size();
 
-        textStyleBuilder.addFeatureCommon(_feat, _rule, true);
+        if (!textStyleBuilder.addFeatureCommon(_feat, _rule, true)) { return; }
```
tallytalwar: :+1:

tallytalwar:
Updating fallback anchor on the same frame does improve the feeling of jumpiness/flickering labels we were seeing before. But its still noticeable. (try with bubble-wrap manhattan area)

core/src/scene/sceneLoader.cpp
```diff
@@ -560,7 +560,7 @@ std::shared_ptr<Texture> SceneLoader::fetchTexture(const std::string& name, cons
                 auto ptr = (unsigned char*)(rawData.data());
                 size_t dataSize = rawData.size();
                 std::lock_guard<std::mutex> lock(m_textureMutex);
-				auto texture = scene->getTexture(name);
+                                auto texture = scene->getTexture(name);
```
tallytalwar: nitpick, can you fix the indentation here? (also line# 807)

hjanetzek:
Did you compare side-by-side eraser-map with master? Could be that you notice it now more while focusing on the labels. It should make no difference as eraser-map doesn't use anchor fallbacks, or does it now?

core/src/scene/sceneLoader.cpp
hjanetzek: ```M-x whitespace-cleanup``` did replace tabs with spaces. @karimnaaji fix your editor! :p

hjanetzek:
Last commit should improve labels flickering on zoom-level jumps when no transition effect is set

core/src/scene/sceneLoader.cpp
nvkelso: Bubble Wrap v4.1.0 does use the proper anchor fallbacks now, but that
version is not yet used in Eraser Map the app (it’s pegged to v3.x).

hjanetzek: Nice! looking good - even though I'm more in favor of four point-placement for better icon-text association :)

@varun: of course there is now more jumping as anchor fallbacks are not coordinated between zoom-levels. definitely something to investigate in future

tallytalwar:
I am fine with merging this now. Changing default to use 4 anchor points instead of 8 makes the transitioning much more smoother.

Going to merge this, if no one has any other issue?
@karimnaaji you would wanna look at some of the last changes here?

karimnaaji:
@tallytalwar hold on this while I'm looking at the last changes.

karimnaaji:
@hjanetzek are you aware of the issue exposed in this [video](https://github.com/tangrams/tangram-es/files/418375/anchor-fallback.zip)? (look at "underground pizza" label). It seems that the fallback does not reset properly the fading transition.

Another thing I noticed; sometimes the fallback involves a fading transition, sometimes not, looking at this more closely.

hjanetzek:
@karimnaaji I think this effect came from the newly added fade_out->fade_in state change. Which was also missing to reset the fade effect.

karimnaaji:
All good to me here!

PR #894, opened 2016-08-02
fix label offset

hjanetzek:


tallytalwar:
LGTM

blair1618:
So an offset was being applied with the opposite sign in 'y'?

hjanetzek:
It was my refactoring from using angles to direction vectors which introduced the issue. I've missed to check the docs - while testing offset parameters positive y-up seemed too natural to me..

hjanetzek:
don't merge yet! now need to fix offset with line labels

blair1618:
Oh yeah, interesting. I think the scene syntax is inconsistent about y-direction in some places :(

hjanetzek:
Now it should be correct. In this case it's probably us having too many different y-axis orientations..

Issue #897, opened 2016-08-03
Expose camera rotation API in obj-c. 


msmollin:

blair1618:
Resolved by #945

Issue #850, opened 2016-07-22
Camera zoom API


msmollin:
settters and getters for camera zoom.

blair1618:
Resolved by #945

PR #898, opened 2016-08-03
Implement camera handling APIs in objective-c

### LINKED ISSUE #849: Map camera position API
setters and getters for camera position

msmollin:
Includes tilt, rotation, zoom, and position management, as well as camera type switching. Fixes #849, #850, #851, #858, #897


ios/src/TGMapViewController.mm
```diff
@@ -1,101 +1,222 @@
...
-    self.renderRequested = true;
-    self.continuous = false;
-
     if (!self.context) {
         NSLog(@"Failed to create ES context");
     }
-
+    self.pixelScale = [[UIScreen mainScreen] scale];
+    self.renderRequested = YES;
+    self.continuous = YES;
```
karimnaaji: For consistency with android we should leave continuous rendering to `NO`.

ios/src/TGMapViewController.mm
```diff
@@ -1,101 +1,222 @@
...
+            return Tangram::EaseType::quint;
+        case TangramEaseTypeCubic:
+            return Tangram::EaseType::cubic;
+        default:
+            return Tangram::EaseType::quint;
+    }
+}
+
+- (void)setZoomEased:(float)zoomLevel withDuration:(float)duration {
+    [self setZoomEased:zoomLevel withDuration:duration withEaseType:TangramEaseTypeQuint];
```
karimnaaji: Default for ease is `Cubic` on android: https://github.com/tangrams/tangram-es/blob/master/android/tangram/src/com/mapzen/tangram/MapController.java#L51

blair1618: Though it's 'quint' in the C++ interface. Maybe it should be linear everywhere or just not have a default, I'm not sure what I was thinking when I made these defaults.

karimnaaji: Either works, it's seems ok to leave defaults for ease (removing the default would involve a method deletion on the [API](https://github.com/tangrams/tangram-es/blob/master/android/tangram/src/com/mapzen/tangram/MapController.java#L238_L257)). we should just have them consistent across platforms.

msmollin: Sooooo linear or cubic? :) I agree with Karim that we should have consistency. I can change to cubic if that's the consensus.

blair1618: Let's just go with cubic for consistency with Android!

msmollin:
Closing this PR and re-opening against master.

PR #918, opened 2016-08-17
Cleanup platform dependent code in FontContext

karimnaaji:
- Remove platform dependent code from core library
- Don't load default font resources twice when loading a remote scene

### DIFF
--- a/android/tangram/src/com/mapzen/tangram/FontFileParser.java
+++ b/android/tangram/src/com/mapzen/tangram/FontFileParser.java
@@ -135,7 +135,14 @@ class FontFileParser {
                                 fallbackFontDict.put(weight, new Vector<String>());
                             }
 
-                            fallbackFontDict.get(weight).add(filename);
+                            // Don't use UI fonts
+                            if (filename.indexOf("UI-") >= 0) {
+                                continue;
+                            }
+
+                            String fullFileName = "system/fonts/" + filename;
+
+                            fallbackFontDict.get(weight).add(fullFileName);
                         } else {
                             skip(parser);
                         }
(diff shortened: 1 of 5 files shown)


tallytalwar:
Reviewing this now.

android/tangram/src/com/mapzen/tangram/FontFileParser.java
```diff
@@ -135,7 +135,14 @@ private void processDocument(XmlPullParser parser) throws XmlPullParserException
                                 fallbackFontDict.put(weight, new Vector<String>());
                             }
 
-                            fallbackFontDict.get(weight).add(filename);
+                            // Don't use UI fonts
+                            if (filename.indexOf("UI-") >= 0) {
+                                continue;
+                            }
+
+                            String fullFileName = "system/fonts/" + filename;
```
tallytalwar: We can probably add a constant for `/system/fonts/` and use that constant everywhere in this file.

karimnaaji: Yep, will update that.

tallytalwar:
Ok, this is good once one small comment is addressed.
Tizen code changes should also be simple to handle here. Will rebase tizen branch once this gets merged.

core/src/text/fontContext.cpp
```diff
@@ -14,98 +14,95 @@
...
+
+            auto addFaces = [&](const char* path) {
+                size_t dataSize;
+                char* data = reinterpret_cast<char*>(bytesFromFile(path, dataSize));
+
+                if (data) {
+                    LOG("Adding bundled font at path %s", path);
+
+                    for (int i = 0, size = BASE_SIZE; i < MAX_STEPS; i++, size += STEP_SIZE) {
+                        m_font[i]->addFace(m_alfons.addFontFace(alfons::InputSource(data, dataSize), size));
```
tallytalwar: So, with this change if default font could not be added, this line will crash. I hit this issue when getting tizen branch upto date with master.

Issue #799, opened 2016-06-17
Marker and perspective 2.5D


yuridiniz:
Hello!

I recently met this project and is really very powerful, but I'm having some difficulties with markers.

When we set the camera to `type : perspective` and added a point, we realized that this marker does not follow the perspective of the camera .

Is there any way to set this behavior?

Thanks

![screenshot_2016-06-17-18-38-31](https://cloud.githubusercontent.com/assets/5683101/16165651/8cfb74ee-34bb-11e6-831e-dedbb4781829.png)

blair1618:
Hi Yuri, thanks for the descriptive issue, diagrams are the best!

Our core team here has been thinking about this problem too. The short answer is: you can't do this yet, but we're working on it. 

The long answer is: we have a layout system for icons and text that's highly optimized for resolving collisions (overlap) between objects in *screen space*, and currently it only works for objects that are oriented to face the screen. Adjusting the geometry of an icon to lay flat against the ground is pretty simple, but we also have to adapt our collision system to deal with some icons being flat, which makes this feature bigger than it seems. But! We hope to add this as part of the revised "Marker" system described in #653 (or soon after that).

Issue #551, opened 2016-02-17
Should to add the point icon may support the direction of icon


czxczq521:
I add the point icon in map, the icon should support direction,but don't find interface to support, how to support gps location icon and it's direction.

![image](https://cloud.githubusercontent.com/assets/16661939/13099453/496a860a-d56d-11e5-942e-1bb53b3c2426.png)

blair1618:
tangram-es doesn't have a way to set the rotation of point geometries right now, but a new approach for managing simple point geometries like this is coming soon!

bcamper:
Actually, Tangram JS does support `angle` (in degrees) on a `point` style geometry, but this is apparently [undocumented](https://github.com/tangrams/tangram/blob/063e72e093771c1d8b45a9b13bd2680f7d9febcd/src/styles/points/points.js#L144) (it is primarily used for internal rotation of labels).

It looks like this may be partially but not fully implemented in ES?

It's in the [vertex layout](https://github.com/tangrams/tangram-es/blob/master/core/src/style/pointStyle.cpp#L36) and [shader](https://github.com/tangrams/tangram-es/blob/master/core/resources/shaders/point.vs#L81)

But `angle` is [not parsed](https://github.com/tangrams/tangram-es/blob/master/core/src/style/pointStyle.cpp#L133-L144) as a `draw` parameter.

@karimnaaji does that sound right or did I miss something?

We should decide either complete support in ES and add to docs, or disable the parsing in JS.

nvkelso:
+1 for supporting this across both platforms, please don't remove it! 

There is data in OSM for viewpoints that includes angle and we could immediately use this feature in our house map styles. It's also interesting to think about for things like tunnel and bridge icons which should be angled along the road.

bcamper:
FWIW angling along a road would be computed dynamically, similar to
labels... but the use cases for peaks based on tile data is interesting.

meetar:
Also for things like vector fields! ➚

blair1618:
Tangram-es *could* parse an 'angle' draw parameter and interpret it the way that tangram-js does for point style geometries, but the meaning of 'angle' becomes more complicated when the map view can change orientation, as in tangram-es. 

In some cases you may want to rotate a geometry in *screen space*, like a tilted billboard. This is the current behavior of the 'angle' draw parameter in tangram-js and the internal angle attribute in tangram-es. 

[...]

PR #925, opened 2016-08-22
orientation and 2.5d labels

karimnaaji:
2.5d projection
---------------
Fixes https://github.com/tangrams/tangram-es/issues/799

Currently labels are always facing the camera, even when the camera orientation (tilt/rotation) changes. Some use cases need to allow 2.5d markers and labels to be viewed projected onto the ground plane (see https://github.com/tangrams/tangram-es/issues/799's description for an example use case). 
This PR adds an option for point labels to be projected onto the ground with `flat` draw rule parameter (defaulting to `false`):

```yaml
layers:
    poi_icons:
         icons:
             size: 20px
             flat: true
```

screen shot 2016-08-25 at 3 12 46 pm
For comparison here is the the same view with `flat:false`
screen shot 2016-08-25 at 3 12 44 pm

Orientation
-----------
Fixes https://github.com/tangrams/tangram-es/issues/551

This PR adds an `angle` style draw rule to modify sprite orientation, specified in floating point degrees. It behaves differently whether the point label is flattened or not: 
- If the label is flattened, the angle will be a negative angle (clockwise order) relative to the world from the north cardinal direction. 
- If the label is not flattened, the angle will be a negative angle relative to the screen from a vertical axis.

**unflattened orientation**
```yaml
layers:
    poi_icons:
         icons:
             size: 20px
             flat: false
             angle: 45
```
screen shot 2016-08-25 at 3 14 59 pm
**flattened orientation**
```yaml
layers:
    poi_icons:
         icons:
             size: 20px
             flat: true
             angle: 45
```
screen shot 2016-08-25 at 3 15 01 pm

cc @bcamper @nvkelso for syntax.

**TODO**
- [x] Update skip transition for `flat` labels
- [x] Update label comparator code
- [x] Update feature selection code 

**FIXME**
- [x] Get scale factor for label markers


karimnaaji:
@bcamper can you give a look at the style syntax here?

bcamper:
Looks good - and as you pointed out, it seems JS regressed and the `angle` there is now being interpreted in radians :/  I will fix that and slate for JS v0.10 release. Thanks!

core/src/tile/tileBuilder.cpp
```diff
@@ -74,7 +73,11 @@ std::shared_ptr<Tile> TileBuilder::build(TileID _tileID, const TileData& _tileDa
         builder.second->addLayoutItems(m_labelLayout);
     }
 
-    m_labelLayout.process();
+    ViewState viewState = m_scene->viewState();
+    viewState.zoom = _tileID.z;
```
hjanetzek: Label placement during tile building should not depend on the current view.

karimnaaji: It still doesn't, the view state is always set to be based on the tile, thus explicit set of zoom and fractional zoom.

hjanetzek: Label::update uses the ViewState's viewPortSize though.

hjanetzek:
@karimnaaji should we merge master into this branch or go through rebase?

karimnaaji:
If you already reviewed that I can start rebasing or merge master into this depending on the conflicts.

core/src/labels/label.h
```diff
@@ -72,71 +84,94 @@ class Label {
...
 
-    Label(Transform _transform, glm::vec2 _size, Type _type, Options _options);
+    Label(WorldTransform _transform, glm::vec2 _size, Type _type, Options _options);
 
     virtual ~Label();
 
-    bool update(const glm::mat4& _mvp, const glm::vec2& _screenSize, float _zoomFract, bool _drawAllLabels = false);
+    // Push the pending transforms to the vbo by updating the vertices
+    virtual void updateVertexBuffer() = 0;
+    virtual glm::vec2 center() const;
```
hjanetzek: Would this be more clear?
`// Add vertices for this label to its Style's shared Mesh`
`pushVertices()`

core/src/labels/label.cpp
```diff
@@ -30,38 +32,80 @@ Label::Label(Label::Transform _transform, glm::vec2 _size, Type _type, Options _
...
-            screenPosition = worldToScreenSpace(_mvp, glm::vec4(p0, 0.0, 1.0),
-                                                _screenSize, clipped);
-            if (clipped) {
-                return false;
-            }
+            if (m_options.flat) {
+                glm::vec2 positions[4];
+                float unitsPerMeter = _viewState.metersPerPixel * _tileInverseScale;
+                glm::vec2 metersDimension = m_dim * unitsPerMeter;
+
```
hjanetzek: Why calling this metersDimension? I think you calculate `unitsPerPixel` here. Keeping this all pixel based would simplify things (for me).

karimnaaji: Yes this is badly named, it's a unit per pixel dimension. Will update.

core/src/labels/label.h
karimnaaji: `push` could be misleading, meaning that we change the size of it. Either `updateVertexBuffer` or `updateVertices` makes sense to me. I can update the comment.

hjanetzek: Yes, `pushVertices` omits to where the vertices are pushed. `add/pushVerticesToMesh` would be more complete. I'm ok with `updateVertices` even though they are not really updated (as in: some place stores a previous state that becomes updated by this call) :)

karimnaaji: Ok this seems accurate with what happens on the dynamic quad mesh, will update for that.

karimnaaji:
Thanks for addressing the fixme @hjanetzek , this is good. @tallytalwar is finishing reviewing and this should be good to go.

core/src/labels/label.cpp
```diff
@@ -30,38 +32,80 @@ Label::Label(Label::Transform _transform, glm::vec2 _size, Type _type, Options _
 
 Label::~Label() {}
 
-bool Label::updateScreenTransform(const glm::mat4& _mvp, const glm::vec2& _screenSize, bool _drawAllLabels) {
-
-    glm::vec2 screenPosition;
+bool Label::updateScreenTransform(const glm::mat4& _mvp, const ViewState& _viewState, bool _drawAllLabels) {
```
karimnaaji: @hjanetzek One last issue I just found; when used as markers, the `mvp` matrix misses a scale component since the point marker has no extent, so the projection to screen goes wrong.

hjanetzek:
shot-2016-09-09_13-44-42

Bad texture sampling - Flat markers must be drawn with perspective projection..

karimnaaji:
This shouldn't have anything to do with the kind of projection.
My guess is that this distortion happens because the screen coverage of the quad triangles is too important for the precision we give to the uvs; the interpolation in the shader might not have enough precision on the floating point uvs attributes.
afca0c44-7695-11e6-81c7-478537e17d1e
If we want to be able to draw big markers like here, I would try increasing the precision for uvs here https://github.com/tangrams/tangram-es/blob/master/core/src/labels/spriteLabel.h#L13.
I'm quite sure the same would happen for any matrix projection.

hjanetzek:
@karimnaaji try it :) the mapzen logo looks like a perfectly fine linear interpolated texture for both triangles (using mediump should be sufficient). I'm pretty sure the texture lookup is missing the appropriate glPosition.zw for the perspective correction.

I just found that there is a `texture2DProj` which might help here. http://stackoverflow.com/questions/29847342/how-to-correctly-map-texture-when-doing-perspective-warping-in-glsl-using-opengl

blair1618:
Yeah I think this is a consequence of projecting vertices into perspective without perspective-correct interpolation (yet another reason to render labels in 3d scene space).

karimnaaji:
@hjanetzek , you're right, I understand the problem, so we should send the perspective factor we get here https://github.com/tangrams/tangram-es/blob/master/core/src/util/geom.cpp#L96, and use `texture2DProj`.

hjanetzek:
shot-2016-09-13_14-56-33

hjanetzek:
@karimnaaji I tried to use texture2DProj but couldn't get it to work with the ortho projection. I think sending the final projected vertices is the currently the best option. Later we may switch to separate meshes and shaders for flat and billboard labels. 
@tallytalwar could you do the final review?

karimnaaji:
@hjanetzek yes I also think it's the best option we have right now so that makes sense, what is happening in the screenshot above? 
Thanks for addressing what we spoke about concerning the scaling issue!

hjanetzek:
@karimnaaji technically the screenshot shows POIs rendered with flat option at size 200 in overlay blend mode. For non-overlay mode we also need to add the layer offset I guess.. 

Though I agree, it looks more like 'The attack of the POI' :)

karimnaaji:
Haha yes :)

Oh ok I thought this was the testing using `texture2DProj`.

karimnaaji:
There seem to be no blocker in merging this if reviewed already. The issue being pointed out is existing but can be considered as an edge case, it shows the limit of our method to draw the labels in screen space, but for reasonably sized sprite textures the issue is in general not visible.

hjanetzek:
OK, I'm going to revert scene.yaml changes and merge it this evening if no one objects.

tallytalwar:
If you both are fine with this, please go ahead with merging this. I will not be able to finish the review today. Will catch-up later on.

PR #926, opened 2016-08-23
Initializing m_resourceLoad seams to fix the problem w Tangram:update…

patriciogonzalezvivo:
Hi! This seams to be the `ushort` responsable of the inconsistency on on `tangram::update()`

Thanks


tallytalwar:
So, looks like this got fixed with https://github.com/tangrams/tangram-es/commit/883885ead3a8e7795e764ef858b724b9080ed70f.
@patriciogonzalezvivo was using an older version of tangram, which had this issue. Going to close this PR.

Thanks

PR #928, opened 2016-08-24
Unified `collide` style param

karimnaaji:
Whether unified child label collide should be defined by the parent collide param.


blair1618:
Well this looks tame to me. Can you help me understand the problem this resolves?

karimnaaji:
Right now if a unified `point` style sets a `collide` parameter the `text` associated with it doesn't inherits from it if absent under `text`.

This fix makes the rule inherit from the parent if the child doesn't explicitly sets it.

blair1618:
LGTM

PR #929, opened 2016-08-24
Fix segfault when bundled font are not found

### LINKED ISSUE #927: FontContext segfaults if no fonts can be found
screen shot 2016-08-23 at 3 33 12 pm

karimnaaji:
Fix #927

### DIFF
--- a/core/src/text/fontContext.cpp
+++ b/core/src/text/fontContext.cpp
@@ -46,13 +46,13 @@ void FontContext::loadFonts() {
             size_t dataSize;
             char* data = reinterpret_cast<char*>(bytesFromFile(DEFAULT, dataSize));
 
-            if (data) {
                 LOG("Loading default font file %s", DEFAULT);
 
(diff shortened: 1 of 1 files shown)


tallytalwar:
@karimnaaji I modified this. Can you give this a look?
Thanks

core/src/text/fontContext.cpp
```diff
@@ -46,13 +46,13 @@ void FontContext::loadFonts() {
             size_t dataSize;
             char* data = reinterpret_cast<char*>(bytesFromFile(DEFAULT, dataSize));
 
-            if (data) {
                 LOG("Loading default font file %s", DEFAULT);
```
karimnaaji: @tallytalwar indentation is off here now

karimnaaji:
@tallytalwar yeah like we spoke about this morning this is good.

core/src/text/fontContext.cpp
tallytalwar: Updated.

PR #930, opened 2016-08-24
'Marker' interface to create and manipulate individual geometries

blair1618:
This adds the concept of 'Markers' to the tangram-es C++ interface. A Marker represents a single feature on the map that can be programmatically added, removed, and manipulated. Markers can be either point, polyline, or polygon geometries. These geometries can be styled using the same options available in the scene file syntax. Users will interact with Markers through new functions in the `Map` class (defined in `tangram.h`): https://github.com/tangrams/tangram-es/blob/marker/core/src/tangram.h#L153-L190

TODO:

 - [x] Compile JS functions defined in marker draw rules
 - [x] ~~Allow user-defined data to be associated with markers~~ (not needed now)
 - [x] ~~Define an ordering among marker labels~~ (`blend_order` is sufficient for current needs)
 - [x] Document Marker and MarkerManager classes
 - [x] Add `markerRemoveAll()` to allow removal of markers whose handles are lost
 - [x] Add `markerSetVisible(Marker* marker, bool visible)` to conveniently toggle marker visibility.

Design questions:

 - Should users get and retain a `Marker*` (currently the case) or an integer handle? An integer handle would remove all concern of holding a pointer to a freed object, but we'd have think about whether integers are recycled and whether that would be confusing. *Resolved: We now use non-decreasing integer handles.*

 - Should location easing be enabled for polyline and polygon Markers? Currently I've only enabled location easing for point Markers, since it's unambiguous what it means to move a point geometry to a destination coordinate. For polylines and polygons, this would currently move the 'origin' of the geometry (the SW corner of the bounding box) to the destination coordinate, which might be more confusing than useful. *Resolved: We only allow easing for point markers, and setting a point ease for a marker with no point geometry will create a point geometry*.

Follow-up work:

 - Add a convenience method for adding custom bitmap data to a Marker
 - Add Android interface for Markers
 - Add iOS interface for Markers

### DIFF
--- a/core/src/marker/markerManager.h
+++ b/core/src/marker/markerManager.h
@@ -2,6 +2,7 @@
 
 #include "scene/styleContext.h"
 #include "scene/drawRule.h"
+#include "util/ease.h"
 #include "util/fastmap.h"
 #include "util/types.h"
 #include <memory>
@@ -27,6 +28,8 @@ public:
     bool setStyling(Marker* marker, const char* styling);
 
     bool setPoint(Marker* marker, LngLat lngLat);
+
+    bool setPointEased(Marker* marker, LngLat lngLat, float duration, EaseType ease);
 
     bool setPolyline(Marker* marker, LngLat* coordinates, int count);
 
--- a/core/src/marker/markerManager.h
+++ b/core/src/marker/markerManager.h
@@ -3,6 +3,7 @@
 #include "scene/styleContext.h"
 #include "scene/drawRule.h"
 #include "util/fastmap.h"
+#include "util/types.h"
 #include <memory>
 #include <vector>
 
@@ -25,11 +26,11 @@ public:
 
     bool setStyling(Marker* marker, const char* styling);
 
-    bool setPoint(Marker* marker, double lng, double lat);
+    bool setPoint(Marker* marker, LngLat lngLat);
 
-    bool setPolyline(Marker* marker, double* coordinates, int count);
+    bool setPolyline(Marker* marker, LngLat* coordinates, int count);
 
-    bool setPolygon(Marker* marker, double** coordinates, int* counts, int rings);
+    bool setPolygon(Marker* marker, LngLat* coordinates, int* counts, int rings);
 
     const std::vector<std::unique_ptr<Marker>>& markers() const;
 
--- a/core/src/marker/markerManager.h
+++ b/core/src/marker/markerManager.h
@@ -1,0 +1,51 @@
+#pragma once
+
+#include "scene/styleContext.h"
+#include "scene/drawRule.h"
+#include "util/fastmap.h"
+#include <memory>
+#include <vector>
+
+namespace Tangram {
+
+class MapProjection;
+class Marker;
+class Scene;
+class StyleBuilder;
+
+class MarkerManager {
+
+public:
+
+    void setScene(std::shared_ptr<Scene> scene);
+
+    Marker* add(const char* styling);
+
+    bool remove(Marker* marker);
+
+    bool setStyling(Marker* marker, const char* styling);
+
+    bool setPoint(Marker* marker, double lng, double lat);
+
+    bool setPolyline(Marker* marker, double* coordinates, int count);
+
+    bool setPolygon(Marker* marker, double** coordinates, int* counts, int rings);
+
+    const std::vector<std::unique_ptr<Marker>>& markers() const;
+
+private:
+
+    bool contains(Marker* marker);
+
+    void build(Marker& _marker);
+
+    DrawRuleMergeSet m_ruleSet;
+    StyleContext m_styleContext;
+    std::shared_ptr<Scene> m_scene;
+    std::vector<std::unique_ptr<Marker>> m_markers;
+    fastmap<std::string, std::unique_ptr<StyleBuilder>> m_styleBuilders;
+    MapProjection* m_mapProjection = nullptr;
+
+};
+
+} // namespace Tangram
--- a/core/src/marker/markerManager.h
+++ b/core/src/marker/markerManager.h
@@ -35,13 +35,15 @@ public:
 
     bool setPolygon(Marker* marker, LngLat* coordinates, int* counts, int rings);
 
+    bool update(int zoom);
+
     const std::vector<std::unique_ptr<Marker>>& markers() const;
 
 private:
 
     bool contains(Marker* marker);
 
-    void build(Marker& _marker);
+    void build(Marker& _marker, int zoom);
 
     DrawRuleMergeSet m_ruleSet;
     StyleContext m_styleContext;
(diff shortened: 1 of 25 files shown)


bcamper:
If we go the route of an integer handle to a marker, I would vote for not recycling them (there are plenty to spare!).

yuridiniz:
This is amazing, congratulations!

yuridiniz:
This allows you to set rotation and flat orientation?

blair1618:
@yuridiniz Not yet, but that's being done here: https://github.com/tangrams/tangram-es/pull/925

blair1618:
screen shot 2016-08-26 at 8 49 20 pm

Ok, so adding arbitrary geometry at run-time with full styling control is kinda neat.

tallytalwar:
This is nice,  did you trace that in?

Tangram ms-paint!!

core/src/tangram.cpp
```diff
@@ -587,6 +603,34 @@ void Map::clearDataSource(DataSource& _source, bool _data, bool _tiles) {
...
+
+bool Map::markerSetPoint(Marker* _marker, LngLat _lngLat) {
+    return impl->markerManager.setPoint(_marker, _lngLat);
+}
+
+bool Map::markerSetPointEased(Marker* _marker, LngLat _lngLat, float _duration, EaseType ease) {
+    return impl->markerManager.setPointEased(_marker, _lngLat, _duration, ease);
+}
+
+bool Map::markerSetPolyline(Marker* _marker, LngLat* _coordinates, int _count) {
```
tallytalwar: Should these `markerSetGeom` methods be doing a `requestRender` in the core, or should the client application be responsible to do that `requestRender` after setting geometry for a marker.

Currently I am doing a `requestRender` from tizen plugin code, on update of a marker.

blair1618: Hm that's a good point. A client application should not need to determine when re-rendering is needed, so I'll add that logic to the core library.

core/src/tangram.h
```diff
@@ -148,6 +150,49 @@ class Map {
...
+
+    // Set the geometry of a marker to a point at the given coordinates; markers can have their
+    // geometry set multiple times with possibly different geometry types; returns true if the
+    // marker handle was found and successfully updated, otherwise returns false.
+    bool markerSetPoint(Marker* _marker, LngLat _lngLat);
+
+    // Set the position of a marker with point geometry to ease to a new location over a duration
+    // in seconds with the given EaseType; returns true if the marker handle was found and has a
+    // point geometry, otherwise returns false.
+    bool markerSetPointEased(Marker* _marker, LngLat _lngLat, float _duration, EaseType _ease);
```
tallytalwar: This function name `markerSetPointEased` is a bit confusing. According to the name it should set the Point and if already set by a previous call, update it using the specified ease.
Already discussed with @blair1618 IRL. We should probably name this to `markerSetEased` and implement easing for all types of marker geometries.

blair1618: It occurs to me now that having an easing position setter for all marker geometries would be kinda weird because there is no comparable non-easing position setter. Instead I'm now leaning towards implementing this function the way that you originally thought it worked, i.e. acting exactly like `setPoint` if the marker does not already have a point set.

tallytalwar:
I have been reviewing this side by side, mostly good to go, will give a look the news commits and if all good will merge. Thanks

blair1618:
Consider the interface here to be frozen, but something bad happens to markers when scene files are changed so I still have to look into that.

hjanetzek:
Looks good overall!
Doesn't all maker custom styling need to be reset when a new scene is set? The markers DrawRuleData may reference parts of the previous scene.

tallytalwar:
Yup you are correct Hannes. We were taking about the same thing yesterday
evening.

core/src/style/textStyleBuilder.cpp
```diff
@@ -43,11 +43,15 @@ void TextStyleBuilder::setup(const Tile& _tile){
...
+    float metersPerTile = 2.f * MapProjection::HALF_CIRCUMFERENCE * exp2(-zoom);
+
+    // In general, a Marker won't cover the same area as a tile, so the effective
+    // "tile size" for building a Marker is the size of a tile in pixels multiplied
+    // by the ratio of the Marker's extent to the length of a tile side at this zoom.
     m_tileSize = 256 * (marker.extent() / metersPerTile);
-    m_tileSize *= m_style.pixelScale();
 
-    // add scale factor to the next zoom-level
+    // (Copied from Tile setup function above, purpose unclear)
```
tallytalwar: I think this is to get the device pixel scale into account when calculating a tile size.

blair1618:
This branch should now be totally functional and ready to merge!

tallytalwar:
One more thing, can we expose a `visbility` function with the marker interface?
`markerSetVisible(MarkerID, bool)`?

Which could internally add a `visible: false` or `visible: true` draw rule to the set styling.

tallytalwar:
Wait its already there: https://github.com/tangrams/tangram-es/pull/930/commits/756dbc2cf3f4458fffb0b480f9535924fa5ec02d. Ignore previous comment.

core/src/marker/markerManager.h
```diff
@@ -0,0 +1,80 @@
...
+
+    DrawRuleMergeSet m_ruleSet;
+    StyleContext m_styleContext;
+    std::shared_ptr<Scene> m_scene;
+    std::vector<std::unique_ptr<Marker>> m_markers;
+    std::vector<std::string> m_jsFnList;
+    fastmap<std::string, std::unique_ptr<StyleBuilder>> m_styleBuilders;
+    MapProjection* m_mapProjection = nullptr;
+    size_t m_jsFnIndex = 0;
+    uint32_t m_idCounter = 0;
```
tallytalwar: Hmm, maybe it makes sense to make this atomic?
It is possible a client application performs a addMarker call from 2 async threads.

blair1618: I've kind of punted on synchronization for markers at present. If we care about thread safety here then there's a lot we'll need to guard - so let's make that a follow-up task.

core/src/marker/markerManager.h
```diff
@@ -0,0 +1,80 @@
...
+    bool update(int zoom);
+
+    // Remove and destroy all markers.
+    void removeAll();
+
+    const std::vector<std::unique_ptr<Marker>>& markers() const;
+
+private:
+
+    Marker* getMarkerOrNull(MarkerID markerID);
```
tallytalwar: `getMarker` should have sufficed. But its fine :D.

blair1618: Glad you commented on this because it's actually intentional and I think it's a very good practice to write explicitly descriptive names like this! It makes the operation trivial to understand when reading it in-line.

core/src/marker/markerManager.h
tallytalwar: Sure, we might hit this sooner than later. Specially when we run it through our biggest test case (Easer Map).

tallytalwar:
Will test it once more on tizen and merge.

PR #942, opened 2016-08-30
fetch submodules explicitly

hjanetzek:


blair1618:
I think I understand the motivations here, but just to be clear, this is trying to:

1. Prevent checking out code that isn't needed for the current build target
2. Prevent developers forgetting to initialize submodules before building

Point 1. hasn't been a major issue so far, but will be more annoying if we add all of our Windows dependencies as submodules. Point 2. seems to be only an occasional problem that's easily corrected, but it does consume some time to remind developers about this step. 

These are real practical problems, however I have conceptual concerns with this approach. I think many developers (myself included) think of a build system as having a distinct role from a source control system. Conflating these two may cause more confusion than it's worth. 

We can take other steps to reduce these problems without changing the nature of our build process, e.g. shallow clones of submodules can reduce the size of downloads by orders of magnitude.

hjanetzek:
For 2. we have a warning that hints to use ```git submodule update --init --recursive```. 
This could also say 'Please run ./get_dependencies.sh [linux|android|....]'. The problem with the global recursive is that one fetches also submodules that we don't need at all, like ```core/include/earcut.hpp/glfw```. That's also a reason why I would prefer to explicitly checkout submodules.

The get_dependencies script could also take care of fetching shallow clones when one just wants to build the tangram. For development I find it comfortable to be able to work on the submodules directly.

PR #944, opened 2016-08-30
min_display_zoom & max_display_zoom

hallahan:
This PR allows you to set a `min_zoom` parameter in your sources block of your scene.yaml.

When a source is given a `min_zoom` parameter, it does not fetch tiles below that min zoom. This includes the raster sub-sources we see with normals, as found in Walkabout and other terrain styles.

As @bcamper mentioned, we might actually want to call this parameter `min_display_zoom`, since we don't under-zoom data at lower zooms--we simply don't provide it.

I'm happy to re-factor that into this PR. Also, we could take a similar approach found here to implement `max_display_zoom`. 

https://github.com/tangrams/tangram/pull/394

Thinking about also implementing `bounds`, we could build that into [`DataSource#isActiveForZoom`](https://github.com/hallahan/tangram-es/blob/f7f2dbfbf60da54a3d085fac2e346e91269f9a2f/core/src/data/dataSource.h#L75-L77), calling it `isActive` instead. 

cc/ @tallytalwar @blair1618 @bcamper

### DIFF
--- a/core/src/data/clientGeoJsonSource.h
+++ b/core/src/data/clientGeoJsonSource.h
@@ -23,7 +23,7 @@ class ClientGeoJsonSource : public DataSource {
 
 public:
 
-    ClientGeoJsonSource(const std::string& _name, const std::string& _url, int32_t _maxZoom = 18);
+    ClientGeoJsonSource(const std::string& _name, const std::string& _url, int32_t _minZoom = 0, int32_t _maxZoom = 18);
     ~ClientGeoJsonSource();
 
     // Add geometry from a GeoJSON string
--- a/core/src/data/clientGeoJsonSource.h
+++ b/core/src/data/clientGeoJsonSource.h
@@ -23,7 +23,8 @@ class ClientGeoJsonSource : public DataSource {
 
 public:
 
-    ClientGeoJsonSource(const std::string& _name, const std::string& _url, int32_t _minDisplayZoom = 0, int32_t _maxZoom = 18);
+    ClientGeoJsonSource(const std::string& _name, const std::string& _url,
+                        int32_t _minDisplayZoom = 0, int32_t _maxDisplayZoom = INT32_MAX, int32_t _maxZoom = 18);
     ~ClientGeoJsonSource();
 
     // Add geometry from a GeoJSON string
--- a/core/src/data/clientGeoJsonSource.h
+++ b/core/src/data/clientGeoJsonSource.h
@@ -23,7 +23,7 @@ class ClientGeoJsonSource : public DataSource {
 
 public:
 
-    ClientGeoJsonSource(const std::string& _name, const std::string& _url, int32_t _minZoom = 0, int32_t _maxZoom = 18);
+    ClientGeoJsonSource(const std::string& _name, const std::string& _url, int32_t _minDisplayZoom = 0, int32_t _maxZoom = 18);
     ~ClientGeoJsonSource();
 
     // Add geometry from a GeoJSON string
--- a/core/src/data/clientGeoJsonSource.h
+++ b/core/src/data/clientGeoJsonSource.h
@@ -24,7 +24,7 @@ class ClientGeoJsonSource : public DataSource {
 public:
 
     ClientGeoJsonSource(const std::string& _name, const std::string& _url,
-                        int32_t _minDisplayZoom = 0, int32_t _maxDisplayZoom = INT32_MAX, int32_t _maxZoom = 18);
+                        int32_t _minDisplayZoom = -1, int32_t _maxDisplayZoom = -1, int32_t _maxZoom = 18);
     ~ClientGeoJsonSource();
 
     // Add geometry from a GeoJSON string
--- a/core/src/data/dataSource.cpp
+++ b/core/src/data/dataSource.cpp
@@ -88,8 +88,8 @@ struct RawCache {
     }
 };
 
-DataSource::DataSource(const std::string& _name, const std::string& _urlTemplate, int32_t _minZoom, int32_t _maxZoom) :
-    m_name(_name), m_minZoom(_minZoom), m_maxZoom(_maxZoom), m_urlTemplate(_urlTemplate),
+DataSource::DataSource(const std::string& _name, const std::string& _urlTemplate, int32_t _minDisplayZoom, int32_t _maxZoom) :
+    m_name(_name), m_urlTemplate(_urlTemplate), m_minDisplayZoom(_minDisplayZoom), m_maxZoom(_maxZoom),
     m_cache(std::make_unique<RawCache>()){
 
     static std::atomic<int32_t> s_serial;
@@ -143,7 +143,7 @@ void DataSource::constructURL(const TileID& _tileCoord, std::string& _url) const
 bool DataSource::equals(const DataSource& other) const {
     if (m_name != other.m_name) { return false; }
     if (m_urlTemplate != other.m_urlTemplate) { return false; }
-    if (m_minZoom != other.m_minZoom) { return false; }
+    if (m_minDisplayZoom != other.m_minDisplayZoom) { return false; }
     if (m_maxZoom != other.m_maxZoom) { return false; }
     if (m_rasterSources.size() != other.m_rasterSources.size()) { return false; }
     for (size_t i = 0, end = m_rasterSources.size(); i < end; ++i) {
--- a/core/src/data/dataSource.cpp
+++ b/core/src/data/dataSource.cpp
@@ -88,8 +88,8 @@ struct RawCache {
     }
 };
 
-DataSource::DataSource(const std::string& _name, const std::string& _urlTemplate, int32_t _maxZoom) :
-    m_name(_name), m_maxZoom(_maxZoom), m_urlTemplate(_urlTemplate),
+DataSource::DataSource(const std::string& _name, const std::string& _urlTemplate, int32_t _minZoom, int32_t _maxZoom) :
+    m_name(_name), m_minZoom(_minZoom), m_maxZoom(_maxZoom), m_urlTemplate(_urlTemplate),
     m_cache(std::make_unique<RawCache>()){
 
     static std::atomic<int32_t> s_serial;
--- a/core/src/data/dataSource.cpp
+++ b/core/src/data/dataSource.cpp
@@ -88,8 +88,10 @@ struct RawCache {
     }
 };
 
-DataSource::DataSource(const std::string& _name, const std::string& _urlTemplate, int32_t _minDisplayZoom, int32_t _maxZoom) :
-    m_name(_name), m_urlTemplate(_urlTemplate), m_minDisplayZoom(_minDisplayZoom), m_maxZoom(_maxZoom),
+DataSource::DataSource(const std::string& _name, const std::string& _urlTemplate,
+                       int32_t _minDisplayZoom, int32_t _maxDisplayZoom, int32_t _maxZoom) :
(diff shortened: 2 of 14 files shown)


hallahan:
At this point, I've refactored `min_zoom` to be `min_display_zoom`. I have also implemented `max_display_zoom`.

hallahan:
Interesting, Travis caught that Android was missing the `#define` for `INT32_MAX`. Perhaps I should just `#ifndef` in the value to the token in the header?

tallytalwar:
Can you rebase this with master? some of the commits you have above have already been merged!

core/src/data/clientGeoJsonSource.h
```diff
@@ -23,7 +23,8 @@ class ClientGeoJsonSource : public DataSource {
 
 public:
 
-    ClientGeoJsonSource(const std::string& _name, const std::string& _url, int32_t _maxZoom = 18);
+    ClientGeoJsonSource(const std::string& _name, const std::string& _url,
+                        int32_t _minDisplayZoom = 0, int32_t _maxDisplayZoom = INT32_MAX, int32_t _maxZoom = 18);
```
tallytalwar: I think its fine to use `18` as the `maxDisplayZoom` default instead of `INT32_MAX`.

core/src/data/dataSource.cpp
```diff
@@ -143,6 +145,7 @@ void DataSource::constructURL(const TileID& _tileCoord, std::string& _url) const
 bool DataSource::equals(const DataSource& other) const {
     if (m_name != other.m_name) { return false; }
     if (m_urlTemplate != other.m_urlTemplate) { return false; }
+    if (m_minDisplayZoom != other.m_minDisplayZoom) { return false; }
```
tallytalwar: Missing check for `m_maxDisplayZoom` for `DataSource` equality.

core/src/data/dataSource.h
```diff
@@ -69,8 +74,13 @@ class DataSource : public std::enable_shared_from_this<DataSource> {
     /* Generation ID of DataSource state (incremented for each update, e.g. on clearData()) */
     int64_t generation() const { return m_generation; }
 
+    int32_t minZoom() const { return m_minDisplayZoom; }
```
tallytalwar: Probably better to rename the accessor to `minDisplayZoom` to avoid any confusion.

core/src/data/dataSource.h
```diff
@@ -105,6 +115,12 @@ class DataSource : public std::enable_shared_from_this<DataSource> {
     // Name used to identify this source in the style sheet
     std::string m_name;
 
+    // Minimum zoom for which tiles will be displayed
+    int32_t m_minDisplayZoom;
```
tallytalwar: nitpick, but why not int16_t?

core/src/data/clientGeoJsonSource.h
tallytalwar: Or a default value of `-1` can be used to ignore any culling of tiles based on min or max display zooms.

hallahan: Doesn't the map actually proxy tile higher zooms than 18? What if we wanted to zoom to 28 and explore a map of the stores in a mall or something?

core/src/tile/tileManager.cpp
```diff
@@ -384,11 +388,24 @@ void TileManager::loadSubTasks(std::vector<std::shared_ptr<DataSource>>& _subSou
...
         auto subTask = subSource->createTask(subTileID, index);
-        if (subTask->isReady()) {
+        // check if we are at valid zoom for source
+        if (!subSource->isActiveForZoom((float)subTileID.z)) {
+            // Right now all subSources are raster, but let's check anyway...
+            if (RasterSource* rasterSubSource = dynamic_cast<RasterSource*>(subSource.get())) {
+                subTasks.insert(it, subTask);
+                // If the subSource isn't active for the zoom, we should just
+                // load an empty texture and move on.
+                rasterSubSource->loadEmptyTexture(std::move(subTask));
```
tallytalwar: This might result in some weird artifacts (black tiles), if the style sheet is using some wrong combination of min/max display zooms. Does it not make sense to include a `subSource`'s min/max display zoom limits also in its main source? This will result in main source's tile not building if the sub/child source tile is not "available" for that display zoom.

cc. @bcamper

bcamper: Yes, after discussing this I amended the logic to limit the "parent" source by any attached raster source's min/max. So only tiles in the intersection of min/max zoom range (and/or bounding box) would be built.

core/src/data/dataSource.cpp
hallahan: Done via https://github.com/tangrams/tangram-es/pull/944/commits/108ab3c46194b8c429ae504f6905db9dd1186ed0

core/src/data/dataSource.h
hallahan: I can switch `m_minDisplayZoom` to be `int16_t`, however, that would be inconsistent with m_maxZoom which has always been `int32_t`. We really could have all these zoom vars be `int8_t` (max value of 127)

core/src/tile/tileManager.cpp
hallahan: When I was implementing this, I was trying things out with Walkabout. My thought was that our normals may not cover a given layer. For example, what if we had normals of zooms 8 - 12, but we wanted to render vectors with 0 - 18? It looks like it works fine as-is--the hillshades just don't get drawn when the sub source doesn't load.

core/src/data/clientGeoJsonSource.h
tallytalwar: Yup in that case what I suggested to use is `-1`. Similar to how js does it and has a `null` as default.

core/src/tile/tileManager.cpp
bcamper: Seems like what you may want to do in that case is attach a raster source with a lower **`max_zoom`** value, and the raster texture will be scaled up at the higher zooms, whereas the `max_display_zoom` parameter was meant to really signify you *don't* want this tile source rendering at a higher zoom level. In our current styles like Walkabout, we set the terrain normals `max_zoom: 15`, and let them be scaled up onto vectors up to z20.

I don't know what is happening with the current behavior (@tallytalwar can illuminate), but it may be difficult to ensure correct drawing in all cases where a raster is attached. We have to provide *some* texture to the GL shader, and with placeholder like a 1x1 pixel image, I would not expect the lighting normal calculations to be correct. You'd want different default textures/behavior for different types of raster attachments (e.g. treating the raster as a direct color texture, vs. treating as a normal map), and there is also the "custom" raster mode where you don't know the intentions of the user.

Am I missing something in the example above, where only the `max_zoom` needs to be set?

core/src/data/dataSource.h
tallytalwar: Ohh I forgot about these. You can leave it 32 bit right now. Though I think all these can be 8 bit but I might have to look as why we had it as 32 in the beginning.

core/src/tile/tileManager.cpp
tallytalwar: The reason it won't render the uber time is because it is not "ready" yet. A main tile only becomes ready to be drawn is when all its dependent source tiles are loaded. 

Along with what Brett mentioned,  there is no point processing a master tile if it's subtIle will never be fetched,  we can ignore all such tiles and save on some processing.

core/src/data/dataSource.h
hallahan: 👍

hallahan:
@tallytalwar - I implemented the proposed changes. Ready for another look. Thanks!

ios/src/TGMapViewController.h
```diff
@@ -12,6 +12,26 @@
 #import <UIKit/UIKit.h>
```
tallytalwar: How come these changes are in this PR?

core/src/data/rasterSource.cpp
```diff
@@ -172,6 +173,11 @@ bool RasterSource::loadTileData(std::shared_ptr<TileTask>&& _task, TileTaskCb _c
     return status;
 }
 
+void RasterSource::loadEmptyTexture(std::shared_ptr<TileTask>&& _task) {
```
tallytalwar: This is not required now.

core/src/tile/tileManager.cpp
```diff
@@ -9,6 +9,7 @@
 #include "glm/gtx/norm.hpp"
 
 #include <algorithm>
+#include <data/rasterSource.h>
```
tallytalwar: We can remove this include now, since `loadEmptyTexture` is not required now.

tallytalwar:
Changes look good, once the comments are addressed and propertly rebased with master will merge.

Thanks @hallahan

ios/src/TGMapViewController.h
hallahan: From my understanding, that's because I did a rebase from tangrams/master to hallahan/minZoom. Git gets kinda weird when dealing with another origin sometimes. It's because I did commits at a point before this iOS work and then rebased yesterday. The nature of rebase is that history gets rewritten, so that commit, as you can see, has a different SHA1. The older problem where we saw old commits that have already been merged is because I branched off of hallahan/master instead of tangrams-es/master. It was up to date, but the fact that one origin merged into another created that effect.

The only way I could resolve this is to make a new branch and cherry pick my commits into a new one and open a new PR. However, I wonder, would it take me out of the commit message once you merge the PR? Since, all of my commits should say it was hallahan with tallytalwar? I can do a new PR if needed. Sorry this is such a headache!

http://stackoverflow.com/questions/32826197/interactive-rebase-after-merging-other-commits-interleaving-mine

core/src/tile/tileManager.cpp
hallahan: Done via https://github.com/tangrams/tangram-es/pull/944/commits/2c778a2bd8ff14a920dd316c43f2f5b2fd495ab5

core/src/data/rasterSource.cpp
hallahan: Done via https://github.com/tangrams/tangram-es/pull/944/commits/2c778a2bd8ff14a920dd316c43f2f5b2fd495ab5

ios/src/TGMapViewController.h
tallytalwar: Hmm.. Probably one more thing you can do here is to take this (c159079af9f57ff0e495a269a4cf850015693aca) commit out.

* `git rebase -i HEAD~12`
* remove the line mentioning this (c159079af9f57ff0e495a269a4cf850015693aca) commit.
* save
* to be sure you are all good, do a diff of this local branch with the remote.

tallytalwar: I also plan to "squash and merge" this PR, so it should say "hallahan with tallytalwar" when merged.

hallahan: Hmm, when I delete c159079 in the interactive rebase, I get conflicts for each commit. When I resolve each, I then diff it with tangrams master, and it removes all of the iOS code. Continuing to look into this...

hallahan: OK, after looking further into it. I don't think there is a solution to removing `cc159079` from the rebased branch without actually removing the diff that that commit applies.

I went ahead and made a new branch and applied a diff of the `minZoom` branch.

https://github.com/hallahan/tangram-es/tree/display_zoom

We can a) open a new PR with that, or b) just merge this PR as-is.

Sorry for the weird rebase.

tallytalwar:
This is good to go, merging when travis goes green.

PR #945, opened 2016-08-31
Expose Camera APIs in iOS SDK

msmollin:


ios/src/TGMapViewController.h
```diff
@@ -22,13 +42,34 @@
...
 @property (assign, nonatomic) BOOL continuous;
-@property (weak, nonatomic) id<TangramGestureRecognizerDelegate> gestureDelegate;
+@property (weak, nonatomic, nullable) id<TangramGestureRecognizerDelegate> gestureDelegate;
+
+// The following are computed property. They return sensible defaults when the above .map property is nil
+@property (assign, nonatomic) TangramCameraType cameraType;
+@property (assign, nonatomic) TangramGeoPoint position;
+@property (assign, nonatomic) float zoom;
+@property (assign, nonatomic) float rotation;
+@property (assign, nonatomic) float cameraTilt;
```
blair1618: `tilt` seems more in line with the other property names than `cameraTilt`. The [MapKit camera](https://developer.apple.com/library/ios/documentation/MapKit/Reference/MKMapCamera_class/index.html#//apple_ref/occ/instp/MKMapCamera/pitch) uses "pitch" for this, and "bearing" for rotation (sort of) but we don't have to match it exactly.

ios/src/TGMapViewController.h
```diff
@@ -22,13 +42,34 @@
...
 
 - (void)renderOnce;
 
+- (void)setPositionEasedAtPosition:(TangramGeoPoint)position withDuration:(float)duration;
```
blair1618: There's gotta be a better way to name this. Is the idea is to make it read like a sentence? What about `startEasingToPosition: ...` or `animateToPosition: ...`

ios/src/TGMapViewController.mm
```diff
@@ -21,6 +21,131 @@ @interface TGMapViewController ()
...
 @implementation TGMapViewController
 
+- (void)setPosition:(TangramGeoPoint)position {
+    if (self.map) {
+        self.map->setPosition(position.longitude, position.latitude);
+    }
+}
+
+- (void)setPositionEasedAtPosition:(TangramGeoPoint)position withDuration:(float)duration {
+    [self setPositionEasedAtPosition:position withDuration:duration withEaseType:TangramEaseTypeQuint];
```
blair1618: I think we decided to make the default ease type "cubic" for consistency with the Android SDK.

ios/src/TGMapViewController.h
```diff
@@ -22,13 +42,34 @@
...
+
+- (void)setPositionEasedAtPosition:(TangramGeoPoint)position withDuration:(float)duration withEaseType:(TangramEaseType)easeType;
+
+- (void)setZoomEased:(float)zoomLevel withDuration:(float)duration;
+
+- (void)setZoomEased:(float)zoomLevel withDuration:(float)duration withEaseType:(TangramEaseType)easeType;
+
+- (void)setRotationEased:(float)rotation withDuration:(float)duration;
+
+- (void)setRotationEased:(float)rotation withDuration:(float)duration withEaseType:(TangramEaseType)easeType;
```
blair1618: Similar to the eased position setter, we could name these along the lines of `startEasingToRotation` or `animateToRotation`, same for zoom. 

There should probably be one of these for tilt as well but I don't see it, am I overlooking it?

ios/src/TGMapViewController.mm
msmollin: We did. Fixed.

ios/src/TGMapViewController.h
msmollin: I like `animateToPosition;` better, agreed. Updated.

ios/src/TGMapViewController.h
msmollin: Agreed, and updated.

As for cameraTilt, yep I missed exposing the eased variants. Good catch. Added.

ios/src/TGMapViewController.h
msmollin: Fair enough. Changed it to `tilt`.

msmollin:
@blair1618 update with changes.

blair1618:
Tweaked some naming; I'm ready to merge this now as soon as builds pass.

PR #947, opened 2016-09-01
Blend mode 'none' -> 'opaque'

blair1618:
The default blend mode is actually denoted "opaque", as we now indicate in the docs: https://github.com/tangrams/tangram-docs/commit/9e96eaf154adb6c180a10fb31df1e1cf62c3db67


bcamper:
Thanks!

Issue #905, opened 2016-08-08
Feature Selection


hallahan:
As discussed in #748, we currently can select labels, but not the underlying points, lines, and polygons. To create an OSM editor or see properties associated with a map feature, we need to be able to select that feature from a tap event.

Looking at Tangram JS, we already have [this functionality](https://github.com/tangrams/tangram/blob/master/src/selection.js). It's an interesting solution, where we have a frame buffer that stores an identifier to a feature in a given pixel. 

Looking at Tangram ES, you have a custom 2d intersection library, [isect2d](https://github.com/tangrams/isect2d) being used for label selection--a really neat approach. This might be useful also for making rendering decisions from feature collision detection?

In the past, for this sort of thing, I've indexed features in an R-Tree or QuadTree, queried, and then done an intersection test.

What are your current thoughts on the best path forward for feature selection?

bcamper:
Thanks @hallahan - the re-use of collision system in ES for feature picking was initially done as a matter of convenience, because we needed a subset of the full feature selection functionality on a short timeline and weren't yet ready to do the implementation for feature picking on every object. But our expectation has always been to extend this to all features.

I think I favor implementing the selection buffer approach from Tangram JS in ES as well, to maintain closer behavioral parity. That said, there are certainly arguments for both approaches, as well as unique needs on mobile that would have to be factored into the selection buffer method, such as a larger "touch area" for smaller features (this *could* be done with a selection buffer).

I will let @blair1618 weigh in with additional considerations.

hallahan:
Regarding a "touch area" for smaller features, I'm thinking of two ways of going about this. One idea is to create geometric buffers around your points, lines, and polygons (making everything a polygon). You could then use those geometries for your selection buffer.

However, one trick I've done in the past that lets you avoid that is to actually create a geometric buffer around the touch point--a touch area polygon. Then, you would query several of the pixels in the selection frame buffer or do a collision test for that touch area polygon. This would be easier and require less computation upfront...

Since several features may be selected from the polygon of a touch area, we'd probably also want to implement selection priorities for features based off of certain things. One simple way of doing that is to priorities points over lines over polygons. Maybe how that is done could be specified in the scene file?

bcamper:
Selection priorities is an interesting idea. I would probably start with using the existing `order` values to determine priority, and then see if that is sufficient, or if an additional selection-specific parameter is needed?

hallahan:
[...]

PR #948, opened 2016-09-01
Feature selection from selection render buffer

karimnaaji:
https://github.com/tangrams/tangram-es/issues/905

### DIFF
--- a/core/src/gl/framebuffer.cpp
+++ b/core/src/gl/framebuffer.cpp
@@ -1,0 +1,111 @@
+#include "framebuffer.h"
+
+#include "gl/texture.h"
+#include "gl/renderState.h"
+#include "gl/error.h"
+
+namespace Tangram {
+
+FrameBuffer::FrameBuffer(bool _colorRenderBuffer) :
+    m_glFrameBufferHandle(0),
+    m_generation(-1),
+    m_valid(false),
+    m_colorRenderBuffer(_colorRenderBuffer) {
+
+}
+
+void FrameBuffer::applyAsRenderTarget(RenderState& _rs,
+                                      unsigned int _rtWidth, unsigned int _rtHeight,
+                                      unsigned int _vpWidth, unsigned int _vpHeight) {
+
+    if (!m_glFrameBufferHandle) {
+        init(_rs, _rtWidth, _rtHeight);
+    }
+
+    if (!m_valid) {
+        return;
+    }
+
+    _rs.depthMask(GL_TRUE);
+
+    // TOOD: get viewport size from render state, re-enable when unbinding
+    GL_CHECK(glViewport(0, 0, _vpWidth, _vpHeight));
+    GL_CHECK(glClear(GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT));
+
+    // TODO: use render state
+    GL_CHECK(glBindFramebuffer(GL_FRAMEBUFFER, m_glFrameBufferHandle));
+}
+
+void FrameBuffer::init(RenderState& _rs, unsigned int _rtWidth, unsigned int _rtHeight) {
+
+    // get previous bound fbo
+
+    GL_CHECK(glGenFramebuffers(1, &m_glFrameBufferHandle));
+
+    // TOOD: use render state for binding
+    GL_CHECK(glBindFramebuffer(GL_FRAMEBUFFER, m_glFrameBufferHandle));
+
+    // Setup color render target
+    if (m_colorRenderBuffer) {
+        GL_CHECK(glGenRenderbuffers(1, &m_glColorRenderBufferHandle));
+        GL_CHECK(glBindRenderbuffer(GL_RENDERBUFFER, m_glColorRenderBufferHandle));
+        GL_CHECK(glRenderbufferStorage(GL_RENDERBUFFER, GL_RGBA,
+                                       _rtWidth, _rtHeight));
+
+        GL_CHECK(glFramebufferRenderbuffer(GL_FRAMEBUFFER, GL_COLOR_ATTACHMENT0,
+                                           GL_RENDERBUFFER, m_glColorRenderBufferHandle));
+    } else {
+        TextureOptions options =
+            {GL_RGBA, GL_RGBA,
+            {GL_NEAREST, GL_NEAREST},
+            {GL_CLAMP_TO_EDGE, GL_CLAMP_TO_EDGE}
+        };
+
+        m_texture = std::make_unique<Texture>(_rtWidth, _rtHeight, options);
+        m_texture->update(_rs, 0);
+
+        GL_CHECK(glFramebufferTexture2D(GL_FRAMEBUFFER, GL_COLOR_ATTACHMENT0,
+                                        GL_TEXTURE_2D, m_texture->getGlHandle(), 0));
+    }
+
+    {
+        // Create depth render buffer
+        GL_CHECK(glGenRenderbuffers(1, &m_glDepthRenderBufferHandle));
+        GL_CHECK(glBindRenderbuffer(GL_RENDERBUFFER, m_glDepthRenderBufferHandle));
+        GL_CHECK(glRenderbufferStorage(GL_RENDERBUFFER, GL_DEPTH_COMPONENT16,
+                                       _rtWidth, _rtHeight));
+
+        GL_CHECK(glFramebufferRenderbuffer(GL_FRAMEBUFFER, GL_DEPTH_ATTACHMENT,
+                                           GL_RENDERBUFFER, m_glDepthRenderBufferHandle));
+    }
+
+    GLenum status = glCheckFramebufferStatus(GL_FRAMEBUFFER);
+    GL_CHECK();
+
+    if (status != GL_FRAMEBUFFER_COMPLETE) {
+        LOGE("Framebuffer status is incomplete");
+    } else {
+        m_valid = true;
+    }
+
+    m_generation = _rs.generation();
+
+    m_disposer = Disposer(_rs);
+}
+
+FrameBuffer::~FrameBuffer() {
+
+    int generation = m_generation;
+    GLuint glHandle = m_glFrameBufferHandle;
+
+    m_disposer([=](RenderState& rs) {
+        if (rs.isValidGeneration(generation)) {
+            // TODO: unset render state for framebuffer binding
+            // rs.framebufferUnset(glHandle);
+
+            GL_CHECK(glDeleteFramebuffers(1, &glHandle));
+        }
+    });
+}
+
+}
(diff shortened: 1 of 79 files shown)


karimnaaji:
This pull request is now ready for review.

Following and upcoming work on this is (separate PRs):
- add a scalable thumb size from the Tangram API
- some work relating to VAO usage that are now being simply deactivated for the selection render pass
- marker selection interface

lygstate:
I suggest add  screenshot function by using framebuffer

karimnaaji:
Hi @lygstate , I'm not quite sure to see any advantage of using a framebuffer for the screenshot functionality. I understand that the pipeline would stall after a call of `glReadPixels`, but this would happen with or without reading from a framebuffer that is not the default one. Maybe you can give more details over your suggestion.

If you find that the screenshot API is a bottleneck for any of your use, I would recommend you to use pixel buffer object when opengl es 3.0 is available; the memory would be managed directly by opengl and you could potentially waste less cpu cycles when reading texture data.

lygstate:
I am using tangram-es in other thread , to move the data from the other thread to the main thread I need making screenshot of the map after the render function, and framebuffer provide that function

karimnaaji:
Can you open an issue with the exact use case and pipeline you have? It's a legitimate request to have a need of making screenshot from a different thread than the render thread, but we might already be able to find a workaround for that with the current API.

core/src/util/featureSelection.cpp
```diff
@@ -29,12 +32,11 @@ uint32_t FeatureSelection::colorIdentifier(const Feature& _feature, const TileID
 
 bool FeatureSelection::clearFeaturesForTile(TileID _tileID) {
 
+    std::lock_guard<std::mutex> guard(m_mutex);
```
karimnaaji: Do we really need to guard the whole block? `find` and `end` on `m_tileFeatures` could use const (`const auto it` can force to use a `const_iterator` on `find`).

hjanetzek: Still the insertion into `m_tileFeatures` would not be synchronized with `find` then. Or `m_tileFeatures.end()` could be changed since `find` returned end.

karimnaaji: Right, I see what you mean.

hjanetzek:
One thought for future optimization: It would be nice to have a separate buffer for selectionColor that gets only bound on selection draw calls.

karimnaaji:
I think what you describe is already what happens, or am I missing something? If by buffer you mean framebuffer then yes, the selectionColor is drawn on a different framebuffer that gets bound only on selection draw calls when a query is pending (Except on debug mode - where selection color is drawn on the main framebuffer)

karimnaaji:
If by buffer you mean vertex buffer, that's a good idea but could get a little complicated on the style side.

hjanetzek:
@karimnaaji yes, I meant to keep the vertex attribute for selection color in a separate buffer.

Looks good to me now. Last commits changed FeatureSelection usage quite a bit. @blair1618 it's your turn :)

karimnaaji:
@hjanetzek thanks I reviewed the changes, this refactor looks good.

core/src/gl/framebuffer.cpp
```diff
@@ -0,0 +1,159 @@
...
+
+    GL::genFramebuffers(1, &m_glFrameBufferHandle);
+
+    _rs.framebuffer(m_glFrameBufferHandle);
+
+    // Setup color render target
+    if (m_colorRenderBuffer) {
+        GL::genRenderbuffers(1, &m_glColorRenderBufferHandle);
+        GL::bindRenderbuffer(GL_RENDERBUFFER, m_glColorRenderBufferHandle);
+        GL::renderbufferStorage(GL_RENDERBUFFER, GL_RGBA8_OES,
```
blair1618: I hate to be the one to say it, but `GL_RGBA8_OES` is _technically_ an extension, so maybe we should default to a color texture target.

core/src/style/pointStyle.h
```diff
@@ -60,8 +62,10 @@ class PointStyle : public Style {
     std::shared_ptr<SpriteAtlas> m_spriteAtlas;
     std::shared_ptr<Texture> m_texture;
 
-    UniformLocation m_uTex{"u_tex"};
-    UniformLocation m_uOrtho{"u_ortho"};
+    struct UniformBlock {
+        UniformLocation uTex{"u_tex"};
+        UniformLocation uOrtho{"u_ortho"};
+    } m_uniforms[2];
```
karimnaaji: Yes good point, this will be more elegant.

core/src/tangram.cpp
```diff
@@ -75,9 +82,11 @@ class Map::Impl {
     TileWorker tileWorker{MAX_WORKERS};
     TileManager tileManager{tileWorker};
     MarkerManager markerManager;
+    FrameBuffer selectionBuffer{256, 256};
```
karimnaaji: This is the size being used in tangram-js. This seems to give enough resolution but I agree that having this a ratio of the actual rendering area would make sense.
@bcamper how did this value was actually chosen in `-js`?

karimnaaji:
@hjanetzek it seems that there is a slight regression coming from the last commits.
Text labels don't get assigned a selection color (only black):

screen shot 2016-10-24 at 6 07 41 pm

core/src/tile/tileBuilder.cpp
```diff
@@ -141,7 +141,7 @@ std::shared_ptr<Tile> TileBuilder::build(TileID _tileID, const TileData& _tileDa
         tile->setMesh(builder.second->style(), builder.second->build());
     }
 
-    tile->setSelectionFeatures(m_selectionFeatures);
+    tile->setSelectionFeatures(std::move((m_selectionFeatures)));
```
hjanetzek: "move does not move" :) 

The copy is intended so that m_selectionFeatures does not get constantly reallocated but is used as a buffer. By copying the selectionFeatures hold by Tile also gets shrinked to the appropriate size.

core/src/tangram.cpp
hjanetzek: Oh, yes. I've added a commit to visualize the actual selection framebuffer for testing

hjanetzek:
Skip drawing of not selectable features should give better results. E.g. here building outlines overdraw the selectable building. (screenshots were taken with a selection buffer with half screen-resolution)
shot-2016-10-25_13-12-37
shot-2016-10-25_13-12-30

shot-2016-10-25_13-10-20
shot-2016-10-25_13-09-21

karimnaaji:
@hjanetzek it is a good idea to skip some drawing, but we should keep on leaving the selection buffer as _what you can see_, otherwise you'll select invisible things.

Here is what can happen with latests commits:

What you actually see:
screen shot 2016-10-25 at 9 48 07 am

What selection buffer gives you:
screen shot 2016-10-25 at 9 48 09 am

core/src/tangram.cpp
karimnaaji: Yes those screenshots you've taken seem to give better results than the debug toggle that was there since you see the actual resolution, we should use that one instead.

hjanetzek:
Hmm, yeah... We could use a `obstacle: true` property to avoid this issue

core/src/tangram.cpp
```diff
@@ -275,6 +275,7 @@ void Map::resize(int _newWidth, int _newHeight) {
     impl->renderState.viewport(0, 0, _newWidth, _newHeight);
 
     impl->view.setSize(_newWidth, _newHeight);
+    impl->selectionBuffer = std::make_unique<FrameBuffer>(_newWidth/2, _newHeight/2, false);
```
hjanetzek: Yes, need to check again. Though it seems that is should be available since GLES 2.0  http://aras-p.info/blog/2012/10/17/non-power-of-two-textures/

core/src/tangram.cpp
hjanetzek: Ok, I'll clean up the debug mode to be mergable.

core/src/tangram.cpp
```diff
@@ -457,6 +457,8 @@ void Map::render() {
 
     impl->labels.drawDebug(impl->renderState, impl->view);
 
+    impl->selectionBuffer.drawDebug(impl->renderState, glm::vec2(impl->view.getWidth(), impl->view.getHeight()));
```
karimnaaji: We can probably leave that commit and replace the toggle, I prefer how the framebuffer is being visualized here.

core/src/gl/framebuffer.cpp
karimnaaji: Yep I overlooked this while getting it to work, I'll update this part of the code to be more consistent over the availability of the extension.

core/src/style/pointStyle.cpp
```diff
@@ -61,14 +63,28 @@ void PointStyle::onBeginDrawFrame(RenderState& rs, const View& _view, Scene& _sc
...
 
     m_textStyle->onBeginDrawFrame(rs, _view, _scene);
 }
 
+void PointStyle::onBeginDrawSelectionFrame(RenderState& rs, const View& _view, Scene& _scene) {
+    m_mesh->upload(rs);
+
+    Style::onBeginDrawSelectionFrame(rs, _view, _scene);
+
+    m_selectionProgram->setUniformMatrix4f(rs, m_uniforms[Style::selectionShaderUniformBlock].uOrtho,
```
karimnaaji: @hjanetzek  here we give a try to give a scale factor to extend the width of polylines, to fix the issue of low resolution framebuffer to select some discarded fragments (specifically for polylines) that you mentioned on slack.

karimnaaji:
@hjanetzek I am experiencing a crash on app exit since recently. I'm focusing on some other things on this branch at the moment. If you can reproduce please give it a look, otherwise I'll look into this next.

```
libc++abi.dylib: terminating with uncaught exception of type std::__1::system_error: mutex lock failed: Invalid argument
```
From `~ShaderProgram`.

core/src/gl/framebuffer.cpp
karimnaaji: Updated in https://github.com/tangrams/tangram-es/pull/948/commits/4692fc724f42c48aed07873ee72d8a737fc63762

core/src/util/color.h
```diff
@@ -16,6 +17,10 @@ struct Color {
     Color(uint32_t _abgr) : abgr(_abgr) {}
     Color(uint8_t _r, uint8_t _g, uint8_t _b, uint8_t _a) : r(_r), g(_g), b(_b), a(_a) {}
 
+    glm::vec4 asVec4() {
```
karimnaaji: Updated in 91fb942e8bfb2d0469299853f5897bf7a7cb7a0b.

karimnaaji:
No changes to come on this branch.

PR #949, opened 2016-09-01
Fix desktop application to destroy map on context destroy

karimnaaji:
So we don't retain the map instance memory when simulating loss of context on desktop apps.

### DIFF
--- a/linux/src/main.cpp
+++ b/linux/src/main.cpp
@@ -269,10 +269,12 @@ void window_size_callback(GLFWwindow* window, int width, int height) {
 void init_main_window(bool recreate) {
 
     // Setup tangram
-    if (!map) {
-        map = new Tangram::Map();
-        map->loadSceneAsync(sceneFile.c_str(), true);
+    if (map) {
+        delete map;
     }
+
+    map = new Tangram::Map();
+    map->loadSceneAsync(sceneFile.c_str(), true);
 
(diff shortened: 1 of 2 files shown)


blair1618:
The situation we want to test is when a map object is unchanged except for the fact that the GL context has been invalidated. If we destroy and re-create the map object whenever we simulate context loss, doesn't that mean we aren't actually testing how the map behaves after a context loss?

karimnaaji:
I agree with the deletion of the app, we would need to find something else; right now, none of the disposer gets assigned a call-back from GL resources destructors since none of the smart pointers for them gets decreased on desktop app when this happens. All we do is recreate and hold the gl objects and memory when simulating this.

blair1618:
I think what you're describing is intended behavior. When the GL context is invalidated the GL resources just disappear, without an application calling "destroy" functions or anything. So we don't need or want disposers to run when the context if invalidated, we only want to re-create the resources we need.

blair1618:
I really wish we didn't have to worry about this scenario, if we had some evidence that this is rare or nonexistent on most current Android hardware then I'd consider removing this behavior completely.

karimnaaji:
Ok I got confused while testing this with the opengl profiler on desktop, I forgot the pipeline wasn't close at all to the Android context loss. We're all good here.

PR #955, opened 2016-09-07
[rfc] GL api wrapper

hjanetzek:
This should allow to cleanly use the evas_glapi on Tizen.

We may also consider to directly use the platform GL headers again. At least for me compile times are not an issue anymore :)


core/src/gl.h
```diff
@@ -270,156 +248,121 @@ typedef char            GLchar;
...
-    GL_APICALL void GL_APIENTRY glStencilMask(GLuint mask);
-    GL_APICALL void GL_APIENTRY glStencilOp(GLenum fail, GLenum zfail, GLenum zpass);
-    GL_APICALL void GL_APIENTRY glClearStencil(GLint s);
-    GL_APICALL void GL_APIENTRY glColorMask(GLboolean red, GLboolean green, GLboolean blue, GLboolean alpha);
-    GL_APICALL void GL_APIENTRY glCullFace(GLenum mode);
-    GL_APICALL void GL_APIENTRY glFrontFace(GLenum mode);
-    GL_APICALL void GL_APIENTRY glClearColor(GLclampf red, GLclampf green, GLclampf blue, GLclampf alpha);
-    GL_APICALL void GL_APIENTRY glGetIntegerv(GLenum pname, GLint *params );
+namespace Tangram {
+struct GL {
```
tallytalwar: @hjanetzek and I were discussing yesterday about `struct` vs `namespace` for `GL` here. The only thing `struct` brings here is future proofing for if we want to add any member variable (maybe some debugging).

tallytalwar:
Tested this on `android` and `osx`. I can not get `ios` to build on my machine, @blair1618 can you test this out, please?

Thanks

tallytalwar:
I am good with this. :+1:

blair1618:
Checkin it out

blair1618:
No issues building and running on iOS. More thoughts on this in a bit.

CMakeLists.txt
```diff
@@ -4,6 +4,7 @@ project(tangram)
 
 # Options
 option(USE_SYSTEM_FONT_LIBS "Use system libraries Freetype, ICU and Harfbuzz via pkgconfig" OFF)
+option(USE_SYSTEM_GLFW_LIBS "Use system libraries for GLFW3 via pkgconfig" OFF)
```
blair1618: Why is this change in this PR? It's not a bad change, just seems unrelated.

core/common/platform_gl.cpp
```diff
@@ -0,0 +1,287 @@
+#include "platform_gl.h"
+
+#include "gl.h"
+#include "gl/error.h"
+
+namespace Tangram {
+
+GLenum GL::getError() {
+    auto result = glGetError();
+    GL_CHECK();
```
blair1618: I think this will cause unbounded recursion into `GL::getError`

core/common/platform_gl.h
```diff
@@ -0,0 +1,67 @@
...
+#ifndef GL_GLEXT_PROTOTYPES
+#define GL_GLEXT_PROTOTYPES 1
+#endif
+
+#include <GLES2/gl2.h>
+#include <GLES2/gl2ext.h>
+typedef void (GL_APIENTRYP PFNGLBINDVERTEXARRAYOESPROC) (GLuint array);
+typedef void (GL_APIENTRYP PFNGLDELETEVERTEXARRAYSOESPROC) (GLsizei n, const GLuint *arrays);
+typedef void (GL_APIENTRYP PFNGLGENVERTEXARRAYSOESPROC) (GLsizei n, GLuint *arrays);
+typedef GLboolean (GL_APIENTRYP PFNGLISVERTEXARRAYOESPROC) (GLuint array);
```
blair1618: I believe these types are already declared in `gl2ext.h`

blair1618:
This is cool. I like that GL error checking is now consolidated to one file and that macros and typedefs are hidden from GL function call sites.

What I'd like to move towards eventually is using the official Khronos GL ES 2 headers and a dynamic function loader like [GLAD](https://github.com/Dav1dde/glad). For the platforms that don't ship with ES 2 we can load ANGLE libraries for a consistent interface. 

In the meantime this looks like a good step that certainly simplifies the Tizen GL interface.

CMakeLists.txt
hjanetzek: I can move it to a separate PR if you prefer.

core/common/platform_gl.cpp
hjanetzek: Good catch!

tallytalwar: Good eye!!

blair1618:
LGTM

PR #956, opened 2016-09-07
Optimize properties get

hjanetzek:


core/src/style/textStyleBuilder.cpp
```diff
@@ -335,8 +335,8 @@ TextStyle::Parameters TextStyleBuilder::applyRule(const DrawRule& _rule,
 
         _rule.get(StyleParamKey::text_anchor, p.labelOptions.anchors);
         if (p.labelOptions.anchors.count == 0) {
-            p.labelOptions.anchors.anchor = { LabelProperty::Anchor::bottom, LabelProperty::Anchor::top,
-                                              LabelProperty::Anchor::right, LabelProperty::Anchor::left };
+            p.labelOptions.anchors.anchor = { {LabelProperty::Anchor::bottom, LabelProperty::Anchor::top,
```
tallytalwar: This syntax might not work with gcc 4.9.

hjanetzek: Travis build still uses 4.9. I know there are some initializer constructs that don't work - but in this case it seems to be ok.

blair1618:
LGTM

PR #957, opened 2016-09-07
Do not render until all texture resources are available

### LINKED ISSUE #814: Postpone scene rendering until all resources are ready

tallytalwar:
Fixes #814

### DIFF
--- a/core/src/scene/sceneLoader.cpp
+++ b/core/src/scene/sceneLoader.cpp
@@ -539,6 +539,9 @@ std::shared_ptr<Texture> SceneLoader::fetchTexture(const std::string& name, cons
 
                     updateSpriteNodes(name, texture, scene);
                     scene->pendingTextures--;
+                    if (scene->pendingTextures == 0) {
+                        requestRender();
+                    }
                 }
             });
         texture = std::make_shared<Texture>(nullptr, 0, options, generateMipmaps);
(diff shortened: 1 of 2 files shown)


hjanetzek:
Does this trigger a render request once all textures are available?

tallytalwar:
Hmm.. good point. I do not think so. Will look.

tallytalwar:
@hjanetzek Updated.

core/src/scene/sceneLoader.cpp
```diff
@@ -596,7 +598,12 @@ std::shared_ptr<Texture> SceneLoader::fetchTexture(const std::string& name, cons
                     }
 
                     updateSpriteNodes(name, texture, scene);
-                    scene->resourceLoad--;
+                    using namespace std::chrono_literals;
+                    std::this_thread::sleep_for(10s);
```
tallytalwar: Ohh yes, updated.

tallytalwar:
Rebased and Updated. Ultra small code change here!

PR #961, opened 2016-09-09
Unified selection buffer technique for all styles

karimnaaji:
Remove old technique to select label with bounding boxes and use a single technique (render to framebuffer with selection ids)
- [x] Update JNI bindings

To go in https://github.com/tangrams/tangram-es/pull/948


bcamper:
🎊

PR #965, opened 2016-09-13
Fix global property updates

tallytalwar:
since global properties are substituded when parsing the yaml scene for the first time, there was no information of
where the global properties were applied when any of these globals were updated.

This fixes the issue, by keeping track of the path of the node(s) which were using the global properties to begin with
so that subsequent global property updates are applied.

### DIFF
--- a/core/src/scene/scene.cpp
+++ b/core/src/scene/scene.cpp
@@ -70,6 +70,21 @@ Scene::Scene(const Scene& _other)
 
 Scene::~Scene() {}
 
+bool Scene::removeGlobalRef(const std::string& key) {
+    bool found = false;
+    for (auto& globalRef : m_referencedGlobals) {
+        if (globalRef.first == key) {
+            globalRef.swap(m_referencedGlobals.back());
+            m_referencedGlobals.pop_back();
+            LOG("Found keys: %s", key.c_str());
+            found = true;
+            break;
+        }
+    }
+
+    return found;
+}
+
 const Style* Scene::findStyle(const std::string& _name) const {
 
     for (auto& style : m_styles) {
--- a/core/src/scene/scene.h
+++ b/core/src/scene/scene.h
@@ -1,7 +1,6 @@
 #pragma once
 
 #include "util/color.h"
-#include "util/fastmap.h"
 #include "view/view.h"
 
 #include <atomic>
@@ -9,6 +8,7 @@
 #include <memory>
 #include <string>
 #include <vector>
+#include <tuple>
 #include <unordered_map>
 
 #include "glm/vec2.hpp"
@@ -74,6 +74,8 @@ public:
     auto& background() { return m_background; }
     auto& fontContext() { return m_fontContext; }
     auto& globals() { return m_globals; }
+    auto& referencedGlobals() { return m_referencedGlobals; }
+    bool removeGlobalRef(const std::string& key);
     Style* findStyle(const std::string& _name);
 
     const auto& path() const { return m_path; }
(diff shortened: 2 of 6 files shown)


tallytalwar:
Will give this a look tomorrow again, but should be ready for review.
Also this is an urgent fix.

tallytalwar:
Found a bug here,  when testing with eraser map. Wait on the review.

tallytalwar:
Updated!

blair1618:
A corner case we'll want to cover before we forget: If a node that previously referred to a global value is explicitly changed to a new value, the link between that global and the node should be broken. For example:
```yaml
global:
  bg_color: cyan
scene:
  background: global.bg_color
```
This now creates a link between `scene.background` and `global.bg_color`. If we apply a scene update that changes `scene.background` to `blue`, we want to destroy that link because the user has now indicated that `scene.background` should no longer be determined by `global.bg_color`.

tallytalwar:
Updated to handle all the cases you mentioned.

core/src/scene/scene.h
```diff
@@ -31,6 +31,8 @@ struct Stops;
  * Scene is a singleton containing the styles, lighting, and interactions defining a map scene
  */
 
+using GlobalUpdate = std::pair<std::string, YAML::Node>;
```
blair1618: Abstraction is cool, but this type definition is used in exactly one place so in this case we're not clarifying anything - let's remove it.

core/src/scene/scene.h
```diff
@@ -74,6 +76,8 @@ class Scene {
     auto& background() { return m_background; }
     auto& fontContext() { return m_fontContext; }
     auto& globals() { return m_globals; }
+    auto& globalUpdates() { return m_globalUpdates; }
```
blair1618: Seems like this is really a collection of global _references_, I think we should call it something like that now that we've figured out exactly how this works.

core/src/scene/sceneLoader.cpp
```diff
@@ -78,31 +78,43 @@ bool SceneLoader::loadConfig(const std::string& _sceneString, Node& root) {
...
-            std::string path;
-            for (const auto& k : keys) { path += "." + k; }
-            LOGW("Cannot update scene - key not found: %s", path.c_str());
-        }
+void SceneLoader::applyUpdates(Scene& scene, const std::vector<SceneUpdate>& updates) {
+    auto& root = scene.config();
+    for (const auto& update : updates) {
+        scene.removeGlobalUpdate(update.keys);
+        auto keys = splitString(update.keys, COMPONENT_PATH_DELIMITER);
+        auto valueNode = YAML::Load(update.value);
```
blair1618: `YAML::Load` throws exceptions on parsing errors, I think we'll need a try/catch here :\

core/src/scene/scene.h
tallytalwar: Yeah I know but the reason I added it because `std::vector<std::pair<std::string, YAML::Node>>` looks ugly. Will remove it though.

tallytalwar:
Updated!

core/src/scene/scene.cpp
```diff
@@ -70,6 +70,21 @@ Scene::Scene(const Scene& _other)
 
 Scene::~Scene() {}
 
+bool Scene::removeGlobalUpdate(const std::string& key) {
```
blair1618: `removeGlobalReference` maybe?

core/src/scene/scene.cpp
```diff
@@ -70,6 +70,21 @@ Scene::Scene(const Scene& _other)
...
+    bool found = false;
+    for (auto& globalUpdate : m_referencedGlobals) {
+        if (globalUpdate.first == key) {
+            globalUpdate.swap(m_referencedGlobals[m_referencedGlobals.size() - 1]);
```
blair1618: I like this trick :) Slightly more concise would be `globalUpdate.swap(m_referencedGlobals.back());`

Also maybe `globalRef` would make a better name for this variable.

tallytalwar:
Last set of nitpicks!! Updated!

PR #966, opened 2016-09-13
Simply print the GL info log when a shader fails to compile

blair1618:
The previous parsing of the error log was unreliable because the logs generally don't conform to a single format.

Having more info about shader errors would still be nice, especially since the shaders get so large. Maybe next we could add an option to write failed shaders to a log file for later analysis.

### DIFF
--- a/core/src/gl/shaderProgram.cpp
+++ b/core/src/gl/shaderProgram.cpp
@@ -223,28 +223,7 @@ GLuint ShaderProgram::makeCompiledShader(const std::string& _src, GLenum _type) 
             infoLog.resize(infoLength);
 
             GL_CHECK(glGetShaderInfoLog(shader, infoLength, NULL, static_cast<GLchar*>(&infoLog[0])));
-            LOGE("Shader compilation failed %s", m_description.c_str());
-
-            std::stringstream ss(source);
-            std::string item;
-            std::vector<std::string> sourceLines;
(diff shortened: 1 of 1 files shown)


hjanetzek:
Would be nice if you could ```#if 0``` out that part for now (or have both). I'll check the different log formats on my devices and update the info parsing soon.

core/src/gl/shaderProgram.cpp
```diff
@@ -223,28 +223,7 @@ GLuint ShaderProgram::makeCompiledShader(const std::string& _src, GLenum _type)
...
-            // Print errors with context
-            std::string line;
-            ss.str(infoLog);
-            while (std::getline(ss, line)) {
-                if (line.length() < 2) { continue; }
-
-                int lineNum = 0;
-                sscanf(line.c_str(), "%*d(%d)", &lineNum);
-                logMsg("\nError on line %d: %s\n", lineNum, line.c_str());
-
```
hjanetzek: @blair1618 it seems one major issue was in the use of stringstream.str(infoLog) instead of creating a new stringstream. Otherwise each line of the infoLog must have been printed here - in any case.

blair1618: I noticed that too. I tried creating a new stringstream and that printed the log, but with the wrong source lines.

blair1618:
@hjanetzek Taking a look at your follow-up PR now.

blair1618:
Closing in favor of #967

PR #967, opened 2016-09-14
fix and improve shader error parsing

hjanetzek:


core/src/gl/shaderProgram.cpp
```diff
@@ -223,26 +223,27 @@ GLuint ShaderProgram::makeCompiledShader(const std::string& _src, GLenum _type)
...
-            ss.str(infoLog);
-            while (std::getline(ss, line)) {
+            std::stringstream logStream(infoLog);
+
+            while (std::getline(logStream, line)) {
                 if (line.length() < 2) { continue; }
 
                 int lineNum = 0;
-                sscanf(line.c_str(), "%*d(%d)", &lineNum);
+                if (!sscanf(line.c_str(), "%*d%*[(:]%d", &lineNum)) { continue; }
```
blair1618: I am very confident that there are formats of shader logs in the wild that we cannot anticipate. We should _always_ print the contents of the log in some way, even if we can't determine a line number from it.

blair1618: Ah I just re-read the diff. Disregard my comment!

blair1618:
Cool. This addresses my concerns from #966. LGTM

PR #969, opened 2016-09-14
workaround old gcc std::regex_iterator issue

hjanetzek:
- and this version is nicer anyway :)

### DIFF
--- a/core/src/gl/shaderProgram.cpp
+++ b/core/src/gl/shaderProgram.cpp
@@ -284,13 +284,10 @@ std::string ShaderProgram::applySourceBlocks(const std::string& source, bool fra
 
     auto sourcePos = source.begin();
 
-    auto linesBegin = std::sregex_iterator(source.begin(), source.end(), pragmaLine);
-    auto linesEnd = std::sregex_iterator();
+    for (auto it = source.begin(); std::regex_search(it, source.end(), sm, pragmaLine);) {
+        it += sm.position() + sm.length();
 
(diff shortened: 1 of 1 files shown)


core/src/gl/shaderProgram.cpp
```diff
@@ -284,13 +284,10 @@ std::string ShaderProgram::applySourceBlocks(const std::string& source, bool fra
 
     auto sourcePos = source.begin();
 
-    auto linesBegin = std::sregex_iterator(source.begin(), source.end(), pragmaLine);
-    auto linesEnd = std::sregex_iterator();
+    for (auto it = source.begin(); std::regex_search(it, source.end(), sm, pragmaLine);) {
+        it += sm.position() + sm.length();
```
tallytalwar: Did we not just check that the `smatch::position` is different in earlier versions of gcc? Is it safe to relly on it then?

Or maybe it gives different results when used with regex_iterators!

hjanetzek: yeah, only std::regex_iterator seems to be broken a rather obscure feature. if regex_search gave the wrong position someone would probably have noticed before release :)

tallytalwar: Plus, it does work on older gcc and newer (ofcourse).
Merging when travis goes green.
Thanks @hjanetzek

blair1618:
Ah, this _is_ nicer! LGTM

PR #975, opened 2016-09-15
Markers can use custom bitmaps

blair1618:
Essentially, this works by splitting draw calls to `DynamicQuadMesh` into batches of vertices grouped by texture. 

The point style pipeline is simplified a little bit too. Instead of a custom "filled circle" shader with a compile-time switch, the point style always uses a texture now. If there's no style texture or marker texture provided, a default "filled circle" texture is used. This default texture is encoded as a data header and managed by `RenderState`.


core/src/marker/markerManager.cpp
```diff
@@ -63,6 +64,19 @@ bool MarkerManager::setStyling(MarkerID markerID, const char* styling) {
     return true;
 }
 
+bool MarkerManager::setBitmap(MarkerID markerID, int width, int height, const unsigned int* bitmapData) {
```
tallytalwar: Can we provide another interface which takes in rawData and use stb_image library internally to convert rawData to bitmap?

tallytalwar: Hmm... https://github.com/tangrams/tangram-es/blob/marker-images/core/src/gl/texture.cpp#L28 **might** already do this for us!

blair1618: Yeah, being able to use compressed image formats would be very useful. The design of that interface will, I think, depend on the use cases that we find in the platform SDKs. For instance, should this function take an array of data or a path to a file? One is simpler but the other allows us to cache repeated images, et cetera. I think this is best left as a follow-up task.

core/src/gl/dynamicQuadMesh.h
```diff
@@ -70,35 +94,64 @@ void DynamicQuadMesh<T>::upload(RenderState& rs) {
...
 
-    size_t vertexOffset = 0;
-    size_t maxVertices = RenderState::MAX_QUAD_VERTICES;
-
-    for (size_t offset = 0; offset < m_nVertices; offset += maxVertices) {
-        size_t nVertices = maxVertices;
-
-        if (offset + maxVertices > m_nVertices) {
-            nVertices = m_nVertices - offset;
+    size_t verticesDrawn = 0;
```
tallytalwar: Great job with updates on variable names here. Makes the code much more clearer.

blair1618:
I've got one more tiny change to make here before merging

core/src/style/pointStyleBuilder.cpp
```diff
@@ -164,6 +169,7 @@ void PointStyleBuilder::addLabel(const Point& _point, const glm::vec4& _quad,
                                                      _params.size,
                                                      _params.labelOptions,
                                                      _params.extrudeScale,
+                                                     _params.texture,
```
tallytalwar: Directly pass `m_texture` here. No need to add it to parameters.

blair1618:
I always hit that stupid button when I'm trying to cancel a comment - pay no mind.

tallytalwar:
Glad you reopened, I had a requested change above!

core/src/style/pointStyleBuilder.cpp
blair1618: Good one - not sure what I was thinking there haha

tallytalwar:
Looks good to merge.

tallytalwar:
We should merge this instead of squashing and merging.

PR #979, opened 2016-09-16
Distinguish JS scalar types for scene globals in functions

blair1618:
Previously, any scalar global value would be created in the JS context as a string. Globals should be created with appropriate JS types to enable correct behavior for cases like this:
```yaml
global:
  bool_value: true
...
      filter: function() { return global.bool_value; }
```

### DIFF
--- a/core/src/scene/styleContext.cpp
+++ b/core/src/scene/styleContext.cpp
@@ -83,14 +83,28 @@ StyleContext::~StyleContext() {
     duk_destroy_heap(m_ctx);
 }
 
+// Convert a scalar node to a boolean, double, or string (in that order)
+// and for the first conversion that works, push it to the top of the JS stack.
+void pushYamlScalarAsJsObject(duk_context* ctx, const YAML::Node& node) {
+    bool booleanValue = false;
+    double numberValue = 0.;
+    if (YAML::convert<bool>::decode(node, booleanValue)) {
+        duk_push_boolean(ctx, booleanValue);
+    } else if (YAML::convert<double>::decode(node, numberValue)) {
+        duk_push_number(ctx, numberValue);
+    } else {
+        duk_push_string(ctx, node.Scalar().c_str());
+    }
+}
+
 void StyleContext::parseSceneGlobals(const YAML::Node& node, const std::string& key, int seqIndex,
                                      duk_idx_t dukObject) {
 
     switch(node.Type()) {
         case YAML::NodeType::Scalar:
             if (key.size() == 0) {
-                duk_push_string(m_ctx, node.Scalar().c_str());
-                duk_put_prop_index(m_ctx, dukObject, seqIndex);
+                pushYamlScalarAsJsObject(m_ctx, node);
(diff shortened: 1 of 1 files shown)


core/src/scene/styleContext.cpp
```diff
@@ -83,14 +83,28 @@ StyleContext::~StyleContext() {
     duk_destroy_heap(m_ctx);
 }
 
+// Convert a scalar node to a boolean, double, or string (in that order)
+// and for the first conversion that works, push it to the top of the JS stack.
+void pushYamlScalarAsJsObject(duk_context* ctx, const YAML::Node& node) {
+    bool booleanValue = false;
+    double numberValue = 0.;
+    if (YAML::convert<bool>::decode(node, booleanValue)) {
```
tallytalwar: You can use `getBool`, `getDouble` defined in `util/yamlHelper.h`.

core/src/scene/styleContext.cpp
```diff
@@ -313,14 +327,10 @@ bool StyleContext::evalFunction(FunctionID id) {
...
-    bool result = false;
-
     if (!evalFunction(_id)) { return false; };
 
-    // check for evaluated value sitting at value stack top
-    if (duk_is_boolean(m_ctx, -1)) {
-        result = duk_get_boolean(m_ctx, -1);
-    }
+    // Evaluate the "truthiness" of the function result at the top of the stack.
+    bool result = duk_to_boolean(m_ctx, -1);
```
tallytalwar: Is it guaranteed for the top of the stack to be a boolean?

core/src/scene/styleContext.cpp
```diff
@@ -508,6 +518,7 @@ duk_ret_t StyleContext::jsGetProperty(duk_context *_ctx) {
     } else {
         duk_push_undefined(_ctx);
     }
+    // FIXME: Distinguish Booleans here as well
```
tallytalwar: makes sense :+1:

tallytalwar:
Super nice catch. Every time I look at duk code, I have to re-wire my brain to work FIFO!

core/src/scene/styleContext.cpp
blair1618: The top of the stack at this point will be the return value of the function that was just evaluated. For a filter, we want to evaluate whether this value is "truthy" or not - since this is what tangram-JS evaluates.

core/src/scene/styleContext.cpp
blair1618: I think this function is preferable, since it avoids using exceptions. We could change those yaml helpers to use these functions in the future.

core/src/scene/styleContext.cpp
tallytalwar: Ok. Makes sense.

core/src/scene/styleContext.cpp
tallytalwar: Yes, less exceptions are better.
Make sense to update the helpers to use `decode`.

PR #980, opened 2016-09-17
Fix conversion of a sequence of filters to an implicit 'any' filter

blair1618:
Previously a sequence of filters, like this:
```yaml
filter:
  - { $zoom: 13, height: { min: 250 } }
  - { $zoom: 13, area: { min: 50000 } }
  - { $zoom: 13, volume: { min: 200000 } }
  - { $zoom: 14, height: { min: 190 } }
```
would be converted into an 'all' filter. These should instead be 'any' filters, as documented here: https://mapzen.com/documentation/tangram/Filters-Overview/#lists-imply-any-mappings-imply-all

### DIFF
--- a/core/src/scene/sceneLoader.cpp
+++ b/core/src/scene/sceneLoader.cpp
@@ -1268,7 +1268,7 @@ Filter SceneLoader::generateFilter(Node _filter, Scene& scene) {
             } else if (key == "any") {
                 filters.push_back(generateAnyFilter(node, scene));
             } else if (key == "all") {
-                filters.push_back(generateFilter(node, scene));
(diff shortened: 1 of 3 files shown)


core/src/scene/sceneLoader.cpp
```diff
@@ -1282,6 +1282,7 @@ Filter SceneLoader::generateFilter(Node _filter, Scene& scene) {
 
     if (filters.size() == 0) { return Filter(); }
     if (filters.size() == 1) { return std::move(filters.front()); }
+    if (_filter.IsSequence()) { return Filter::MatchAny(std::move(filters)); }
```
tallytalwar: Nice catch here.

Issue #978, opened 2016-09-16
FontContext::fetch segfaults in Android in Scene Update


hallahan:
While working on MBTiles Support (#960), I've found that I can very nicely add an MBTiles file to a data source dynamically using the scene update mechanism. This works great on OS X, but it segfaults on Android.

After doing a bit of debugging, I've pinpointed that this is when we re-fetch a font if it has been specified in the `scene.yaml`. In specific, it happens right when we do the `startUrlRequest` call. It makes sense that this happens only one one platform, since this is platform specific code.

``` cpp
void FontContext::fetch(const FontDescription& _ft) {

    const static std::regex regex("^(http|https):/");
    std::smatch match;

    if (std::regex_search(_ft.uri, match, regex)) {
        // Load remote
        resourceLoad++;
        startUrlRequest(_ft.uri, [&, _ft](std::vector<char>&& rawData) {
            if (rawData.size() == 0) {
                LOGE("Bad URL request for font %s at URL %s", _ft.alias.c_str(), _ft.uri.c_str());
            } else {
                std::lock_guard<std::mutex> lock(m_mutex);
                char* data = reinterpret_cast<char*>(rawData.data());

                for (int i = 0, size = BASE_SIZE; i < MAX_STEPS; i++, size += STEP_SIZE) {
                    auto font = m_alfons.getFont(_ft.alias, size);
                    font->addFace(m_alfons.addFontFace(alfons::InputSource(data, rawData.size()), size));
                }
            }

            resourceLoad--;
        });
    } else {
    ...
```

If I comment out the font block and the references to it in the `scene.yaml`, my scene update activating MBTiles works.

hallahan:
In specific, I did some log messages in Android's `startUrlRequest`.

```cpp
bool startUrlRequest(const std::string& _url, UrlCallback _callback) {
    LOG("android native startUrlRequest _url: %s", _url.c_str());
    jstring jUrl = jniRenderThreadEnv->NewStringUTF(_url.c_str());
    LOG("jstring jUrl = jniRenderThreadEnv->NewStringUTF(_url.c_str());");
    // This is probably super dangerous. In order to pass a reference to our callback we have to convert it
    // to a Java type. We allocate a new callback object and then reinterpret the pointer to it as a Java long.
    // In Java, we associate this long with the current network request and pass it back to native code when
    // the request completes (either in onUrlSuccess or onUrlFailure), reinterpret the long back into a
[...]

Issue #960, opened 2016-09-08
Core MBTiles Support


hallahan:
As discussed in #931, an important feature many would like to add is offline map tile support. After a bit of further discussion, it looks like the consensus is to build MBTiles support into the data sources.

MBTiles is a [spec](https://github.com/mapbox/mbtiles-spec) that has been maintained since 2011, primarily as a tile storage and transport container. Because it is a widely used method to store tiles, both vector and image, it seems only logical to continue using this schema to store vector tiles offline.

The spec itself describes the SQLite database schema that is required for the file to be considered _MBTiles_. Right now the spec [only describes the contents](https://github.com/mapbox/mbtiles-spec/blob/ded3ecc24b5076bb8242dc1aae333ee04e8ed8bc/1.2/spec.md#content-1) of a tile to be `png` or `jpg`. @pnorman, however, has [submitted a PR](https://github.com/mapbox/mbtiles-spec/pull/46) that adds `geojson`, `topojson`, `o5m`, and `mvt` to the list of valid formats. In addition, there will be a `compression` field in the MBTiles metadata table that declares the codec in which tile contents may be compressed.

In the wild, you can find MBTiles storing vector tiles in several places. The most notable example is [osm2vectortiles.org](http://osm2vectortiles.org/).

My thought is that we want to use MBTiles in two key ways: 1) As a tile store and transport mechanism, and 2) as a cache. 
## Tile Store

There are currently several apps and map SDKs that store raster and vector tiles offline, however, the approach is often only of a caching nature. You will specify a bbox and zoom range in the application, and the application will fetch tiles from an HTTP endpoint, one-by-one. This approach is cumbersome and slow, particulary on a poor internet connection. An alternative workflow would be for Mapzen and others to generate and store pre-populated MBTiles on the server. This approach is consistent with osm2vectortiles, and it can potentially lend well as an addition to [Mapzen Metro Extracts](https://mapzen.com/data/metro-extracts/).
## Cache

[...]

Issue #931, opened 2016-08-25
Offline data storage


matteblair:
_This is a general thread to discuss adding offline storage for map data in tangram-es._

**Motivations**

A mapping engine may be tasked with functioning offline in many ways for many reasons:
- A person is using it on a subway with no network signal
- A person is using it in a remote region of the world with no internet access
- A person has internet access, but wants to avoid using bandwidth for reasons of cost or time
- A person needs to view a set of data that cannot be hosted online for security reasons

_Et cetera._

**Current status**

At present, tangram-es must have network access to retrieve data for new tiles whenever the map view changes. Aside from the currently visible tiles, only a small cache of data is kept in memory within the engine (roughly one screen's worth of data). In the Android and iOS tangram SDKs, the networking layers are able to cache some HTTP requests, but depending on the expiration policy in request headers the networking layer may not be permitted to re-use a response after a certain time (or at all). Some tangram users on Android have found ways to store network data indefinitely with OkHTTP, but the process is cumbersome and platform-specific (https://github.com/tangrams/tangram-es/issues/623). 

**Possible goals**

There are many ways think about solving this problem. 
1. _Allow client applications to provide offline storage._ Instead of tangram-es itself storing data for offline use, the library could instead enable client applications to access and redirect all of the network requests it makes, leaving any storage logic up to the client. This allows the most flexibility for application developers, but also requires substantial work from them. 
2. _Provide offline storage in the platform SDKs._ The tangram SDK for each supported platform could implement its own approach to offline storage, using APIs appropriate to that platform. This allows client applications to more easily interact with the caching logic, but also introduces potential divergence in behavior across platforms.
3. _Provide offline storage in the core C++ library._ The cross-platform core of tangram-es could implement offline storage using a platform-agnostic C++ solution. This would ensure consistent behavior across platforms and may provide better performance than the other options, but would also provide the least flexibility to application developers.

bcamper:
We also want to think about how to best support pre-rendered *tile bundles* for offline use, either shipped with an app as a resource, or downloaded at run-time (see apps like Offmaps as an example, where in-app purchase is used to download new tile bundles). We should look at common bundle formats such as MBTiles, as well as more bare-metal approaches like a simple archive of tile files (that could respect a tile URL pattern, with a `file://` protocol or other local access mechanism).

cc @nvkelso

hallahan:
[...]

PR #982, opened 2016-09-19
MBTiles Support Sprint

hallahan:
I think the MBTiles work is ready for review. This PR is the implementation of #960 & #931. In short, you can now have your tiles cached and stored in an MBTiles SQLite3 database for offline usage.
## Overview

You can declare a data source to have an `mbtiles` file path both programmatically and via your `scene.yaml`. 

If you want to add tiles to your cache online, you can put both a `url` and `mbtiles` in your data source, and http tile requests will be written to your MBTiles db. If the MBTiles file does not exist, Tangram ES will attempt to create a fresh MBTiles file all set up with the proper schema at that path given.

``` yaml
sources:
    osm:
        type: GeoJSON
        url:  https://vector.mapzen.com/osm/all/{z}/{x}/{y}.json
        mbtiles: /Users/njh/osm-data/mbtiles/winters-mapzen-geojson.mbtiles
        max_zoom: 16
        url_params:
            api_key: vector-tiles-tyHL4AY
```

You don't need a URL, however. You can have your data source simply fetch tiles from the MBTiles and not request anything from the network. 

``` yaml
sources:
    osm:
        type: GeoJSON
        mbtiles: /Users/njh/osm-data/mbtiles/winters-mapzen-geojson.mbtiles
        max_zoom: 16
```

This works nicely when you have full knowledge and access to your filesystem. On Android, however, you might need to get a file path to your MBTiles programmatically first.

``` java
File storageDir = Environment.getExternalStorageDirectory();
File mbtilesFile = new File(storageDir, "tangram-geojson-cache.mbtiles");
map.setMBTiles("osm", mbtilesFile);
```

This would set the file `tangram-geojson-cache.mbtiles` to be your MBTiles in the `osm` datasource. It will use a URL from your `scene.yaml` to populate your cache just like before (or not). Similarly, `Map::setMBTiles` is [exposed from the native](https://github.com/hallahan/tangram-es/blob/8232af10e4c7ccc5582f790d6eeee29046d8d04e/core/src/tangram.cpp#L252-L256) C++ `Map` interface as well. We will also want to add Objective C bindings.
## MBTiles Spec

The MBTiles database schema is built to adhere to @pnorman 's [MBTiles 2.0 specification](https://github.com/pnorman/mbtiles-spec/blob/2.0/2.0/spec.md). The metadata table is [being populated with the required values](https://github.com/hallahan/tangram-es/blob/8232af10e4c7ccc5582f790d6eeee29046d8d04e/core/src/util/mbtiles.h#L144-L175) declared. In addition to adhering to the spec itself, I'm also creating MBTiles databases with [identical schema](https://github.com/mapbox/node-mbtiles/blob/4bbfaf991969ce01c31b95184c4f6d5485f717c3/lib/schema.sql) to the most widely used reference implementation, node-mbtiles. You can see the creation of the database in [mbtiles.h](https://github.com/hallahan/tangram-es/blob/8232af10e4c7ccc5582f790d6eeee29046d8d04e/core/src/util/mbtiles.h). One addition that I'd like to make is that we check the `metadata` table and make sure that the data format is consistent with what we have declared in our `scene.yaml`. 
## SQLiteCpp

I'm quite happy with the use of [SQLiteCpp](https://github.com/SRombauts/SQLiteCpp) C++ SQLite3 wrapper library. It integrates easily with the CMake build process. It works without any issue on Mac OS X and Android, and it's usage is part of the core library. One thing that sets this MBTiles implementation apart from the others is that it is done in the core C++ library. We do not re-invent the wheel and re-implement for Android, iOS, etc.
## MD5 Checksum Hash

One clever feature found in `node-mbtiles` that is not mentioned in the spec is that the underlying database schema has a `map` and `images` table that are joined to get you your tile data. The key that joins these tables is an MD5 checksum of the actual tile data. If you have many tiles with duplicate or empty data structures, rather than having a repeat entry in the images table, the checksum will join you to a single entry in `images` to all of the tiles pointing to that same data.

Rather than integrating a big encryption library like OpenSSL, I just wanted the MD5 function. So, after a bit of searching, I found a C++ hash library that was built to be portable and easy to integrate. I was able to bring in only the MD5 function, therefore saving us from adding extra clutter to the code base. I have a [readme](https://github.com/hallahan/tangram-es/blob/8232af10e4c7ccc5582f790d6eeee29046d8d04e/core/include/hash-library/README.md) that goes into greater detail. I did manually test the hash values with several other MD5 generators and found that the output values matched with the same input.
## MBTiles Tile Task

You'll notice that reads and writes to the database are being done inside of an [MBTilesTileTask](https://github.com/hallahan/tangram-es/blob/8232af10e4c7ccc5582f790d6eeee29046d8d04e/core/src/tile/mbtilesTileTask.h). This class is a subclass of `DownloadTileTask`, and it is run a tile worker thread, so we can do our blocking call to the database without trouble.
## FontContext::fetch Segfault Bug

The way in which we are adding an MBTiles file to a data source is via a [scene update](https://github.com/hallahan/tangram-es/blob/8232af10e4c7ccc5582f790d6eeee29046d8d04e/core/src/tangram.cpp#L252-L256). One bug I kept running into is that if we have a url font resource in the scene.yaml, we get segfaults on android when `startUrlRequest` is called to fetch a remote font. This will always happen in a scene update unless we remove that remote font URL. See #978 .
### Compression

I held off on implementing compression of the vector data in the database for now. It's definitely something that I'd like to have on by default, but it isn't an essential functionality. I'm thinking it will be good to wait until we've decided on a compression library to work with multi-file scene archives. I'd like to use the same library to compress vector data here.
### Documentation

We definitely will want a tutorial and documentation to explain the usage of this feature. It will be important to explain how the `mbtiles` parameter behaves in the `scene.yaml`. Also, we'll want an example in the `tangram-android-examples` project.
### MBTiles Spec Merge

It sounds like there still needs to be more discussion to convince Mapbox to merge the pull request. We may need to do some tweaks to our implementation if the spec has to change.
### Unit Tests

It would be a good idea to include unit tests, particularly as this feature matures. I've used Catch in projects before, and it is should be pretty straightforward to add some tests with an MBTiles fixture. This should become more important when there is a Mapzen backend providing MBTiles files with data of various formats. Checking to make sure that we are reading and writing data in the format that we think it is will become important. Also, making sure that the data matches what is being stated in the metadata will become more important.
### iOS Objective C API

We will want to expose this functionality in the `TGMapViewController` as well. That should be simple.
### Mapzen MBTiles Metro Extracts

One exciting endeavor will be to include MBTiles with Mapzen's Metro Extracts. I believe that this will be a valuable resource for people to obtain offline maps.

cc/ @tallytalwar @blair1618 @pnorman @bcamper


nvkelso:
@hallahan We're hoping to have some support for offline tile packages thru Metro Extracts as you describe in 2016Q4, thanks for proposing the Tangram changes to support this on the client side!

hallahan:
Looks like the Linux setup on Travis is failing. I just tried it out on my Ubuntu 16.04 box, and it's working fine. 

Looking at SQLiteCpp's CMakeLists.txt, it looks like they add some special logic for Linux in the example build:

```
if (UNIX)
        target_link_libraries(SQLiteCpp_example1 pthread)
        if (NOT APPLE)
            target_link_libraries(SQLiteCpp_example1 dl)
        endif ()
        ...
```
https://github.com/SRombauts/SQLiteCpp/blob/b6512c4c6fbb964c6889387234d8a13a71b2a817/CMakeLists.txt#L237-L241

Would we want to put this in the `linux.cmake`? Looking at the `linux.cmake`, it seems to include the dynamic link library, but in a different way:

```
target_link_libraries(${EXECUTABLE_NAME}
    ${CORE_LIBRARY}
    -lcurl glfw
    # only used when not using external lib
    -ldl
    -pthread
    ${GLFW_LIBRARIES}
    ${OPENGL_LIBRARIES})
```

hjanetzek:
Hey Nicholas -- thanks for pushing offline support forward!
I just had a quick look over the changes and it looks all good so far.

At this point I think `DataSource` needs some restructuring that I want to discuss with the team when they are back next week.

To fix the linux build, adding `dl` to core (instead of tangram executable) should fix it.

hallahan:
🎉 Travis is happy now! 🎉 

@hjanetzek - Thanks for taking a look. Adding `dl` to the core CMake for UNIX and NOT MAC works, but I bet I could have specified this in a cleaner way. https://github.com/tangrams/tangram-es/pull/982/commits/6d7b7cbbdab1e5dc104ffad1ad8e12b97add0176

hallahan:
Does anyone have any thoughts on how to move forward? Is this pull request something Mapzen would like to merge?

tallytalwar:
@hallahan Most of us were out for last 2 weeks, including 1 week of all mapzen conference.

We will be having a discussion on this, this week and get back to you.

bcamper:
Yes, thanks for all your work here @hallahan. We are definitely interested, it's just been a matter of time available to think through the architecture (probably refactoring data sources, considering a plugin system, etc.). We expect to be able to support this one way or another though.

nvkelso:
We're also moving forward with offline vector tile package support via our Metro Extracts product. 

Please for @iandees to post a link to the New York City sample here.

hallahan:
I agree that some refactoring of the data source architecture might clean things up. In particular, some of the constructors are getting pretty gnarly.

core/src/data/dataSource.cpp
```diff
@@ -105,7 +107,12 @@ DataSource::~DataSource() {
 }
 
 std::shared_ptr<TileTask> DataSource::createTask(TileID _tileId, int _subTask) {
-    auto task = std::make_shared<DownloadTileTask>(_tileId, shared_from_this(), _subTask);
+    std::shared_ptr<DownloadTileTask> task;
+    if (hasMBTiles()) {
+        task = std::make_shared<MBTilesTileTask>(_tileId, shared_from_this(), _subTask);
```
tallytalwar: :+1:

core/src/tile/mbtilesTileTask.cpp
```diff
@@ -0,0 +1,105 @@
...
+ * we have rawData. If not, we query the SQLite MBTiles database for the rawData that we seek.
+ *
+ * @param _tileBuilder
+ */
+void MBTilesTileTask::process(TileBuilder &_tileBuilder) {
+    if (hasData()) {
+        // If the data did not come from the in-memory cache, it is new.
+        // It should be added to the MBTiles.
+        if (!dataFromCache) {
+            putMBTilesData();
```
tallytalwar: Generic question, how much of performance gain we get when we grab the date from cache, vs when we run the query to grab the tile data every time. Just debating on memory vs performance benefit here?

hjanetzek:
Hey @hallahan I've started the DataSource refactoring and also ported your MBTiles code over to the new interface at https://github.com/tangrams/tangram-es/pull/1015. Please have a look if you find something missing or have ideas for improvement.
https://github.com/tangrams/tangram-es/pull/1014

core/src/tile/mbtilesTileTask.cpp
hallahan: Good question. I implemented it as you see because I didn't want to change the in-memory functionality.

lygstate:
I suggest using the following schema:
```
sources:
    osm:
        type: GeoJSON
        # for posix
        url: mbtiles:///Users/njh/osm-data/mbtiles/winters-mapzen-geojson.mbtiles
        # for Windows
        #url: mbtiles://C:/Users/njh/osm-data/mbtiles/winters-mapzen-geojson.mbtiles
        max_zoom: 16
```

tallytalwar:
Probably make sense to close this PR in favor of #1015. #1015 already squashes all the commits in here into a single commit: b5c4dc1dcbb4e1137367b50bfe70683661c53a19.

Nice description though in this PR :D.

PR #987, opened 2016-09-26
Updates, Fixes, and Changes for iOS 10 and Xcode 8 framework

msmollin:
Most of the changes here lie within the xcode project file, and involve removing additional hard links to the iOS 9 SDK, so now the project file references `iPhoneOS.sdk` rather than `iPhoneOS9.3.sdk`. It also updates the project to be compliant with changes that occurred in the main codebase, specifically pulling the harfbuzz requirement out of alfons and up into the main dependencies, and the removal of the alfons-wrap library.

Previously, changes were made that exposed the internal instance of the Tangram `Map` object that TGMapViewController uses. The issue here is that by exposing that, it requires the framework implementation to then expose nearly the entire tangram-es public interface, since the Map object depends upon almost the entirety of the core interface to function. 

However, the underlying tangram-es core project uses relative paths to header files that the Xcode framework doesn't replicate. Merely adding the headers to be exposed via the framework's public headers isn't good enough because the include statements fail. The easy fix (included here) is to just remove the public `Map` reference and make it private for now. However, we should really address the relative path issues in the core framework, maybe as part of the ongoing build improvements.

### DIFF
--- a/ios/src/TGMapViewController.h
+++ b/ios/src/TGMapViewController.h
@@ -7,8 +7,6 @@
 //  Copyright (c) 2016 Mapzen. All rights reserved.
 //
 
-#include "tangram.h"
-
 #import <UIKit/UIKit.h>
 #import <GLKit/GLKit.h>
 
@@ -47,7 +45,6 @@ NS_ASSUME_NONNULL_END
 struct TileID;
 @interface TGMapViewController : GLKViewController <UIGestureRecognizerDelegate>
 
-@property (assign, nonatomic, nullable) Tangram::Map* map;
 @property (assign, nonatomic) BOOL continuous;
 @property (weak, nonatomic, nullable) id<TangramGestureRecognizerDelegate> gestureDelegate;
 
--- a/ios/src/TGMapViewController.mm
+++ b/ios/src/TGMapViewController.mm
@@ -6,6 +6,7 @@
 //  Updated by Matt Smollinger on 7/29/16.
 //  Copyright (c) 2016 Mapzen. All rights reserved.
 //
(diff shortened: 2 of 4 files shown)


ecgreb:
Does making the Tangram map instance private prevent client apps from interacting with it directly?

Our experience with the Android SDK has been that exposing the underlying `MapController` instance can be useful to allow for advanced use cases that require a higher degree of customization.

msmollin:
For Objective-C, yes it does prevent that. Swift can't call directly into C++ objects or functions without an objective-c interface, so it doesn't hamper that. Because of that, future work will expose the rest of the methods at the map controller level that would be available on the private `Map` instance, so ultimately it won't cause too many issues for all but the most advanced users.

blair1618:
Hiding the `Tangram::Map*` member of `TGMapViewController` is probably the right move. This member is the equivalent of the pointer to the native map object in `MapController` in the tangram-es Android SDK, not the `MapController` itself. We'll want to wrap all of the functions of `Tangram::Map` with more idiomatic signatures in Obj-C/Swift for `TGMapViewController`.

ecgreb:
:+1:

msmollin:
@blair1618 or others, any issues with merging this then?

blair1618:
@msmollin We're probably good since the travis build is passing, but I'll give it a try today on my machine and make sure nothing's out of order.

karimnaaji:
@msmollin This PR seems obsolete with https://github.com/tangrams/tangram-es/pull/1003, feel free to close this one and submit additional changes to the new PR if needed.

PR #996, opened 2016-10-04
Flip texture coordinate for polygon extrusion

tallytalwar:
- to sync with JS behavior

TINY TINY CHANGE!


blair1618:
LGTM

You know I bet this behaved the same as texture coordinates in JS when we first implemented it, but then we started flipping textures on load a little while ago and that reversed the meaning of vertical texture coordinates!

PR #999, opened 2016-10-05
Fix an infinite loop in DynamicQuadMesh for very large meshes

blair1618:
When texture batches are absent, the number of vertices considered 'textured' should
equal the number of vertices 'indexed'. The previous code would cause an infinite
loop of draw calls with zero vertices when meshes without texture batches had more
than RenderState::MAX_QUAD_VERTICES vertices.

Thanks @karimnaaji!


karimnaaji:
👍 LGTM

PR #1000, opened 2016-10-05
Scene 'global' refactor

blair1618:
- Add YamlPath struct
 - Use YamlPath to store locations of global values and nodes that reference them
 - Remove function to delete global references, to match behavior of tangram-js
 - Simplify SceneLoader functions for applying global values
 - Allow scene updates to operate on nodes that didn't previously exist (this was
   not intentional but seems to be the most natural behavior).
 - Fix issue where global references within sequences would not be updated
 - Simplify and clarify sceneUpdateTests

### DIFF
--- a/core/src/scene/styleContext.cpp
+++ b/core/src/scene/styleContext.cpp
@@ -85,7 +85,7 @@ StyleContext::~StyleContext() {
 
 // Convert a scalar node to a boolean, double, or string (in that order)
 // and for the first conversion that works, push it to the top of the JS stack.
-void pushYamlScalarAsJsObject(duk_context* ctx, const YAML::Node& node) {
+void pushYamlScalarAsJsPrimitive(duk_context* ctx, const YAML::Node& node) {
     bool booleanValue = false;
     double numberValue = 0.;
     if (YAML::convert<bool>::decode(node, booleanValue)) {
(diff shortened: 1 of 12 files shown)


tallytalwar:
Reviewing this now!

blair1618:
There's one more piece of clean-up work I'll add here in a few minutes. I'll do some profiling too and make sure this doesn't make anything too slow.

tallytalwar:
Yeah profiling for this should be good, I doubt it will be substantially slow compared to what it was before.

blair1618:
Profiled on an iPhone 4S. `SceneLoader::applyGlobals` is slower than I would like, but overall there's no significant change to total scene loading time. 

@tallytalwar review and merge when ready.

core/src/scene/styleContext.cpp
```diff
@@ -97,76 +97,64 @@ void pushYamlScalarAsJsObject(duk_context* ctx, const YAML::Node& node) {
...
+    }
+}
 
+void StyleContext::parseSceneGlobals(const YAML::Node& node) {
     switch(node.Type()) {
-        case YAML::NodeType::Scalar:
-            if (key.size() == 0) {
-                pushYamlScalarAsJsObject(m_ctx, node);
-                duk_put_prop_index(m_ctx, dukObject, seqIndex); // push
+    case YAML::NodeType::Scalar:
```
tallytalwar: Can you add an indent for the `case`s here, to follow our code styling?

core/src/scene/styleContext.cpp
```diff
@@ -97,76 +97,64 @@ void pushYamlScalarAsJsObject(duk_context* ctx, const YAML::Node& node) {
...
-                        duk_pop(m_ctx); //pop key
-                        // push property as a string
-                        duk_push_string(m_ctx, nodeValue.c_str());
-                        duk_put_prop_string(m_ctx, dukObject, key.c_str());
-                    }
-                } else {
-                    pushYamlScalarAsJsObject(m_ctx, node);
-                    duk_put_prop_string(m_ctx, dukObject, key.c_str());
-                }
+                pushYamlScalarAsJsPrimitive(m_ctx, node);
```
tallytalwar: Nicely done here.

tests/unit/sceneUpdateTests.cpp
```diff
@@ -12,147 +12,160 @@ using namespace Tangram;
...
-    updates.push_back({"cameras.perspective-camera.active", "true"});
-    // Tangram apply scene updates, reload the scene
-    SceneLoader::applyUpdates(*scene, updates);
-    root = scene->config();
-    SceneLoader::parseGlobals(root["global"], scene);
-    SceneLoader::applyGlobalRefUpdates(root, scene);
-    SceneLoader::applyGlobalProperties(root, scene);
-    REQUIRE(root["cameras"]["iso-camera"]["active"].Scalar() == "false");
-    REQUIRE(root["cameras"]["perspective-camera"]["active"].Scalar() == "true");
+// This was previously enforced but it didn't seem clearly useful or desirable,
```
blair1618: This test fails with the current code. My point here is that I don't think this is a bad thing to allow. There's not a clear parallel with tangram-js to compare with since the update interfaces are totally different.

core/src/scene/styleContext.cpp
blair1618: We don't indent these consistently, in `styleParam.cpp` alone you can see both indented and un-indented case labels. However, our "official" style is to _not_ indent case labels (https://github.com/tangrams/tangram-es/blob/master/.clang-format#L27)

tests/unit/sceneUpdateTests.cpp
```diff
@@ -12,147 +12,160 @@ using namespace Tangram;
...
+TEST_CASE("Apply scene update to a new map entry") {
+    // Setup.
+    Scene scene;
+    REQUIRE(SceneLoader::loadConfig(sceneString, scene.config()));
+    // Add an update.
+    std::vector<SceneUpdate> updates = {{"map.c", "new_value"}};
+    // Apply scene updates, reload scene.
+    SceneLoader::applyUpdates(scene, updates);
+    const Node& root = scene.config();
+    REQUIRE(root["map"]["c"].Scalar() == "new_value");
```
blair1618: I agree that we would ideally ensure that only the nodes we expect to change are actually changed, but that's a bit of a problem. What do we check? One node? Which one? Why only that one? All other nodes? This is a general problem with testing operations on yaml documents. One solution I've considered is string-ifying the modified document and comparing it to the output yaml you would expect, but there is no one "correct" way to write a yaml structure so it's a meaningless effort. 

There is an infinity of arbitrary truths that we could choose to evaluate for any of these functions. Since our tests can be read as a form of documentation, I'd prefer to keep them logically focused.

tests/unit/sceneUpdateTests.cpp
tallytalwar: A user might apply a scene update on a non-existent node (say a typo happens), what should the code do in that case?

blair1618:
I think all comments are addressed now.

tallytalwar:
LGTM. Will merge when travis goes green.

PR #1001, opened 2016-10-06
Fix tizen url worker

tallytalwar:
This partially solves the timeouts in the tizen unit tests.

There are more test failures/time outs in the `map_view` APIs, but those seemed to be linked to `readyCb` work!

### DIFF
--- a/tizen/src/urlWorker.cpp
+++ b/tizen/src/urlWorker.cpp
@@ -120,5 +120,8 @@ void UrlWorker::stop() {
     for (auto& worker : m_workers) {
         worker->join();
     }
+
+    m_workers.clear();
+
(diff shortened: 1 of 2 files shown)


tizen/src/urlWorker.cpp
```diff
@@ -34,7 +34,7 @@ void UrlWorker::run() {
     CURL* curlHandle;
 
     {
-        std::lock_guard<std::mutex> lock(m_mutex);
+        std::lock_guard<std::mutex> lock(m_mutexProxyAddress);
         curlHandle = curl_easy_init();
```
hjanetzek: I would call it ```m_mutexInitCurl```

Issue #917, opened 2016-08-17
Continue improving build system for iOS Framework


msmollin:

blair1618:
Continuation of discussions in https://github.com/tangrams/tangram-es/pull/891

PR #1003, opened 2016-10-06
Build Tangram ios framework

karimnaaji:
Testing a cmake configuration to build tangram ios framework.

It seems to be conventionally adopted to name frameworks with the first uppercase letter, so I think `Tangram` is a more suitable name over `tangram_framework`.

To build the binary framework run the following:
```sh
make cmake-ios-framework
open build/ios-framework/tangram.xcodeproj/
```
Then select the `Tangram` target in the target selector for generic ios devices:

screen shot 2016-10-06 at 10 04 52 am

The framework will be under `/build/ios/framework/lib/Debug/`.

Some advantages over a configurable xcodeproject: 
- No need to have to manually update the xcode project configuration, when adding/removing core library files or external dependencies files (the master xcode project is currently broken because of file removal from the alfons library after when it has been configured)
- Possibility for automated builds from continuous integration if invoked with xcodebuild

@msmollin please let me know if any of this works and give a a try before I go any further (adding explicit resources as fonts into the framework), so we can see if it is a possible alternative over configurable xcode project.

Here is a generated framework (without `.dSYM` symbols), I am quite unsure about its validity after incorporating it in another project:
[Tangram.framework.zip](https://github.com/tangrams/tangram-es/files/514115/Tangram.framework.zip)

Related issue https://github.com/tangrams/tangram-es/issues/917


tallytalwar:
@karimnaaji Can you also update the `README` when this is finalized? Thanks

karimnaaji:
Yes none of that is final, I just want to make sure it is a valid and usable framework before using this alternative.

msmollin:
Good first step. Couple things you need:
* There's no umbrella public header. If you take a look at the xcodeproj file I have in the `ios` folder, it has a public header that imports the TGMapViewController. This is needed to get swift imports to work correctly. It should match the name of the framework, so `Tangram.h`. Here's the umbrella header being used currently: https://github.com/tangrams/tangram-es/blob/master/ios/tangram_framework/tangram_framework.h

* Looks like `TGMapViewController.h` is currently scoped to project-level visibility, so it can't be imported. That'll need to be made public as well so objective-c folks can use it / import it directly.

karimnaaji:
Interesting, if I look at the configuration of the generated project I can see them under `Public`:

screen shot 2016-10-06 at 11 49 54 am

Is that what you would set as configuration here?

msmollin:
I tried the latest changes. It's now working and importing correctly in Objective-C projects, but not in Swift. I'll take a look in a bit to see what changed in the build script. I have some suspicions.

karimnaaji:
@msmollin added what is needed for bridging with swift here https://github.com/tangrams/tangram-es/pull/1003/commits/5ad2a8c2d2acede9dd535a6d8c6b0d7e00085dad.

Resources are also being added to the framework. Please give this a review when you have the time!

scripts/travis/script_build.sh
```diff
@@ -19,6 +19,8 @@ if [[ ${PLATFORM} == "ios" ]]; then
     # Build ios project
     echo "Building ios project (simulator)"
     make ios-sim
```
msmollin: Looks like you probably should disable the ios-sim build task here since it's causing travis builds to go too long and terminate the job.

karimnaaji:
@msmollin the last few commits are to test travis build for the framework, I will remove them. It seems to be failing because of code signing, we can add this later when you know more about code signing from CI.

msmollin:
I tested out the framework build process. It correctly imports into both Swift and Objective-C projects now. However, it fails with an error `The application bundle does not contain a valid identifier.` when xcode goes to copy the framework bundle over to the device (I know its the framework because if I make xcode not copy the framework to the device, I don't get this error... but then I get a crash because it can't find the framework!) This normally happens because something inside the application bundle is conflicting with some system-level directory. This can include any frameworks and such.

Just comparing the framework the manual xcode project generates versus this cmake one, I'm seeing two additional folders: `img` and `import` that are being generated. Might be worth seeing if not including those would make Xcode/iOS happy.

I also noticed that the cmake-generated framework is not including a `_CodeSignature` folder, which I don't believe is necessary but just wanted to note it incase we can't fix this issue otherwise.

toolchains/iOS.framework.cmake
```diff
@@ -47,7 +47,10 @@ set(FRAMEWORK_HEADERS
     ${PROJECT_SOURCE_DIR}/ios/framework/TangramMap.h
     ${PROJECT_SOURCE_DIR}/ios/src/TGMapViewController.h)
 
-add_library(${FRAMEWORK_NAME} SHARED ${SOURCES} ${HEADERS})
+add_bundle_resources(RESOURCES "${PROJECT_SOURCE_DIR}/scenes" ".")
```
karimnaaji: @msmollin Hm I see, Can you try the following:
- comment out this line
- regenerate the whole project
- build the framework again

This line injects the scene resources into the framework under `TangramMap/framework`, to see if your issue is still happening?

karimnaaji:
OK so the test we have done this morning comparing the framework you built and this one were helpful.

Concerning the issue `The application bundle does not contain a valid identifier` issue, it was a missing identifier from the `.plist` framework file, it's fixed from https://github.com/tangrams/tangram-es/pull/1003/commits/d9f623958b6461436ec864adf000fef0c05386ed. The run-time issue was more complex but the `RPATH` within the shared library were set statically:

```
▶ otool -D build/ios-framework/lib/Debug/TangramMap.framework/TangramMap 
build/ios-framework/lib/Debug/TangramMap.framework/TangramMap (architecture armv7):
/Users/karim/dev/tangram-es/build/ios-framework/lib/Debug/TangramMap.framework/TangramMap
```

OSX and iOS needs it to be relative using `@rpath` syntax (fixed by https://github.com/tangrams/tangram-es/pull/1003/commits/d6d5d429fa65e969bc172fec3cce8d36e46a1195):

```
▶ otool -D build/ios-framework/lib/Debug/TangramMap.framework/TangramMap 
build/ios-framework/lib/Debug/TangramMap.framework/TangramMap (architecture armv7):
@rpath/TangramMap.framework/TangramMap
``` 

So that when the application starts and loads the binary gets loaded into memory it does not crash looking to load the library from this static path.

Makefile
```diff
@@ -302,10 +302,10 @@ cmake-ios-framework:
 	cmake ../.. ${IOS_FRAMEWORK_CMAKE_PARAMS}
 
 ios-framework: cmake-ios-framework
-	xcodebuild -target ${IOS_FRAMEWORK_TARGET} -project ${IOS_FRAMEWORK_BUILD_DIR}/${IOS_FRAMEWORK_XCODE_PROJ}
+	xcodebuild -target ${IOS_FRAMEWORK_TARGET} -project ${IOS_FRAMEWORK_BUILD_DIR}/${IOS_FRAMEWORK_XCODE_PROJ} | xcpretty
 
 ios-sim: ${IOS_SIM_BUILD_DIR}/${IOS_XCODE_PROJ}
-	xcodebuild -target ${IOS_TARGET} -project ${IOS_SIM_BUILD_DIR}/${IOS_XCODE_PROJ}
+	xcodebuild -target ${IOS_TARGET} -project ${IOS_SIM_BUILD_DIR}/${IOS_XCODE_PROJ} | xcpretty
```
msmollin: Won't this cause issues if you don't have xcpretty installed on your machine?

karimnaaji: Isn't `xcpretty` a default OS X installed command tool when xcode command line tools are installed? Otherwise yes. I added this for the convenience of reading travis builds, outputs there are pretty verbose :/

msmollin: I wish. xcpretty is an excellent third party utility:
https://github.com/supermarin/xcpretty

karimnaaji: Too bad, will remove last commit once reviewed.

toolchains/iOS.framework.cmake
```diff
@@ -0,0 +1,91 @@
...
+    FRAMEWORK_VERSION ${FRAMEWORK_VERSION}
+    MACOSX_FRAMEWORK_IDENTIFIER com.mapzen.tangramMap
+    MACOSX_FRAMEWORK_INFO_PLIST ${IOS_FRAMEWORK_RESOURCES}
+    #VERSION 1.0.0
+    #SOVERSION 1.0.0
+    PUBLIC_HEADER "${FRAMEWORK_HEADERS}"
+    RESOURCE "${IOS_FRAMEWORK_RESOURCES}"
+    )
+
+set_xcode_property(${FRAMEWORK_NAME} CODE_SIGN_IDENTITY "iPhone Developer")
```
tallytalwar: We are currently testing removing this to see if travis builds the framework without code signing.

Locally I was able to build the framework without code signing and import it into a sample app. And I was able to code sign this sample app and deploy it on a device.

toolchains/iOS.framework.cmake
```diff
@@ -0,0 +1,91 @@
...
+
+set(FRAMEWORK_NAME TangramMap)
+set(FRAMEWORK_VERSION "1.0")
+
+set(BUILD_IOS_FRAMEWORK TRUE)
+
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}
+    -fobjc-abi-version=2
+    -fobjc-arc
+    -std=c++1y
```
tallytalwar: Does xcode clang allow `c++14` now? If so we can move from `c++1y`.

tallytalwar: Yup looks like `c++14` works now.

toolchains/iOS.framework.cmake
```diff
@@ -0,0 +1,91 @@
...
+    MACOSX_FRAMEWORK_IDENTIFIER com.mapzen.tangramMap
+    MACOSX_FRAMEWORK_INFO_PLIST ${IOS_FRAMEWORK_RESOURCES}
+    #VERSION 1.0.0
+    #SOVERSION 1.0.0
+    PUBLIC_HEADER "${FRAMEWORK_HEADERS}"
+    RESOURCE "${IOS_FRAMEWORK_RESOURCES}"
+    )
+
+set_xcode_property(${FRAMEWORK_NAME} CODE_SIGN_IDENTITY "iPhone Developer")
+set_xcode_property(${FRAMEWORK_NAME} ENABLE_BITCODE "YES")
```
tallytalwar: Just for comment and info gathering from @msmollin 

Hey @msmollin you mentioned on slack that, using bitcode is the new way of doing things, since we can take advantage of app thinning, etc etc. But you also said the size of the framework went from 17m to 78m for you! Both me and @karimnaaji have ~31m as the framework size. Can you check and explain how was it 78m for you?

toolchains/iOS.framework.cmake
karimnaaji: Yep will update this flag.

toolchains/iOS.framework.cmake
karimnaaji: https://travis-ci.org/tangrams/tangram-es/builds/169242554

This build was successful with `xcodebuild` option `CODE_SIGNING_REQUIRED=NO` see commit https://github.com/tangrams/tangram-es/commit/cb549ea4a54e6db81414090f7be91c409e5eb25a for details (we also need to use `xcpretty` to reduce log size otherwise travis is failing due to large log size, over 4Mb).

toolchains/iOS.framework.cmake
tallytalwar: This is cool. Also happy with the time ~8mins to build 5 architectures.

tallytalwar:
Waiting on `ios` travis builds to go green. Works all locally.

PR #1005, opened 2016-10-07
MarkerManager functions should gracefully return if scene loading failed

tallytalwar:
Even though this was a side effect of commenting out scene loading in the tizen plugin, it does make sense for marker manager to not crash (uninitialized members and nullptrs) when a scene loading fails.

### DIFF
--- a/core/src/marker/markerManager.cpp
+++ b/core/src/marker/markerManager.cpp
@@ -87,6 +87,9 @@ bool MarkerManager::setVisible(MarkerID markerID, bool visible) {
 }
 
 bool MarkerManager::setPoint(MarkerID markerID, LngLat lngLat) {
+
+    if (!m_scene) { return false; }
+
     Marker* marker = getMarkerOrNull(markerID);
     if (!marker) { return false; }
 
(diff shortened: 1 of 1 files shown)


tallytalwar:
@blair1618 can you give this a look. Thanks

core/src/marker/markerManager.h
```diff
@@ -68,7 +68,7 @@ class MarkerManager {
 
     DrawRuleMergeSet m_ruleSet;
     StyleContext m_styleContext;
-    std::shared_ptr<Scene> m_scene;
+    std::shared_ptr<Scene> m_scene = nullptr;
```
blair1618: Redundant, `shared_ptr` is initialized to `nullptr` in its constructor.

core/src/marker/markerManager.cpp
```diff
@@ -87,6 +87,9 @@ bool MarkerManager::setVisible(MarkerID markerID, bool visible) {
 }
 
 bool MarkerManager::setPoint(MarkerID markerID, LngLat lngLat) {
+
+    if (!m_mapProjection) { return false; }
```
blair1618: I'd prefer checking for `m_scene` in all these conditionals for consistency, though practically it shouldn't matter.

core/src/marker/markerManager.h
tallytalwar: Sorry I can not find the constructor... :mag:

tallytalwar:
Updated with requested changes.

PR #1007, opened 2016-10-07
Ensure that loadMapAsync triggers render when done

### LINKED ISSUE #963: Map view waits for an input event before rendering
This can be observed with the demo applications on OS X and Android (Linux does not exhibit this and iOS hasn't been checked yet). It looks like the Map is not fetching tiles or rendering until an input event of some kind. On OS X any input will trigger, even a mouse movement.

blair1618:
Resolves #963


hjanetzek:
`applySceneUpdates` needs the same change. Or we could instead add the renderRequest to `JobQueue::add`

blair1618:
@hjanetzek You're right, I've added the same change to `Map::applySceneUpdates`.

PR #1010, opened 2016-10-10
Fix crashes on teardown

tallytalwar:
Fixes 2 instances when tangram will crash on teardown:

1. An async task starts running when the `Map`'s destructor has been invoked by another thread. This causes certain states in the `Map::Impl` to be invalid and hence cause a crash. Plus, there is no point running these async scene loading tasks if map is being destroyed. (1724989)
2. Currently due to the order in which members of `Map::Impl` are destroyed, there are deleted references of `Impl` members which are used by the `jobs` of the `Map::Impl::JobQueue`. It makes sense for these jobs (`Map::Impl`'s jobs) to be run from within its destructor, to make sure all member references are valid during the running of these jobs. (e9737d5)

### DIFF
--- a/core/src/scene/filters.cpp
+++ b/core/src/scene/filters.cpp
@@ -122,7 +122,6 @@ int Filter::filterCost() const {
         // Most expensive filter should be checked last
         return 1000;
     }
-    assert(false);
     return 0;
 }
 
--- a/core/src/tangram.cpp
+++ b/core/src/tangram.cpp
@@ -62,15 +62,19 @@ public:
     Labels labels;
     std::unique_ptr<AsyncWorker> asyncWorker = std::make_unique<AsyncWorker>();
     InputHandler inputHandler{view};
-    TileWorker tileWorker{MAX_WORKERS};
-    TileManager tileManager{tileWorker};
-    MarkerManager markerManager;
 
     std::vector<SceneUpdate> sceneUpdates;
     std::array<Ease, 4> eases;
 
     std::shared_ptr<Scene> scene = std::make_shared<Scene>();
     std::shared_ptr<Scene> nextScene = nullptr;
+
+    // NB: Destruction of (managed and loading) tiles must happen
+    // before implicit destruction of 'scene' above!
+    // In particular any references of Labels and Markers to FontContext
+    TileWorker tileWorker{MAX_WORKERS};
+    TileManager tileManager{tileWorker};
+    MarkerManager markerManager;
 
     bool cacheGlState;
 
--- a/core/src/tangram.cpp
+++ b/core/src/tangram.cpp
@@ -60,7 +60,7 @@ public:
     JobQueue jobQueue;
     View view;
     Labels labels;
-    AsyncWorker asyncWorker;
+    std::unique_ptr<AsyncWorker> asyncWorker = std::make_unique<AsyncWorker>();
     InputHandler inputHandler{view};
     TileWorker tileWorker{MAX_WORKERS};
     TileManager tileManager{tileWorker};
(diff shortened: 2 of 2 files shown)


blair1618:
This appears to be a more complicated problem than originally suspected. 

In addition to the crashes seen on Linux, both `master` and this branch crash on OS X when I set the demo window to close immediately after creating the `Map`. The crash occurs in a different location though. It crashes in `~Map` -> `~Impl` -> `~JobQueue` where a lambda function from `loadSceneAsync` tries to dereference `impl`, which is a `unique_ptr` to null: https://github.com/tangrams/tangram-es/blob/master/core/src/tangram.cpp#L188

The `Map::Impl` object is held in a `unique_ptr` and destroyed by going out of scope. From examining the stack during the crash, I can see that the `~unique_ptr` implementation used in the OS X build works in two steps:

1. Set the raw pointer member to null
2. Delete the held object

This means that while members of `Impl` are being destroyed, any references to `Impl` via its `unique_ptr` (such as references captured in lambdas) will evaluate to null!

This doesn't happen on Linux because that demo is built using the GNU libstdc++ standard library (the OS X build uses LLVM libc++), which has a different implementation of `~unique_ptr`! Specifically, the two steps above occur in the opposite order in libstdc++ and libc++.

We can fix this crash on OS X with additional checks for `impl != nullptr`, but I would prefer to find a resolution for this that doesn't depend on implementation details of the standard library.

tallytalwar:
Tried the same with xcode 8, and got the same error on osx.

I agree, as discussed IRT, `(!impl)` check does resolve the issue on osx, but we should try to find a generic solution for both gcc and clang builds.

hjanetzek:
Would it help to remove the running of jobs from JobQueue destructor? If the destructor should do anything then it is to cancel/free the jobs in my opinion, with the requirement that the free function may not access any captured context.

hjanetzek:
I think the safest way to handle this would be to teardown all threads explicitly before implicit destruction of Impl members takes place.

I guess the crash on osx looks like:
```
1. ~Impl
2. explicit jobQueue.runJobs(); - ok
3. implicit ~AsyncWorker -> finishes scene loading task, adds job which access impl
4. implicit ~JobQueue ->  runs job with half destructed impl 
```

And, yes, the libstdc++ unique_ptr destructor implementation is nicer since the crash will be more reliable :)

tallytalwar:
Yeah I think the best place to tear these down will be from map destructor.

tallytalwar:
@hjanetzek https://github.com/tangrams/tangram-es/compare/fix-crash-on-teardow2 is exactly what I have. Can you push that commit here?

I think we still need the `valid` check since there is no point doing actual scene loading etc during a teardown, we can exit early.

core/src/tangram.cpp
```diff
@@ -90,16 +98,21 @@ static std::bitset<8> g_flags = 0;
...
+    impl->valid.store(false);
     TextDisplay::Instance().deinit();
     Primitives::deinit();
 }
 
 void Map::Impl::setScene(std::shared_ptr<Scene>& _scene) {
+
+    if (!valid.load()) { return; }
+
     {
```
hjanetzek: I think the `valid` check is prone to races - why calling `impl->setScene` when `impl` may be invalid here?

tallytalwar: Hmm Partially agree with this.
1. `valid` is atomic so should not race.
2. different implementations of unique_ptr makes `impl` valid or invalid. Its not guaranteed to be invalid because of that, as we saw yesterday.

tallytalwar:
Ok, these are good.

I can not reproduce any crashes now.

tallytalwar:
Going to merge this when travis goes green.

core/src/scene/filters.cpp
```diff
@@ -122,7 +122,6 @@ int Filter::filterCost() const {
         // Most expensive filter should be checked last
         return 1000;
     }
-    assert(false);
```
blair1618: :clap:

PR #1013, opened 2016-10-12
iOS framework bundle resources path resolution

karimnaaji:
Update platform layer for iOS to fallback to tangram framework bundle when main executable bundle path resolution fails. This allows us dynamically resolve the absolute path of the resources put within the `.framework` file.

### DIFF
--- a/ios/src/platform_ios.mm
+++ b/ios/src/platform_ios.mm
@@ -12,6 +12,7 @@
 #include "TGMapViewController.h"
 
 static TGMapViewController* viewController;
+static NSBundle* tangramFramework;
 NSURLSession* defaultSession;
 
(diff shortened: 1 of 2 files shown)


toolchains/iOS.cmake
```diff
@@ -1,6 +1,7 @@
 include(${CMAKE_SOURCE_DIR}/toolchains/iOS.toolchain.cmake)
 
 add_definitions(-DPLATFORM_IOS)
+add_definitions(-DTANGRAM_FRAMEWORK=\"\")
```
tallytalwar: Do we even need to define this for direct `iOS` builds?

karimnaaji: This is a leftover from a previous approach. I will remove.

PR #836, opened 2016-07-14
Fixes #240 (Adds Windows build with MSVC, no clang or gcc support )

lygstate:
Try to getting tangram-es works under win32.

PR #1015, opened 2016-10-14
MBTiles DataSource

hjanetzek:


hjanetzek:
Testing with OSM2Vectortiles http://osm2vectortiles.org/downloads/
shot-2016-10-18_13-19-20

hallahan:
This is awesome! I will be looking through this (and trying out with osm2vectortiles).

core/src/data/mbtilesDataSource.cpp
```diff
@@ -0,0 +1,422 @@
...
+        stmt.bind(3, y);
+
+        if (stmt.executeStep()) {
+            SQLite::Column column = stmt.getColumn(0);
+            const char* blob = (const char*) column.getBlob();
+            const int length = column.getBytes();
+
+            // OSM2Vectiles does not contain a 'compression' field
+            // Try deflate by default..
+            if (inflate(blob, length, _data) != 0) {
```
hallahan: 👍

hallahan: Maybe when we open the database we should look and see if there is a compression field. If there is no compression field, I agree to try deflate by default and fall back to identity.

core/src/util/inflateBuffer.h
```diff
@@ -0,0 +1,10 @@
+#pragma once
+
+#include <vector>
+#include <string.h>
+
+namespace Tangram {
+
+int inflate(const char* _data, size_t _size, std::vector<char>& dst);
```
hallahan: This is awesome! Perhaps we should have a deflate function as well for when we populate the cache? As I was developing, I noticed that the cache got large pretty quickly.

core/src/scene/sceneLoader.cpp
```diff
@@ -896,11 +899,34 @@ void SceneLoader::loadSource(const std::string& name, const Node& source, const
...
-    std::string url = source["url"].Scalar();
+    std::string url;
+    std::string mbtiles;
     int32_t minDisplayZoom = -1;
     int32_t maxDisplayZoom = -1;
     int32_t maxZoom = 18;
 
+    std::string mime;
+
+    if (type == "GeoJSON") {
```
hallahan: Perhaps we can discern the mime type automatically from the `RawDataSoure` by just returning the mime type of the `DataSource`? We will know the mime type of the `DataSource` based on the implementation of the `DataSource`. Then, you can put the logic of figuring out the mime type for raster sources in `RasterSource` instead of in here?

core/src/util/inflateBuffer.h
hallahan: Also, should `inflate` be in the Tangram global namespace?

core/src/data/mbtilesDataSource.cpp
hjanetzek: Sounds good!

core/src/data/mbtilesDataSource.cpp
```diff
@@ -0,0 +1,422 @@
...
+
+    // REPLACE INTO statement in map table
+    SQLite::Statement putMap;
+
+    // REPLACE INTO statement in images table
+    SQLite::Statement putImage;
+
+    MBTilesQueries(SQLite::Database& _db, bool _offline)
+        : getTileData(_db, "SELECT tile_data FROM tiles WHERE zoom_level = ? AND tile_column = ? AND tile_row = ?;"),
+          putMap(_db, _offline ? "REPLACE INTO map (zoom_level, tile_column, tile_row, tile_id) VALUES (?, ?, ?, ?);" : "SELECT 1;" ),
```
hjanetzek: This is a quick workaround I put in yesterday to test with our Mapzen tilepacks which do not have a `map` table - needs cleanup.

core/src/data/mbtilesDataSource.cpp
```diff
@@ -0,0 +1,422 @@
...
+
+
+namespace Tangram {
+
+/**
+ * The schema.sql used to set up an MBTiles Database.
+ *
+ * https://github.com/mapbox/node-mbtiles/blob/4bbfaf991969ce01c31b95184c4f6d5485f717c3/lib/schema.sql
+ */
+const char* SCHEMA = R"SQL_ESC(BEGIN;
```
hjanetzek: no, just missing a `static`..

core/src/util/inflateBuffer.h
hjanetzek: :+1:, also moving to a separate namespace

core/src/data/mbtilesDataSource.cpp
```diff
@@ -0,0 +1,422 @@
...
+
+void MBTilesDataSource::setupMBTiles() {
+    try {
+
+        auto mode = SQLite::OPEN_READONLY;
+
+        if (m_offlineMode) {
+            // Need to explicitly open a SQLite DB with OPEN_READWRITE
+            // and OPEN_CREATE flags to make a file and write.
+            mode = SQLite::OPEN_READWRITE | SQLite::OPEN_CREATE;
```
hjanetzek: yeah, there are a few modes to consider.  

1. only use existing tiles
2. try existing tiles, download and save
3. try download, fallback to existing tiles
4. try download and save to mbtiles, fallback to existing tiles

So far I intended to recreate the logic of the core-mbtiles branch. 

oh - in this case I just wanted to make sure not to overwrite mbtiles packs :)

core/src/data/mbtilesDataSource.cpp
```diff
@@ -0,0 +1,422 @@
...
+                                     std::string _mime, bool _offlineCache)
+    : m_name(_name), m_path(_path), m_mime(_mime), m_offlineMode(_offlineCache) {
+
+    m_worker = std::make_unique<AsyncWorker>();
+
+    setupMBTiles();
+}
+
+MBTilesDataSource::~MBTilesDataSource() {
+    // TODO teardown db?
```
hjanetzek: :+1:

core/src/data/dataSource.h
```diff
@@ -128,13 +132,10 @@ class DataSource : public std::enable_shared_from_this<DataSource> {
...
 
-    // URL template for requesting tiles from a network or filesystem
-    std::string m_urlTemplate;
-
-    std::unique_ptr<RawCache> m_cache;
-
     /* vector of raster sources (as raster samplers) referenced by this datasource */
     std::vector<std::shared_ptr<DataSource>> m_rasterSources;
+
+    std::unique_ptr<RawDataSource> m_sources;
```
hjanetzek: Yes, I'm also not pleased with the `RawDataSource` and `-DataSource` naming. I thought of renaming `DataSource` to `TileDataSource` since the result of `DataSource::parse` is a `TileData` object. Still open for suggestions and thinking about further refactoring of DataSource

core/src/scene/sceneLoader.cpp
hjanetzek: yes, this should be improved - I was not sure which way to go: when passing `mime` to MBTilesDataSource constructor `*Source::getMime(string url)` must be static.

core/src/data/dataSource.h
hallahan: `TileSource` sounds good.

scenes/scene.yaml
```diff
@@ -80,10 +88,11 @@ styles:
 sources:
     osm:
         type: MVT
-        url:  https://vector.mapzen.com/osm/all/{z}/{x}/{y}.mvt
+        #url:  https://vector.mapzen.com/osm/all/{z}/{x}/{y}.json
+        mbtiles: nyc.mbtiles
```
hallahan: Well, we might want to cache non-Mapzen sources too. What if I had a web service with vector tiles I wanted cached?

What if we had some sort of field in the metadata saying what the origin is? @pnorman might have some insight.

hallahan: Disable cross-origin tile sharing 😉

core/src/scene/sceneLoader.cpp
hallahan: Mime type for a tile source should be standard and well-known. Perhaps just have `*Source::getMime(string url)` return a constant string?

scenes/scene.yaml
hjanetzek: Having the tile url in the metadata would be an option to allow updating of externally created files. At the moment I'm adding the check to only write to the db when it contains `description:MBTiles tile container created by Tangram ES.` :)

lygstate:
Wonderful.

lygstate:
We do not need add the mbtiles attribute,
just support for a new URL schema,
url:  mbtiles://file_path;
That's would reduce much code effort.
And also suggest merge all the UrlWorker into a single place.
Do not implement them in platform code.
but in tangram-es/core directly.
I''ve done that in 
https://github.com/tangrams/tangram-es/pull/836

blair1618:
Hi @lygstate 

I don't think we want to invent a new non-standardized URL scheme. An 'mbtiles' scheme wouldn't be usable with existing libraries the way that 'http' and 'file' are. A separate 'mbtiles' entry also allows us to add optional caching in a backwards-compatible way.

The design of UrlWorker isn't strictly related to this PR, but I'd gladly review those changes separately. (By the way, #836 hasn't been forgotten - thanks for staying active in the project!)

hallahan:
Hi @hjanetzek - how is progress on this PR coming?

hjanetzek:
Hey @hallahan, I think the branch is in a good shape. DataSource refactoring needs to be reviewed first though and everybody is busy with things we need for the mobile SDKs - I hope we will get back to this soon.

scenes/scene.yaml
pnorman: Yes, you'll need to track that a mbtiles file has the same source if you want to avoid writing garbage to some other mbtiles file.

pnorman:
This includes code for md5 sums. Is there something already included in one of the other libraries required that could be used, like sha256?

How can we make sure we only write to mbtiles files which are written in a way we understand? There is no guarantee that we can write to an arbitrary mbtiles file, even if it is storing vector tiles from the same source, because they could implement the mbtiles tables interface in a different way.

bcamper:
Good questions. @blair1618 will make a proposal as to Tangram syntax for handling the different read/write/fallback scenarios.

hallahan:
Hey, how are things going? Any word on this PR and MBTiles support?

tallytalwar:
Yes, @hallahan. We were busy with a project which we finished in december. Will be giving this a good shot now, expect more action on this now :D.

Thanks for the patience and sorry for the delay on this.

matteblair:
I have not reviewed all of this yet, but so far it's looking mostly solid.

One problem I've noticed is that SQLiteCpp/sqlite3 seems to have its own implementation for opening and reading a file path. This has two results for us:

 - We need to resolve the mbtiles path from a scene file into an absolute path before we give it to sqlite (paths in the scene file might be absolute, but more often they are relative to the scene file's location).
 - Android applications currently can't bundle an mbtiles file. Android application bundles are effectively zip archives, and so there is no file path that identifies any individual file within the bundle. One could work around this by sneakily copying a bundled mbtiles database to a location on disk at run-time (given suitable permissions and with a start-up delay).

matteblair:
Configuration seems to still be an open question here. I see that suggestions have been made for new parameters in the scene file, which is the typical way that we introduce new features, but in this instance I feel that some of these configuration options go beyond "styling a scene" and enter the territory of application implementation. For example, the location of an mbtiles file will probably need to be different on Android and iOS because of their storage mechanisms (as in my previous comment). Even more generally, I don't know whether we would expect to ever support mbtiles files as a data source in the JS engine, though @bcamper may have thoughts here. 

With these considerations in mind, I would suggest that an interface for configuring MBTiles data sources/caches should be part of the tangram-es API and not part of the scene file.

nvkelso:
There are offline situations where this could be useful for Tangram JS, too.

tallytalwar:
We do have an tangram ES level API to add an MBTile DS in this PR. But yes it utilizes updating the scene ultimately adds a DS via scene yaml load!

If JS has a use case of having offline MBTile Datasource, I am fine keeping it in the scene file description only.

matteblair:
@nvkelso I can imagine that an offline data cache of some kind would be useful for tangram-js, but it is unclear to me whether a SQLite database file is a good fit for this purpose. It seems undesirable to add a SQL library to the JS engine for just this use. And even if tangram-js did support reading MBTiles files, the issue would remain that it would likely use a different configuration (file location, timeouts, etc) than, say, an Android application.

nvkelso:
@matteblair I recall @bcamper's opinion being similar.  But it remains a valid use case that I think we should either support out of the box or provide documentation around the suggested workaround (e.g. running a local tileserver reading from the SQLite DB and masquerading it as a local tile endpoint).

matteblair:
@nvkelso Totally! It is a valid use case. I think there was also talk of using zip archives for offline use - tangram-js already has a zip reader so that might get us to a solution faster.

@pnorman 's concern about different access modes to an MBTiles file seems sensible to me - we will want our interface to encourage/enforce safe usage patterns.

I can currently see two use patterns for MBTiles files:

1. Pre-downloaded data set, read only.

2. Data cache created by tangram, read+write.

As long as we keep these roles distinct, we can be confident that we won't accidentally write incompatible entries to a database. The files used for these roles could be specified in clearly different ways.

Are there other use patterns that I'm not thinking of?

matteblair:
As I'm thinking through using an MBTiles file as a cache mechanism, it occurs to me that we can only store one set of tiles in a single database (or else I misunderstand the MBTiles format). If that's true, then if we wanted to cache the data from a scene like Walkabout (which uses a vector tile set and a raster tile set) then we would need two separate MBTiles files to store the combined data. 

In general, it seems that we'll need one database file per network data source to effectively cache map data, which is a minor bummer.

core/src/data/mbtilesDataSource.cpp
```diff
@@ -0,0 +1,506 @@
...
+}
+
+void MBTilesDataSource::openMBTiles() {
+
+    try {
+        auto mode = SQLite::OPEN_READONLY;
+        if (m_cacheMode) {
+            // Need to explicitly open a SQLite DB with OPEN_READWRITE
+            // and OPEN_CREATE flags to make a file and write.
+            mode = SQLite::OPEN_READWRITE | SQLite::OPEN_CREATE;
```
tallytalwar: "- only write to MBTiles if the db was created by tangram" from the commit comments.. I should read it as "only write to MBTiles if the db was created by tangram with cacheMode true".

core/src/data/mbtilesDataSource.cpp
```diff
@@ -0,0 +1,506 @@
...
+
+    if (_task->rawSource == this->level) {
+
+        m_worker->enqueue([this, _task, _cb](){
+            TileID tileId = _task->tileId();
+
+            auto& task = static_cast<BinaryTileTask&>(*_task);
+            task.rawTileData = std::make_shared<std::vector<char>>();
+
+            getTileData(tileId, *task.rawTileData);
```
tallytalwar: Probably make sense to put this under an `if` block and check for validity.

core/src/data/mbtilesDataSource.cpp
```diff
@@ -0,0 +1,506 @@
...
+
+        } else if (m_offlineMode) {
+            LOGW("try fallback tile: %s, %d", _task->tileId().toString().c_str());
+
+            m_worker->enqueue([this, _task, _cb](){
+
+                auto& task = static_cast<BinaryTileTask&>(*_task);
+                task.rawTileData = std::make_shared<std::vector<char>>();
+
+                getTileData(_task->tileId(), *task.rawTileData);
```
tallytalwar: Probably make sense to put this under an `if` block and check for validity.

core/src/data/mbtilesDataSource.cpp
```diff
@@ -0,0 +1,506 @@
...
+
+            if (task.hasData()) {
+                LOGW("loaded tile: %s, %d", tileId.toString().c_str(), task.rawTileData->size());
+
+                _cb.func(_task);
+
+            } else if (next) {
+
+                // Don't try this source again
+                _task->rawSource = next->level;
```
tallytalwar: Hmm... by this it means the mbtile sounce will not be tried again, however if `cacheMode` is enabled, mbtiles database will get relevant tile data to be used next time? Or you want to fallback on NetworkDataSource's cache at that point?

tallytalwar:
@matteblair you are correct in saying that we will need one database file per network datasource for caching with mbtiles. It looks like its exactly what is implemented in this PR with when caching mode is enabled for an mbtiles datasource.

Way I see it being used for caching with all network datasources is by defining an MBTiles datasource for each network source and enabling `offline` and `caching` mode for the corresponding MBTiles datasource.

core/src/data/mbtilesDataSource.cpp
tallytalwar: Ignore this, makes sense now.

tallytalwar:
> I can currently see two use patterns for MBTiles files:
> 
> 1. Pre-downloaded data set, read only.
> 
> 2. Data cache created by tangram, read+write.
> 
> As long as we keep these roles distinct, we can be confident that we won't accidentally write incompatible entries to a database. The files used for these roles could be specified in clearly different ways.

So I think currently the core MbTileDatasource code handles this gracefully, and won't accidentally write incompatible entries to a db. 
Currently, how it is implemented via `SceneLoader` all MBTileDataSource will have read+write support. We will have to aptly exposed via an API, and let the user define the behavior when setting a MBTile Datasource.

core/src/data/mbtilesDataSource.cpp
matteblair: If we can't allocate an empty vector of bytes here then crashing seems like an appropriate response.

core/src/data/mbtilesDataSource.cpp
matteblair: Can you explain your question more?

core/src/data/mbtilesDataSource.cpp
tallytalwar: nopes, in that case (issues with file read from MBTiles most likely) one would want to fallback on the network `next` datasource.

core/src/data/mbtilesDataSource.cpp
tallytalwar: I just wanted to clarify my understanding of ""- only write to MBTiles if the db was created by tangram".

This is only when MBTileDataSource is created with `cacheMode` enabled.

matteblair:
I think `MBTilesDataSource` is able to follow the policy we want, but it's important that the user interface makes it easy to set up both offline sources and caches correctly, rather than relying on run-time checks.

tallytalwar:
Exactly my point :D

core/src/data/mbtilesDataSource.cpp
matteblair: Mmmm I think that if creating this object fails then we're _out of memory_ and should do nothing but exit the process.

matteblair: Or I don't know what you want to check the validity of?

core/src/data/rasterSource.h
```diff
@@ -36,7 +36,7 @@ class RasterSource : public TileSource {
     // TODO Is this always PNG or can it also be JPEG?
     virtual const char* mimeType() override { return "image/png"; };
 
-    void loadTileData(std::shared_ptr<TileTask> _task, TileTaskCb _cb);
+    virtual void loadTileData(std::shared_ptr<TileTask> _task, TileTaskCb _cb);
```
hjanetzek: Since c++11 one can use `override` to have the compiler check that a function actually overrides a virtual function of the base class. I think this is much better than using `virtual` in the derived class just as a hint for the programmer.

tallytalwar: Removed the commit, and lets try to make it consistent as and when one touches that appropriate code.

core/src/scene/sceneLoader.cpp
```diff
@@ -988,17 +1025,13 @@ void SceneLoader::loadSource(const std::string& name, const Node& source, const
...
                                                    minDisplayZoom, maxDisplayZoom, maxZoom,
                                                    options, generateMipmaps);
-    } else {
-        LOGW("Unrecognized data source type '%s', skipping", type.c_str());
     }
 
-    if (sourcePtr) {
-        _scene->tileSources().push_back(sourcePtr);
-    }
+    _scene->tileSources().push_back(sourcePtr);
```
hjanetzek: The source type is now checked at the top of the function.

matteblair:
If we were to use the scene file to configure MBTiles sources, the most sensible approach I can think of is to allow a new type of `sources` entry where the `url` field contains the location of a file with the `.mbtiles` extension and the `type` field is not used (tile format is deduced from the MBTiles metadata). e.g.
```yaml
sources:
  _my_source:
    url: /storage/emulated/0/mapdata/newyork.mbtiles
    max_zoom: 16
```
For the caching use pattern, we can use a new field for `sources` entries, tentatively called `cache`, that specifies the location of a file with the `.mbtiles` extension. The cache file is permitted to not yet exist, but the path to the file must exist. e.g.
```yaml
sources:
  _my_source:
    type: MVT
    url: https://tile.mapzen.com/mapzen/vector/v1/all/{z}/{x}/{y}.mvt
    url_params:
      api_key: xxxxxxx
    cache: /storage/emulated/0/mapdata/cache.mbtiles
```
Some things to note about this approach:
 - The paths in these examples are typical for local file storage on Android. I used these paths because unlike most file paths we specify in scene files, these cannot be relative paths from the scene file (at least on Android). The caching use pattern will require (different) absolute file paths on both Android and iOS. The application developer will also need to request any needed permissions to write a file to disk. This illustrates my concerns that configurations using this approach are _not portable_.
 -  The read/write access to MBTiles files should be evident with this syntax. The file in `url` is always read-only and the file in `cache:` is always read-write.
 - There is much more that could be specified about caching behavior, e.g. the timeout duration on the primary data source, the expiration of cached data, etc. Many of these parameters are application-dependent, they would differ between a mobile and desktop application. This makes me hesitate to put them in the scene file, which aims to be as platform-agnostic as possible.

vicb:
I'm really interested in using this feature so I'll add some comments - I don't know a lot about this project and some might not be relevant:

### Using .mbtiles files

> tile format is deduced from the MBTiles metadata

May be it could be good to add a `type: mbtiles` anyway for extensibility purpose ?

### cache 

```yaml
sources:
  _my_source:
    cache: /storage/emulated/0/mapdata/cache.mbtiles
```

App devs probably care about a fs backed cache, do they really care about where the cache will be stored, ie should the path be optional ?
Should you also provision a `type: mbtile` entry here for extensibility - it could probably be the default for now.

Anyway kudos for this projectand can't wait for the offline support to be merged !

matteblair:
Hi @vicb, thanks for the rapid feedback!

> May be it could be good to add a type: mbtiles anyway for extensibility purpose ?

I was thinking about this too. The reason I didn't suggest it is that the other options for `type` are vector tile data formats, whereas MBTiles is a database that can contain tiles in any of the other formats, so it's not really consistent.

> App devs probably care about a fs backed cache, do they really care about where the cache will be stored, ie should the path be optional ?

Fair point. The path will be important for some cases, e.g. if an app wants to use a temporary folder for the cache or be able to pull the file off the device later. For other cases it might not matter. I believe that we can access a reasonable "default" location for saving files on all of our supported platforms. To configure this we could make the `cache` field a mapping, allowing for more options later:
```yaml
cache: 
  url: /storage/emulated/0/mapdata/cache.mbtiles # optional!
```
or we could use a Boolean `true` to represent "unspecified caching":
```yaml
cache: true
```
Again, I see both of these as going beyond the intended purpose of the scene file :\

hjanetzek:
@matteblair sounds all pretty much good to me.  Could you add these changes?
I think it will generally makes sense to pass mbtiles options as `SceneUpdate` on scene loading (left to the application developer or Mapzen SDK) - No need to consider portability of mbtiles paths imo right now.  
Another thing to keep in mind/take care of: When mbtiles are used for raster sources they have to be registered via scene files since the initial tile sources are checked to build the shaders.

matteblair:
@hjanetzek Yep, I will start implementing these suggestions, beginning with the read-only usage since that seems the most clear.

vicb:
@matteblair 

```
sources:
  _my_source:
    url: /storage/emulated/0/mapdata/newyork.mbtiles
    max_zoom: 16
```

From the [docs](https://mapzen.com/documentation/tangram/sources/#type) the type is required w/o default.
I think that omitting the `type` here would harm extensibility - how would you implement a `mztiles` format for example ?
I understand that the tile themselves could be in different format but still you need to know how to read the container, don't you ?

matteblair:
@vicb Yep, `type` is currently required for all data sources but that could change with the addition of MBTiles sources. What I was proposing was that we use the file extension of the URL to signify that we are using an MBTiles source (or, in the future, `.mztiles` for a hypothetical MZTiles source).

The drawback I see to mandating `type: MBTiles` is that it is redundant information and it could invite errors where a `type` is given that conflicts with the URL (e.g. someone mistakenly uses the vector tile type for an MBTiles source). That said, with MVT tile sources we have a similar tradeoff, since MVT could be inferred from the file extension, and we still require a `type` for those sources. 

I now agree that it would be more consistent to require that MBTiles sources specify `type: MBTiles`.

matteblair:
I'm still working on making the user experience here make sense, these are my current thoughts.

When I'm using an existing MBTiles database as a data source, there should be no need to specify the format of the tiles in the database because the database itself holds that information as metadata. If I download an MBTiles file without knowing exactly what it contains, I am not sure how I would determine the format of its contents (short of running SQL queries on it). I wouldn't know what to put in the `type` parameter in this case. 

This led us to deciding on the `type: MBTiles` option, which lets tangram-es deduce the tile format from the MBTiles file itself. 

There are still cases where the user must know something about the tile format in the database. When I use an MBTiles file as a raster tile source, I must know that the contents are raster tiles so that I can set the `rasters` parameter of other data sources that use it. When tangram-es is initializing data sources it must also be aware that a source with `type: MBTiles` _might be_ a raster source, but also _might not be_, which is a little weird.

In code terms, the `type: MBTiles` approach also implies that the "type" of a tile source isn't known until after it is initialized, which is difficult with our current inheritance-based approach to tile sources. I'm currently working on resolving this by refactoring for composition instead of inheritance.

placz:
Just a couple thoughts since I'm interested in seeing this feature.

It sounds like while technically the more specific type can be inferred, there are some cases in the yaml where it would be an error to have a mismatch, such as the rasters example. Perhaps the type for mbtiles can be used as an assertion that the data source is this type rather than choosing a parser.  Thus, specifying 'type:MBTiles-raster' would assert that it's a raster and can be used as a raster in other sources.  'type:MBTiles' would not be able to be used as a raster in other sources.

matteblair:
Alright - here's where I've landed on this.

The scene file now supports only read-only MBTiles sources using existing files. When the `url` of a `sources` entry ends with the `.mbtiles` extension, it is recognized as an MBTiles source. The `type` field remains unchanged, that is it should be set to the type of vector or raster tile expected from the data source. For example:
```yaml
sources:
  _savedTiles: #This is an example MBTiles data source.
    type: TopoJSON #We expect TopoJSON data from the source.
    url: /path/to/my/file.mbtiles #Only absolute file paths supported for now.
    max_zoom: 16 #Other optional parameters are applied as usual.
```
I think we should merge this now to provide the core functionality and then add caching and additional configuration later on as we figure out usage.

tallytalwar:
Will give the updates a look in a bit and merge if all good.

core/src/scene/sceneLoader.cpp
```diff
@@ -954,29 +975,44 @@ void SceneLoader::loadSource(const std::shared_ptr<Platform>& platform, const st
...
+        url.find("{x}") != std::string::npos &&
         url.find("{y}") != std::string::npos &&
         url.find("{z}") != std::string::npos;
 
+    bool isMBTilesFile = false;
+    {
+        const char* extStr = ".mbtiles";
+        const size_t extLength = strlen(extStr);
+        const size_t urlLength = url.length();
+        isMBTilesFile = urlLength > extLength && (url.compare(urlLength - extLength, urlLength, extStr) == 0);
```
tallytalwar: Hmm I think this should be:
`url.compare(urlLength - extLength, extLength, extStr) == 0`

To check if substring [arg1, arg1+arg2) in `url` is same as arg3!

bcamper:
Agreed this is the right initial scope, thanks!

core/src/scene/sceneLoader.cpp
tallytalwar: Also probably make sense for `tiled` check here. If a url defines a `tiled` datasource (specified x,y,z) it can not be a mbtiles datasource.

core/src/data/mbtilesDataSource.cpp
```diff
@@ -0,0 +1,508 @@
...
+    SQLite::Statement putImage;
+
+    MBTilesQueries(SQLite::Database& _db, bool _cache)
+        : getTileData(_db, "SELECT tile_data FROM tiles WHERE zoom_level = ? AND tile_column = ? AND tile_row = ?;"),
+          putMap(_db, _cache ? "REPLACE INTO map (zoom_level, tile_column, tile_row, tile_id) VALUES (?, ?, ?, ?);" : ";" ),
+          putImage(_db, _cache ? "REPLACE INTO images (tile_id, tile_data) VALUES (?, ?);" : ";") {}
+
+};
+
+    MBTilesDataSource::MBTilesDataSource(std::shared_ptr<Platform> _platform, std::string _name,
```
matteblair: Fixed. That's what I get for editing in Xcode haha

core/src/scene/sceneLoader.cpp
matteblair: You are correct, fixed!

tallytalwar:
Will be merging this when CI builds are done.

matteblair:
Rebasing - hang tight

core/src/scene/sceneLoader.cpp
matteblair: Right - your comment about `tiled`. In the case of something like `url: /w/{x}/{y}/{z}.mbtiles` I would lean towards treating it as a path to an MBTiles file - curly braces are valid in paths and are somewhat common on Windows. I would not make it a recommended practice though ;)

tallytalwar: Sure, again good to document.

horvayev:
Does it work for Android and iOS? I was able to render map from mbtiles on Linux. But with the same scene.yaml I cann't make it work on Android. Im also using file picker to get path of mbtiles.

`sources:
    osm:
        type: MVT
        url: /sdcard/Download/prague.mbtiles
        max_zoom: 14`

Picker in code below returns correct path "/sdcard/Download/prague.mbtiles".
`        dialog.setDialogSelectionListener(new DialogSelectionListener() {
            @Override
            public void onSelectedFilePaths(String[] files) {
                List<SceneUpdate> updates = new ArrayList<SceneUpdate>();
                updates.add(new SceneUpdate("sources.osm.url", files[0]));
                view.getMapAsync(MainActivity.this, "sceneMbtiles.yaml", updates);
            }
        });

        dialog.show();`

rwrx:
Is this MBTiles sources support usable for offline maps on Android? Have you tested it? I am just asking, I didn't tried nor looked into it. Thank you.

tallytalwar:
Hi @rwrx

Using MBTiles as an offline source of map tiles is working on all platforms we support. There are other usecases of MBTiles discussed in the (merged) PR which are not fully supported.

To use a MBTiles set in your scene you can refer to [this](https://github.com/tangrams/tangram-es/pull/1015#issuecomment-278452805) comment above.

rwrx:
It is also possible to change that path to mbtiles file at runtime? I mean not only hardcode it into yaml scene file. And it is possible to have multiple mbtiles paths defined? I mean a use case where user has for example downloaded multiple mbtiles for countries and I need to show them on the map.

kueda:
Has this functionality been documented anywhere, or better yet implemented in a demo? Would be great to have a working example.

matteblair:
@kueda There's an open issue to add this in our docs repo: https://github.com/tangrams/tangram-docs/issues/192, thanks for the reminder to add this!

PR #1019, opened 2016-10-17
Point markers can draw in a specified order

blair1618:
Point markers can now have a `drawOrder` specified which determines their relative visual ordering among other point markers (in a painter's algorithm sense). 

This enables a tizen maps plugin feature: https://github.com/mapzen/tizen-maps-plugin-mapzen/issues/46


blair1618:
I believe this is working as desired now. I would still like to do some profiling with this change because I changed label sorting from `std::sort` to `std::stable_sort` which will have different performance characteristics.

blair1618:
Profiling on an iPhone 4S shows no measurable difference in performance.

Issue #1027, opened 2016-10-20
Objective-C/Swift Framework API for Tangram-ES


karimnaaji:
This issue will show Tangram-ES API progress of exposed functionalities through the tangram framework for Objective-C and Swift bindings.
- [x] Queue scene updates (#851)
- [x] Screen coords to lat/lon conversions (#859)
- [x] Delegate for event `onSceneLoaded`
- [x] Delegate for event `onMapReady` (#856)
- [x] Delegate for event `onFeatureSelected` (#855)
- [x] ~~Dynamic API Key update~~ (Should be done at SDK level) (#853)
- [x] Dynamic style replacement (#852)
- [x] Marker API
- [x] System fonts (#825)

cc @msmollin, @ecgreb

PR #1029, opened 2016-10-20
iOS API

karimnaaji:
This adds interface for the following:
- Load scene and load scene async
- Scene updates
- Feature picking
- Long/Lat to screen position
- Map view delegate for feature picking and scene loading callbacks

https://github.com/tangrams/tangram-es/issues/1027


ios/src/TGMapViewController.mm
```diff
@@ -23,26 +25,116 @@ @interface TGMapViewController ()
...
+    if (items.size() == 0) { return; }
+
+    const auto& result = items[0];
+    const auto& properties = result.properties;
+    CGPoint position = CGPointMake(result.position[0], result.position[1]);
+
+    NSMutableDictionary* dictionary = [[NSMutableDictionary alloc] init];
+
+    for (const auto& item : properties->items()) {
+        NSString* key = [NSString stringWithUTF8String:item.key.c_str()];
```
msmollin: Are you guaranteed to have valid items/properties/results the whole way here? If any of these are null/nil at any point you'll get a crash trying to insert into the dictionary here.

karimnaaji:
@msmollin please give this a review.

ios/src/TGMapViewController.mm
```diff
@@ -9,11 +9,13 @@
 
 #import "TGMapViewController.h"
 #include "platform_ios.h"
+#include "data/propertyItem.h"
 #include "tangram.h"
 
 @interface TGMapViewController ()
 
-@property (nullable, strong, nonatomic) EAGLContext *context;
+@property (nullable, nonatomic) NSString* scenePath;
```
msmollin: This property declaration should be `(nullable, copy, nonatomic)` to avoid the NSString being changed out from underneath you by external references.

msmollin: Related to Strings, Apple has generally moved to using `NSURL` for paths on disk. The use of strings is considered a legacy implementation at this point. Any issue with change this to using a NSURL and then updating the delegate to use a URL for the path?

One thing to note is you'll want to use the `fileUrl` versions. Otherwise, NSURL defaults to HTTP-esque semantics and will attempt to percent encode the path as if it was an HTTP URL, which is generally bad for file paths :)

ios/demo/src/ViewController.h
```diff
@@ -0,0 +1,15 @@
...
+//  ViewController.h
+//
+//  Created by Karim Naaji on 10/12/16.
+//  Copyright © 2016 Karim Naaji. All rights reserved.
+//
+
+#import <UIKit/UIKit.h>
+#import "TGMapViewController.h"
+
+@interface ViewController : TGMapViewController
```
msmollin: You should name this something other than "ViewController" as that's *really* close to apple's stuff. Maybe MapViewController

ios/src/TGMapViewController.h
```diff
@@ -46,31 +47,50 @@ struct TileID;
...
+
+- (void)queueSceneUpdate:(NSString*)componentPath withValue:(NSString*)value;
+
+NS_ASSUME_NONNULL_END
+
+- (void)pickFeaturesAt:(CGPoint)screenPosition;
+
+- (void)requestRender;
+
+- (CGPoint)lngLatToScreenPosition:(TGGeoPoint)lngLat;
```
msmollin: These can probably be assumed non-null as well, due to the fact you're returning a fully formed CGPoint/TGGeoPoint with NAN as the entries, although see the review comments on that.

ios/src/TGMapViewController.mm
```diff
@@ -23,26 +25,116 @@ @interface TGMapViewController ()
...
+}
+
+- (void)requestRender {
+    if (!self.map) { return; }
+
+    self.renderRequested = YES;
+}
+
+- (CGPoint)lngLatToScreenPosition:(TGGeoPoint)lngLat {
+    static const CGPoint nullCGPoint = {(CGFloat)NAN, (CGFloat)NAN};
```
msmollin: Hmmmmm. I'm not sure about this approach. I generally prefer if the function "fails", you should return `nil`, which is effectively what you're doing here but with values of NAN instead, which means now you have to check the internal values before using them.

ios/src/TGMapViewController.mm
karimnaaji: `NSString` type is ok for tangram, we perform path resolution within the library for bundle and urls.

ios/src/TGMapViewController.mm
karimnaaji: Returning `nil` would mean either returning an objective-c object or a pointer to a `CGPoint` (CGPoint is a c-type struct). I also prefer the `nil` option but here `NaN` seems to be the only alternative if we want to keep on using `CGPoint`.

ios/src/TGMapViewController.h
karimnaaji: Those are c-type structs, assuming them non-null with the macro doesn't seem to make too much sense since those value would get copied.

karimnaaji:
@msmollin changes updated, please give this another look.

ios/src/TGMapViewController.mm
blair1618: We _could_ use `NSURL` here and just get the string representation for calls into C++. This variable often refers to files inside the current application bundle, is that easy with `NSURL`?

karimnaaji: The problem is that even if we use `NSURL` from the public interface and store it here, we would only end up using `[NSURL absolutePath]` and convert this to a `char*` string for use in the Tangram API. Then, later on this path would get resolved by the iOS platform layer using the bundle resolution; where we would loose what this `NSURL` object has to offer for tangram.

Considering what `NSURL` can do and what we would end up using within this wrapper, it seems that we can keep on using `NSString`. This wouldn't stop the user from using it at a higher level (in the SDK for example).

msmollin: @blair1618 indeed bundle resolution was what I was thinking of - the `NSBundle` class provides access to the main application bundle and provides a `NSURL` set of methods and an older implementation using `NSStrings`. To be clear the string-returning functions aren't deprecated or anything. It's just common to prefer `NSURL` as it provides more safeguards and also NSURL is used by other processes to securely access bundles outside of the application sandbox for extensions. 

I see here https://github.com/tangrams/tangram-es/blob/tangram-ios-api/ios/src/platform_ios.mm#L65 that we're also using the legacy string methods for doing the bundle resolution that @karimnaaji mentioned, which looks like something I didn't manage to get to review before I went on vacation, otherwise I probably would've said the same thing there.

This is fine for now. I will probably add an extension on top of this in the Mapzen iOS SDK to add NSURLs as an option for passing along scene file locations if we want to go that route.

msmollin:
@karimnaaji changes look good.

PR #1032, opened 2016-10-21
iOS demo app update

karimnaaji:
Same as https://github.com/tangrams/tangram-es/pull/1012 but now uses the tangram binary framework.

- [x] Update README instructions.


msmollin:
@karimnaaji rebase against master here and I'll review this.

karimnaaji:
Rebased, @blair1618 please also give a look at build system changes.

msmollin:
So on my machine, I got this error:
```Check dependencies
"tangram" isn't code signed but requires entitlements. It is not possible to add entitlements to a binary without signing it.
Code signing is required for product type 'Application' in SDK 'iOS 10.0'```

My guess is trying to auto-build the app target is making this unhappy. We could probably avoid this by just building the framework and not the app target, as anyone running the tangram IOS demo is going to have to open xcode anyway and build and run, at which point code signing can be resolved IF they want to target a device.

msmollin:
Discussed this with @karimnaaji and apparently I forgot to reset my origin branch on my local machine due to the force push. Fixed that and now everything works. @blair1618 can you review this for build configuration changes?

karimnaaji:
Merging once travis go green.

PR #1033, opened 2016-10-21
iOS fix gesture delegate

karimnaaji:
The implementation of our gesture delegate/responders is diverging from our gesture listeners on Android.
- _Tap_ and _Double Tap_ gestures should not handle their corresponding native`handleTapGesture` and `handleDoubleTapGesture`. (Common use case pattern where this break is feature selection - We couldn't plug a delegate to the API and select a feature without changing the map location).
- _Pan_, _Pinch_, _Rotation_ and _Shove_ should call their corresponding native `handle*` if no delegate is provided.
- For convenience we should add the `TGMapViewController` view as parameters to the delegate methods to be able to perform more action on the view once a gesture even has been received.
- [x] add handle for long press gesture.


msmollin:
@karimnaaji rebase against master and I'll review this.

ios/demo/src/MapViewController.m
```diff
@@ -1,28 +1,74 @@
...
+{
     [super viewDidLoad];
-    // Do any additional setup after loading the view, typically from a nib.
-}
 
+    self.mapViewDelegate = [[MapViewControllerDelegate alloc] init];
+    self.gestureDelegate = [[MapViewControllerRecognizerDelegate alloc] init];
 
-- (void)didReceiveMemoryWarning {
+    [super loadSceneFileAsync:@"https://tangrams.github.io/walkabout-style/walkabout-style.yaml"];
```
msmollin: Instead of starting this here, you should override `viewWillAppear:` and start the network load there. That way you avoid any issues where the device is slow to finish initializing the graphics stack.

ios/src/TGMapViewController.h
```diff
@@ -56,6 +56,7 @@ struct TileID;
 @property (assign, nonatomic) float zoom;
 @property (assign, nonatomic) float rotation;
 @property (assign, nonatomic) float tilt;
+@property (readonly, nonatomic) CGFloat pixelScale;
```
msmollin: nitpick, you should also declare `assign` here. I like to be explicit with memory management designation even though the compiler will figure out that CGFloat is a struct and should be assign anyway.

ios/src/TGMapViewController.h
```diff
@@ -33,12 +33,12 @@ NS_ASSUME_NONNULL_BEGIN
...
-- (void)recognizer:(UIGestureRecognizer *)recognizer didRecognizePanGesture:(CGPoint)location;
-- (void)recognizer:(UIGestureRecognizer *)recognizer didRecognizePinchGesture:(CGPoint)location;
-- (void)recognizer:(UIGestureRecognizer *)recognizer didRecognizeRotationGesture:(CGPoint)location;
-- (void)recognizer:(UIGestureRecognizer *)recognizer didRecognizeShoveGesture:(CGPoint)location;
+- (void)mapView:(TGMapViewController *)view recognizer:(UIGestureRecognizer *)recognizer didRecognizeSingleTap:(CGPoint)location;
+- (void)mapView:(TGMapViewController *)view recognizer:(UIGestureRecognizer *)recognizer didRecognizeDoubleTap:(CGPoint)location;
+- (void)mapView:(TGMapViewController *)view recognizer:(UIGestureRecognizer *)recognizer didRecognizePanGesture:(CGPoint)location;
+- (void)mapView:(TGMapViewController *)view recognizer:(UIGestureRecognizer *)recognizer didRecognizePinchGesture:(CGPoint)location;
+- (void)mapView:(TGMapViewController *)view recognizer:(UIGestureRecognizer *)recognizer didRecognizeRotationGesture:(CGPoint)location;
+- (void)mapView:(TGMapViewController *)view recognizer:(UIGestureRecognizer *)recognizer didRecognizeShoveGesture:(CGPoint)location;
```
msmollin: 👍  very iOS friendly.

ios/demo/src/MapViewController.m
karimnaaji: Ok perfect, I'll override this one. 

Concerning `update:` on the view, I think that `TGMapViewController` should expose publicly the update method so override of this view can directly call `[super update]`. If that makes sense I'll also update that one.

msmollin: I am not sure I have a good enough understanding to have an opinion on that topic. I don't believe it would pose an issue, but @blair1618 might have a better idea.

karimnaaji: If we don't expose it, an override of the TGMapViewController `update:` from the implementation view won't let the user able to propagate that update to tangram.

ios/src/TGMapViewController.h
karimnaaji: Removed that one, figured the user won't have to deal with that from the discussion we had in the meeting, so it's always in logical pixel from the public interface.

PR #1037, opened 2016-10-26
Small sceneloader cleanups

hjanetzek:


core/src/tangram.cpp
```diff
@@ -180,27 +187,32 @@ void Map::loadScene(const char* _scenePath, bool _useScenePosition) {
 void Map::loadSceneAsync(const char* _scenePath, bool _useScenePosition, MapReady _platformCallback, void *_cbData) {
     LOG("Loading scene file (async): %s", _scenePath);
 
+    std::shared_ptr<Scene> nextScene;
     {
         std::lock_guard<std::mutex> lock(impl->sceneMutex);
         impl->sceneUpdates.clear();
         impl->nextScene = std::make_shared<Scene>(_scenePath);
         impl->nextScene->useScenePosition = _useScenePosition;
+        nextScene = impl->nextScene;
```
tallytalwar: What's the conceptual difference here? In both the cases `impl->nextScene` is passed down to the async task.

core/src/tangram.cpp
```diff
@@ -180,27 +187,32 @@ void Map::loadScene(const char* _scenePath, bool _useScenePosition) {
...
-            jobQueue.add([scene, ok, _platformCallback, _cbData, this]() {
+            impl->jobQueue.add([nextScene, ok, _platformCallback, _cbData, this]() {
                     {
                         std::lock_guard<std::mutex> lock(impl->sceneMutex);
-                        if (scene == impl->nextScene) {
+                        if (nextScene == impl->nextScene) {
                             impl->nextScene.reset();
-                        } else { return; }
+                        } else {
+                            // loadScene[Async] was called in the meantime.
```
tallytalwar: :+1:

core/src/tangram.cpp
tallytalwar: Ahh ... I see the race!

Issue #457, opened 2016-01-08
Android native debugging toolchain has CMake errors


blair1618:
Trying to build the android-debug target currently results in a CMake configuration error:
```
CMake Error at toolchains/android.gdb.cmake:83 (get_property):
  The LOCATION property may not be read from target "tangram".  Use the
  target name directly with add_custom_command, or use the generator
  expression $<TARGET_FILE>, as appropriate.

Call Stack (most recent call first):
  toolchains/android.cmake:57 (android_ndk_gdb_debuggable)
  CMakeLists.txt:60 (include)
```
This is due to a policy change in CMake 3.0: https://cmake.org/cmake/help/v3.3/policy/CMP0026.html

softlxf:
I inserted following codes to make it work temporarily.
```
    if(POLICY CMP0026)
        cmake_policy(SET CMP0026 OLD)
    endif()
```

hjanetzek:
Thank you! @softlxf

msgrasser:
Experienced the same issue on 0.4.1-android tag last night and dropped the code above into `toolchains/android.gdb.cmake` just above line 83; glad this note happens to be here, but it would be really great to see @softlxf's suggestion incorporated into the repo!

hjanetzek:
finally.. android debug builds are back

PR #1039, opened 2016-10-26
fix GL_CHECK with debug builds

hjanetzek:
I wonder how this worked previously for debug builds..


hjanetzek:
https://github.com/tangrams/tangram-es/issues/457

karimnaaji:
LGTM

PR #1044, opened 2016-10-28
iOS markers API

karimnaaji:
Add markers to public framework API and objective-c classes for polygon, points and polylines.

Updated demo app with example usage.

https://github.com/tangrams/tangram-es/issues/1027


karimnaaji:
@msmollin you can start review this.

PR #1053, opened 2016-11-01
Return unormalized position in feature queries

karimnaaji:


karimnaaji:
@hjanetzek can you look at this?

blair1618:
Convo IRL: There's a potential inconsistency in the screen position of a feature when the query is requested and when it is completed - the map could move between those events. Let's approach this problem later!

PR #1055, opened 2016-11-01
iOS markers api (rebased)

karimnaaji:
Rebase of https://github.com/tangrams/tangram-es/pull/1044

### DIFF
--- a/ios/src/TGGeoPoint.h
+++ b/ios/src/TGGeoPoint.h
@@ -1,0 +1,27 @@
+//
+//  TGGeoPoint.h
+//  tangram
+//
+//  Created by Karim Naaji on 10/27/16.
+//
+//
+
+#ifndef TGGeoPoint_h
+#define TGGeoPoint_h
+
+struct TGGeoPoint {
+    double longitude;
+    double latitude;
+};
+
+typedef struct TGGeoPoint TGGeoPoint;
+
+static inline TGGeoPoint TGGeoPointMake(double lat, double lon)
+{
+    TGGeoPoint p;
+    p.latitude = lat;
+    p.longitude = lon;
+    return p;
+}
+
+#endif /* TGGeoPoint_h */
--- a/ios/src/TGGeoPoint.h
+++ b/ios/src/TGGeoPoint.h
@@ -6,8 +6,7 @@
 //
 //
 
-#ifndef TGGeoPoint_h
-#define TGGeoPoint_h
+#pragma once
 
 struct TGGeoPoint {
     double longitude;
@@ -24,4 +23,3 @@ static inline TGGeoPoint TGGeoPointMake(double lat, double lon)
     return p;
 }
 
-#endif /* TGGeoPoint_h */
--- a/ios/src/TGGeoPolygon.mm
+++ b/ios/src/TGGeoPolygon.mm
@@ -1,0 +1,72 @@
+//
+//  TGGeoPolygon.mm
+//  tangram
+//
+//  Created by Karim Naaji on 10/27/16.
+//
+//
+
+#import "TGGeoPolygon.h"
+
+#include <vector>
+
+using Ring = std::vector<TGGeoPoint>;
+
+@interface TGGeoPolygon () {
+    std::vector<Ring> polygons;
+    unsigned int points;
+}
+
+@end
+
+@implementation TGGeoPolygon
+
+- (id)init
+{
+    self = [super init];
+
+    if (self) {
+        points = 0;
+    }
+
+    return self;
+}
+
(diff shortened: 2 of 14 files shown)


ios/src/TGGeoPolygon.mm
```diff
@@ -0,0 +1,72 @@
...
+//
+
+#import "TGGeoPolygon.h"
+
+#include <vector>
+
+using Ring = std::vector<TGGeoPoint>;
+
+@interface TGGeoPolygon () {
+    std::vector<Ring> polygons;
```
karimnaaji: Yes, that's a good call, it seems that we can change that without modifying the current interface of this class and we wouldn't have to rearrange the data before giving it to `markerSetPolygon`.

karimnaaji:
@blair1618 , @msmollin , comments have been addressed, please give this a last eye and merge when ready.

msmollin:
LGTM - waiting on @blair1618 before merging.

ios/src/TGGeoPoint.h
```diff
@@ -0,0 +1,27 @@
+//
+//  TGGeoPoint.h
+//  tangram
+//
+//  Created by Karim Naaji on 10/27/16.
+//
+//
+
+#ifndef TGGeoPoint_h
+#define TGGeoPoint_h
```
blair1618: Is this the only header here that uses include guards? If we need these then we should prefer `#pragma once`, but if we can use `#import` for headers on iOS then it seems that these shouldn't be needed.

ios/src/TGGeoPolygon.mm
```diff
@@ -0,0 +1,80 @@
...
+//
+//
+
+#import "TGGeoPolygon.h"
+
+#include <vector>
+
+@interface TGGeoPolygon () {
+    std::vector<TGGeoPoint> coordinates;
+    std::vector<int> rings;
```
blair1618: @msmollin Out of curiosity, is there a CF/NS type that does the job of `std::vector` in Objective-C land? (i.e. a dynamic array with contiguous memory)

msmollin: That would be `NSMutableArray` but it can't hold `int` so you would have to convert to using `NSNumber` which is probably not something you'd want. You could always get fancy with Objective-C++ and use C++ arrays like `int *rings = new int[...]` but given how we're using them that might not be advantageous. Also I can't remember if that style of dynamic array ensures contiguous memory allocation.

ios/src/TGGeoPoint.h
karimnaaji: It's just for portability. I used the `#ifndef` syntax to make it c-compliant, if someones uses `include` syntax in objective-c for c headers this would still work without using objective-c++ with modern c++ syntax. 
Also relevant: http://stackoverflow.com/questions/23696115/is-pragma-once-part-of-the-c11-standard#23699893

blair1618: As far as I know, `#pragma once` works in C for every compiler I've ever heard of - have you found that to be untrue?

ios/src/TGGeoPolygon.mm
blair1618: Weeeird. How would one create a "dynamic array" of integers in Objective-C then?

ios/src/TGGeoPoint.h
msmollin: FWIW, this is iOS specific code, but even on Mac, LLVM/Clang is the standard compiler toolchain and pragma once is supported. GCC is "supported" as well, but not installed anymore by xcode, which also supports pragma once. So, your portability requirements are limited.

That said, if you look at CoreFoundation code, it's littered with `#ifndef` statements that I'm assuming predate GCC support of pragma once, and definetly predate clang. So, using it is not without precedent, but I wouldn't concern yourself highly with portability in this particular instance.

ios/src/TGGeoPolygon.mm
msmollin: Convert them to NSNumbers and use NSMutableArray if you want to stay purely within the scope of foundation classes. There are several shorthand ways of generating NSNumbers, like `NSNumber *num = @9;`. So you could do something like `NSMutableArray *array = @[@9, @10];` 

That said, a major selling point of Objective-C is its interoperability with C and C++, so you don't have to confine yourself like that.

ios/src/TGGeoPoint.h
karimnaaji: I might have used old versions of GCC in the past where this wasn't supported, but it seems that we're good with with the xcode toolchain on that end.

Udpated for `#pragma` syntax.

PR #1056, opened 2016-11-02
Android Marker API

sarahlensing:
- Adds native marker method declarations to `MapController`
- Creates methods for adding/removing markers to `MapController`
- Adds a new class, `Marker`, for customizing point, polyline, polygon, bitmap markers

### DIFF
--- a/android/tangram/src/com/mapzen/tangram/MapController.java
+++ b/android/tangram/src/com/mapzen/tangram/MapController.java
@@ -658,8 +658,8 @@ public class MapController implements Renderer {
      */
     public Marker addMarker() {
         checkPointer(mapPointer);
-        long pointer = nativeMarkerAdd(mapPointer);
-        return new Marker(mapView.getContext(), pointer, this);
+        long markerId = nativeMarkerAdd(mapPointer);
+        return new Marker(mapView.getContext(), markerId, this);
     }
 
     /**
@@ -669,7 +669,7 @@ public class MapController implements Renderer {
      */
     public boolean removeMarker(Marker marker) {
         checkPointer(mapPointer);
-        checkPointer(marker.markerId);
+        checkId(marker.markerId);
         return nativeMarkerRemove(mapPointer, marker.markerId);
     }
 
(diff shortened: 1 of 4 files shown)


android/tangram/src/com/mapzen/tangram/Marker.java
```diff
@@ -0,0 +1,179 @@
...
+import com.mapzen.tangram.geometry.Polygon;
+import com.mapzen.tangram.geometry.Polyline;
+
+/**
+ * Class used to display points, polylines, and bitmaps dynamically on a map. Do not create one of
+ * these objects directly, instead use {@link MapController#addMarker()}.
+ */
+public class Marker {
+
+    Context context;
```
ecgreb: Private scope?

android/tangram/src/com/mapzen/tangram/Marker.java
```diff
@@ -0,0 +1,179 @@
...
+    long markerId = 0;
```
ecgreb: How does `FeaturePickListener` work with the new marker API? Is the `markerId` returned as a property? Do we need a public getter for `markerId` so we can match which feature is clicked in the callback?

android/tangram/src/com/mapzen/tangram/Marker.java
```diff
@@ -0,0 +1,179 @@
...
+        this.context = context;
+        this.markerId = markerId;
+        this.map = map;
+    }
+
+    /**
+     * Sets the styling to be used to display either a point, polyline, or bitmap for this marker.
+     * If the marker is going to be used to display a bitmap, a 'points' style must be set.
+     *
+     * ie. { style: 'points', color: 'white', size: [50px, 50px], order: 2000, collide: false }
```
ecgreb: It would be nice to add some formatting tags here otherwise these examples will display as one long paragraph when the javadoc is parsed.

```
   * <ul>
   * <li>{ style: 'points', color: 'white', size: [50px, 50px], order: 2000, collide: false }</li>
   * <li>{ style: 'lines', color: '#06a6d4', width: 5px, order: 2000 }</li>
   * <li>{ style: 'polygons', color: '#06a6d4', width: 5px, order: 2000 }</li>
   * </ul>
```

android/tangram/src/com/mapzen/tangram/Marker.java
sarahlensing: Updated

android/tangram/src/com/mapzen/tangram/Marker.java
sarahlensing: Updated

android/tangram/src/com/mapzen/tangram/MapController.java
```diff
@@ -737,6 +765,60 @@ void checkPointer(long ptr) {
         }
     }
 
+    void checkId(long id) {
+        if (id <= 0) {
+            throw new RuntimeException("Tried to perform an operation on an invalid id! This means you may have used an object that has been disposed and is no longer valid.");
```
blair1618: I'm not sure that we should apply the same sort of check to marker IDs that we do to the map pointer. For the map pointer, we decided to throw an exception if the pointer is null because it shouldn't happen during correct use and there's no sensible way to recover from it. For markers though, it's not unreasonable to accidentally use a marker ID that was previously removed (we return a boolean mostly to express that case). Passing an invalid marker ID into the native code here is an expected circumstance, so let's just do that.

blair1618:
Things got weird for a second while I was bringing this branch up to date - I've changed this PR to be based against master now and I think everything is cool now. *Nothing to see here.*

android/tangram/src/com/mapzen/tangram/Marker.java
```diff
@@ -0,0 +1,189 @@
...
+    }
+
+    /**
+     * Sets the drawable to be used to load a bitmap. When displaying a drawable, a
+     * 'points' style must also be set on the marker (see {@link Marker#setStyling(String)}.
+     *
+     * @param drawable the drawable
+     * @return whether the drawable's bitmap was successfully set
+     */
+    public boolean setDrawable(Drawable drawable) {
```
ecgreb: It might be nice to also have a public method exposed to set a raw bitmap. Raw bitmaps would be useful mostly for downloaded images where as drawables lend themselves to working with bundled resources.

This would be mostly for convenience so the client app would not need to convert the `Bitmap` to a `BitmapDrawable` just to have Tangram convert it back to a `Bitmap`. But maybe there is also an advantage to forcing developers to convert it to a `BitmapDrawable` so we can properly set target density? Although this should happen automatically [if the proper constructor is used](https://developer.android.com/reference/android/graphics/drawable/BitmapDrawable.html#BitmapDrawable(android.content.res.Resources, android.graphics.Bitmap)).

ecgreb: Marking as approved since all changes look good. We can always come back and add this method later if we decide it would be useful.

android/tangram/src/com/mapzen/tangram/MapController.java
```diff
@@ -654,6 +652,36 @@ public void pickFeature(float posX, float posY) {
...
+
+    /**
+     * Removes the passed in {@link Marker} from the map.
+     * @param marker to remove from the map.
+     * @return whether or not the marker was removed
+     */
+    public boolean removeMarker(Marker marker) {
+        checkPointer(mapPointer);
+        checkId(marker.getMarkerId());
+        return nativeMarkerRemove(mapPointer, marker.getMarkerId());
```
tallytalwar: does it make sense to set the `markerID` of this `marker` as `0` if the `nativeMarkerRemove` succeeds?

blair1618: That's a good point, it probably makes sense to invalidate the Java Marker object.

Issue #1057, opened 2016-11-02
Point marker setVisible not working


karimnaaji:
Point marker visibility can't be updated from the API. 
Those markers are not using the style draw method to be drawn, and this is where visibility is used to skip render https://github.com/tangrams/tangram-es/blob/master/core/src/style/style.cpp#L400. It doesn't seem that point labels use that marker property anywhere else.

cc @sarahlensing

blair1618:
Resolved by #1058

PR #1058, opened 2016-11-02
Adjust visibility of marker points

karimnaaji:
#1057

### DIFF
--- a/core/src/labels/label.h
+++ b/core/src/labels/label.h
@@ -174,11 +174,11 @@ public:
 
     bool offViewport(const glm::vec2& _screenSize);
 
+    void setAlpha(float _alpha);
+
 private:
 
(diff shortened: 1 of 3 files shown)


blair1618:
@karimnaaji Can you think of a way to do this without RTTI?

blair1618:
Actually I'm gonna try something - hold on.

karimnaaji:
What we can do is set a property in the mesh class `bool isLabelMesh()` or something in those lines, and `static_cast` on it. A lot of those casts are made in the codebase for this particular purpose, that would be good to remove it!

blair1618:
Yeah, I'll do this with a separate change. 

It seems that the only RTTI we use in the core code is for checking style types and we can easily replace that with an enum member.

PR #1060, opened 2016-11-03
faster applySourceBlocks

hjanetzek:


core/src/gl/shaderProgram.cpp
```diff
@@ -281,34 +278,39 @@ std::string ShaderProgram::applySourceBlocks(const std::string& source, bool fra
...
-        it += sm.position() + sm.length();
+    while (std::getline(sourceIn, line)) {
+        if (line.empty()) {
+            continue;
+        }
 
-        std::string pragmaName = sm[1];
-        bool unique;
-        std::tie(std::ignore, unique) = pragmas.emplace(std::move(pragmaName));
+        sourceOut << line << '\n';
```
karimnaaji: Just tested, and this is definitely an improvement!

Do you think we can apply the "Replace blackslash+newline with spaces" part at this point on `line` variable.
If we find `\\` on `line` variable directly and not inject `\n` on the stream at this point?

core/src/gl/shaderProgram.cpp
```diff
@@ -286,7 +294,15 @@ std::string ShaderProgram::applySourceBlocks(const std::string& source, bool fra
             continue;
         }
 
-        sourceOut << line << '\n';
+        size_t start = 0;
+
+        // Replace blackslash+newline with spaces
+        if ((start = line.find("\\", start)) != std::string::npos) {
+            line.replace(start, 1, " ");
```
hjanetzek: Better check the last character explicitly

karimnaaji:
We could also strip line comments from our script that builds the shaders to header files.

blair1618:
@karimnaaji @hjanetzek Are you still working on this? Seems pretty good to me.

karimnaaji:
@blair1618 , we spoke about adding one more commit to remove all our `\` from `core/shaders` and making sure we don't use them anywhere else than `rasters.glsl`. 
Coming shortly!

PR #1064, opened 2016-11-03
Degenerate point vertices when selection color = 0

karimnaaji:


blair1618:
So this is the same block of code that was added to `polygon.vs` and `polyline.vs`, did we just forget to add it to the points shader?

karimnaaji:
Yes we didn't update that one.

PR #1067, opened 2016-11-06
Add robust URL handling and use it for scene imports

blair1618:
- Adds the `Tangram::Url` class, which represents URLs and data URIs, compliant with IETF RFCs 1808 and 3986. (I'm pretty happy with how this turned out, parsing and resolving URLs gets complicated but it's all done without allocating additional memory).
 - Adds a ton of tests for `Tangram::Url`, using examples from IETF RFC 3986.
 - Uses `Url` for resolving resource locations in `Importer`.
 - Fixes up and clarifies the tests for scene importing.

### DIFF
--- a/core/src/scene/importer.cpp
+++ b/core/src/scene/importer.cpp
@@ -70,8 +70,7 @@ Node Importer::applySceneImports(const Url& scenePath, const Url& resourceRoot) 
 
     LOGD("Processing scene import Stack:");
     std::vector<Url> sceneStack;
-    std::unordered_set<std::string> globalTextures;
-    importScenes(root, rootScenePath, sceneStack, globalTextures);
+    importScenesRecursive(root, rootScenePath, sceneStack);
 
     return root;
 }
@@ -88,11 +87,9 @@ void Importer::processScene(const Url& scenePath, const std::string &sceneString
     try {
         auto sceneNode = YAML::Load(sceneString);
 
-        resolveSceneUrls(sceneNode, scenePath);
-
         m_scenes[scenePath] = sceneNode;
 
-        for (const auto& import : getScenesToImport(sceneNode)) {
+        for (const auto& import : getResolvedImportUrls(sceneNode, scenePath)) {
             m_sceneQueue.push_back(import);
             m_condition.notify_all();
         }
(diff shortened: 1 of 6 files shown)


core/src/util/url.cpp
```diff
@@ -0,0 +1,487 @@
...
+        t.buffer.append(b.buffer, b.parts.location.start, b.parts.location.count);
+    }
+    t.parts.location.count = t.buffer.size() - t.parts.location.start;
+
+    // Resolve the path.
+    t.parts.path.start = t.buffer.size();
+    if (r.hasNetLocation()) {
+        // If the relative URL has a netLocation, use its path as well.
+        t.buffer.append(r.buffer, r.parts.path.start, r.parts.path.count);
+    } else if (r.hasPath()) {
```
hjanetzek: How can a URL have a netlocation without having a scheme (being 'absolute')?

core/src/util/url.cpp
```diff
@@ -0,0 +1,487 @@
...
+    if (buffer.compare(parts.scheme.start, parts.scheme.count, "http") == 0 ||
+        buffer.compare(parts.scheme.start, parts.scheme.count, "https") == 0) {
+        flags |= HAS_HTTP_SCHEME;
+    } else if (buffer.compare(parts.scheme.start, parts.scheme.count, "file") == 0) {
+        flags |= HAS_FILE_SCHEME;
+    }
+
+    // If '//' is next in the string, then the substring up to the following '/' is the network location.
+    if (buffer.compare(start, 2, "//") == 0) {
+        start += 2;
```
hjanetzek: This should only be checked when a scheme was found and it's not a `FILE_SCHEME`

core/src/util/url.cpp
```diff
@@ -0,0 +1,487 @@
...
+    // Implements https://tools.ietf.org/html/rfc3986#section-5.2.4
+    // with in-place manipulation instead of making a new buffer.
+
+    size_t end = start + count;
+    size_t pos = start; // 'input' position.
+    size_t out = pos; // 'output' position.
+
+    while (pos < end) {
+        if (str[pos] == '.' && str[pos + 1] == '.' && str[pos + 2] == '/') {
+            pos += 3;
```
hjanetzek: `pos + 1` could be out of bounds read

hjanetzek: I think we could also remove consecutive slashes in this loop

blair1618:
Thanks for giving this a look!

About the handling of `file://` in https://github.com/tangrams/tangram-es/pull/1067/commits/105d395e78090577d9ebd77ffa550335eb15a780:

According to the specifications I've read and the behavior of both NSURL and cURL, any part of a URL that follows  `://` but precedes the next `/` is part of the "net location" or "authority" and not part of the path. That is `file:///Users/matt/hi.txt` locates a file at `/Users/matt/hi.txt` and `file://Users/matt/hi.txt` locates a file at `/matt/hi.txt`. I think we should follow the spec here.

Removing consecutive `/`s in https://github.com/tangrams/tangram-es/pull/1067/commits/0f046d321ea3060e5d03a3233b54769819f328be:

I haven't read anything from the IETF about this being part of a normalization process, and neither NSURL nor cURL perform this operation when resolving URLs. Why would we add this step?

hjanetzek:
Ah I see now - `file:///` is short for `file://localhost/` https://en.wikipedia.org/wiki/File_URI_scheme. That part was ok - still in `//a/b/c` `a` should not be considered as netlocation.  I'll update that commit

Removing superfluous slashes was part of the path normalization lib I skimmed over https://github.com/cpp-netlib/uri/blob/master/src/detail/uri_normalize.cpp. yes, not necessary I guess but looks nicer.

blair1618:
Why would we not consider that `a` to be the "net location"? NSURL decomposes URLs like that with `a` as the "host", and the RFC is pretty clear about URL structure: https://tools.ietf.org/html/rfc3986#section-3

blair1618:
Right, so properly speaking `//a/b/c` is what RFC 3986 calls a relative reference, which _still_ has an "authority" component: https://tools.ietf.org/html/rfc3986#section-4.2

hjanetzek:
ok - all good! updating tomorrow

core/src/scene/importer.cpp
```diff
@@ -82,141 +69,112 @@ Node Importer::applySceneImports(const std::string& scenePath, const std::string
...
+                    textureUrlNode = Url(textureUrlNode.Scalar()).resolved(base).string();
                 }
                 m_globalTextures.insert(texture.first.Scalar());
             }
         }
     }
 
+    // Resolve inline texture URLs.
+
     if (Node styles = root["styles"]) {
```
hjanetzek: No - probably only the m_globalTextures for the imports up to this point should be considered

tests/unit/sceneImportTests.cpp
```diff
@@ -15,183 +15,178 @@ class TestImporter : public Importer {
...
+    CHECK(uniformsA["u_tex3"].Scalar() == "tex3");
+
+    auto styleB = root["styles"]["styleB"];
+
+    CHECK(styleB["texture"].Scalar() == "/root/imports/in_imports.png");
+
+    auto uniformsB = styleB["shaders"]["uniforms"];
+
+    CHECK(uniformsB["u_tex1"].Scalar() == "/root/imports/in_imports.png");
+    CHECK(uniformsB["u_tex2"].Scalar() == "tex2");
```
hjanetzek: I think this one is not correct: It would mean than the imported scene-node may reference textures from the importing scene-node (or any import level above).

core/src/util/url.cpp
```diff
@@ -214,51 +214,51 @@ Url Url::resolve(const Url& b, const Url& r) {
...
+        t.append(b, b.parts.scheme);
         t.buffer.append("://");
     }
 
     // Resolve the netLocation.
-    t.parts.location.start = t.buffer.size();
     if (r.hasNetLocation()) {
         // If the relative URL has a netLocation, use it for the new URL.
-        t.buffer.append(r.buffer, r.parts.location.start, r.parts.location.count);
+        t.parts.location = { t.buffer.size(), r.parts.location.count };
```
blair1618: I had something like this originally, but this initialization syntax isn't supported in GCC 4.9.2 unless you write a constructor for it, so I just set the members separately.

hjanetzek:
Oh that 'save' commit went back in by mistake - wanted to drop that

hjanetzek:
@blair1618 are you ok with these changes? Looks good to mere to me.

blair1618:
The changes in https://github.com/tangrams/tangram-es/pull/1067/commits/3efd2a9adde840fbd0d36d42a4335b9ca283ad72 don't make sense to me. Why would we resolve/normalize the URLs for textures in a different place than for fonts or data sources?

hjanetzek:
This is necessary if a yaml fragment A imports fragment B and defines a style that references a global texture that is defined in fragment B. `m_globalTextures` did not contain the texture from B when resolveTextureUrl was called for A.

blair1618:
Ah sorry, I wasn't clear. You are correct that we should resolve texture URLs after a scene has been merged with all of it's own imports. I am suggesting that we should normalize all of the URLs at that point, not just the texture URLs.

hjanetzek:
The `Importer` code could be refactored into `resolveSceneUrls` and `resolveImports` though

blair1618:
Oh right, the import URLs! Ok I understand now. What if we resolve the import URLs in `getScenesToImport` and then resolve the other URLs after merging using `resolveSceneUrls`?

hjanetzek:
yeah, that should work

core/src/scene/importer.cpp
```diff
@@ -82,140 +69,89 @@ Node Importer::applySceneImports(const std::string& scenePath, const std::string
...
 }
 
-void Importer::processScene(const std::string &scenePath, const std::string &sceneString) {
+void Importer::processScene(const Url& scenePath, const std::string &sceneString) {
 
-    LOGD("Process: '%s'", scenePath.c_str());
+    LOGD("Process: '%s'", scenePath.string().c_str());
+
+    // Don't load imports twice
+    if (m_scenes.find(scenePath) != m_scenes.end()) {
```
tallytalwar: I guess this is redundant since same check happens in Line# 50 above, before `processScene` gets called.

hjanetzek: Good point! it should be redundant. At least we should have a better solution :) 

The problem is that the same download path can be enqueued more than once. The first finished download adds the processed scene node. This only solves that the second yaml gets discarded.

tallytalwar: Ahh.. I see. Nice catch in that case :+1:

core/src/util/url.h
```diff
@@ -0,0 +1,158 @@
+#pragma once
+
+#include <string>
+
+namespace Tangram {
+
+// This class is based on the URL concept specified by IETF RFC 1808
+// (https://tools.ietf.org/html/rfc1808) and the URI concept specified by
+// IETF RFC 3986 (https://tools.ietf.org/html/rfc3986) In particular this
```
tallytalwar: Thanks for the references here. :+1:

PR #1069, opened 2016-11-07
Misc fixes

hjanetzek:
Some issues found by valgrind when testing with raster sources

### DIFF
--- a/core/src/tangram.cpp
+++ b/core/src/tangram.cpp
@@ -109,7 +109,10 @@ Map::~Map() {
     // The unique_ptr to Impl will be automatically destroyed when Map is destroyed.
     impl->tileWorker.stop();
     impl->asyncWorker.reset();
-    impl->jobQueue.runJobs();
+
+    // Make sure other threads are stopped before calling stop()!
+    // All jobs will be executed immediately on add() afterwards.
+    impl->jobQueue.stop();
 
     TextDisplay::Instance().deinit();
     Primitives::deinit();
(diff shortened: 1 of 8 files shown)


core/src/tangram.cpp
```diff
@@ -109,7 +109,10 @@ Map::~Map() {
     // The unique_ptr to Impl will be automatically destroyed when Map is destroyed.
     impl->tileWorker.stop();
     impl->asyncWorker.reset();
-    impl->jobQueue.runJobs();
+
+    // Make sure other threads are stopped before calling stop()!
+    // All jobs will be executed immediately on add() afterwards.
+    impl->jobQueue.stop();
```
blair1618: @hjanetzek what's the bug that this fixes?

hjanetzek: Future bugs :) This should make things more robust though. E.g. in our current use Impl::renderState could be destructed before Impl::jobQueue without bugs.

The actual bug I was looking at was the static UrlWorkers still holding onto RasterTileTask-\>RasterSource-\>Textures where the texture disposer called into garbage.

blair1618: Yeah, the destruction process relies on the ordering of members right now which isn't ideal.

The case of holding onto tasks that invoke disposers does seem pretty risky, this change sounds like a good idea.

core/src/tile/tileWorker.cpp
```diff
@@ -130,6 +130,8 @@ void TileWorker::stop() {
     for (auto& worker : m_workers) {
         worker->thread.join();
     }
+
+    m_queue.clear();
```
tallytalwar: :+1:

Issue #1074, opened 2016-11-09
When the text is Chinese, the text_wrap parameter does not work. Because the words in Chinese are not separated by spaces.


ligai322:
for example:  I come from Beijing.
But This sentence write in Chinese, like '我来自北京'.

nvkelso:
We'll need to add special implementation logic for this, there are a few standards to reference.

ligai322:
@nvkelso  We can refer to the implementation of CSS,  like word-wrap,word-break and text-wrap.

nvkelso:
**See also:**

- https://en.wikipedia.org/wiki/Line_breaking_rules_in_East_Asian_languages
- Line breaking conventions can be found in [JLREQ](https://www.w3.org/TR/css-text-3/#JLREQ) and [JIS4051](https://www.w3.org/TR/css-text-3/#JIS4051) for Japanese, [ZHMARK](https://www.w3.org/TR/css-text-3/#ZHMARK) for Chinese, and in [UAX14](https://www.w3.org/TR/css-text-3/#UAX14) for all scripts in Unicode.
- [JIS4051] Formatting rules for Japanese documents (『日本語文書の組版方法』). Japanese Standards Association. 2004. JIS X 4051:2004. In Japanese
- [JLREQ] Yasuhiro Anan; et al. Requirements for Japanese Text Layout. 3 April 2012. W3C Working Group Note. URL: http://www.w3.org/TR/2012/NOTE-jlreq-20120403/
- [UAX14] Asmus Freytag. Line Breaking Properties. 17 January 2012. Unicode Standard Annex #14. URL: http://www.unicode.org/unicode/reports/tr14/
- [ZHMARK] 标点符号用法 (Punctuation Mark Usage). 1995. 中华人民共和国国家标准

westnordost:
From Asian languages, this also at least applies to Thai, Lao and Burmese.

hjanetzek:
It's possible that liblinebreak that we use requires the correct language to be specified. It worked before when we applied liblinebreak to individual script/language runs. This was changed a while ago to detect linebreaks without language set on the full string to first split on explicit line breaks.

karimnaaji:
Fixed by https://github.com/tangrams/tangram-es/issues/1075.

PR #1075, opened 2016-11-09
Text line breaks

hjanetzek:
https://github.com/tangrams/tangram-es/issues/1074


hjanetzek:
@karimnaaji can you look at this? The bug was that the check started a new line only when there are more than _minLineChars since the last possible break. Since liblinebreak seems to consider all Chinese characters to be breakable `charCount - lastChar >= _minLineChars` was never true.

PR #1076, opened 2016-11-09
More infos for label feature selection

karimnaaji:
Add interface to get more infos when selecting POIs/Icons and text labels.

Generic feature selection makes it hard to adjust a marker on top of an icon since the interface returns back the screen position of the created feature selection query. Putting a marker at this  position makes it nearly impossible to add a marker sticking to an existing label on screen.

This PR adds the set of touched label created for that feature with their corresponding lat/lon.

Example:

Selecting any of the point or text in this POI:
screen shot 2016-11-09 at 2 52 22 pm

Returns:
screen shot 2016-11-09 at 2 52 36 pm

cc @sarahlensing @ecgreb for the Android interface.

TODO:
- [x] JNI bindings
- [x] Treat line labels

### DIFF
--- a/android/tangram/jni/platform_android.cpp
+++ b/android/tangram/jni/platform_android.cpp
@@ -91,7 +91,7 @@ void setupJniEnv(JNIEnv* jniEnv, jobject _tangramInstance, jobject _assetManager
         jniEnv->DeleteGlobalRef(touchLabelClass);
     }
     touchLabelClass = (jclass)jniEnv->NewGlobalRef(jniEnv->FindClass("com/mapzen/tangram/TouchLabel"));
-    touchLabelInitMID = jniEnv->GetMethodID(touchLabelClass, "<init>", "([DILjava/util/Map;)V");
+    touchLabelInitMID = jniEnv->GetMethodID(touchLabelClass, "<init>", "(DDILjava/util/Map;)V");
 
     if (hashmapClass) {
         jniEnv->DeleteGlobalRef(hashmapClass);
@@ -322,7 +322,7 @@ void setCurrentThreadPriority(int priority) {
     setpriority(PRIO_PROCESS, tid, priority);
 }
 
-void labelsPickCallback(jobject listener, const std::vector<Tangram::TouchLabel>& labels) {
+void labelPickCallback(jobject listener, const Tangram::LabelPickResult* labelPickResult) {
 
     JniThreadBinding jniEnv(jvm);
 
@@ -330,12 +330,11 @@ void labelsPickCallback(jobject listener, const std::vector<Tangram::TouchLabel>
 
     jobject touchLabel = nullptr;
 
-    if (labels.size() > 0) {
-        auto label = labels[0];
-        auto properties = label.touchItem.properties;
+    if (labelPickResult) {
+        auto properties = labelPickResult->touchItem.properties;
 
-        position[0] = label.touchItem.position[0];
-        position[1] = label.touchItem.position[1];
+        position[0] = labelPickResult->touchItem.position[0];
+        position[1] = labelPickResult->touchItem.position[1];
 
         jobject hashmap = jniEnv->NewObject(hashmapClass, hashmapInitMID);
 
@@ -346,28 +345,26 @@ void labelsPickCallback(jobject listener, const std::vector<Tangram::TouchLabel>
         }
 
         jdoubleArray darr = jniEnv->NewDoubleArray(2);
-        jniEnv->SetDoubleArrayRegion(darr, 0, 2, reinterpret_cast<jdouble*>(&label.coordinate));
-        touchLabel = jniEnv->NewObject(touchLabelClass, touchLabelInitMID, darr, label.type, hashmap);
+        touchLabel = jniEnv->NewObject(touchLabelClass, touchLabelInitMID, labelPickResult->coordinate.longitude,
+            labelPickResult->coordinate.latitude, labelPickResult->type, hashmap);
     }
 
     jniEnv->CallVoidMethod(listener, onLabelPickMID, touchLabel, position[0], position[1]);
     jniEnv->DeleteGlobalRef(listener);
 }
 
-void featurePickCallback(jobject listener, const std::vector<Tangram::TouchItem>& items) {
+void featurePickCallback(jobject listener, const Tangram::FeaturePickResult* featurePickResult) {
 
     JniThreadBinding jniEnv(jvm);
 
     jobject hashmap = jniEnv->NewObject(hashmapClass, hashmapInitMID);
     float position[2] = {0.0, 0.0};
 
-    if (items.size() > 0) {
-        auto result = items[0];
-        auto properties = result.properties;
-        // auto labels = result.labels;
+    if (featurePickResult) {
+        auto properties = featurePickResult->properties;
 
-        position[0] = result.position[0];
-        position[1] = result.position[1];
+        position[0] = featurePickResult->position[0];
+        position[1] = featurePickResult->position[1];
 
         for (const auto& item : properties->items()) {
             jstring jkey = jniEnv->NewStringUTF(item.key.c_str());
--- a/android/tangram/jni/platform_android.cpp
+++ b/android/tangram/jni/platform_android.cpp
@@ -47,10 +47,16 @@ static jmethodID cancelUrlRequestMID = 0;
 static jmethodID getFontFilePath = 0;
 static jmethodID getFontFallbackFilePath = 0;
 static jmethodID onFeaturePickMID = 0;
+static jmethodID touchLabelInitMID = 0;
 
+static jclass touchLabelClass = nullptr;
 static jclass hashmapClass = nullptr;
 static jmethodID hashmapInitMID = 0;
 static jmethodID hashmapPutMID = 0;
+
+static jclass listClass = nullptr;
+static jmethodID listAddMID = 0;
+static jmethodID listInitMID = 0;
 
 static AAssetManager* assetManager = nullptr;
 
@@ -80,7 +86,20 @@ void setupJniEnv(JNIEnv* jniEnv, jobject _tangramInstance, jobject _assetManager
     setRenderModeMethodID = jniEnv->GetMethodID(tangramClass, "setRenderMode", "(I)V");
 
     jclass featurePickListenerClass = jniEnv->FindClass("com/mapzen/tangram/MapController$FeaturePickListener");
-    onFeaturePickMID = jniEnv->GetMethodID(featurePickListenerClass, "onFeaturePick", "(Ljava/util/Map;FF)V");
+    onFeaturePickMID = jniEnv->GetMethodID(featurePickListenerClass, "onFeaturePick", "(Ljava/util/Map;Ljava/util/List;FF)V");
+
+    if (touchLabelClass) {
+        jniEnv->DeleteGlobalRef(touchLabelClass);
+    }
+    touchLabelClass = (jclass)jniEnv->NewGlobalRef(jniEnv->FindClass("com/mapzen/tangram/TouchLabel"));
+    touchLabelInitMID = jniEnv->GetMethodID(touchLabelClass, "<init>", "([DI)V");
+
+    if (listClass) {
+        jniEnv->DeleteGlobalRef(listClass);
+    }
+    listClass = (jclass)jniEnv->NewGlobalRef(jniEnv->FindClass("java/util/ArrayList"));
+    listAddMID = jniEnv->GetMethodID(listClass, "add", "(Ljava/lang/Object;)Z");
+    listInitMID = jniEnv->GetMethodID(listClass, "<init>", "()V");
 
     if (hashmapClass) {
         jniEnv->DeleteGlobalRef(hashmapClass);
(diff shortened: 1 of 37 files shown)


core/src/labels/label.h
```diff
@@ -186,6 +191,8 @@ class Label {
 
     int m_anchorIndex;
 
+    uint32_t m_selectionColor;
```
hjanetzek: You can use the selection color from Sprite/TextLabel `m_vertexAttrib.selectionColor`

hjanetzek: I think we don't need that member at all when `getSelectionColor() { return m_vertexAttribs.selectionColor; }` is defined for Sprite/TextLabels :)

karimnaaji: I went on this path, but this need a pure virtual methods reimplemented on both, it's a little simpler as it is now.

karimnaaji:
@blair1618 can you give a look on this one?

blair1618:
I think we can make this interface much more practical. While it is useful to provide additional information about labels in selection results, I don't think returning a list of labels is the best approach.

As I understand this interface now, if a user makes a query that hits one label of a feature then they receive a result containing all labels of that feature, even those that their query point didn't hit. This means that if an app wanted to place a marker at the location of a selection query result, it would need to iterate over the list of labels and somehow determine which one is closest to the location that was queried. This seems unnecessarily burdensome. 

I think the interface would be more useful if we returned at most one label result. To implement this, we may need to generate additional selection colors for labels so that they are all distinct from each other. 

I would also strongly consider creating two separate selection functions, one for labels and one for general features, like we have discussed in meetings.

karimnaaji:
The interface has been refactored, this can be given another look.

android/tangram/src/com/mapzen/tangram/MapController.java
```diff
@@ -70,11 +71,23 @@
...
         void onFeaturePick(Map<String, String> properties, float positionX, float positionY);
     }
+    /**
+     * Interface for a callback to receive information about labels picked from the map
+     */
+    public interface LabelPickListener {
+        /**
+         * Receive information about labels found in a call to {@link #pickLabels(float, float)}
+         * @param label The {@link TouchLabel} that has been selected
+         * @param positionX The horizontal screen coordinate of the tapped location
```
ecgreb: Is `positionX` and `positionY` the raw position of the single tap input or the center point of the selected feature?

karimnaaji: `positionX` and `positionY` are the position that have been used to create the query, the user can use this position to know which query is related to the callback. The position of the feature should be retrieved by using `label.getCoordinate()`. 

To be able to place a pin at the center the selected feature this coordinate should be used to place a point marker.

android/tangram/src/com/mapzen/tangram/MapController.java
```diff
@@ -70,11 +71,23 @@
...
+         * @param positionY The vertical screen coordinate of the tapped location
+         */
+        void onLabelPick(TouchLabel label, float positionX, float positionY);
```
ecgreb: Most of the POIs we care about in Eraser Map have both and icon and a text label. Are these considered separate labels? Will this method be invoked once for each? Or will it be invoked for whichever one occupies the pixel where that tap is centered? And if so will the `positionX` and `positionY` returned differ depending on whether the icon or text is tapped?

android/tangram/src/com/mapzen/tangram/MapController.java
```diff
@@ -650,6 +663,26 @@ public void pickFeature(float posX, float posY) {
...
+        labelPickListener = listener;
+    }
+
+    /**
+     * Query the map for labeled features at the given screen coordinates; results will be returned
+     * in a callback to the object set by {@link #setLabelPickListener(LabelPickListener)}
+     * @param posX The horizontal screen coordinate
+     * @param posY The vertical screen coordinate
+     */
+    public void pickLabels(float posX, float posY) {
```
karimnaaji: This one is necessary the same way `pickFeatureAt` is used, and can be used on a callback of a single tap if you want, but not only.

In the demo app it's called on a `singleTapConfirmed` https://github.com/tangrams/tangram-es/blob/feature-selection-label-infos/android/demo/src/com/mapzen/tangram/android/MainActivity.java#L136

android/tangram/src/com/mapzen/tangram/MapController.java
karimnaaji: It will be called only once for the one that have been selected, either text or label. Once we have a scalable thumb size parameter it would return the closest.

android/tangram/src/com/mapzen/tangram/MapController.java
karimnaaji: We should probably remove the plural for clarity (`pickLabel`) to reflect the behavior.

core/src/tangram.h
```diff
@@ -11,12 +11,27 @@ namespace Tangram {
...
 
+using FeatureSelectionCallback = std::function<void(const std::vector<TouchItem>&)>;
+
+struct TouchLabel {
+    LabelType type;
+    LngLat coordinate;
+    TouchItem touchItem;
+};
+
+using LabelSelectionCallback = std::function<void(const std::vector<TouchLabel>&)>;
```
blair1618: Is it possible that more than one label could be selected for a query at a single pixel? If not then we could return just one label pointer or nullptr.

core/src/tangram.h
```diff
@@ -11,12 +11,27 @@ namespace Tangram {
...
+
 struct TouchItem {
     std::shared_ptr<Properties> properties;
     float position[2];
     float distance;
 };
 
+using FeatureSelectionCallback = std::function<void(const std::vector<TouchItem>&)>;
+
+struct TouchLabel {
```
blair1618: I'm not sure `TouchLabel` is the best name, since nothing about this requires a touch interface (same goes for `TouchItem`). Maybe `LabelResult` or `LabelQueryResult`?

android/tangram/jni/platform_android.cpp
```diff
@@ -311,6 +322,38 @@ void setCurrentThreadPriority(int priority) {
...
+        jobject hashmap = jniEnv->NewObject(hashmapClass, hashmapInitMID);
+
+        for (const auto& item : properties->items()) {
+            jstring jkey = jniEnv->NewStringUTF(item.key.c_str());
+            jstring jvalue = jniEnv->NewStringUTF(properties->asString(item.value).c_str());
+            jniEnv->CallObjectMethod(hashmap, hashmapPutMID, jkey, jvalue);
+        }
+
+        jdoubleArray darr = jniEnv->NewDoubleArray(2);
+        jniEnv->SetDoubleArrayRegion(darr, 0, 2, reinterpret_cast<jdouble*>(&label.coordinate));
```
blair1618: Probably easier to just use 2 double arguments here rather than instantiate an array.

android/tangram/src/com/mapzen/tangram/TouchLabel.java
```diff
@@ -0,0 +1,47 @@
...
+    public enum LabelType {
+        ICON,
+        TEXT,
+    }
+
+    private LngLat coordinate;
+    private LabelType type;
+    private Map<String, String> properties;
+
+    public TouchLabel(double[] coordinates, int type, Map<String, String> properties) {
```
blair1618: Recommend making this package-private, I don't think SDK users should be creating these objects manually.

android/tangram/src/com/mapzen/tangram/TouchLabel.java
```diff
@@ -0,0 +1,47 @@
+package com.mapzen.tangram;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * {@code TouchLabel} represents labels that can be selected on the screen
+ */
+public class TouchLabel {
```
blair1618: `LabelResult` seems like a clearer name to me.

core/src/labels/labels.cpp
```diff
@@ -63,6 +63,32 @@ void Labels::processLabelUpdate(const ViewState& viewState,
     }
 }
 
+
+bool Labels::getLabel(const std::vector<std::unique_ptr<Style>>& _styles,
+                                      const std::vector<std::shared_ptr<Tile>>& _tiles,
+                                      uint32_t _selectionColor, Label*& _label, Tile*& _tile) {
```
blair1618: `*&` make me irrationally uncomfortable.

core/src/tangram.cpp
```diff
@@ -425,25 +437,59 @@ void Map::render() {
...
+            float x = labelQuery.position.x / impl->view.getWidth();
+            float y = (1.f - (labelQuery.position.y / impl->view.getHeight()));
+
+            // TODO: read with a scalable thumb size and iterate over the read colors
+            GLuint color = impl->selectionBuffer->readAt(x, y);
+
+            if (color != 0) {
+                // Retrieve the label for this selection color
+                if (impl->labels.getLabel(impl->scene->styles(), impl->tileManager.getVisibleTiles(), color, label, tile)) {
+                    std::vector<TouchLabel> touchLabels;
```
blair1618: Seems unused?

core/src/tangram.h
```diff
@@ -11,12 +11,27 @@ namespace Tangram {
...
 
+enum LabelType {
+    icon,
+    text,
+};
+
 struct TouchItem {
     std::shared_ptr<Properties> properties;
     float position[2];
     float distance;
```
blair1618: Doesn't seem like `distance` should be here any more - it's always zero for a feature result so we could make it a member of the label result instead. Or maybe we don't care that much about it?

core/src/tangram.h
blair1618: To really be consistent in naming, we could have `FeaturePickCallback`, `FeaturePickResult`, `LabelPickCallback`, and `LabelPickResult`.

core/src/labels/labels.cpp
karimnaaji: We would use `**` here, I just feel like pointer indirection is hard to read, but I can change for that.

android/tangram/src/com/mapzen/tangram/TouchLabel.java
karimnaaji: 👍

android/tangram/jni/platform_android.cpp
karimnaaji: Yes, leftover from previous implementation, this would be cleaner.

core/src/tangram.h
karimnaaji: `LabelPickResult` seems more explicit, along with `LabelPickCallback` and al.
We can also reflect that in the android interface.

core/src/tangram.h
karimnaaji: I left it for now because of the potential thumb size that we would introduce. Right now we only return one item since the feature querying is pixel perfect, but we would potentially have a distance from the center of thumb.

I will clean that up and we can rethink this more when we introduce that option.

android/tangram/src/com/mapzen/tangram/MapController.java
ecgreb: Currently there aren't any calls to `pickFeatureAt(...)` in the Android SDK or Eraser Map code yet the `FeaturePickListnener.onFeaturePick(...)` callback still gets invoked on single tap. Is that going to change with this pull request?

android/tangram/src/com/mapzen/tangram/MapController.java
ecgreb: Ok cool makes sense. Follow up question about coordinates below with regards to text vs. icon.

android/tangram/src/com/mapzen/tangram/MapController.java
ecgreb: Would `label.getCoordinate()` return the same longitude and latitude values for the icon label and the text label no matter which one is selected?

I'm just trying to avoid the case where the pin gets dropped in a different location depending on whether the icon or text is tapped.

karimnaaji: Yes, since those have been created for the same feature, they would return the same coordinate.

android/tangram/src/com/mapzen/tangram/MapController.java
karimnaaji: It seems that this is done here in the android SDK https://github.com/mapzen/android/blob/5eb635120a2ea3388c5c66f40c51ec5a276ec3ad/library/src/main/java/com/mapzen/android/graphics/MapzenMap.java#L54, isn't this what triggers the whole selection in Eraser Map?

core/src/tangram.h
blair1618: Generic features don't have a fixed "position" that we could compare against during selection, so I'm not sure how we would come up with a "distance" even if we had a pick radius > 0.

android/tangram/src/com/mapzen/tangram/MapController.java
karimnaaji: So what you would need to do is switch to `pickLabel` instead, and parse the results of `label.getProperties()`, along with `getCoodinate()` to retrieve the longitude/latitude.

android/tangram/src/com/mapzen/tangram/MapController.java
ecgreb: 👍

PR #1079, opened 2016-11-11
Add interface to apply scene updates

karimnaaji:


msmollin:
👍  once the build finishes.

PR #1084, opened 2016-11-14
Import: resolve urls only once

hjanetzek:
Only resolve the urls for the current scenePath on sceneNode (instead of on merged root)
Tested with loading tron as relative path `./tangram -f tron/tron.yaml`

@blair1618 This should also fix the issue with Android assets.

### DIFF
--- a/core/src/scene/importer.cpp
+++ b/core/src/scene/importer.cpp
@@ -70,6 +70,7 @@ Node Importer::applySceneImports(const Url& scenePath, const Url& resourceRoot) 
 
     LOGD("Processing scene import Stack:");
     std::vector<Url> sceneStack;
+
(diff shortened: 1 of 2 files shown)


hjanetzek:
Good that Importer has do many unit test - I'll dig more into it tomorrow.

blair1618:
Yeah, the change I made for Android assets isn't sufficient - the same problem happens for the relative paths used in iOS/macOS app bundles. However, I was leaning towards giving more things an absolute URL.

hjanetzek:
I agree with the `asset://` scheme approach to make URL resolution idempotent. Thought it would be nicer if URL resolution could be applied only once in the first place.

PR #1085, opened 2016-11-14
Url: remove duplicate slashes from path

hjanetzek:
Another option would be to change removeLastSegmentFromRange
to drop duplicate slashes

### DIFF
--- a/core/src/util/url.cpp
+++ b/core/src/util/url.cpp
@@ -455,7 +455,11 @@ size_t Url::removeDotSegmentsFromRange(std::string& str, size_t start, size_t co
     size_t out = pos; // 'output' position.
 
     while (pos < end) {
-        if (pos + 2 < end &&
+        if (pos + 1 < end &&
+                   str[pos] == '/' &&
+                   str[pos + 1] == '/') {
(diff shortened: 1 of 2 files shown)


blair1618:
Does this fix some undesirable behavior?

hjanetzek:
yes - exemplified in the test case

blair1618:
I'm not sure that this is the correct way to deal with duplicated slashes. I think that `//` represents a path segment containing an empty string and is distinct from no path segment. cURL seems to agree and resolves `file:///Users/matt/Downloads//../Desktop/little.txt` to `/Users/matt/Downloads/Desktop/little.txt`

hjanetzek:
It might be shell user common sense but not part of url dots removal spec. Or an option for a nitpick PR on curl.

blair1618:
Mmmmm I'm not convinced - NSURL and cURL both treat it as an empty path segment, and they are designed for handling generic URLs whereas bash/zsh is not.

hjanetzek:
Yeah ok - it seems to be the rule for URLs to have `..` replace empty path segments

PR #1086, opened 2016-11-14
Use scene update value

karimnaaji:


msmollin:
👍  - annoyed I missed that the first time around :)

PR #1089, opened 2016-11-15
Prevent 'global' references from being resolved as URLs

blair1618:
Currently, if a `global.` reference is used in a scene node that represents a URL, that reference will be resolved as a relative URL and not replaced with the appropriate global value.

This change makes `global.` a reserved prefix that will prevent values from being treated like URLs. This means that global values will now behave as documented for values that expect URLs. 

This also implies that you cannot use `global.png` or similar file names/paths. If you _really_ need to call your file "global.png" then you can use the equivalent path `./global.png`

### DIFF
--- a/core/src/scene/importer.cpp
+++ b/core/src/scene/importer.cpp
@@ -98,10 +98,19 @@ void Importer::processScene(const Url& scenePath, const std::string &sceneString
     }
 }
 
-bool nodeIsTextureUrl(const Node& node, const Node& textures) {
-
+bool nodeIsPotentialUrl(const Node& node) {
     // Check that the node is scalar and not null.
     if (node.IsNull() || !node.IsScalar()) { return false; }
+
+    // Check that the node does not contain a 'global' reference.
+    if (node.Scalar().compare(0, 7, "global.") == 0) { return false; }
+
+    return true;
+}
+
+bool nodeIsTextureUrl(const Node& node, const Node& textures) {
+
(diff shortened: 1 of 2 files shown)


bcamper:
What about scene `import`?

blair1618:
@bcamper Gooood question - I'm not sure how to handle that. I don't think we can allow `global.` references in `import` statements because imported paths are used _before_ `global.`s can be replaced.

bcamper:
Oh hehe right, makes sense. Good policy to stick to then.

blair1618:
@hjanetzek Does this look alright to you?

hjanetzek:
@blair1618 can't `global.png` be a path unless `global: {png: ... }` is defined? This would be the same logic as for global textures.

blair1618:
@hjanetzek The trick is that we don't know the total set of values in `global: {}` until the merging is done, so we can't assume that a value won't be defined by a "parent" scene.

hjanetzek:
I see - so one would have to make a first pass to get all global blocks before merging imports (and resolving URLs). 

I'm ok with merging this. We can consider some special cases or the import logic later.

blair1618:
Oh yeah that's possible - we can iterate over the DAG of imports and generate the total set of `global:` values before actually merging the scenes, then check potential URLs against that set before resolving them. Let's keep that in mind in case this becomes a problem in practice!

PR #1092, opened 2016-11-15
System fonts

karimnaaji:
- Add iOS system font interface.
- Refactor how system fonts are being handled on all the platforms and simplify interface.

Dependent work on alfons https://github.com/hjanetzek/alfons/pull/9

To test carefully!


ios/src/TGMapViewController.mm
```diff
@@ -519,6 +519,7 @@ - (void)respondToShoveGesture:(UIPanGestureRecognizer *)shoveRecognizer
 - (void)viewDidLoad
 {
     [super viewDidLoad];
+    NSLog(@"Font families: %@", [UIFont familyNames]);
```
msmollin: If we want to log from the SDK side, we should probably add a macro to not output the log statements from release builds of the framework.

Otherwise you should remove this.

ios/src/platform_ios.mm
```diff
@@ -126,14 +132,115 @@ bool isContinuousRendering() {
...
+
+    return data;
+}
+
+std::vector<FontSourceHandle> systemFontFallbacksHandle() {
+    NSArray* fallbacks = [UIFont familyNames];
+
+    std::vector<FontSourceHandle> handles;
+
+    for (id fallback in fallbacks) {
```
msmollin: There a reason this is an `id` type here? https://developer.apple.com/reference/uikit/uifont/2091412-familynames?language=objc

Seems to indicate objects inside the return array will be of type `NSString`.

ios/src/platform_ios.mm
```diff
@@ -1,15 +1,21 @@
 #ifdef PLATFORM_IOS
 
 #import <Foundation/Foundation.h>
```
msmollin: Since you're importing UIKit below, you can remove the foundation import (UIKit includes it).

ios/src/CGFontConverter.mm
```diff
@@ -0,0 +1,154 @@
...
+
+    size_t* tableSizes = (size_t*)malloc(sizeof(size_t) * tableCount);
+    memset(tableSizes, 0, sizeof(size_t) * tableCount);
+
+    BOOL containsCFFTable = NO;
+
+    size_t totalSize = sizeof(FontHeader) + sizeof(TableEntry) * tableCount;
+
+    for (int index = 0; index < tableCount; ++index) {
+
```
msmollin: I'm concerned about the performance implications about iterating the same iterator set twice. Any way to combine this loop and the one at L128?

ios/src/platform_ios.mm
```diff
@@ -1,15 +1,21 @@
...
+#import <UIKit/UIKit.h>
 #import <utility>
 #import <cstdio>
 #import <cstdarg>
 #import <fstream>
 #import <regex>
 
+#include <iostream>
```
msmollin: You can probably swap all of these over to using import rather than include - it's safer. Also, it looks like you're importing <fstream> twice?

ios/src/platform_ios.mm
```diff
@@ -126,14 +132,115 @@ bool isContinuousRendering() {
...
+        {"condensed", UIFontDescriptorTraitCondensed},
+        {"monospace", UIFontDescriptorTraitMonoSpace},
+    };
+
+    UIFont* font = [UIFont fontWithName:[NSString stringWithUTF8String:_name.c_str()] size:1.0];
+
+    if (font == nil) {
+        // Get the default system font
+        if (_weight.empty()) {
+            font = [UIFont systemFontOfSize:0.0];
```
msmollin: You should make this `1.0` per the docs: https://developer.apple.com/reference/uikit/uifont/1619042-systemfontofsize?language=objc

"The size (in points) to which the font is scaled. This value must be greater than 0.0."

ios/src/platform_ios.mm
```diff
@@ -126,14 +132,115 @@ bool isContinuousRendering() {
...
+        } else {
+            int weight = std::atoi(_weight.c_str());
+
+            // Default to 400 boldness
+            weight = (weight == 0) ? 400 : weight;
+
+            // Map weight value to range [100..900]
+            weight = std::min(std::max(100, (int)floor(weight / 100.0 + 0.5) * 100), 900);
+
+            font = [UIFont systemFontOfSize:0.0 weight:weightTraits[weight]];
```
msmollin: Same thing as above about 0.0

ios/src/platform_ios.mm
```diff
@@ -126,14 +132,115 @@ bool isContinuousRendering() {
...
+
+        auto it = fontTraits.find(_face);
+        if (it != fontTraits.end()) {
+            traits = it->second;
+
+            // Create a new descriptor with the symbolic traits
+            descriptor = [descriptor fontDescriptorWithSymbolicTraits:traits];
+
+            if (descriptor != nil) {
+                font = [UIFont fontWithDescriptor:descriptor size:0.0];
```
msmollin: Also here re: 0.0

toolchains/iOS.toolchain.cmake
```diff
@@ -48,8 +48,8 @@ endif (CMAKE_UNAME)
 
 # Force the compilers to gcc for iOS
 include (CMakeForceCompiler)
-CMAKE_FORCE_C_COMPILER (/usr/bin/clang Apple)
-CMAKE_FORCE_CXX_COMPILER (/usr/bin/clang++ Apple)
+# CMAKE_FORCE_C_COMPILER (/usr/bin/clang Apple)
```
msmollin: I'm assuming these are commented out since Apple doesn't ship anything but clang now?

ios/src/platform_ios.mm
```diff
@@ -126,14 +132,115 @@ bool isContinuousRendering() {
...
+            }
+        }
+    }
+
+    return loadUIFont(font, _size);
```
msmollin: Will whatever is calling into this thing be expecting null pointers?

ios/src/CGFontConverter.h
```diff
@@ -0,0 +1,18 @@
...
+//  tangram
+//
+//  Created by Karim Naaji on 11/01/16.
+//
+//
+
+#import <Foundation/Foundation.h>
+#import <CoreGraphics/CoreGraphics.h>
+
+@interface CGFontConverter : NSObject
```
karimnaaji: Yep, you mentioned that to me but I forgot to update it! Will fix.

ios/src/CGFontConverter.mm
karimnaaji: I haven't seen this showing up in the profiler on my testing. This one first loop is necessary to know how much memory to allocate for the `stream` to be filled up later on.

ios/src/platform_ios.mm
karimnaaji: Yes, null pointer is expected on the usage of this.

toolchains/iOS.toolchain.cmake
karimnaaji: Yes those were giving warnings, we can remove it.

ios/src/platform_ios.mm
karimnaaji: No specific reason for the `id`, is this harmful in any way?

ios/src/platform_ios.mm
karimnaaji: Yep, we can clean up those a bit and use `import` instead, even though those are in the same compilation unit so there should be no risk of multiple inclusion.

ios/demo/src/MapViewController.m
```diff
@@ -24,8 +24,12 @@ - (void)mapView:(TGMapViewController *)mapView didLoadSceneAsync:(NSString *)sce
     newYork.longitude = -74.00976419448854;
     newYork.latitude = 40.70532700869127;
 
-    [mapView setZoom:16];
-    [mapView setPosition:newYork];
+    TGGeoPoint cairo;
```
tallytalwar: maybe make sense to keep all demo apps to be centered on manhattan coordinates?

nitpick though!

linux/src/platform_linux.cpp
```diff
@@ -99,15 +107,32 @@ unsigned char* bytesFromFile(const char* _path, size_t& _size) {
...
+    handles.push_back(getFontHandle(FONT_JA));
+    handles.push_back(getFontHandle(FALLBACK));
+
+    return handles;
 }
 
-// No system fonts fallback implementation (yet!)
-std::string systemFontFallbackPath(int _importance, int _weightHint) {
-    return "";
+unsigned char* systemFont(const std::string& _name, const std::string& _weight, const std::string& _face, size_t* _size) {
```
tallytalwar: Can you add that comment of adding system fonts for linux? (yet!)

ios/src/TGFontConverter.mm
```diff
@@ -0,0 +1,154 @@
...
+
+    memset(stream, 0, totalSize);
+    char* dataStart = (char*)stream;
+    char* dataPtr = dataStart;
+
+    // compute font header entries
+    uint16_t entrySelector = 0;
+    uint16_t searchRange = 1;
+
+    while (searchRange < tableCount >> 1) {
```
karimnaaji: This part gets the closest power of two after tableCount and gets the 4th next power of two after it.

This is described here in the _Organization of an OpenType Font_ https://www.microsoft.com/typography/otspec/otff.htm.

It is described as "(Maximum power of 2 <= numTables) x 16."
So you can see the x16 as `searchRange <<= 4;` and the `(Maximum power of 2 <= numTables)` as the while loop.

ios/src/TGFontConverter.mm
```diff
@@ -0,0 +1,154 @@
...
+{
+    const uint32_t* dataBuff = (const uint32_t *)CFDataGetBytePtr(dataRef);
+    uint32_t dataLength = (uint32_t)CFDataGetLength(dataRef);
+    return CalcTableCheckSum(dataBuff, dataLength);
+}
+
+@implementation TGFontConverter
+
+//Reference:
+//http://skia.googlecode.com/svn-history/r1473/trunk/src/ports/SkFontHost_mac_coretext.cpp
```
blair1618: This, I think: https://skia.googlesource.com/skia/+/master/src/ports/SkFontHost_mac.cpp

karimnaaji: Yes, this reference is actually https://github.com/google/skia/blob/2563601fc2b0505619f905f86bd249ae630197cc/src/ports/SkFontHost_mac.cpp#L1706.

ios/src/TGFontConverter.mm
tallytalwar: Great. Glad I asked.

Can you put that `microsoft` link as reference here?

karimnaaji: Yep, will update those references.

tallytalwar:
I just have to test the workings on tizen and we should be good to go with this one.

msmollin:
Yep. Sorry thought I had clicked approved :)

PR #1096, opened 2016-11-17
do not allow to merge map with non-map nodes

hjanetzek:
this may modify a sequence into a map and in particular
break filters


hjanetzek:
:+1:

PR #1104, opened 2016-11-23
Rebuild all markers when pixel density is set

blair1618:
Resolves an issue that is very similar to https://github.com/mapzen/tizen-maps-plugin-mapzen/issues/72 (have not yet tested on Tizen).


tallytalwar:
This does seem to resolve the tizen issue also.

Issue #1062, opened 2016-11-03
Expose NSURLSession/NSURLCache from platform.mm for configuration


msmollin:
On iOS we currently have no way to configure caching options or modify outbound requests to the tile services. This would prevent people from adding things like a Do-Not-Track header (used in EraserMap). We also shoud be able to alter how cache is configured and used by the URL request framework, say in the event we want longer or shorter durations between invalidations, or to change the size/location of the cache.

PR #1107, opened 2016-11-28
iOS http handler

karimnaaji:
Fixes https://github.com/tangrams/tangram-es/issues/1062


karimnaaji:
@msmollin can you look over this? Thanks.

Issue #1099, opened 2016-11-21
iOS Marker API not setting markers when referencing resources from a scene file


msmollin:
Basic setup:
Scene file: bubble wrap from the no-api-key branch

Test Code (Swift):

```
import UIKit
import TangramMap
class TangramVC:  TGMapViewController{

    override func viewDidLoad() {
        super.viewDidLoad()
        self.loadSceneFile("scene.yaml")
    }

    override func viewDidAppear(animated: Bool) {
        super.viewDidAppear(animated)
        let marker = markerAdd()
        let point = TGGeoPoint(longitude: -122.44880676269531, latitude: 37.76155490343394)
        markerSetStyling(marker, styling: "{ style: ux-location-gem-overlay, interactive: true, sprite: ux-current-location, size: 36px, collide: false }")
        markerSetPoint(marker, coordinates: point)
        markerSetVisible(marker, visible: true)
        animateToPosition(point, withDuration: 2.0)
        animateToZoomLevel(15, withDuration: 2.0)
    }
}
```

Note that if we use a style string not dependent on a style sheet, for example `{ style: 'points', color: 'white', size: [25px, 25px], order:500, collide: false }` then everything works and a 25px white circle appears on the map.

blair1618:
On Android, the mechanism for loading a scene file is to launch an asynchronous task and then receive a callback when loading is complete and scene resources are available. We may want a similar mechanism for iOS so that we can rely on having scene resources available.

Markers can also use an arbitrary user-supplied bitmap, which would obviate the scene dependency.

msmollin:
@blair1618 totally agreed, and we already have that functionality in the form of a `TGMapViewDelegate` callback. However, this particular problem is a little more vexing. Regardless of using either the "synchronous" method above, or the async method, neither get the gem to appear.

What's odd is that by applying a marker to the map, there seems to be a scene reload occurring, at least according to the logs I'm seeing. Perhaps there's a race condition where we're attempting to reference the scene file that has not yet completed parsing?

As for user supplied bitmaps, we haven't had a chance to expose that yet in the Objective-C wrapper.

Issue #1113, opened 2016-11-29
Default scene pois fail to load


karimnaaji:
Our current scene loader fails to load the poi atlas texture on desktop for our default scene https://github.com/tangrams/tangram-es/blob/master/scenes/scene.yaml.

The Url gets replaced from `img/poi_icons_32.png` to `/img/poi_icons_32.png` in the scene importer while resolving scene urls, and fails to load later on in the scene loading pipeline. 

Possibly related issue https://github.com/tangrams/tangram-es/issues/1099.

blair1618:
_Actually_ resolved by https://github.com/tangrams/tangram-es/pull/1174

PR #1116, opened 2016-11-29
Fix absolute resource resolution on on iOS and OS X

karimnaaji:
Let bundle resolve absolute paths from the bundle root on iOS and OS X.

Resolve part of https://github.com/tangrams/tangram-es/issues/1113, Linux would still have issues relating to that.

Should fix https://github.com/tangrams/tangram-es/issues/1099.


msmollin:
👍

PR #1117, opened 2016-11-29
null or undefined jsfunction evaluations are legit

### LINKED ISSUE #1108: Inconsistency with JS on text_source fallback
We discovered in https://github.com/tangrams/paparazzi/issues/12 that there is an inconsistency between JS and ES around the handling of `text_source` values when not using the default. This is causing a significant rendering bug on styles like Refill under ES.

It seems like Tangram ES is evaluating as if it is: `text_source: [name:short, name]` even though I've specified it as: `text_source: name:short`.

@blair1618 noted that this seems like a bug in ES as specifying `text_source: 'name:short'` (which should be equivalent here to `text_source: name:short`) produces the same behavior as in JS.

Specifically, on Refill most the light grey region text labels on the map shouldn't be displayed at zoom 4.

**JS correct**
JS correct

**ES bug, look in Columbia**
ES bug

See also:

- https://github.com/tangrams/refill-style-more-labels/blob/gh-pages/refill-style-more-labels.yaml#L4121-L4133
- The visible on Tangram ES region names source their text from name:short via this global:
  -https://github.com/tangrams/refill-style-more-labels/blob/gh-pages/refill-style-more-labels.yaml#L22-L25
- In ES code we are setting the text's source default as the name property.
https://github.com/tangrams/tangram-es/blob/master/core/src/style/textStyleBuilder.cpp#L292
- https://github.com/tangrams/tangram-docs/blob/1d0d68a14b6548c71778cd2aa529aa3ddea0399e/pages/draw.md#text_source

tallytalwar:
Style params evaluating a jsFunction as null or undefined will not be marked "inactive". These are
legal values for a js function to return!

fixes #1108

### DIFF
--- a/core/src/scene/styleContext.cpp
+++ b/core/src/scene/styleContext.cpp
@@ -460,6 +460,7 @@ void StyleContext::parseStyleResult(StyleParamKey _key, StyleParam::Value& _val)
     } else if (duk_is_null_or_undefined(m_ctx, -1)) {
         // Ignore setting value
         LOGD("duk evaluates JS method to null or undefined.");
+        _val = Undefined();
     } else {
         LOGW("Unhandled return type from Javascript style function for %d.", _key);
     }
--- a/core/src/scene/styleParam.h
+++ b/core/src/scene/styleParam.h
@@ -134,7 +134,7 @@ struct StyleParam {
         }
     };
 
-    using Value = variant<none_type, bool, float, uint32_t, std::string, glm::vec2, Width, LabelProperty::Anchors>;
+    using Value = variant<Undefined, none_type, bool, float, uint32_t, std::string, glm::vec2, Width, LabelProperty::Anchors>;
 
     StyleParam() :
         key(StyleParamKey::none),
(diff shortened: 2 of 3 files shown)


hjanetzek:
Can't we use the none_type for this?

tallytalwar:
We can not use `none_type` because of the check in `evalStyle` https://github.com/tangrams/tangram-es/blob/89e222cde99f259572a9399eda0acc8058c0f21c/core/src/scene/styleContext.cpp#L340.

Currently (in master) the value returned is `none_type` only.

tallytalwar:
`struct undefined` is also a good one. What issues do you see with the `nullptr`? I was initially suspicious of its usage!

hjanetzek:
yeah let's have our own. programming is all about making up types :)

tallytalwar:
Nice call @hjanetzek. Our own `undefined` type makes more sense here and less risk prone.

Updated!

blair1618:
This still produces the problem described in https://github.com/tangrams/tangram-es/issues/1108 for me. Investigating...

blair1618:
Ah wait, false alarm. I had some modifications in my local copy of refill-style.

tallytalwar:
Ahh! Did not see this change being pushed to master. Will land on tizen in our next update (coming soon with yaml-cpp optimizations)

PR #1119, opened 2016-11-30
store feature properties only once

hjanetzek:
Currently it Properties is stored three times if a feature has interactive icon and text style. This change only keeps the featureId to its tiles.m_selectionFeatures.

### DIFF
--- a/core/src/labels/label.cpp
+++ b/core/src/labels/label.cpp
@@ -14,9 +14,9 @@ namespace Tangram {
 
 const float Label::activation_distance_threshold = 2;
 
-Label::Label(Label::WorldTransform _worldTransform, glm::vec2 _size, uint32_t _selectionColor, Type _type, Options _options)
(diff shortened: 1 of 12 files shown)


hjanetzek:
@karimnaaji could you have a look?

karimnaaji:
Yep code looks good to me! Will test on device shortly, otherwise you can go ahead and merge when travis is ready.

hjanetzek:
yes, I also need to do some more testing

karimnaaji:
@hjanetzek can we put 99f4844 in another PR, it needs a little more work to update the android and iOS interface.

tallytalwar:
@karimnaaji https://github.com/tangrams/tangram-es/tree/loadSceneWithUpdates

karimnaaji:
👍 , removing that commit from here.

karimnaaji:
@hjanetzek merging this if you're done with testing? Everything runs fine on device.

PR #1125, opened 2016-12-01
Load scene with updates

tallytalwar:
NOTE: ios and android will be done in a subsequent PR. Need this in master for tizen.

### DIFF
--- a/core/src/tangram.cpp
+++ b/core/src/tangram.cpp
@@ -250,6 +250,11 @@ void Map::queueSceneUpdate(const char* _path, const char* _value) {
     impl->sceneUpdates.push_back({_path, _value});
 }
 
+void Map::queueSceneUpdate(std::vector<SceneUpdate> sceneUpdates) {
+    std::lock_guard<std::mutex> lock(impl->sceneMutex);
+    impl->sceneUpdates.insert(impl->sceneUpdates.end(), sceneUpdates.begin(), sceneUpdates.end());
+}
+
 void Map::applySceneUpdates() {
 
     std::shared_ptr<Scene> nextScene;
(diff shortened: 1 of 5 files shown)


tallytalwar:
cc. @karimnaaji for ios work.

core/src/tangram.h
```diff
@@ -55,15 +55,18 @@ class Map {
...
     // Load the scene at the given absolute file path synchronously
-    void loadScene(const char* _scenePath, bool _useScenePosition = false);
+    void loadScene(const char* _scenePath, bool _useScenePosition = false,
+                   const std::vector<SceneUpdate>& sceneUpdates = {});
 
     // Request an update to the scene configuration; the path is a series of yaml keys
     // separated by a '.' and the value is a string of yaml to replace the current value
     // at the given path in the scene
     void queueSceneUpdate(const char* _path, const char* _value);
+    void queueSceneUpdate(std::vector<SceneUpdate> sceneUpdates);
```
karimnaaji: We can remove `void queueSceneUpdate(const char* _path, const char* _value);` since the new call contains the previous API too.

core/src/tangram.cpp
```diff
@@ -214,9 +219,9 @@ void Map::loadSceneAsync(const char* _scenePath, bool _useScenePosition, MapRead
         nextScene = impl->nextScene;
     }
 
-    runAsyncTask([nextScene, _platformCallback, _cbData, this](){
+    runAsyncTask([nextScene, _platformCallback, _cbData, _sceneUpdates, this](){
```
karimnaaji: Let's simply copy the data from here, we can't rely on the user reference to stay valid when using it in the async task.

core/src/tangram.h
tallytalwar: I think we can leave this for the time being. It might be good to have an overloaded single update additions. We can refactor it later.

core/src/tangram.cpp
tallytalwar: Need to check this. Based on the documentation for lambda capture this is a capture of the value!! But the original `_sceneUpdates` is a `const reference`, not sure what will happen. Will test it.

tallytalwar:
@karimnaaji @blair1618 ping on this. We can send an update for `tizen` if we merge this.

PR #1127, opened 2016-12-01
faster stringFromFile

hjanetzek:
- avoid temporary buffer
- about twice as fast, tested on desktop linux and Z3
- 4ms vs 8ms on Z3 for loading bubble-wrap

### DIFF
--- a/android/tangram/jni/platform_android.cpp
+++ b/android/tangram/jni/platform_android.cpp
@@ -65,6 +65,10 @@ PFNGLBINDVERTEXARRAYOESPROC glBindVertexArrayOESEXT = 0;
 PFNGLDELETEVERTEXARRAYSOESPROC glDeleteVertexArraysOESEXT = 0;
 PFNGLGENVERTEXARRAYSOESPROC glGenVertexArraysOESEXT = 0;
 
+// Android assets are distinguished from file paths by the "asset" scheme.
+static const char* aaPrefix = "asset:///";
+static const size_t aaPrefixLen = 9;
+
 void bindJniEnvToThread(JNIEnv* jniEnv) {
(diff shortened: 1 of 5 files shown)


android/tangram/jni/platform_android.cpp
```diff
@@ -233,80 +237,83 @@ bool isContinuousRendering() {
 
 }
 
-unsigned char* bytesFromAssetManager(const char* _path, size_t& _size) {
+struct ByteBuffer{
```
blair1618: Unused?

android/tangram/jni/platform_android.cpp
```diff
@@ -233,80 +237,83 @@ bool isContinuousRendering() {
...
 
     _size = 0;
+    unsigned char* data = nullptr;
 
-    // Android assets are distinguished from file paths by the "asset" scheme.
-    const char* aaPrefix = "asset:///";
-    const size_t aaPrefixLen = strlen(aaPrefix);
+    auto allocator = [&](size_t size) {
+        _size = size;
+        data = (unsigned char*) malloc(sizeof(char) * size);
```
blair1618: Bit of an edge case, but if a file exists and has a size of zero then the call to `malloc(0)` could return `NULL`. We use `NULL` to indicate that the file could not be found, which is slightly different from the file having zero size.

android/tangram/jni/platform_android.cpp
hjanetzek: yup

android/tangram/jni/platform_android.cpp
blair1618: Ok, I guess we don't need to address this case right at this moment - something to be aware of though.

PR #1128, opened 2016-12-01
optimize getDouble

hjanetzek:
300ms vs 1100ms on Z3 for building layers - especially filters

- yes we may loose some infinity/nan formats but we don't handle them explicitly anyway

### DIFF
--- a/core/src/util/yamlHelper.cpp
+++ b/core/src/util/yamlHelper.cpp
@@ -64,30 +64,44 @@ glm::vec4 getColorAsVec4(const Node& node) {
 }
 
 std::string parseSequence(const Node& node) {
+    if (node.IsSequence()) { return ""; }
+
     std::stringstream sstream;
     for (const auto& val : node) {
-        try {
-            sstream << val.as<double>() << ",";
-        } catch (const BadConversion& e) {
-            try {
-                sstream << val.as<std::string>() << ",";
-            } catch (const BadConversion& e) {
-                LOGE("Float or Unit expected for styleParam sequence value");
-            }
+        if (!val.IsScalar()) {
+            // LOG
+            return "";
+        }
+
+        double value;
+        if (getDouble(val, value)) {
+            sstream << value << ",";
+        } else {
+            sstream << val.as<std::string>() << ",";
         }
     }
     return sstream.str();
 }
 
-bool getDouble(const Node& node, double& value, const char* name) {
-    try {
-        value = node.as<double>();
-        return true;
-    } catch (const BadConversion& e) {}
+bool getDouble(const Node& node, double& value) {
 
-    if (name) {
-        LOGW("Expected a floating point value for '%s' property.:\n%s\n", name, Dump(node).c_str());
+    if (node.IsScalar()) {
+        const std::string& s = node.Scalar().c_str();
+        char* pos;;
+        value = strtod(s.c_str(), &pos);
+        if (pos == s.c_str() + s.length()) {
+            return true;
+        }
     }
+
+    return false;
+}
+
+bool getDouble(const Node& node, double& value, const char* name) {
+
+    if (getDouble(node, value)) { return true; }
+
+    LOGW("Expected a floating point value for '%s' property.:\n%s\n", name, Dump(node).c_str());
     return false;
 }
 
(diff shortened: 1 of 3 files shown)


karimnaaji:
Is there any unit tests on this already? Otherwise we can probably add some, it seems pretty isolated.

core/src/util/yamlHelper.cpp
```diff
@@ -64,30 +64,44 @@ glm::vec4 getColorAsVec4(const Node& node) {
 }
 
 std::string parseSequence(const Node& node) {
+    if (node.IsSequence()) { return ""; }
```
blair1618: Missing negation?

core/src/util/yamlHelper.cpp
```diff
@@ -64,30 +64,44 @@ glm::vec4 getColorAsVec4(const Node& node) {
...
-            } catch (const BadConversion& e) {
-                LOGE("Float or Unit expected for styleParam sequence value");
-            }
+        if (!val.IsScalar()) {
+            // LOG
+            return "";
+        }
+
+        double value;
+        if (getDouble(val, value)) {
```
blair1618: I don't understand this, we're converting a string to a double so that we can convert it back to a string? Shouldn't we just concatenate the `Scalar()` values?

core/src/util/yamlHelper.cpp
```diff
@@ -64,30 +64,44 @@ glm::vec4 getColorAsVec4(const Node& node) {
...
-    try {
-        value = node.as<double>();
-        return true;
-    } catch (const BadConversion& e) {}
+bool getDouble(const Node& node, double& value) {
 
-    if (name) {
-        LOGW("Expected a floating point value for '%s' property.:\n%s\n", name, Dump(node).c_str());
+    if (node.IsScalar()) {
+        const std::string& s = node.Scalar().c_str();
```
blair1618: This allocates a copy of the string - just take a reference to `node.Scalar()`.

core/src/util/yamlHelper.cpp
hjanetzek: I also wondered why this was done.

core/src/util/yamlHelper.cpp
hjanetzek: sure, that's what I intended - learned one more weirdness of c++ today. Wouldn't have expected the above code to work :)

Some 'rationale'  https://herbsutter.com/2008/01/01/gotw-88-a-candidate-for-the-most-important-const/

core/src/util/yamlHelper.cpp
hjanetzek: ack

core/src/util/yamlHelper.cpp
blair1618: Let's just remove the decode-to-double step, I can't see any reason for it.

hjanetzek:
@blair1618 ping?

karimnaaji:
@hjanetzek suggestion ignored?

hjanetzek:
@karimnaaji we can make another PR for unit tests.

karimnaaji:
Okay sounds good!

hjanetzek:
Started with unit tests and found that we dont support all YAML integer types as we ask for floating point :) 

http://yaml.org/spec/1.2/spec.html#id2761292 `.as<double>()` doesn't handle hex or octal.

blair1618:
Ah, good to know. Sounds like this isn't a regression though?

(For what it's worth, y2j handles hex and octal integers)

tallytalwar:
Looks like this breaks at least the buildings (on bubble-wrap).
screenshot from 2016-12-05 20-36-58

core/src/util/yamlHelper.cpp
tallytalwar: Ok this is something required. Seems to fix the issue I pointed out in this PR. Any color which is a sequence of doubles, get bad values and hence messes up the colors. Also visible in the default scene file.

screenshot from 2016-12-05 20-57-49

tallytalwar: Ahh that extra `,` at the end!

tallytalwar: Ok this is super weird... its not happy with `.65` and needs `0.65`! Which is what explicit conversion to `double` and then to `string` is getting us.

PR #1130, opened 2016-12-01
optimize applyGlobals

hjanetzek:
- YamlPath strings were allocated for each sequence and map node. This only allocates strings for
   the global-reference nodes.
- YamlPath::get should not create nodes along the way

### DIFF
--- a/core/src/scene/sceneLoader.cpp
+++ b/core/src/scene/sceneLoader.cpp
@@ -48,6 +48,8 @@ const std::string DELIMITER = ":";
 // TODO: make this configurable: 16MB default in-memory DataSource cache:
 constexpr size_t CACHE_SIZE = 16 * (1024 * 1024);
 
+static const std::string GLOBAL_PREFIX = "global.";
+
 std::mutex SceneLoader::m_textureMutex;
 
 bool SceneLoader::loadScene(std::shared_ptr<Scene> _scene) {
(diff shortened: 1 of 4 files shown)


core/src/scene/sceneLoader.cpp
```diff
@@ -71,16 +73,11 @@ bool SceneLoader::loadScene(std::shared_ptr<Scene> _scene) {
 void SceneLoader::applyUpdates(Scene& scene, const std::vector<SceneUpdate>& updates) {
     auto& root = scene.config();
     for (const auto& update : updates) {
-        Node value;
         try {
-            value = YAML::Load(update.value);
-        } catch (YAML::ParserException e) {
+            YamlPath(update.path).create(root) = YAML::Load(update.value);
```
tallytalwar: Maybe good to check if the `update.path` exists in the current `scene.config` before trying the update. Maybe this also fixes the tests.

tests/unit/sceneUpdateTests.cpp
```diff
@@ -187,3 +187,17 @@ TEST_CASE("Apply and propogate repeated global value updates") {
     CHECK(root["seq"][1].Scalar() == "global_a_value");
     CHECK(root["map"]["b"].Scalar() == "newer_global_b_value");
 }
+
+TEST_CASE("Regression: scene update requesting a sequence from a scalar") {
```
tallytalwar: Good check!

blair1618:
Hm. I actually intentionally allowed `YamlPath` to create nodes along the path when I last updated this code. The confusion caused by the previous behavior seemed to outweigh the potential error-checking feedback that it provided.

blair1618:
Err, can we have some consensus here? This is a change that could break current usage of the library. At the very least this PR should have included documentation to `Map::queueSceneUpdate` and `MapController$queueSceneUpdate`.

tallytalwar:
How is this a breaking change? Rather this will make sure code does not crash on an exception of invalid scene update being parsed. Can you elaborate on "could break current usage of the library"?

blair1618:
> YamlPath::get should not create nodes along the way

That's the part that could break current usage. But looking at the changes now, it doesn't seem that `YamlPath::get` was changed at all so I'm even more confused now.

tallytalwar:
Ohh your confusion is legitimate. @hjanetzek removed that code. We had some discussion on it and decided to move some of that code to another PR.

Must have missed updating the description.

blair1618:
Ahh I see. I think I'm clear on this now and it all seems sensible.

PR #1131, opened 2016-12-02
Use glKView content scale factor instead UIScreen pixel scale

karimnaaji:
After some time investigating an issue on some errors while positioning pins on some iOS phones, It seems that pixel scale was not giving the fractional scale at which we should translate coordinates from logical pixel to points, a small error was noticeable on the iPhone 6 plus when selecting features and performing a `screenToLatLon` to place a marker.

Apple docs concerning this are not that clear ([content scale factor](https://developer.apple.com/reference/uikit/uiview/1622657-contentscalefactor) and [UIScreen scale](https://developer.apple.com/reference/uikit/uiscreen/1617836-scale)); but after reading over this https://www.paintcodeapp.com/news/iphone-6-screens-demystified, it seems that the iPhone 6 plus actually performs a downsampling of 1.15 on the framebuffer. For a natural pixel scale of 3, this gives us 3/1.15, so ~2.6; exactly what `[glKView contentScaleFactor]` returns.

### DIFF
--- a/ios/demo/src/MapViewController.m
+++ b/ios/demo/src/MapViewController.m
@@ -38,11 +38,6 @@
     // Not feature selected
     if (feature.count == 0) {
 
-        // Convert the 2d screen position to the lat lon
-        TGGeoPoint latlon = [mapView screenPositionToLngLat:position];
-
-        // Set the map position
-        [mapView setPosition:latlon];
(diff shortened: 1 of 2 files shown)


blair1618:
`why@2.6x.png`

PR #1132, opened 2016-12-02
iOS label selection

karimnaaji:


karimnaaji:
@msmollin can you review this next?

ios/demo/src/MapViewController.m
```diff
@@ -32,22 +32,40 @@ - (void)mapView:(TGMapViewController *)mapView didLoadSceneAsync:(NSString *)sce
...
+        NSLog(@"\t%@ -- %@", key, [[labelPickResult properties] objectForKey:key]);
+
+        if ([key isEqualToString:@"name"]) {
+             UIAlertView *alert = [[UIAlertView alloc] initWithTitle:@"Label selection callback"
+                                                             message:[[labelPickResult properties] objectForKey:key]
+                                                            delegate:nil
+                                                   cancelButtonTitle:@"OK"
+                                                   otherButtonTitles:nil];
+             [alert show];
+             [alert release];
```
msmollin: Shouldn't have to call release in ARC code. This is ARC compiled, yeah?

msmollin: Can also remove it below too.

ios/src/TGLabelPickResult.h
```diff
@@ -0,0 +1,24 @@
...
+typedef NS_ENUM(NSInteger, TGLabelType) {
+    TGLabelTypeIcon = 0,
+    TGLabelTypeText,
+};
+
+@interface TGLabelPickResult : NSObject
+
+- (id) initWithCoordinates:(TGGeoPoint)coordinates type:(TGLabelType)type properties:(NSDictionary *)properties;
+
+@property (readonly, nonatomic) TGGeoPoint coordinates;
```
msmollin: Nitpick: I like to be explicit with properties memory management. So this would read
`+@property (readonly, assign, nonatomic) TGGeoPoint coordinates;`

whereas the Dictionary below would read
`+@property (readonly, strong, nonatomic) NSDictionary* properties;`

This becomes important for NSStrings and a few other classes where the default, because it's an object, is `strong`, but the convention is to always declare `copy` for them to avoid weirdness with class clusters.

ios/src/TGLabelPickResult.h
```diff
@@ -0,0 +1,24 @@
...
+#import "TGGeoPoint.h"
+
+typedef NS_ENUM(NSInteger, TGLabelType) {
+    TGLabelTypeIcon = 0,
+    TGLabelTypeText,
+};
+
+@interface TGLabelPickResult : NSObject
+
+- (id) initWithCoordinates:(TGGeoPoint)coordinates type:(TGLabelType)type properties:(NSDictionary *)properties;
```
msmollin: You'll want to define whether this return is Nullable or not, because Swift. Looks like you'd want `nonnull` Also, use `instancetype` instead of `id` for safer checking and language features.

ios/src/TGMapViewController.h
```diff
@@ -45,12 +46,17 @@ NS_ASSUME_NONNULL_BEGIN
...
 @end
 
+NS_ASSUME_NONNULL_END
+
 @protocol TGMapViewDelegate <NSObject>
 @optional
-- (void)mapView:(TGMapViewController*)mapView didLoadSceneAsync:(NSString*)scene;
-- (void)mapView:(TGMapViewController*)mapView didSelectFeature:(NSDictionary*)feature atScreenPosition:(CGPoint)position;
+- (void)mapView:(TGMapViewController *)mapView didLoadSceneAsync:(nonnull NSString *)scene;
+- (void)mapView:(TGMapViewController *)mapView didSelectFeature:(NSDictionary *)feature atScreenPosition:(CGPoint)position;
```
msmollin: Since you can now return nil here, you'll want to mark the NSDictionary * parameter here as `nullable` . Swift I think assumes `nullable` by default

msmollin: We probably should just go thru and annotate everywhere what is nullable vs nonnull in areas we don't use the `NS_ASSUME_NONNULL` macros.

ios/src/TGMapViewController.mm
```diff
@@ -149,6 +149,47 @@ - (void)pickFeaturesAt:(CGPoint)screenPosition
     });
 }
 
+- (void)pickLabelAt:(CGPoint)screenPosition
+{
+    if (!self.map && !self.mapViewDelegate) { return; }
```
msmollin: Why don't we want to call this code if we lack a delegate?

ios/src/TGMapViewController.h
karimnaaji: Since there's a need for a `NS_ASSUME_NONNULL`, my guess was that the default behavior for those were `nullable`, I don't want bloat signatures of function parameters with `nullable` everywhere except if it happens to actually be `nonnullable` by default.

msmollin: Yeah so if you don't define `nullable` or `nonnull` then they get exposed to Swift as implicitly unwrapped optionals, which is not terribly safe to work with. Marking this explicitly as `nullable` makes them get imported into Swift as `optional` which is far better.

ios/src/TGLabelPickResult.h
karimnaaji: Same argument here for `nullable`, if this happens to not be the default, we can start to explicitly add them.

Good call on `instancetype, I will update all the other initializers.

ios/demo/src/MapViewController.m
karimnaaji: Yes I will the ARC flags and update accordingly.

ios/demo/src/MapViewController.m
```diff
@@ -32,22 +32,40 @@ - (void)mapView:(TGMapViewController *)mapView didLoadSceneAsync:(NSString *)sce
...
+- (void)mapView:(TGMapViewController *)mapView didSelectLabel:(TGLabelPickResult *)labelPickResult atScreenPosition:(CGPoint)position
 {
+    if (!labelPickResult) { return; }
 
-    // Not feature selected
-    if (feature.count == 0) {
+    NSLog(@"Picked label:");
 
-        return;
+    for (id key in [labelPickResult properties]) {
```
karimnaaji: I've been using vim so I was not aware of that, but will update too if that makes code completion easier.

ios/src/TGMapViewController.mm
karimnaaji: I guess if the user hasn't provided a delegate there's no need to overhead tangram with feature selection, I will make sure this is something mentioned in the documentation of this interface.

ios/src/TGMapViewController.h
karimnaaji: I see, then I'll go through more than just what is new in this PR and will update it. It seems fair to make it another pull request though.

ios/src/TGLabelPickResult.h
karimnaaji: Why the `coordinates` should be `readonly` and `assign` at the same time?
Updating for 
```
+@property (readonly, copy, nonatomic) TGGeoPoint coordinates;
```

karimnaaji:
@msmollin Updated, needs rebase.

ios/src/TGMapViewController.mm
msmollin: Agreed yeah. If that's the intended behavior that's fine. We should just document it.

PR #1134, opened 2016-12-02
Fix stops for size style param

tallytalwar:
Size parameter is special and can be a 1d or a 2d. Unlike width or offsets.
This creates a different handling for size style param for stops.

Note: `stops` definitely need some refactoring love. Can definitely be made much concise. FUTURE PR!

### DIFF
--- a/core/src/scene/stops.cpp
+++ b/core/src/scene/stops.cpp
@@ -58,7 +58,7 @@ auto Stops::FontSize(const YAML::Node& _node) -> Stops {
         float key = frameNode[0].as<float>();
 
         if (lastKey > key) {
-            LOGW("Invalid stop order: key %f > %f\n", lastKey, key);
+            LOGW("Invalid stop order: key %f > %f", lastKey, key);
             continue;
         }
 
@@ -88,7 +88,7 @@ auto Stops::Sizes(const YAML::Node& _node, const std::vector<Unit>& _units) -> S
         float key = frameNode[0].as<float>();
 
         if (lastKey > key) {
-            LOGW("Invalid stop order: key %f > %f\n", lastKey, key);
+            LOGW("Invalid stop order: key %f > %f", lastKey, key);
             continue;
         }
         lastKey = key;
@@ -101,8 +101,7 @@ auto Stops::Sizes(const YAML::Node& _node, const std::vector<Unit>& _units) -> S
             if (StyleParam::parseValueUnitPair(frameNode[1].Scalar(), start, sizeValue)) {
                 for (auto& unit : _units) {
                     if (sizeValue.unit != unit) {
-                        LOGW("Size StyleParam can only take in pixel values.");
-                        return {};
+                        LOGW("Size StyleParam can only take in pixel values in: %s", Dump(_node).c_str());
                     }
                 }
 
@@ -119,8 +118,7 @@ auto Stops::Sizes(const YAML::Node& _node, const std::vector<Unit>& _units) -> S
                 if (StyleParam::parseValueUnitPair(sequenceNode.Scalar(), 0, sizeValue)) {
                     for (auto& unit : _units) {
                         if (sizeValue.unit != unit) {
-                            LOGW("Size StyleParam can only take in pixel values.");
-                            return {};
+                            LOGW("Size StyleParam can only take in pixel values in: %s", Dump(_node).c_str());
                         }
                     }
                     sizeValues.push_back(sizeValue);
@@ -147,7 +145,7 @@ auto Stops::Offsets(const YAML::Node& _node, const std::vector<Unit>& _units) ->
         float key = frameNode[0].as<float>();
 
         if (lastKey > key) {
-            LOGW("Invalid stop order: key %f > %f\n", lastKey, key);
+            LOGW("Invalid stop order: key %f > %f", lastKey, key);
             continue;
         }
         lastKey = key;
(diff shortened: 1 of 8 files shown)


core/src/style/pointStyleBuilder.cpp
```diff
@@ -138,11 +138,17 @@ auto PointStyleBuilder::applyRule(const DrawRule& _rule, const Properties& _prop
     _rule.get(StyleParamKey::angle, p.labelOptions.angle);
 
     auto sizeParam = _rule.findParameter(StyleParamKey::size);
-    if (sizeParam.stops && sizeParam.value.is<float>()) {
-        float lowerSize = sizeParam.value.get<float>();
-        float higherSize = sizeParam.stops->evalWidth(m_zoom + 1);
```
tallytalwar: Don't know why `evalWidth` was used before here? I have switched it to use `evalFloat` both in `pointStyleBuilder` and `Stops::evalSize`.
Or I am missing something?

hjanetzek: @tallytalwar evalWidth does the exponential interpolation which is required to interpolate properly relative to zoomlevels.

say for the stops [0, 1px],[2, 4px] the result should be 2 at zoom 1, not 3 (x=(1-0)/(2-0), 1*(1-x)+4*x).

tallytalwar: Ahh, perfect. Will revert back to use `evalWidth` here. Thanks

tallytalwar: Actually these should be renamed aptly.

hjanetzek: yes - also I think we should use exponential interpolation for all cases that involve zoomlevels

tallytalwar: Sure, will probably have a separate PR for that. In this just renaming `evalWidth` and `evalFloat` to apt names.

tallytalwar: Hmm ... shouldn't `evalVec2` also have the exponential interpolation in that case? Will change this also.

tallytalwar:
Updated!

core/src/style/pointStyleBuilder.cpp
tallytalwar: Updated to do exponential interpolation for `size` style parameter. Probably `offset` should also be using the exponential interpolation, but leaving that for another PR.

core/src/scene/stops.cpp
```diff
@@ -268,6 +323,32 @@ auto Stops::evalColor(float _key) const -> uint32_t {
...
+
+    const glm::vec2& lowerVal = lower->value.get<glm::vec2>();
+    const glm::vec2& upperVal = upper->value.get<glm::vec2>();
+
+    return glm::vec2(lowerVal.x * (1 - lerp) + upperVal.x * lerp,
+                     lowerVal.y * (1 - lerp) + upperVal.y * lerp);
+
+}
+
 auto Stops::evalVec2(float _key) const -> glm::vec2 {
```
karimnaaji: Is that used anymore?

tallytalwar: In `offset`. According to the math I did and discussion above I think all these interpolations should be using exponential specially when interpolating on zoom. But I only modified the usage for `size` and `width` was using exponential from before.

karimnaaji: Yes, icon/text offsets still need linear interpolation.

core/src/scene/stops.cpp
```diff
@@ -75,6 +75,61 @@ auto Stops::FontSize(const YAML::Node& _node) -> Stops {
...
 }
 
+auto Stops::Sizes(const YAML::Node& _node, const std::vector<Unit>& _units) -> Stops {
+    Stops stops;
+    if (!_node.IsSequence()) {
+        return stops;
+    }
+
+    for (auto& unit : _units) {
+        if (unit != Unit::pixel) {
```
karimnaaji: Those units come from the https://github.com/tangrams/tangram-es/blob/master/core/src/scene/styleParam.cpp#L93 for the specific style param key, not from the style parsed data, it seems that can't ever happen unless we change what's in the static variable.

core/src/scene/stops.cpp
```diff
@@ -75,6 +75,61 @@ auto Stops::FontSize(const YAML::Node& _node) -> Stops {
...
+        }
+        lastKey = key;
+
+        if (frameNode[1].IsScalar()) {
+            StyleParam::ValueUnitPair sizeValue;
+            sizeValue.unit = Unit::pixel;
+            size_t start = 0;
+
+            if (StyleParam::parseValueUnitPair(frameNode[1].Scalar(), start, sizeValue)) {
+                stops.frames.emplace_back(key, sizeValue.value);
```
karimnaaji: This is the point where the units should be tested against `_units` since it's the user parsed data, `sizeValue.unit` must be any `_units` value.

tallytalwar: Ahh you are right. I forgot to update this code. Will fix.

karimnaaji: I have a fix along with unit tests on this, will update this shortly.

tallytalwar:
@karimnaaji :+1: on the unit tests here.

core/src/scene/stops.cpp
hjanetzek: @karimnaaji  if the change should be linear with map scale this means it exponential with 'zoomlevel'

PR #1136, opened 2016-12-05
Add listener for map ready

karimnaaji:


ios/src/TGMapViewController.h
```diff
@@ -49,6 +49,7 @@ NS_ASSUME_NONNULL_BEGIN
 @optional
 - (void)mapView:(TGMapViewController*)mapView didLoadSceneAsync:(NSString*)scene;
 - (void)mapView:(TGMapViewController*)mapView didSelectFeature:(NSDictionary*)feature atScreenPosition:(CGPoint)position;
+- (void)didCompleteView:(TGMapViewController *)mapView;
```
msmollin: Nitpick: just to be consistent with the rest of Cocoa, this should probably read more like:
`- (void)viewDidCompleteLoading(TGMapViewController*)mapView`

karimnaaji:
@msmollin Updated.

tallytalwar:
:+1:

PR #1138, opened 2016-12-05
Catch BadSubscript exceptions

karimnaaji:
Prevent crashes on exception when trying to access missing YAML keys.


hjanetzek:
I think we should rather check the nodes. I have a 'more const' commit ready for submit that handles these cases in importer.

hjanetzek:
Basically if you search a key in a const Node& you have to check if the result is an undefined Node - on a non const Node the index will be created.

karimnaaji:
Okay I see! Yes we would rather check the nodes then. Closing this one.

Issue #1063, opened 2016-11-03
Add an optional constructor/initializer to allow for pre-setting certain stylesheet keys before parsing


msmollin:
In certain scenarios it would be advantageous for Tangram-es to allow pre-setting certain keys before parsing of a stylesheet begins in order to avoid re-parsing the entire style sheet later just to add URL query parameters. The current system leans towards developers use the synchronous scene loading functions in order to preempt things like tiles downloading so we can configure items such as URL query parameters or modify the stylesheet based on runtime configuration. The issue with using synchronous scene loading is this impacts performance on main application thread which on mobile OSes (iOS, Android, etc.) is generally frowned upon as it can impact UI performance and could lead to watchdog processes force-closing your application.

tallytalwar:
Should be valid for all platforms tangram-es is supported. We should remove `android`, `ios` tags here!

tallytalwar:
I believe we do provide a way to apply scene updates during initial scene load now, which should take care of this.

PR #1143, opened 2016-12-05
iOS interface with new scene Update API

karimnaaji:
Follows https://github.com/tangrams/tangram-es/pull/1125
Steps towards https://github.com/tangrams/tangram-es/issues/1063

### DIFF
--- a/ios/src/TGSceneUpdate.h
+++ b/ios/src/TGSceneUpdate.h
@@ -1,0 +1,18 @@
+//
+//  TGSceneUpdate.h
+//  tangram
+//
+//  Created by Karim Naaji on 11/05/16.
+//
+//
+
+#import <Foundation/Foundation.h>
+
+@interface TGSceneUpdate : NSObject
+
+- (id)initWithPath:(NSString *)path value:(NSString *)value;
+
+@property (strong, nonatomic) NSString* path;
+@property (strong, nonatomic) NSString* value;
+
+@end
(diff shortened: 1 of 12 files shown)


ios/src/TGSceneUpdate.h
```diff
@@ -0,0 +1,18 @@
...
+//
+//  Created by Karim Naaji on 11/05/16.
+//
+//
+
+#import <Foundation/Foundation.h>
+
+@interface TGSceneUpdate : NSObject
+
+- (id)initWithPath:(NSString *)path value:(NSString *)value;
```
msmollin: `instancetype` instead of `id` and mark as nullable/nonnull

ios/src/TGSceneUpdate.h
```diff
@@ -0,0 +1,18 @@
...
+
+@property (strong, nonatomic) NSString* path;
```
msmollin: these should be 
`+@property (copy, nonatomic) NSString* path;` due to how NSStrings are implemented under the covers.

ios/src/TGSceneUpdate.mm
```diff
@@ -0,0 +1,25 @@
...
+//
+//  Created by Karim Naaji on 11/05/16.
+//
+//
+
+#import "TGSceneUpdate.h"
+
+@implementation TGSceneUpdate
+
+- (id)initWithPath:(NSString *)path value:(NSString *)value;
```
msmollin: instancetype here as well.

karimnaaji:
@msmollin Updated

karimnaaji:
@tallytalwar can you gives this a review?

tallytalwar:
Will get to it today.

PR #1144, opened 2016-12-06
Fix polygon centroid calculation

### LINKED ISSUE #1140: Centroid function produces incorrect results
https://github.com/tangrams/tangram-es/blob/0ffcabe797aa086e8eae69bed2c604f87e93c6e0/core/src/util/geom.cpp#L104-L123

The centroid of a polygon is *not* the mean position of its vertices, as calculated here. 

https://en.wikipedia.org/wiki/Centroid#Centroid_of_polygon

Related: https://github.com/tangrams/tangram-es/issues/1133

blair1618:
The previous implementation was not correctly calculating the centroid of polygons, which we use for some label placement.

Tested by adding draw rules to the `water` layer in the default test scene:
```yaml
            lines:
                order: 9
                color: 'red'
                width: '2px'
                tile_edges: true
            points:
                order: 10
                color: 'green'
                centroid: true
                collide: false
```

The equivalent function in tangram-js: https://github.com/tangrams/tangram/blob/master/src/geo.js#L181-L198

Resolves #1140

### DIFF
--- a/core/src/util/geom.cpp
+++ b/core/src/util/geom.cpp
@@ -102,24 +102,27 @@ glm::vec2 worldToScreenSpace(const glm::mat4& _mvp, const glm::vec4& _worldPosit
 }
 
 glm::vec2 centroid(const std::vector<std::vector<glm::vec3>>& _polygon) {
+
+    if (_polygon.empty()) {
+        return glm::vec2();
+    }
+
+    // We will consider only the outer ring of the polygon when calculating
+    // the centroid - holes will be ignored. It would be possible to transform
+    // a polygon with holes into a polygon that covers the same set of points
+    // without any holes, but that seems overly complex for our needs.
+    const auto& ring = _polygon.front();
     glm::vec2 centroid;
-    int n = 0;
+    float area = 0.f;
 
-    for (auto& l : _polygon) {
-        for (auto& p : l) {
-            centroid.x += p.x;
-            centroid.y += p.y;
-            n++;
-        }
+    for (auto curr = ring.begin(), prev = ring.end() - 1; curr != ring.end(); prev = curr, ++curr) {
+        float a = (prev->x * curr->y - curr->x * prev->y);
+        centroid.x += (prev->x + curr->x) * a;
+        centroid.y += (prev->y + curr->y) * a;
+        area += a;
     }
 
-    if (n == 0) {
-        return centroid;
-    }
-
-    centroid /= n;
-
-    return centroid;
+    return centroid / (3.f * area);
 }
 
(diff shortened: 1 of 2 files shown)


core/src/util/geom.cpp
```diff
@@ -102,24 +102,27 @@ glm::vec2 worldToScreenSpace(const glm::mat4& _mvp, const glm::vec4& _worldPosit
...
-    centroid /= n;
+    for (auto curr = ring.begin(), prev = ring.end() - 1; curr != ring.end(); prev = curr, ++curr) {
+        float a = (prev->x * curr->y - curr->x * prev->y);
+        centroid.x += (prev->x + curr->x) * a;
+        centroid.y += (prev->y + curr->y) * a;
+        area += a;
+    }
 
-    return centroid;
+    return centroid / (3.f * area);
```
tallytalwar: Note: this calculated `area` is actually twice the area of the polygon. But the final centroid result is accurate.

blair1618: Correct, I thought about naming this differently but I couldn't easily express what this represented haha

tallytalwar: Thats fine. The end result is the same. + we save on 2 division operations :D

westnordost:
In the [implementation of it in Leaflet](https://github.com/Leaflet/Leaflet/blob/c10f405a112142b19785967ce0e142132a6095ad/src/layer/vector/Polygon.js#L78) on which the tangram-JS version of it says to be based on, I see a check to prevent a possible division by zero. This is missing in both tangram-JS and tangram-ES.

Perhaps it makes sense to include that? In the wild, there could always be area-type map features that consist of only one point or of several points but all of these are on the same spot. Surely, this is a mistake in the data, but it should not send tangram crashing.

westnordost:
Same for the centroid of polylines, probably

matteblair:
That's a fair point. I believe a polygon with zero area would produce a centroid of { NAN, NAN } in tangram-es currently (rather than a crash). We _could_ return the first point in the first ring in this case, but that's not necessarily the "center" of the polygon. The triangle { { 0, 0 }, { 1, 0 }, { 0, 0 } } has an area of zero but the first point is not the "center". Even so, it is probably more useful to return _some_ point on the polygon than _no_ point. I'll make this change. @bcamper we may want to do the same in tangram-js.

westnordost:
oh, right. So if area is 0, it could mean two things. To distinguish between the two, I guess one would need this (pseudo code):
```
if area = 0 
  if length of ring = 0
    return ring[0]
  else
    return point(0,0)
```

matteblair:
https://github.com/tangrams/tangram-es/pull/1221

PR #1146, opened 2016-12-06
Fix color parsing for numbers starting with non-digits

blair1618:
A recent change to our scene loading code that converts strings to doubles has exposed this bug in a color parsing function. `double`s need not start with a character that passes `isdigit()`.


blair1618:
Yeah, it's a pretty involved pipeline for all of the style parameters right now. I think it could help to pass the un-modified "node" type further down the pipe before making conversions.

PR #1149, opened 2016-12-06
Fix resolution of absolute file paths in iOS and macOS

blair1618:
In the iOS and macOS platform code that resolves file paths, absolute file paths were not being resolved correctly. The previous implementation simply appended the relative path to the end of the bundle resource path. This is correct if the file has a relative path, but for absolute paths the bundle path should not be used. NSURL is able to perform the correct resolution process. On iOS we do not typically access files using absolute paths, but on macOS our demo application uses absolute paths when a file argument is provided with `-f`, so that feature is currently broken.

### DIFF
--- a/ios/src/platform_ios.mm
+++ b/ios/src/platform_ios.mm
@@ -54,22 +54,31 @@ bool isContinuousRendering() {
 
 NSString* resolvePath(const char* _path) {
 
-    NSString* path = [NSString stringWithUTF8String:_path];
+    NSString* pathString = [NSString stringWithUTF8String:_path];
 
-    NSString* resources = [[NSBundle mainBundle] resourcePath];
-    NSString* fullBundlePath = [resources stringByAppendingPathComponent:path];
+    NSURL* resourceFolderUrl = [[NSBundle mainBundle] resourceURL];
+
+    NSURL* resolvedUrl = [NSURL URLWithString:pathString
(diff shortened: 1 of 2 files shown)


ios/src/platform_ios.mm
```diff
@@ -54,22 +54,33 @@ bool isContinuousRendering() {
 
 NSString* resolvePath(const char* _path) {
 
-    NSString* path = [NSString stringWithUTF8String:_path];
+    NSString* pathString = [NSString stringWithUTF8String:_path];
+
+    NSURL* resourceFolderUrl = [NSURL fileURLWithPath:[[NSBundle mainBundle] resourcePath]
```
blair1618: I see! If there aren't established practices against it, I would recommend that we just require all scene-related files be located in 'Resources', similar to how we require those files to be in 'assets' on Android.

msmollin: You can simplify this by calling `[[NSBundle mainbundle] resourceURL]` which produces a file-based URL to the directory.

msmollin: As for requiring Resources, this is fine for a start. Eventually we're going to need to deal with Asset Catalog usage which is not stored in Resources, but that's farther down the path.

blair1618:
Good catch @msmollin - updated!

PR #1152, opened 2016-12-08
Build Android native libraries through Gradle

matteblair:
Android Studio 2.2 and the associated Gradle components now include a mechanism to execute "external native builds" via CMake. This enables us to build all pieces of our Android SDK, including the native libraries, from a Gradle command. This also allows us to use the latest versions of the Android NDK, which are incompatible with our old Android build system.


blair1618:
Ignore this PR for now - I've got some wrestling to do with Travis CI.

android/tangram/build.gradle
```diff
@@ -14,44 +14,65 @@ version = VERSION_NAME
...
+      }
+    }
+  }
+
+  productFlavors {
+    slim {
+      // Default configuration
+    }
+    full {
+      externalNativeBuild.cmake.abiFilters 'armeabi', 'arm64-v8a', 'x86'
```
tallytalwar: Should we also be including `armv7` or its picked up by default?

blair1618: I said to ignore this PR! ;)

I was unsure about this too, but empirically it seems to pick up `armeabi-v7a` from the default configuration.

tallytalwar: Can not hold a man for his curiosity :).

blair1618:
Ready for review now! Before merging, though, we should merge https://github.com/tangrams/yaml-cpp/pull/8 and use the normal `tangram` branch of `yaml-cpp` here. I've also dropped Travis builds for iOS and macOS, in anticipation of moving those to CircleCI.

matteblair:
Ping @tallytalwar & @hjanetzek for review

scripts/travis/script_deploy_android_snapshot.sh
```diff
@@ -8,16 +8,8 @@ if [ "${PLATFORM}" = "android" ] && [ "${TRAVIS_PULL_REQUEST}" = "false" ] && [
...
 
-    # Build all android architectures (armeabi-v7a already build)
-    make android-native-lib ANDROID_ARCH=x86
-    make android-native-lib ANDROID_ARCH=armeabi
-    make android-native-lib ANDROID_ARCH=arm64-v8a
-    #### Currently Not building the following architectures
-    #make android-native-lib ANDROID_ARCH=x86_64
-    #make android-native-lib ANDROID_ARCH=mips
-    #make android-native-lib ANDROID_ARCH=mips64
+    make android-sdk
```
tallytalwar: So previously `arm7l` was done during the default build process and other architecture were build only during deployment.

It seems this will build, everything again including arm7l (in other words, arm7l will be builds twice).

android/tangram/build.gradle
```diff
@@ -14,44 +14,65 @@ version = VERSION_NAME
...
-    if (packageTask) { packageTask.include(['**/gdbserver', '**/gdb.setup']) }
+  buildTypes {
+    release {
+      externalNativeBuild {
+        cmake.arguments '-DCMAKE_BUILD_TYPE=Release'
+      }
+    }
+    debug {
+      externalNativeBuild {
+        cmake.arguments '-DCMAKE_BUILD_TYPE=Debug'
```
tallytalwar: Do these take care of debugging for demo app and also debugging on the device, which was disabled here: https://github.com/tangrams/tangram-es/pull/1152/commits/b859576c1f7268b35ce509f0d10c195ea2fdd489

android/tangram/build.gradle
```diff
@@ -14,44 +14,65 @@ version = VERSION_NAME
...
+    targetSdkVersion 25
     versionCode buildVersionCode()
     versionName VERSION_NAME
+    externalNativeBuild {
+      cmake {
+        targets 'tangram'
+        arguments '-DPLATFORM_TARGET=android',
+                  '-DANDROID_TOOLCHAIN=clang',
+                  '-DANDROID_STL=c++_shared'
+        cppFlags '-std=c++14',
```
tallytalwar: https://developer.android.com/ndk/guides/cpp-support.html

android/tangram/build.gradle
```diff
@@ -14,44 +14,65 @@ version = VERSION_NAME
...
     minSdkVersion 15
-    targetSdkVersion 24
+    targetSdkVersion 25
     versionCode buildVersionCode()
     versionName VERSION_NAME
+    externalNativeBuild {
+      cmake {
+        targets 'tangram'
+        arguments '-DPLATFORM_TARGET=android',
+                  '-DANDROID_TOOLCHAIN=clang',
```
matteblair: `clang` became the default very recently - I would prefer to leave this here in case someone is using an NDK from before that change.

android/tangram/build.gradle
matteblair: I suppose it couldn't hurt, though there are no runtime errors produced by using things from the `<atomic>` headers so it's unclear whether it is necessary.

tallytalwar: Yup I saw when I build, it did not complain. But I would link atomic only because documentation is suggesting it.

android/demo/build.gradle
```diff
@@ -1,37 +1,27 @@
...
-    manifest.srcFile 'AndroidManifest.xml'
-    java.srcDirs = ['src']
-    resources.srcDirs = ['src']
-    res.srcDirs = ['res']
     assets.srcDirs = ['../../scenes']
   }
-  buildTypes {
-    debug {
-      // applicationIdSuffix ".debug"
-      jniDebuggable true
```
matteblair: This is a good point about the "debuggable" flags, I'm not sure of the impact of these flags on either the library or demo module under the new build system, but I can try both and see.

android/tangram/build.gradle
matteblair: `atomic` is linked in `android.cmake` now.

matteblair:
I've set up debug configurations for the library and demo app in a way that _should_ allow debugging native code - that is, I've tested the same configuration in a minimal test project and was able to debug native code fro Android studio. However, I have not been able to successfully debug native code from Android Studio yet. Any further ideas are welcome!

matteblair:
I could squash more of these commits together, but there are actually quite a lot of distinctly meaningful changes here, despite the low number of lines added. I'd rather have many commits that describe these changes in accurate detail.

matteblair:
~ Ping for review approval ~

tallytalwar:
All this is good, I just want to play with the debugging settings in android studio, to see if I can figure out something here. Else will merge it this week.

Is this blocking you from something?

matteblair:
Not blocking any new features, I just enjoy using the new build system much more than the old one.

karimnaaji:
Are we moving to android ndk13b now and API25? It would be worth mentioning those somewhere if not already.

CMakeLists.txt
```diff
@@ -25,11 +25,6 @@ endif()
 
 include(${PROJECT_SOURCE_DIR}/toolchains/utils.cmake)
 
-# cmake output configuration
-set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
-set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
-set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin)
```
karimnaaji: If we remove that from here we need to update the readme in several places. Could this be android specific if only useful here?

Makefile
```diff
@@ -339,9 +269,9 @@ ios-framework-sim: cmake-ios-framework-sim
 
 ios-framework-universal: ios-framework ios-framework-sim
 	@mkdir -p ${IOS_FRAMEWORK_UNIVERSAL_BUILD_DIR}/${CONFIG}
-	@cp -r ${IOS_FRAMEWORK_BUILD_DIR}/lib/${CONFIG}/TangramMap.framework ${IOS_FRAMEWORK_UNIVERSAL_BUILD_DIR}/${CONFIG}
-	lipo ${IOS_FRAMEWORK_BUILD_DIR}/lib/${CONFIG}/TangramMap.framework/TangramMap \
-		${IOS_FRAMEWORK_SIM_BUILD_DIR}/lib/${CONFIG}/TangramMap.framework/TangramMap \
+	@cp -r ${IOS_FRAMEWORK_BUILD_DIR}/${CONFIG}-iphoneos/TangramMap.framework ${IOS_FRAMEWORK_UNIVERSAL_BUILD_DIR}/${CONFIG}
```
karimnaaji: For this, the readme would be obsolete on where to find the build targets.

CMakeLists.txt
matteblair: It's unclear to me why we would want to use these custom output locations. I'd prefer to keep this change but you are correct that I need to update the README as well.

scripts/travis/before_install.sh
```diff
@@ -5,18 +5,31 @@ set -o pipefail
...
-    echo "Cloning mindk..."
-    git clone --quiet --depth 1 --branch deploy-all-arch https://github.com/tangrams/mindk.git
-    export ANDROID_NDK=$PWD/mindk/android-ndk-r10e
+    # Note: the right way to download these packages is through the Android Studio SDK manager,
+    # these steps should be removed when/if ndk-bundle and cmake become available from the
+    # command-line SDK update tool.
+
+    # Download android ndk
+    echo "Downloading ndk..."
+    curl -L https://dl.google.com/android/repository/android-ndk-r13b-linux-x86_64.zip -o ndk.zip
```
karimnaaji: Is there any way we would be able to cache this in between builds?

tallytalwar: We can I think, will require to update travis configuration to use their caching, but based on the email they send me, it seems they copy the contents to a s3 bucket and download it from their in the next travis run, which will take equally long. So no use as such!

android/tangram/build.gradle
```diff
@@ -14,44 +7,51 @@ version = VERSION_NAME
...
+  externalNativeBuild {
+    cmake {
+      path '../../CMakeLists.txt'
+    }
   }
-}
 
-// Add gdb server to apk
-afterEvaluate {
-    Sync packageTask = project.getTasks().findByName("packageReleaseJniLibs")
```
matteblair: I don't believe these are in use, I'll remove those targets.

CMakeLists.txt
karimnaaji: Yeah CMake is a little annoying for outputs on desktop platforms, everything would at the same level (libraries and executable with resources), but it's ok to remove it, just makes things more clear on what is the executable (thinking of linux and raspberry pi mainly).

scripts/travis/before_install.sh
matteblair: Unfortunately there is no useful way to cache this: https://docs.travis-ci.com/user/caching/#Things-not-to-cache

karimnaaji: Aah that's annoying :/

matteblair:
This does effectively change our required NDK version, however we do not require an exact version of the NDK now. Instead we require the 'ndk bundle' package that is installed through Android Studio. 

This also updates our target Android API level to 25 (i.e. Android 7.1) but we don't currently specify our target API level in the documentation and it should not pose any constraints to users.

tallytalwar:
Agreed on he ndk version number @matteblair. But will be good to mention min android studio requirements which adds `ndk bundle` packages.

android/tangram/build.gradle
karimnaaji: I think the file we had can go away with it too, `toolchains/android.gdb.cmake`.

tallytalwar:
screenshot from 2016-12-21 13-51-05

Maybe we need some cmake configuration to bundle these libraries appropriately.

CMakeLists.txt
matteblair: Ah I see, that is a bit inconvenient. It turns out that these specific lines don't actually impact the Android build, so I'll revert this change.

Makefile
matteblair: This is reverted as well.

matteblair:
@tallytalwar Can you check over the updated README and merge this when you get a chance?

scripts/travis/before_install.sh
lygstate: I would suggest use appveyor for Android， would speed up a lot

PR #1157, opened 2016-12-09
Fix non-square bitmap images flipping

### LINKED ISSUE #1156: Marker setDrawable
wheh I call function setDrawable, I have exception 

> java.lang.ArrayIndexOutOfBoundsException: length=1152; index=1152

tallytalwar:
- Fixes #1156

### DIFF
--- a/core/src/marker/markerManager.cpp
+++ b/core/src/marker/markerManager.cpp
@@ -72,6 +72,11 @@ bool MarkerManager::setBitmap(MarkerID markerID, int width, int height, const un
     texture->setData(bitmapData, size);
 
     marker->setTexture(std::move(texture));
+
+    // The geometry is unchanged, but the mesh must be rebuilt because DynamicQuadMesh contains
+    // texture batches as part of its data.
+    buildGeometry(*marker, m_zoom);
+
     return true;
 }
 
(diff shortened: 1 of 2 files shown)


core/src/marker/markerManager.cpp
```diff
@@ -72,6 +72,11 @@ bool MarkerManager::setBitmap(MarkerID markerID, int width, int height, const un
     texture->setData(bitmapData, size);
 
     marker->setTexture(std::move(texture));
+
+    // The geometry is unchanged, but the mesh must be rebuilt because DynamicQuadMesh contains
+    // texture batches as part of its data.
+    buildGeometry(*marker, m_zoom);
```
tallytalwar: can you do a `return buildGeometry(*marker, m_zoom)` here?

blair1618: I wrote that at first, but I think it could actually be kind of misleading. For instance, if you call `setBitmap` on a marker before you set the styling or geometry, then the call would return `false` even though the bitmap was successfully assigned to the marker.

blair1618: Hmm I'd rather have the failure communicated from another function. To fully specify a marker, an application has to call `setStyling` and `setPoint`, `setPolyline`, or `setPolygon`, so to me it makes more sense to return an indication of failed geometry from the geometry setting functions.

blair1618: And this whole thing has reinforced to me how convoluted it can be to put a marker on the map. There's gotta be something better we can do :\

tallytalwar: Sure, sounds good.

PR #1160, opened 2016-12-09
Require BOOL returns in TGRecognizerDelegate protocol methods

### LINKED ISSUE #1159: Add BOOL return type requirement to TGRecognizerDelegate protocol methods
This will allow us to determine if we should continue the default interaction processing or not.

msmollin:
This allows us to know when to continue default interaction processing or not.

Fixes #1159

### DIFF
--- a/ios/src/TGMapViewController.h
+++ b/ios/src/TGMapViewController.h
@@ -48,13 +48,26 @@ NS_ASSUME_NONNULL_BEGIN
 
 @protocol TGRecognizerDelegate <NSObject>
 @optional
-- (BOOL)mapView:(TGMapViewController *)view recognizer:(UIGestureRecognizer *)recognizer didRecognizeSingleTapGesture:(CGPoint)location;
-- (BOOL)mapView:(TGMapViewController *)view recognizer:(UIGestureRecognizer *)recognizer didRecognizeDoubleTapGesture:(CGPoint)location;
-- (BOOL)mapView:(TGMapViewController *)view recognizer:(UIGestureRecognizer *)recognizer didRecognizeLongPressGesture:(CGPoint)location;
-- (BOOL)mapView:(TGMapViewController *)view recognizer:(UIGestureRecognizer *)recognizer didRecognizePanGesture:(CGPoint)location;
-- (BOOL)mapView:(TGMapViewController *)view recognizer:(UIGestureRecognizer *)recognizer didRecognizePinchGesture:(CGPoint)location;
-- (BOOL)mapView:(TGMapViewController *)view recognizer:(UIGestureRecognizer *)recognizer didRecognizeRotationGesture:(CGPoint)location;
-- (BOOL)mapView:(TGMapViewController *)view recognizer:(UIGestureRecognizer *)recognizer didRecognizeShoveGesture:(CGPoint)location;
+- (BOOL)mapView:(TGMapViewController *)view recognizer:(UIGestureRecognizer *)recognizer shouldRecognizeSingleTapGesture:(CGPoint)location;
+- (void)mapView:(TGMapViewController *)view recognizer:(UIGestureRecognizer *)recognizer didRecognizeSingleTapGesture:(CGPoint)location;
+
+- (BOOL)mapView:(TGMapViewController *)view recognizer:(UIGestureRecognizer *)recognizer shouldRecognizeDoubleTapGesture:(CGPoint)location;
+- (void)mapView:(TGMapViewController *)view recognizer:(UIGestureRecognizer *)recognizer didRecognizeDoubleTapGesture:(CGPoint)location;
+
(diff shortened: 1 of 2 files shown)


ios/src/TGMapViewController.mm
```diff
@@ -546,10 +549,11 @@ - (void)respondToShoveGesture:(UIPanGestureRecognizer *)shoveRecognizer
     // don't trigger shove on single touch gesture
     if ([shoveRecognizer numberOfTouches] == 2) {
         if (self.gestureDelegate && [self.gestureDelegate respondsToSelector:@selector(recognizer:didRecognizeShoveGesture:)]) {
-            [self.gestureDelegate mapView:self recognizer:shoveRecognizer didRecognizeShoveGesture:displacement];
-        } else {
-            self.map->handleShoveGesture(displacement.y);
+            if (![self.gestureDelegate mapView:self recognizer:shoveRecognizer didRecognizeShoveGesture:displacement]) {
```
karimnaaji: Should this one be out of the test `if ([shoveRecognizer numberOfTouches] == 2) {`? Implementation developer might still want to decide on that.

karimnaaji: Exactly, otherwise there is a restriction on the number of touches made by this controller.

msmollin: Gotcha. So that would change how this one particular gesture delegate works compared to the others, and I think might make it a little confusing: in some instances returning `true` to the delegate function, tangram wouldn't do processing, whereas in other cases it would. 

Maybe a better solution is to allow the number of touches to be customized? I think it would be preferable to have a unified delegate callback system wherein there code only sends the delegate callbacks when tangram also could process it internally. Thoughts?

blair1618:
Can we document the meaning of the return value from these functions? On Android returning `true` means "I've handled this event, nothing else needs to be done" and `false` means "This event isn't fully handled, continue handling it". It seems like the return values here might mean the opposite of that? It's not totally clear, hence the request for documentation ;)

ios/src/TGMapViewController.mm
karimnaaji: I would say that the idea behind this is more:
Implementation developer: "Subscribes to the view to know whenever a gesture is recognized, decides whether tangram should follow up on the gesture internally" -> Tangram: "Handle or not the gesture internally".

msmollin: Ah I am following now. Part of the issue I'm having here is that a "shove" gesture is not a gesture built into iOS, so users aren't necessarily going to know how to react to what a shove gesture actually is. 

For that kind of a pattern, Apple normally separates it into separate delegate function calls. A good example is `UITableViewDelegate's` different `should*` functions versus `did*` functions - https://developer.apple.com/reference/uikit/uitableviewdelegate

So maybe that's something we should do here as well? We can have a "shouldRecognizeShoveGesture" and a "didRecognizeShoveGesture". The `shouldRecognize` functions would pass along things like "numberOfTouches` and other important parameters, and expect a boolean return. If the user returns `true` or doesn't implement the `shouldRecognize` then Tangram will handle the recognition, and then call the "didRecognize".

I think this will also clear up @blair1618's question below.

msmollin:
Updated with new `shouldRecognize` variants of the existing codebase. @blair1618 / @karimnaaji take a look at the update. Also note @karimnaaji that I rebased this against master so you'll want to make sure you reset your local git accordingly.

ios/src/TGMapViewController.h
```diff
@@ -48,12 +48,25 @@ NS_ASSUME_NONNULL_BEGIN
...
+- (BOOL)mapView:(TGMapViewController *)view recognizer:(UIGestureRecognizer *)recognizer shouldRecognizeSingleTapGesture:(CGPoint)location;
 - (void)mapView:(TGMapViewController *)view recognizer:(UIGestureRecognizer *)recognizer didRecognizeSingleTapGesture:(CGPoint)location;
+
+- (BOOL)mapView:(TGMapViewController *)view recognizer:(UIGestureRecognizer *)recognizer shouldRecognizeDoubleTapGesture:(CGPoint)location;
 - (void)mapView:(TGMapViewController *)view recognizer:(UIGestureRecognizer *)recognizer didRecognizeDoubleTapGesture:(CGPoint)location;
+
+- (BOOL)mapView:(TGMapViewController *)view recognizer:(UIGestureRecognizer *)recognizer shouldRecognizeLongPressGesture:(CGPoint)location;
 - (void)mapView:(TGMapViewController *)view recognizer:(UIGestureRecognizer *)recognizer didRecognizeLongPressGesture:(CGPoint)location;
+
+- (BOOL)mapView:(TGMapViewController *)view recognizer:(UIGestureRecognizer *)recognizer shouldRecognizePanGesture:(CGPoint)location;
```
karimnaaji: This is taking a location from the interface but is called with a `displacement`.

msmollin: Hmmm I can certainly rename that, although that is what was previously implemented?

msmollin:
@karimnaaji updated per comment.

Issue #1066, opened 2016-11-04
Feature selection for markers


karimnaaji:
We currently don't return markers while selecting features. We need to draw the markers in the feature selection draw pass and return their id in the feature selection callback.

PR #1161, opened 2016-12-09
Marker selection

karimnaaji:
Fixes https://github.com/tangrams/tangram-es/issues/1066
Fixes https://github.com/tangrams/tangram-es/issues/1163


core/src/marker/markerManager.cpp
```diff
@@ -335,13 +336,35 @@ bool MarkerManager::buildGeometry(Marker& marker, int zoom) {
...
 
+    marker.setSelectionColor(selectionColor);
     marker.setMesh(styler->style().getID(), zoom, styler->build());
 
     return true;
 }
 
+const Marker* MarkerManager::getMarkerOrNullBySelectionColor(uint32_t selectionColor) const {
+    for (const auto& marker : m_markers) {
+        if (marker->selectionColor() == selectionColor) {
```
westnordost: Why is it selection **color**? I'd expect the id of a marker to be of relevance for selection, not the color.

westnordost: Ah, okay

core/src/selection/selectionQuery.h
```diff
@@ -0,0 +1,43 @@
...
+                 const TileManager& _tileManager, const Labels& _labels) const;
+
+    QueryType type() const { return m_type; }
+
+private:
+    glm::vec2 m_position;
+    QueryCallback m_queryCallback;
+    QueryType m_type;
+
+    static std::map<size_t, uint32_t> s_colorCache;
```
matteblair: This doesn't seem like the right data structure for this job. Static storage could also create problems with multiple map instances in a single process (even though it is regularly cleared). Is reading the framebuffer really so slow that we need to cache these results?

core/src/tangram.cpp
```diff
@@ -432,8 +427,16 @@ void Map::render() {
     // Run render-thread tasks
     impl->renderState.jobQueue.runJobs();
 
+    bool hasGeometryFeatureQuery = drawSelectionBuffer;
+    bool hasMarkerQuery = drawSelectionBuffer;
+
+    for (const auto& selectionQuery : impl->selectionQueries) {
+        if (selectionQuery.type() == QueryType::feature) { hasGeometryFeatureQuery = true; }
```
matteblair: Need a case for `QueryType::label` I believe.

core/src/tangram.cpp
```diff
@@ -442,82 +445,28 @@ void Map::render() {
...
-            auto label = impl->labels.getLabel(color);
-            if (!label.first) {
-                labelQuery.onLabelSelection(nullptr);
-                continue;
-            }
-            auto props = label.second->getSelectionFeature(label.first->options().featureId);
-            if (!props) {
-                labelQuery.onLabelSelection(nullptr);
-                continue;
+            if (hasMarkerQuery) {
```
matteblair: This is an interesting and substantial design decision. If a client makes a query for a feature or label (but not a marker) on a point that is covered by a marker, the result can be a feature or label that is _under_ the marker. This is potentially useful, but diverges from the WYSIWYG behavior of this feature.

I'm not sure yet whether this is good or bad.

core/src/selection/selectionQuery.cpp
```diff
@@ -0,0 +1,109 @@
...
+        }
+
+        auto marker = _markerManager.getMarkerOrNullBySelectionColor(color);
+
+        if (!marker) {
+            cb(nullptr);
+            return;
+        }
+
+        MarkerPickResult markerResult(marker->id(), {{m_position.x, m_position.y}});
```
matteblair: It may be more useful here to return the origin or the bounding-box-center of the marker as the position.

core/src/selection/selectionQuery.h
```diff
@@ -0,0 +1,43 @@
...
+
+    void process(const View& _view, const FrameBuffer& _framebuffer, const MarkerManager& _markerManager,
+                 const TileManager& _tileManager, const Labels& _labels) const;
+
+    QueryType type() const { return m_type; }
+
+private:
+    glm::vec2 m_position;
+    QueryCallback m_queryCallback;
+    QueryType m_type;
```
matteblair: The type member may be redundant with the variant - you can check the type of a variant with `is<Type>()`.

matteblair: Alternatively you could keep the type member and use a simple union for the query instead :)

karimnaaji: True! Will update that.

core/src/tangram.cpp
karimnaaji: The problem with `QueryType::label` is that they are draw in `onBeginDrawFrame`, that is commonly needed by all the styles.

core/src/selection/selectionQuery.h
karimnaaji: `glReadPixels` is quite known for being slow (mainly why PBO started to exist), but haven't benchmarked it in here; I just assumed that reducing calls to it would be good practice. A common pattern would be querying the system for the same position for the three kind of queries.
You're right about the static storage, but why isn't it the right data structure though? It's keyed by the hash of a 2d screen position.

matteblair: I see, then it is probably prudent to cache the results. 

It seems like a poor data structure because in all of our current use cases the map will be storing exactly one entry. I would probably use a vector or forward list instead and do a linear search over the entries. 

As for making the cache non-static, we can allow the call site to create a cache object on the stack and pass it into the processing calls as it loops over the queries.

core/src/tangram.cpp
karimnaaji: Thanks for bringing that up, to me the most consistent selection is whatever makes us the closest to WYSIWYG behavior, so I'll just remove those conditionals.

core/src/selection/selectionQuery.cpp
```diff
@@ -0,0 +1,121 @@
...
+            cb(nullptr);
+            return;
+        }
+
+        glm::dvec2 bbCenter = marker->bounds().center();
+        glm::dvec2 lonLat = _view.getMapProjection().MetersToLonLat(bbCenter);
+        bool clipped;
+        glm::vec2 screenPosition = _view.lonLatToScreenPosition(lonLat.x, lonLat.y, clipped);
+
+        if (clipped) {
```
karimnaaji: Hmmm, I don't know, if a polygon centroid goes out of screen for example, I'm not sure what value we should return in that case, because the clipped one would be quite irrelevant :/

core/src/selection/selectionQuery.cpp
```diff
@@ -0,0 +1,121 @@
...
+            cb(nullptr);
+            return;
+        }
+
+        MarkerPickResult markerResult(marker->id(), {{screenPosition.x, screenPosition.y}});
```
karimnaaji: Yep good point! And this solves the issue mentioned above, since in that case the `lnglatToScreen` returns false when clipped.

karimnaaji: Actually, it is quite misleading to return the `LngLat`, for polylines and polygons, the center of the bounding box is not that relevant. And for the frame delay, the `LngLat` can be retrieved from the marker objects holding those position on the platform layer on Android.

matteblair: Polyline and polygon markers are seeming more and more distant in functionality from point markers, I am tempted to make them a separate type of object - but that is for another time.

I think it is important to return the `LngLat` in the case of point markers for the same reason we return it for labels, a client cannot get the "anchor" for a point marker by simply converting the screen coordinates that they queried. We could return both sets of coordinates.

matteblair:
:+1: Cool, I'll check this out today.

PR #1165, opened 2016-12-12
Run iOS builds on CircleCI

blair1618:
Currently, CircleCI provides much lower latency for starting builds on macOS machines than Travis.

Our build process requires quite a few tweaks to the Circle build configuration, but it seems to work in principle.

This allows us to remove the iOS and macOS build jobs from our Travis configuration, ~~however I've also added a new linux configuration on Travis to build with the LLVM toolchain and ensure that we catch any compiler divergences (it may not be necessary but we can try it).~~

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,4 +1,30 @@
-general:
-  branches:
-    only:
-      - /circle-.*/
+machine:
+  xcode:
+    version: 7.3
+  environment:
+    XCODE_PROJECT: build/tangram.xcodeproj
+    XCODE_SCHEME: phony # Dummy value, Circle won't run without a scheme.
+checkout:
+  post:
+    - git submodule sync
+    - git submodule update --init --recursive
+test:
+  pre:
+    - set -o pipefail &&
+      mkdir build &&
+      cd build &&
+      cmake ..
+      -DPLATFORM_TARGET=ios.framework
+      -DIOS_PLATFORM=SIMULATOR
+      -DCMAKE_TOOLCHAIN_FILE=toolchains/iOS.toolchain.cmake
+      -DBUILD_IOS_FRAMEWORK=TRUE
+      -G Xcode
+  override:
+    - set -o pipefail &&
+      xcodebuild
+      -project build/tangram.xcodeproj
+      -target TangramMap
+      -configuration Release |
+      xcpretty
+  post:
+    - mv build/lib/Release/TangramMap.framework ${CIRCLE_ARTIFACTS}/
--- a/circle.yml
+++ b/circle.yml
@@ -2,14 +2,15 @@ machine:
   xcode:
     version: 8.1
   environment:
+    # Dummy values, Circle won't run without a project and scheme.
     XCODE_PROJECT: build/tangram.xcodeproj
-    XCODE_SCHEME: phony # Dummy value, Circle won't run without a scheme.
+    XCODE_SCHEME: phony
 checkout:
   post:
     - git submodule sync
     - git submodule update --init --recursive
 test:
   override:
-    - make ios | xcpretty
+    - make ios
   post:
     - mv build/ios-framework-universal/Release/TangramMap.framework ${CIRCLE_ARTIFACTS}/
--- a/circle.yml
+++ b/circle.yml
@@ -1,6 +1,6 @@
 machine:
   xcode:
-    version: 7.3
+    version: 8.1
   environment:
     XCODE_PROJECT: build/tangram.xcodeproj
     XCODE_SCHEME: phony # Dummy value, Circle won't run without a scheme.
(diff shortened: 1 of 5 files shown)


blair1618:
I've decided against adding a linux job using LLVM on Travis. Due to the software on Travis containers being a bit antiquated, it seems to be extraordinarily difficult to get a working C++ toolchain with clang and libc++. LLVM _does_ host apt packages for clang and libc++ but Travis is just using outdated URLs in their PPA listing (see https://github.com/travis-ci/apt-source-whitelist/pull/288).

tallytalwar:
Thats fine, but I thought we had clang builds on travis linux at some point.

circle.yml
```diff
@@ -1,4 +1,30 @@
-general:
-  branches:
-    only:
-      - /circle-.*/
+machine:
+  xcode:
+    version: 7.3
```
msmollin: Apple just released Xcode 8.2 - we should probably test this against Xcode 8.1 at the very least. Circle normally has some instability in their containers when they upgrade to latest xcode so thus why I'm not saying to use 8.2... yet. I don't think we're planning on supporting anything earlier than iOS 9 with the iOS SDK, or OS X 10.11, so Xcode 8.1 should work just fine.

toolchains/iOS.toolchain.cmake
```diff
@@ -114,15 +114,11 @@ else (${IOS_PLATFORM} STREQUAL "OS")
 endif (${IOS_PLATFORM} STREQUAL "OS")
 
 # Setup iOS developer location unless specified manually with CMAKE_IOS_DEVELOPER_ROOT
-# Note Xcode 4.3 changed the installation location, choose the most recent one available
-set (XCODE_POST_43_ROOT "/Applications/Xcode.app/Contents/Developer/Platforms/${IOS_PLATFORM_LOCATION}/Developer")
```
msmollin: 👍

blair1618:
@tallytalwar We might have, yes. Compiling with Clang is quite easy, the Linux containers have clang-3.4 installed by default! However they only have libstdc++ for the stdlib. Installing libc++ turns out to be nearly impossible and our build is now incompatible with the installed version of libstdc++.

circle.yml
```diff
@@ -1,4 +1,30 @@
...
+    - git submodule update --init --recursive
+test:
+  pre:
+    - set -o pipefail &&
+      mkdir build &&
+      cd build &&
+      cmake ..
+      -DPLATFORM_TARGET=ios.framework
+      -DIOS_PLATFORM=SIMULATOR
+      -DCMAKE_TOOLCHAIN_FILE=toolchains/iOS.toolchain.cmake
```
msmollin: If you're targeting the ios.framework target, would you want iOS.framework.cmake here instead?

circle.yml
```diff
@@ -1,4 +1,30 @@
...
+    XCODE_PROJECT: build/tangram.xcodeproj
+    XCODE_SCHEME: phony # Dummy value, Circle won't run without a scheme.
+checkout:
+  post:
+    - git submodule sync
+    - git submodule update --init --recursive
+test:
+  pre:
+    - set -o pipefail &&
+      mkdir build &&
```
karimnaaji: It's not possible to run the `make` command that we already have for this? `make ios-framework-simulator`

circle.yml
```diff
@@ -1,4 +1,30 @@
...
+      -G Xcode
+  override:
+    - set -o pipefail &&
+      xcodebuild
+      -project build/tangram.xcodeproj
+      -target TangramMap
+      -configuration Release |
+      xcpretty
+  post:
+    - mv build/lib/Release/TangramMap.framework ${CIRCLE_ARTIFACTS}/
```
karimnaaji: Is it possible to put the artifact of the universal framework in any way? Have you tried running our command `make ios` on ci?

blair1618: If we build the universal framework then yes, we can put that among the artifacts. 

I tried running `make ios` at first and was not successful, so for my own sanity I wrote the commands explicitly. I just tried it again and it still runs into problems: https://circleci.com/gh/tangrams/tangram-es/77. If you think this can be easily fixed then I'd definitely prefer writing a single `make` command!

circle.yml
blair1618: `iOS.toolchain.cmake` is the toolchain file, which locates the sysroot, compilers, etc. `iOS.framework.cmake` is the build script, which contains the steps of the build. We end up using both through this command.

circle.yml
karimnaaji: @blair1618 it seems to be going through in https://github.com/tangrams/tangram-es/pull/1172, the SDK correctly found when using iPhone OS.

msmollin:
Just FYI - I approved changes from previous just to clear the review - waiting on the merge in from @karimnaaji's branch.

PR #1166, opened 2016-12-12
fix InvalidNode exception

hjanetzek:
- font is taken as const& so looking up 'url'
  in it may return an invalid node.

@karimnaaji sorry this took me a while to get to.. tomorrow comes the full 'more const&' branch


hjanetzek:
There is also an issue for yaml-cpp regarding problems with Node::Type() throwing on invalid nodes https://github.com/jbeder/yaml-cpp/issues/451

PR #1170, opened 2016-12-13
Faster texture flipping + Disable some image formats

hjanetzek:
Maybe we could keep BMP for live texture editing with Paintbrush - the others are unlikely to be supported by JS
https://en.wikipedia.org/wiki/Comparison_of_web_browsers#Image_format_support

### DIFF
--- a/core/src/gl/texture.cpp
+++ b/core/src/gl/texture.cpp
@@ -8,6 +8,12 @@
 #include "tangram.h"
 #include "log.h"
 
+// Enable only JPEG, PNG, GIF, TGA and PSD
+#define STBI_NO_BMP
+#define STBI_NO_HDR
+#define STBI_NO_PIC
+#define STBI_NO_PNM
+
 #define STB_IMAGE_IMPLEMENTATION
 #include "stb_image.h"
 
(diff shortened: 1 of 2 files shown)


core/src/gl/texture.cpp
```diff
@@ -8,6 +8,14 @@
 #include "tangram.h"
 #include "log.h"
 
+// Only JPEG, PNG and GIF
+#define STBI_NO_BMP
+#define STBI_NO_PSD
+#define STBI_NO_TGA
```
tallytalwar: Should we though? I agree that for compatibility with JS, we can. But if tangram-es is being used by a game developer exporting textures from photoshop or other softwares, PSD, BMP, TGA are some of the textures artists, game developers, students use all the time.

karimnaaji: If they are not supported in JS, supporting them here breaks scene file compatibility because of file extensions.

tallytalwar: Yes I agree with that. But this can be a ES only thing. Users of ES may not be users of JS.

hjanetzek: TGA might be a good candidate to keep as it's likely faster to decode than PNG. For the other formats I don't see any real benefit for having them and it should be easy enough to export to our supported formats.

blair1618: From discussion in slack: let's keep PSD and TGA  and remove BMP, HDR, PIC, and PNM.

hjanetzek: @blair1618 ok, updated

tallytalwar:
Do we get a significant speed-up by not using stb flipping api?

hjanetzek:
150ms on Z3

PR #1172, opened 2016-12-13
Pick SDK for iPhone OS from toolchain

karimnaaji:


blair1618:
Nice. One issue: the output from xcodebuild is so verbose that it gets truncated in the CircleCI output. Can we just pipe the output of `make` into `xcpretty`?

blair1618:
Cool, seems to format the xcode output correctly - we lose the CMake output though :\

I'll merge this into the other branch and maybe fiddle with the output formatting a bit more.

karimnaaji:
Yep I noticed that :/, we can probably split up the commands as it is in the original branch and only give `xcodebuild` to `xcpretty`.

PR #1173, opened 2016-12-13
Add Android API for scene loading with scene updates

tallytalwar:
- MapView provides an overloaded `getMapAsync` which accepts an ArrayList of SceneUpdates (new
Class).
- MapController::loadSceneFile provides an overloaded method to accept an ArrayList of SceneUpdates
- MapController::queueSceneUpdates provides an overloaded method to accept an ArrayList of
SceneUpdates

### DIFF
--- a/android/tangram/jni/jniExports.cpp
+++ b/android/tangram/jni/jniExports.cpp
@@ -110,29 +110,21 @@ extern "C" {
         delete map;
     }
 
-    JNIEXPORT void JNICALL Java_com_mapzen_tangram_MapController_nativeLoadScene(JNIEnv* jniEnv, jobject obj, jlong mapPtr, jstring path) {
+    JNIEXPORT void JNICALL Java_com_mapzen_tangram_MapController_nativeLoadScene(JNIEnv* jniEnv, jobject obj, jlong mapPtr, jstring path, jobjectArray jcomponentPaths, jobjectArray jcomponentValues) {
         assert(mapPtr > 0);
         auto map = reinterpret_cast<Tangram::Map*>(mapPtr);
         const char* cPath = jniEnv->GetStringUTFChars(path, NULL);
-        map->loadScene(resolveScenePath(cPath).c_str());
-        jniEnv->ReleaseStringUTFChars(path, cPath);
-    }
-
-    JNIEXPORT void JNICALL Java_com_mapzen_tangram_MapController_nativeLoadSceneWithUpdates(JNIEnv* jniEnv, jobject obj, jlong mapPtr, jstring path, jobjectArray jsceneUpdatePaths, jobjectArray jsceneUpdateValues) {
-        assert(mapPtr > 0);
-        size_t n_sceneUpdates = (jsceneUpdatePaths == NULL)? 0 : jniEnv->GetArrayLength(jsceneUpdatePaths);
+        size_t n_sceneUpdates = (jcomponentPaths == NULL)? 0 : jniEnv->GetArrayLength(jcomponentPaths);
 
         std::vector<Tangram::SceneUpdate> sceneUpdates;
         for (size_t i = 0; i < n_sceneUpdates; i++) {
-            jstring path = (jstring) (jniEnv->GetObjectArrayElement(jsceneUpdatePaths, i));
-            jstring value = (jstring) (jniEnv->GetObjectArrayElement(jsceneUpdateValues, i));
+            jstring path = (jstring) (jniEnv->GetObjectArrayElement(jcomponentPaths, i));
+            jstring value = (jstring) (jniEnv->GetObjectArrayElement(jcomponentValues, i));
             sceneUpdates.emplace_back(stringFromJString(jniEnv, path), stringFromJString(jniEnv, value));
             jniEnv->DeleteLocalRef(path);
             jniEnv->DeleteLocalRef(value);
         }
 
-        auto map = reinterpret_cast<Tangram::Map*>(mapPtr);
-        const char* cPath = jniEnv->GetStringUTFChars(path, NULL);
         map->loadScene(resolveScenePath(cPath).c_str(), false, sceneUpdates);
         jniEnv->ReleaseStringUTFChars(path, cPath);
     }
@@ -473,6 +465,25 @@ extern "C" {
         map->queueSceneUpdate(cPath, cValue);
         jnienv->ReleaseStringUTFChars(path, cPath);
         jnienv->ReleaseStringUTFChars(value, cValue);
+    }
+
+    JNIEXPORT void JNICALL Java_com_mapzen_tangram_MapController_nativeQueueSceneUpdates(JNIEnv* jniEnv, jobject obj, jlong mapPtr, jobjectArray jcomponentPaths, jobjectArray jcomponentValues) {
+        assert(mapPtr > 0);
+        size_t n_sceneUpdates = (jcomponentPaths == NULL)? 0 : jniEnv->GetArrayLength(jcomponentPaths);
+
+        std::vector<Tangram::SceneUpdate> sceneUpdates;
+        for (size_t i = 0; i < n_sceneUpdates; i++) {
+            jstring path = (jstring) (jniEnv->GetObjectArrayElement(jcomponentPaths, i));
+            jstring value = (jstring) (jniEnv->GetObjectArrayElement(jcomponentValues, i));
+            sceneUpdates.emplace_back(stringFromJString(jniEnv, path), stringFromJString(jniEnv, value));
+            jniEnv->DeleteLocalRef(path);
+            jniEnv->DeleteLocalRef(value);
+        }
+
+        if (sceneUpdates.empty()) { return; }
+
+        auto map = reinterpret_cast<Tangram::Map*>(mapPtr);
+        map->queueSceneUpdate(sceneUpdates);
     }
 
     JNIEXPORT void JNICALL Java_com_mapzen_tangram_MapController_nativeApplySceneUpdates(JNIEnv* jnienv, jobject obj, jlong mapPtr) {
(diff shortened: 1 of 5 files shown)


android/tangram/src/com/mapzen/tangram/MapController.java
```diff
@@ -747,6 +768,18 @@ public void queueSceneUpdate(String componentPath, String value) {
...
 
     /**
+     * Enqueue a scene component update with its corresponding YAML node value
+     * @param componentPath The YAML component path delimited by a '.' (example "scene.animated")
+     * @param value A YAML valid string (example "{ property: true }" or "true")
+     */
+    public void queueSceneUpdate(ArrayList<SceneUpdate> sceneUpdates) {
+        checkPointer(mapPointer);
+        for(SceneUpdate sceneUpdate : sceneUpdates) {
+            nativeQueueSceneUpdate(mapPointer, sceneUpdate.path, sceneUpdate.value);
```
karimnaaji: Don't we want to make it a single call here `nativeQueueSceneUpdates(long mapPtr, String[] sceneUpdatePaths, String[] sceneUpdateValues)`? It would reduce jumping between platform code and core library.

android/tangram/src/com/mapzen/tangram/SceneUpdate.java
```diff
@@ -0,0 +1,24 @@
...
+    public String path;
+    public String value;
+
+    public SceneUpdate() { this("", ""); }
+
+    public SceneUpdate(String path, String value) {
+        set(path, value);
+    }
+
+    public SceneUpdate set(String p, String v) {
```
karimnaaji: If we make this public we should be more explicit on the names variable names `p` and `v`. 
I'm not sure we need the default constructor `SceneUpdate()` and this setter since the variable are public.

android/tangram/src/com/mapzen/tangram/SceneUpdate.java
```diff
@@ -0,0 +1,24 @@
...
+        path = p;
+        value = v;
+        return this;
```
karimnaaji: Why the return?

android/tangram/src/com/mapzen/tangram/MapController.java
tallytalwar: Good point,  will update.

android/tangram/src/com/mapzen/tangram/SceneUpdate.java
tallytalwar: I was trying something else before.  Will update this.

android/tangram/jni/jniExports.cpp
```diff
@@ -118,6 +118,25 @@ extern "C" {
         jniEnv->ReleaseStringUTFChars(path, cPath);
     }
 
+    JNIEXPORT void JNICALL Java_com_mapzen_tangram_MapController_nativeLoadSceneWithUpdates(JNIEnv* jniEnv, jobject obj, jlong mapPtr, jstring path, jobjectArray jsceneUpdatePaths, jobjectArray jsceneUpdateValues) {
```
blair1618: Seems like we could merge this with the method above and just pass a null array when no updates are given.

android/tangram/src/com/mapzen/tangram/MapController.java
```diff
@@ -229,6 +230,26 @@ public void loadSceneFile(String path) {
     }
 
     /**
+     * Load a new scene file
+     * @param path Location of the YAML scene file within the application assets
+     * @param sceneUpdates List of path:value pairs to be applied when loading this scene
+     */
+    public void loadSceneFile(String path, ArrayList<SceneUpdate> sceneUpdates) {
```
blair1618: These methods should take a `List<SceneUpdate>`, the `List` interface should be preferred to concrete types when we don't need the specific type.

blair1618: We should also check this for `null`.

android/tangram/src/com/mapzen/tangram/SceneUpdate.java
```diff
@@ -0,0 +1,24 @@
...
+/**
+ * {@code SceneUpdate} Represents a DataStructure to specify a yaml path and the corresponding value for a Scene Update.
+ */
+
+public class SceneUpdate {
+
+    public String path;
+    public String value;
+
+    public SceneUpdate() { this("", ""); }
```
blair1618: Recommend removing this constructor.

android/tangram/src/com/mapzen/tangram/SceneUpdate.java
```diff
@@ -0,0 +1,24 @@
+package com.mapzen.tangram;
+
+/**
+ * {@code SceneUpdate} Represents a DataStructure to specify a yaml path and the corresponding value for a Scene Update.
+ */
+
+public class SceneUpdate {
+
+    public String path;
+    public String value;
```
blair1618: Recommend making these members private and adding getters.

android/tangram/src/com/mapzen/tangram/SceneUpdate.java
blair1618: Recommend removing the setter and only setting members in the constructor.

android/tangram/src/com/mapzen/tangram/MapController.java
blair1618: Recommend using the `SceneUpdate` type in the older `queueSceneUpdate` method as well.

android/tangram/src/com/mapzen/tangram/SceneUpdate.java
```diff
@@ -0,0 +1,24 @@
...
+
+    public SceneUpdate() { this("", ""); }
+
+    public SceneUpdate(String path, String value) {
```
blair1618: Recommend adding documentation explaining the meaning of these arguments.

android/tangram/src/com/mapzen/tangram/MapView.java
```diff
@@ -65,7 +81,11 @@ public void getMapAsync(@NonNull final OnMapReadyCallback callback,
             @SuppressWarnings("WrongThread")
             protected Boolean doInBackground(Void... params) {
                 mapInstance.init();
-                mapInstance.loadSceneFile(sceneFilePath);
+                if (sceneUpdates.size() > 0) {
```
blair1618: This branch should be unnecessary if we allow empty or `null` lists as the second argument to `loadSceneFile`.

tallytalwar:
Addressed all comments here: 7353bbe

karimnaaji:
Comments got wiped out of the pull request: https://github.com/tangrams/tangram-es/commit/7353bbe7d80eef064b37178c02ab4dcaf75c46d2#commitcomment-20205669. Merging once updated.

tallytalwar:
Ohh, thanks @karimnaaji . Updated.

karimnaaji:
@ecgreb this one needs your approval! It's good to merge otherwise.

android/tangram/src/com/mapzen/tangram/MapController.java
```diff
@@ -738,12 +760,27 @@ public void setDebugFlag(DebugFlag flag, boolean on) {
...
-        nativeQueueSceneUpdate(mapPointer, componentPath, value);
+        nativeQueueSceneUpdate(mapPointer, sceneUpdate.getPath(), sceneUpdate.getValue());
+    }
+
+    /**
+     * Enqueue a scene component update with its corresponding YAML node value
+     * @param sceneUpdates List of {@code SceneUpdate}
+     */
+    public void queueSceneUpdate(List<SceneUpdate> sceneUpdates) {
+        checkPointer(mapPointer);
```
ecgreb: It would be nice to provide some input validation on `sceneUpdates` param. Perhaps throw an `IllegalArgumentException` if it is empty or `null`?

android/tangram/src/com/mapzen/tangram/MapController.java
```diff
@@ -222,9 +224,29 @@ static MapController getInstance(GLSurfaceView view) {
...
+     * Load a new scene file
+     * @param path Location of the YAML scene file within the application assets
+     * @param sceneUpdates List of {@code SceneUpdate}
+     */
+    public void loadSceneFile(String path, List<SceneUpdate> sceneUpdates) {
+
+        String[] updateStrings = null;
+
+        if (sceneUpdates != null) {
+            updateStrings = new String[sceneUpdates.size() * 2];
```
ecgreb: The logic to build `updateStrings` is repeated again below in `queueSceneUpdate(...)`. Could be extracted into a private method?

android/tangram/src/com/mapzen/tangram/MapController.java
tallytalwar: Hmm, tangram will be handling a `null` or `empty` `sceneUpdates` gracefully, through the jni code here: https://github.com/tangrams/tangram-es/blob/4b543616be5d1bebc1cfc25b56e4d8037ad61088/android/tangram/jni/jniExports.cpp#L470-L483.

Am I missing java thingy, which will help here?

android/tangram/src/com/mapzen/tangram/MapController.java
tallytalwar: Addressed here: https://github.com/tangrams/tangram-es/pull/1173/commits/c3622852bf4af7e4d05d160f42c7ba56c4dd0080

android/tangram/src/com/mapzen/tangram/MapController.java
ecgreb: No the code won't break but it could be helpful to the developer if we provided some explicit feedback rather than failing silently in the case of invalid input.

Also passing data across the JNI is expensive. If we can avoid that by checking the input in Java first that would be preferable.

ecgreb: 👍

tallytalwar:
With an update to arguments for `queueSceneUpdates`, this now is a API breaking change (for the next release).

Will rebase this with master and merge.

PR #1174, opened 2016-12-13
Relative paths resolved against relative paths remain relative.

blair1618:
This is a bit weird and should maybe be treated as an error later, but it solves an immediate bug on macOS and iOS.

Technically this doesn't violate RFC 1808 or RFC 3986 because neither covers this specific case.

### DIFF
--- a/core/src/util/url.cpp
+++ b/core/src/util/url.cpp
@@ -241,18 +241,14 @@ Url Url::resolve(const Url& b, const Url& r) {
         } else {
             // Merge the relative path with the base path.
             if (b.hasNetLocation() && !b.hasPath()) {
-                t.buffer.append("/");
+                t.buffer += '/';
                 t.buffer.append(r.buffer, r.parts.path.start, r.parts.path.count);
             } else {
                 // Add the base URL's path.
                 t.buffer.append(b.buffer, b.parts.path.start, b.parts.path.count);
-                // Remove the last segment of the base URL's path.
-                size_t end = removeLastSegmentFromRange(t.buffer, t.parts.path.start, t.buffer.size());
-                t.buffer.resize(end);
+                // Remove the last segment of the base URL's path, up to the last '/'.
+                while (!t.buffer.empty() && t.buffer.back() != '/') { t.buffer.pop_back(); }
                 // Add the relative URL's path.
-                if (r.buffer[r.parts.path.start] != '/') {
-                    t.buffer += '/';
-                }
                 t.buffer.append(r.buffer, r.parts.path.start, r.parts.path.count);
             }
         }
--- a/tests/unit/urlTests.cpp
+++ b/tests/unit/urlTests.cpp
@@ -174,6 +174,27 @@ TEST_CASE("Resolve a URL against a relative base URL", "[Url]") {
 
 }
 
+TEST_CASE("Resolve a relative URL against an empty base URL", "[Url]") {
+
+    Url base("");
+
+    CHECK(Url("g:h").resolved(base).string() == "g:h");
+    CHECK(Url("g").resolved(base).string() == "g");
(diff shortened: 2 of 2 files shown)


blair1618:
@msmollin can you check that this fixes the problems with finding scene resource files on iOS?

msmollin:
So I built a release version of the iOS framework off this branch's code and put it here - https://github.com/tangrams/ios-framework/tree/resourceLoadingFix

A couple things I noticed / thought about:

1) Yes it does appear to resolve the issue we were seeing before, at least on the surface. Here's a Screenshot off my phone of manhattan with building grids and POI icons:
screen shot 2016-12-13 at 10 05 21 pm

2) However, it seems to have made my 25/25 White "Find Me" button disappear:
screen shot 2016-12-13 at 10 05 34 pm

There should be a 25 px circle in the center of the screen there. Now, what I'm unsure of, is my styling for the marker looks like `{ style: 'ux-location-gem-overlay', color: 'white', size: [25px, 25px], order:500, collide: false }`, so could Tangram be attempting to load the associated sprite that's attached to this thing now? 

It could also be something entirely unrelated. There's about 10 commits that have gone into tangram-es master since I built the framework build in that's in the framework repo's master branch. I haven't looked to see where this particular code branched from, but most of them are @karimnaaji's work so maybe he knows what might've happened?

blair1618:
Wheee whack-a-bug!

I'd guess that yes, this means the location icon is actually (almost) loading now. That marker should have a texture associated with it now but probably needs a `sprite`, since the POIs are in a sprite atlas. Try adding `sprite: ux-current-location` to the marker styling string.

tallytalwar:
What @blair1618 said. But if you want your white circle, you will have to update the scene to remove `texture` from `ux-location-gem-overlay` style. But who likes a white circle! :D.

msmollin:
So, changing to that line, I now get a crash coming out of Tangram core :) 

`{ style: ux-location-gem-overlay, interactive: true, sprite: ux-current-location, size: 36px, collide: false }`

screen shot 2016-12-14 at 9 23 57 am

msmollin:
The crash is "EXEC_BAD_ACCESS" which indicates trying to access a point in memory that's been deallocated.

karimnaaji:
@msmollin I catched this crash and fixed it in https://github.com/tangrams/tangram-es/pull/1161. 

What happens is you set the marker as interactive (not a feature yet), and the object to obtain new feature selection color is not set for text and icon markers. We shouldn't crash on that for sure, but this is going to be fixed in the incoming merges.

A temporary fix is to simply remove `interactive: true` and the crash will go away.

msmollin:
Sure enough, that fixes it.

So with `{ style: ux-location-gem-overlay, sprite: ux-current-location, size: 36px, collide: false }`

We now have:
screen shot 2016-12-14 at 10 35 42 am

Very exciting.

nvkelso:
nice!

can you populating data into the data source and have the stylesheet do
this for you automatically, though?

```
mz_current_location:
        type: GeoJSON
       …
````

File: core/src/util/url.cpp [hunk 1]
```diff
@@ -250,7 +250,10 @@ Url Url::resolve(const Url& b, const Url& r) {
                 size_t end = removeLastSegmentFromRange(t.buffer, t.parts.path.start, t.buffer.size());
                 t.buffer.resize(end);
                 // Add the relative URL's path.
-                if (r.buffer[r.parts.path.start] != '/') {
+                if (!t.buffer.empty()) {
```
karimnaaji: Is this covered on the associated unit tests?

blair1618:
@nvkelso This is using the `Marker` interface, which does styling a little differently than data in layers. With `Marker`s we can add, remove, and even smoothly interpolate individual point features, so it's much better suited to a very dynamic datum like current location.

PR #1175, opened 2016-12-14
Android marker selection

karimnaaji:
Marker selection for Android.

To go in https://github.com/tangrams/tangram-es/pull/1161

### DIFF
--- a/android/demo/src/main/java/com/mapzen/tangram/android/MainActivity.java
+++ b/android/demo/src/main/java/com/mapzen/tangram/android/MainActivity.java
@@ -12,11 +12,13 @@ import com.mapzen.tangram.LngLat;
 import com.mapzen.tangram.MapController;
 import com.mapzen.tangram.MapController.FeaturePickListener;
 import com.mapzen.tangram.MapController.LabelPickListener;
+import com.mapzen.tangram.MapController.MarkerPickListener;
 import com.mapzen.tangram.MapController.ViewCompleteListener;
 import com.mapzen.tangram.MapData;
 import com.mapzen.tangram.Marker;
 import com.mapzen.tangram.MapView;
 import com.mapzen.tangram.MapView.OnMapReadyCallback;
+import com.mapzen.tangram.Marker;
 import com.mapzen.tangram.TouchInput.DoubleTapResponder;
 import com.mapzen.tangram.TouchInput.LongPressResponder;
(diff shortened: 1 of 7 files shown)


android/tangram/src/com/mapzen/tangram/MapController.java
```diff
@@ -697,13 +726,30 @@ public void pickLabel(float posX, float posY) {
...
+    /**
      * Adds a {@link Marker} to the map which can be used to dynamically add points and polylines
      * to the map.
      * @return Newly created {@link Marker} object.
      */
     public Marker addMarker() {
         checkPointer(mapPointer);
         long markerId = nativeMarkerAdd(mapPointer);
+
+        Marker marker = new Marker(mapView.getContext(), markerId, this);
```
ecgreb: Should this `Marker` instance now be the return value below instead of returning a separate instance?

android/tangram/src/com/mapzen/tangram/MapController.java
```diff
@@ -90,6 +94,21 @@
...
+    /**
+     * Interface for a callback to receive the picked {@link Marker}
+     * Triggered after a call of {@link #pickMarker(float, float)}
+     * Listener should be set with {@link #setMarkerPickListener(MarkerPickListener)}
+     */
+    public interface MarkerPickListener {
+        /**
+         * Receive the picked {@link Marker}
+         * @param marker The {@link Marker} the marker that has been selected
+         * @param positionX The horizontal screen coordinate of the tapped location
```
ecgreb: Should we update the description "The horizontal screen coordinate of the tapped location" to "The horizontal screen coordinate of the picked location" since `pickMarker(...)` can be invoked at any time even on non-tap events?

android/tangram/src/com/mapzen/tangram/MapController.java
```diff
@@ -90,6 +94,21 @@
...
+         * @param positionY The vertical screen coordinate of the tapped location
+         */
+        void onMarkerPick(Marker marker, float positionX, float positionY);
```
ecgreb: Is there any way to get the current `LngLat` that was used to set the marker position from this callback?

https://github.com/tangrams/tangram-es/blob/master/android/tangram/src/main/java/com/mapzen/tangram/Marker.java#L96

PR #1176, opened 2016-12-14
wip: Font loading accelerator

hjanetzek:


ios/src/platform_ios.mm
```diff
@@ -153,20 +153,23 @@ bool isContinuousRendering() {
...
 
-        UIFont* font = [UIFont fontWithName:fallback size:1.0];
+        handles.emplace_back([fallback]() {
 
-        size_t dataSize = 0;
-        auto cdata = loadUIFont(font, &dataSize);
+             __block std::vector<char> data;
 
-        if (!cdata) { continue; }
+             dispatch_sync(dispatch_get_main_queue(), ^{
```
msmollin: I realize this is marked [WIP] currently, but just FYI calling dispatch_sync on the main queue in iOS/OS X can quickly lead to a deadlock situation if the call-site calling this function doesn't know that's what this is doing.

You should probably do some defensive programming to check to see if you're already on the main thread, kind of like this answer here: http://stackoverflow.com/a/10341532/490180

hjanetzek: yes, that seems sensible. it's my first lines objective-c code, so I would have certainly asked for some testing and advice :)

msmollin: Haha excellent, and no worries! Just helping out :)

hjanetzek:
@karimnaaji could you check platform_ios.mm changes? Copying fallback into the lambda is probably not safe

tallytalwar:
> But it only depends on what we want; faster scene/font loading with possible lag since the loading is deferred, or slower scene loading with a more constant framerate? Knowing that those fonts could possibly never be loaded too!

How significant will be these delays?

hjanetzek:
@karimnaaji font loading happens only on the TileWorker threads. I think compared to the time for downloading tiles, styling, etc the font loading time doesn't contribute much to the tile building.

karimnaaji:
@tallytalwar it's quite significant, much enough for you to perceive it. I would suggest you to run on device. 

For the numbers:
```
TANGRAM platform_ios.mm:172: LOADING FONT TIME 16.653000ms
LOADED Font: Copperplate Regular size: 40
TANGRAM platform_ios.mm:172: LOADING FONT TIME 426.117004ms
[...]
LOADED Font: Kohinoor Bangla Regular size: 40
TANGRAM platform_ios.mm:172: LOADING FONT TIME 3373.363770ms
Font is broken or irrelevant...
LOADED Font: Zapfino Regular size: 40
TANGRAM platform_ios.mm:172: LOADING FONT TIME 3379.856689ms
LOADED Font: Bodoni 72 Oldstyle Book size: 40
```

Containing (decoding of TTF + loading), knowing that we also copy the data to an alfons container from a C data, the data comes from a [NSData] that can only be viewed as a C array. 

@hjanetzek yep I know it happens here, I suspect that there might be some sync with the main thread going on in iOS while accessing those fonts.

karimnaaji:
@hjanetzek just saw your last commit with the sync on texture, will give it another try and see if it helps.

hjanetzek:
what? so you are waiting more than 7 seconds at for fonts loading on startup :D

hjanetzek:
How large are the ios system fonts that we need? Maybe it's better to convert them on the first startup and save them to a temp folder so we can access them faster afterwards (and don't have to hold them in memory at all). also suggested here  http://stackoverflow.com/questions/3692812/on-ios-can-i-access-the-system-provided-fonts-ttf-file

karimnaaji:
@hjanetzek no it never has been that much, I would have noticed it.

Here is master's numbers:

```
TANGRAM tangram.cpp:311: resize: 1080 x 1920
TANGRAM tangram.cpp:211: Loading scene file (async): https://tangrams.github.io/walkabout-style/walkabout-style.yaml
TANGRAM platform_ios.mm:175: LOADING FONT TIME 334.455994
WARNING sceneLoader.cpp:915: Duplicate DataSource: normals
WARNING sceneLoader.cpp:221: Parsing style: 'yaml-cpp: error at line 724, column 16: bad conversion':
```

screen shot 2016-12-20 at 1 25 24 pm

I suspect that the copy of data here https://github.com/tangrams/tangram-es/pull/1176/files#diff-63bdd101ebd93e44ff36af6d14172e8eR164 is the bottleneck, I would need to profile more.

ios/src/platform_ios.mm
```diff
@@ -153,20 +153,18 @@ bool isContinuousRendering() {
...
-        FontSourceHandle fontSourceHandle = [cdata, dataSize](size_t* _size) -> unsigned char* {
-            *_size = dataSize;
-
-            return cdata;
-        };
-
-        handles.push_back(fontSourceHandle);
+            UIFont* font = [UIFont fontWithName:fallback size:1.0];
+            size_t dataSize = 0;
+            auto cdata = loadUIFont(font, &dataSize);
```
hjanetzek: yeah - I thought so too, and did it, but loadUIFont is used below where it needs `unsigned char'*`. In a separate commit this can be changed to vector in both cases. 

In any way copying some memory is unlikely to take more than a few ms.

hjanetzek:
@karimnaaji maybe the 3 seconds per font is from a debug/instrumentation build?

karimnaaji:
It was run on instrumentation build. The time is not per font though! It's the total time:

screen shot 2016-12-20 at 2 28 49 pm

After speaking with @msmollin , it seems possible that behind the scene a synchronization is being made while accessing the UIFont from threads, where this would potentially slow down when done from threads.

hjanetzek:
yeah - it's quite certain to take longer as at that time other worker threads and rendering run in parallel.

karimnaaji:
I am still doing the testing on the first changes to have better comparisons with the initial results, but will test the separate mutex in a few!

ios/src/platform_ios.mm
karimnaaji: Just for the sake of documenting this PR, after some digging with @hjanetzek here are some numbers will full optimization we've found with the different available options:
```
1199.566895ms for assign with reserve
1202.317993ms for assign without reserve
1327.153198ms for std::copy with reserve
1450.960205ms for std::copy without reserve
1263.508057ms for insert with reserve
889.072998ms for insert without reserve
395.985931ms without the copy
```

So, further work here to improve interfacing with alfons would save us half the time.

PR #1180, opened 2016-12-15
tizen: download cancellation

hjanetzek:


core/src/style/textStyleBuilder.cpp
```diff
@@ -27,7 +27,9 @@ const static std::string key_name("name");
 
 TextStyleBuilder::TextStyleBuilder(const TextStyle& _style)
     : StyleBuilder(_style),
-      m_style(_style) {}
+      m_style(_style) {
+    m_textLabels = std::make_unique<TextLabels>(m_style);
```
tallytalwar: with this initialization of `m_textLabels` either is redundant or should be done a `reset` with a newer `TextLabels` instance.

tallytalwar:
@hjanetzek wanna give this a look?

tallytalwar:
All other commits related to attribution work has been moved to #1188

tizen/src/urlWorker.cpp
```diff
@@ -71,6 +80,10 @@ void UrlWorker::run() {
         stream.seekp(0);
         CURLcode result = curl_easy_perform(curlHandle);
 
+        if (thread->canceled) {
```
tallytalwar: This is good, but we can probably also do the same check before actually making the url request through `curl_easy_perform`. Basically avoid doing a network call if that url is not required now.

tallytalwar:
"Never click the "Update Branch" button"... It added an extra merge master commit! :|

PR #1181, opened 2016-12-15
iOS marker selection

karimnaaji:
Marker selection for iOS.

To go in #1161


ios/demo/src/MapViewController.m
```diff
@@ -32,6 +32,22 @@ - (void)mapView:(TGMapViewController *)mapView didLoadSceneAsync:(NSString *)sce
     [mapView setPosition:newYork];
 }
 
+- (void)mapView:(TGMapViewController *)mapView didSelectMarker:(TGMapMarkerId)markerId atScreenPosition:(CGPoint)position
+{
+    if (markerId == 0) {
```
msmollin: Just curious - would we ever want to call the delegate callback if the markerId is zero? Seems like something I wouldn't expect as an implementing developer.

ios/src/TGMapViewController.mm
```diff
@@ -149,6 +149,32 @@ - (void)pickFeaturesAt:(CGPoint)screenPosition
...
+    if (!self.map && !self.mapViewDelegate) { return; }
+
+    screenPosition.x *= self.contentScaleFactor;
+    screenPosition.y *= self.contentScaleFactor;
+
+    self.map->pickMarkerAt(screenPosition.x, screenPosition.y, [screenPosition, self](const Tangram::MarkerPickResult* markerPickResult) {
+        CGPoint position = CGPointMake(0.0, 0.0);
+
+        if (!markerPickResult) {
+            if ([self.mapViewDelegate respondsToSelector:@selector(mapView:didSelectMarker:atScreenPosition:)]) {
```
msmollin: Oh I see. Hmmm. What value does this add calling with a marker value of zero?

msmollin:
Does this need a rebase to get the circle builds working?

ios/demo/src/MapViewController.m
karimnaaji: We want to always have a callback, it's a nicer behavior. As the implementing developer querying the system without knowing if the query has ever completed is misleading; you actually don't know if it actually performed the query at all. So you never really know if it failed for other reasons than not finding a pick result.

karimnaaji:
Yes it does, in the process of rebasing the whole thing :)

PR #1182, opened 2016-12-15
iOS marker selection

karimnaaji:
Rebase of https://github.com/tangrams/tangram-es/pull/1181
Going in https://github.com/tangrams/tangram-es/pull/1161

### DIFF
--- a/ios/src/TGMapViewController.mm
+++ b/ios/src/TGMapViewController.mm
@@ -205,6 +205,32 @@ __CG_STATIC_ASSERT(sizeof(TGGeoPoint) == sizeof(Tangram::LngLat));
     });
 }
 
+- (void)pickMarkerAt:(CGPoint)screenPosition
+{
+    if (!self.map && !self.mapViewDelegate) { return; }
+
+    screenPosition.x *= self.contentScaleFactor;
+    screenPosition.y *= self.contentScaleFactor;
+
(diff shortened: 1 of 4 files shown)


ios/src/TGMapViewController.mm
```diff
@@ -205,6 +205,32 @@ - (void)pickFeatureAt:(CGPoint)screenPosition
...
+
+    screenPosition.x *= self.contentScaleFactor;
+    screenPosition.y *= self.contentScaleFactor;
+
+    self.map->pickMarkerAt(screenPosition.x, screenPosition.y, [screenPosition, self](const Tangram::MarkerPickResult* markerPickResult) {
+        CGPoint position = CGPointMake(0.0, 0.0);
+
+        if (!markerPickResult) {
+            if ([self.mapViewDelegate respondsToSelector:@selector(mapView:didSelectMarker:atScreenPosition:)]) {
+                [self.mapViewDelegate mapView:self didSelectMarker:0 atScreenPosition:position];
```
msmollin: After thinking about this for a bit, what are your thoughts on having a "success" and "error" callback here, rather than reusing the same method and passing along a zero? We could still pass along the marker id of zero if we want, but I feel like it would be more obvious to the implementing developers what actually happened.

karimnaaji: That's another solution, but I'm a little anxious to give to the user more implementation callback to override from the default interface (3 more in that case if we want to be consistent between selectable components). That can be easily forwarded that way in the SDK though.

Issue #1167, opened 2016-12-13
Unable to get callback for client data source point with properties


ecgreb:
Currently when adding a point with properties to the map using a client data source it is not possible to retrieve those properties using `pickFeature()` or `pickLabel()`.

1. When calling `MapController.pickFeature()` on the visible point marker the `FeaturePickListener` is invoked but the `properties` map is empty.

2. When calling `MapController.pickLabel()` on the visible point marker the `LabelPickListener` is invoked but the `labelPickResult` param is `null` (which should contain the properties).

This issue also manifests in Eraser Map search result pins https://github.com/mapzen/eraser-map/issues/800.

**Steps to reproduce**
1. Run sample activity below.
2. Zoom in and tap on current location gem ([marked `interactive` in the stylesheet](https://github.com/tangrams/bubble-wrap/blob/gh-pages/bubble-wrap.yaml#L1155)).
3. Notice properties are not available in either callback.
4. Tap on another built-in map POI (ex. Hill Country Barbeque Market).
5. Properties are available in label pick listener (but not feature pick listener).

**Sample code**
```java
public class MainActivity extends AppCompatActivity implements MapView.OnMapReadyCallback {

  @Override protected void onCreate(Bundle savedInstanceState) {
    super.onCreate(savedInstanceState);
    setContentView(R.layout.activity_main);

    final MapView mapView = (MapView) findViewById(R.id.map);
    mapView.getMapAsync(this, "styles/bubble-wrap/bubble-wrap.yaml");
  }

  @Override public void onMapReady(final MapController mapController) {
    final LngLat lngLat = new LngLat(-73.9903, 40.7443); // Samsung NEXT
    final MapData mapData = mapController.addDataLayer("mz_current_location");
    final HashMap<String, String> properties = new HashMap<>();

    mapController.setPosition(lngLat);
    mapController.setZoom(17);

    properties.put("Property 1", "Samsung");
    properties.put("Property 2", "NEXT");
    mapData.addPoint(lngLat, properties);
    mapController.requestRender();

    mapController.setTapResponder(new TouchInput.TapResponder() {
      @Override public boolean onSingleTapUp(float x, float y) {
        return false;
      }

      @Override public boolean onSingleTapConfirmed(float x, float y) {
        mapController.pickFeature(x, y);
        mapController.pickLabel(x, y);
        return true;
      }
    });

    mapController.setFeaturePickListener(new MapController.FeaturePickListener() {
      @Override
[...]

PR #1184, opened 2016-12-15
Fix selection color applied to point labels

blair1618:
Resolves https://github.com/tangrams/tangram-es/issues/1167

This solves the observed problem, but I'm not sure that this is the intended data flow :\

### DIFF
--- a/core/src/style/pointStyleBuilder.cpp
+++ b/core/src/style/pointStyleBuilder.cpp
@@ -175,7 +175,7 @@ void PointStyleBuilder::addLabel(const Point& _point, const glm::vec4& _quad,
     uint32_t selectionColor = 0;
 
     if (_params.interactive) {
-        selectionColor = _rule.featureSelection->nextColorIdentifier();
+        selectionColor = _rule.selectionColor;
     }
 
     m_labels.push_back(std::make_unique<SpriteLabel>(glm::vec3(glm::vec2(_point), m_zoom),


karimnaaji:
This actually breaks the unique selection color that each label would be broken. After some debugging I think I found where the issue lives.

Several things that makes this bug appear:
- On the issue https://github.com/tangrams/tangram-es/issues/1167, it is mentioned that the draw rule under the layer `layers.mz_current_location_gem.draw.ux-location-gem-overlay.interactive` is marked `interactive`, but it is actually not. I had to add that on my debugging code:
```c++
        std::vector<Tangram::SceneUpdate> updates = {
            {"layers.mz_current_location_gem.draw.ux-location-gem-overlay.interactive", "true"},
        };
        map->loadSceneAsync("https://raw.githubusercontent.com/tangrams/bubble-wrap/gh-pages/bubble-wrap.yaml",
            true, nullptr, nullptr, updates);
```
cc @ecgreb on that to confirm the value of the `interactive` flag.
- Some recent optimization uses the set `m_labels` within `labels.cpp` to retrieve the [label associated to a specific selection color](https://github.com/tangrams/tangram-es/blob/master/core/src/labels/labels.cpp#L67) read from the framebuffer. The bug is that this set is **only** added labels that can **not** occlude, which is true from the stylesheet [here](https://github.com/tangrams/bubble-wrap/blob/gh-pages/bubble-wrap.yaml#L1159).

I guess the would be to have a separate vector in `labels.cpp` that contains all labels (including the ones that can't occlude) to be able to retrieve the matching selectionColor -> label.

karimnaaji:
@blair1618 pushed on this branch https://github.com/tangrams/tangram-es/tree/hot-fix-1167.

hjanetzek:
@karimnaaji  ah right! missing these interactive labels. We could add only the labels with selection color to m_selectionLabels instead of collecting all.

karimnaaji:
@hjanetzek , right maintaining that set from `labels.cpp` would be the right way to handle that!

blair1618:
Ah good catch, I forgot that each label needs it's own selection identifier and it can't just use the same one as the feature. I'll test that fix you pushed.

tallytalwar:
@karimnaaji I see what you are mentioning.

1. Current location gem should be marked non interactive in the style sheet. It might be for the purposes of the example, @ecgreb could have referenced it. But yes I see the point of the bug :D.

2. I see the bug of non colliding labels are not checked when retrieving the labels for selection color. Instead of putting all visible labels for selection color checking, does it make sense to only put labels with "interactive" property set for selection color checking?

ecgreb:
In the version of the stylesheet I was using at the time current location gem was marked interactive. However after [this commit](https://github.com/tangrams/bubble-wrap/commit/9738570b31b0d01746b0711082919ffca6c1d645#diff-b8806f7b52d569e5a54f86d92d7b4064L1095) was merged that is now no longer the case.

But yes I did verify the layer was interactive when reproducing the issue. And I agree the current location gem should NOT be interactive moving forward I was just using it for the test case.

blair1618:
Yes, bubble-wrap was very recently revised to reduce the interactive layers.

Closing in favor of #1185

nvkelso:
Some apps allow interactivity, we can modify our behavior to indicate say "current location" or "my location" on tap of the gem.

PR #1185, opened 2016-12-15
Hot fix for https://github.com/tangrams/tangram-es/issues/1167

blair1618:
This resolves the issue as far as I can tell. The surrounding code is too complex for me to tell whether this is an egregious hack. If it isn't, I recommend merging this so that we can provide the Mapzen Android SDK with an updated release.

### DIFF
--- a/core/src/labels/labels.cpp
+++ b/core/src/labels/labels.cpp
@@ -60,13 +60,14 @@ void Labels::processLabelUpdate(const ViewState& viewState,
             m_needUpdate |= label->evalState(dt);
             label->addVerticesToMesh();
         }
+        m_allLabels.emplace_back(label.get(), tile, isProxy);
     }
 }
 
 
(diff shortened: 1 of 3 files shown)


core/src/labels/labels.cpp
```diff
@@ -60,13 +60,14 @@ void Labels::processLabelUpdate(const ViewState& viewState,
             m_needUpdate |= label->evalState(dt);
             label->addVerticesToMesh();
         }
+        m_allLabels.emplace_back(label.get(), tile, isProxy);
```
karimnaaji: Yep that's a good point! @hjanetzek also proposed it, I will clean up that branch since it's a lot of debug code and will update for that.

karimnaaji:
@blair1618 this is not an egregious hack, this is the actual fix for the problem explained here https://github.com/tangrams/tangram-es/pull/1184#issuecomment-267441623.

PR #1188, opened 2016-12-19
Attribution

tallytalwar:
Adds an interface for Tangram Map to add attribution, using yaml styling for the attribution text.

Breaking this work from https://github.com/tangrams/tangram-es/pull/1180

@hjanetzek some rebasing made you the main owner of all the Attribution interface.

This also has a fix on when no draw rules are specified, draw rule evaluation should return `false`.

### DIFF
--- a/core/src/scene/attribution.cpp
+++ b/core/src/scene/attribution.cpp
@@ -1,0 +1,54 @@
+#include "data/tileData.h"
+#include "labels/labels.h"
+#include "marker/markerManager.h"
+#include "marker/marker.h"
+#include "scene/attribution.h"
+#include "scene/scene.h"
+#include "style/textStyle.h"
+
+namespace Tangram {
+
+void Attribution::setup(Scene& scene, const std::string& attributionText, const std::string& attributionStyling) {
+
+    m_style = std::make_unique<TextStyle>("attribution-overlay", scene.fontContext(), true);
+    m_style->build(scene);
+    m_builder = m_style->createBuilder();
+
+    m_marker = std::make_unique<Marker>(0);
+    auto feature = std::make_unique<Feature>();
+    feature->geometryType = GeometryType::points;
+    feature->props.set("name", attributionText);
+    auto blah = feature->props.get("name").get<std::string>();
+    feature->points.emplace_back();
+    m_marker->setFeature(std::move(feature));
+
+    m_marker->setStylingString(attributionStyling);
+}
+
+void Attribution::draw(RenderState& rs, const View& view, Scene& scene) {
+    m_style->onBeginFrame(rs);
+    m_style->onBeginDrawFrame(rs, view, scene);
+    m_style->onEndDrawFrame();
+}
+
+void Attribution::update(const View& view) {
+    if (!m_marker->mesh()) { return; }
+
+    auto* mesh = dynamic_cast<const LabelSet*>(m_marker->mesh());
+    m_style->onBeginUpdate();
+
+    for (auto& l : mesh->getLabels()) {
+        l->setScreenPosition({10, view.getHeight()});
+        l->enterState(Label::State::visible);
+        l->evalState(0);
+        l->addVerticesToMesh();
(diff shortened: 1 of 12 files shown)


core/src/scene/attribution.h
```diff
@@ -0,0 +1,30 @@
...
+class RenderState;
+class StyleBuilder;
+class Scene;
+class TextStyle;
+class View;
+
+class Attribution {
+    std::unique_ptr<TextStyle> m_style = nullptr;
+    std::unique_ptr<StyleBuilder> m_builder = nullptr;
+    std::unique_ptr<Marker> m_marker = nullptr;
```
hjanetzek: StyleBuilder does not need to be a member, it's only used once during setup

core/src/tangram.cpp
```diff
@@ -379,6 +384,9 @@ bool Map::update(float _dt) {
         }
     }
 
+    if (impl->attribution.isSetup()) {
+        impl->attribution.update(impl->view);
+    }
     FrameInfo::endUpdate();
```
hjanetzek: I would use `std::unique_ptr<Attribution> attribution`, passing setup parameters to the constructor: when it is set it's already set up

core/src/scene/attribution.cpp
```diff
@@ -0,0 +1,54 @@
...
+
+    m_style = std::make_unique<TextStyle>("attribution-overlay", scene.fontContext(), true);
+    m_style->build(scene);
+    m_builder = m_style->createBuilder();
+
+    m_marker = std::make_unique<Marker>(0);
+    auto feature = std::make_unique<Feature>();
+    feature->geometryType = GeometryType::points;
+    feature->props.set("name", attributionText);
+    auto blah = feature->props.get("name").get<std::string>();
```
hjanetzek: möööp

core/src/scene/attribution.cpp
```diff
@@ -0,0 +1,54 @@
...
+
+    auto* mesh = dynamic_cast<const LabelSet*>(m_marker->mesh());
+    m_style->onBeginUpdate();
+
+    for (auto& l : mesh->getLabels()) {
+        l->setScreenPosition({10, view.getHeight()});
+        l->enterState(Label::State::visible);
+        l->evalState(0);
+        l->addVerticesToMesh();
+    }
```
hjanetzek: one optimization: I guess update is only necessary on init, when viewport changed or gl context was recreated

core/src/scene/attribution.cpp
tallytalwar: Yeah was for some testing. WIll remove :D

matteblair:
This still strikes me as the sort of thing that would be vastly easier to implement using the UI frameworks of each target platform. On Android and iOS it will be much better to just provide access to a platform-native text view that a client application can set, style, and position. Tizen is sort of the odd case here because we're building a plugin for an interface that doesn't account for information strings like this. It would probably still be less code to implement this at the Tizen plugin layer. 

We established that we can't add a child text object to the image object where we draw the map, but there must be some way to rasterize the contents of an evas object onto a canvas or composite it with an image. @hjanetzek any ideas?

msmollin:
Just weighing in for iOS: this is really easy to add at the SDK-specific layer. GLKit views are UIView subclasses and supports rendering whatever UIKit controls you want.

tallytalwar:
@karimnaaji brought up an interesting point that we can handle attribution via this interface for all platforms other than ios and android, where instead of exposing this interface we expose platform specific UI traits.

But does it make sense to have core rendering engine do this via tangram styling for all other platforms?

hjanetzek:
Since we are bound on tizen to use an image object (i.e. we draw to some surface) we have to use the drawing methods available for that OpenGL surface. I guess we could create an offscreen evas canvas and use that as texture on our surface - sounds as if it would involve more code though.

tallytalwar:
Closing this for now.

PR #1190, opened 2016-12-20
evalRuleForContext should return false when no draw rules are set

tallytalwar:
How I found this (Story):
During the generic attribution code (using markers), I was adding a feature with its text property set to the attribution string. But during testing when I specified no 

You add a marker with its underlying feature having a `name` property, and added some garbage rules for marker styling. I expected no text to render, but it did.

### DIFF
--- a/core/src/scene/drawRule.cpp
+++ b/core/src/scene/drawRule.cpp
@@ -163,6 +163,7 @@ bool DrawRuleMergeSet::evaluateRuleForContext(DrawRule& rule, StyleContext& ctx)
         }
 
         bool valid = true;
+        bool drawRuleSet = false;
         for (size_t i = 0; i < StyleParamKeySize; ++i) {
 
             if (!rule.active[i]) {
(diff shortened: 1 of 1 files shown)


matteblair:
Do you have a story or some context for this? :)

tallytalwar:
Updated description with some scenario, I found this. 

Basically if no drawrules are specified for a feature draw, it should have invalid evaluation for draw rules.

matteblair:
I don't know what you mean by "garbage rules". Can you show the example you used?

matteblair:
If a style is specified but given no draw rules, I think I would expect the draw rule defaults to apply, which may result in a feature being drawn.

tallytalwar:
It was some random text as the marker styling ... "garbage styling" :D

matteblair:
Invalid YAML should be handled before this step, in `MarkerManager` I believe.

tallytalwar:
Hmm.. If that is the case ... then default for `text_source` is `name` property of the feature. In which case it was correctly rendering the text. So maybe this is not needed.

tallytalwar:
Sorry I meant a valid yaml "garbage: styling".

PR #1194, opened 2016-12-22
More font loading speedups

karimnaaji:
Reduce data reinterpretation between platform layer and alfons.
Moving from a total of 946.282ms to 425.619ms on iPhone 6s, other platforms need to be updated.


hjanetzek:
@karimnaaji I think we could add a move constructor to `alfons::InputSource` with this PR - just in case the compiler does not optimize the copy out.

karimnaaji:
Yeah we checked with @tallytalwar comparing data address to see if the RVO was actually propagating from the platform layer to alfons, but it would be worth to force the compiler to do that.

PR #1197, opened 2016-12-24
Make it possible to break on breakpoints set in tangram cpp code.

Guillaume227:
When an android studio module (e.g. tangram-es/demo) is calling into a shared lib built by another one (e.g. tangram-es/tangram), it needs to declare the location of the shared lib for breakpoints to work in native code.
Here is a way to make it work, based on example found there 👍 https://github.com/googlesamples/android-ndk/tree/master/hello-libs

Note : this really is something that needs to be added in the build.gradle of any module integrating tangram.
Ideally Android Studio would have a way to just resolve native deps in a way that would make this redundant but it's not the case in the current version.

A twist on that change that could be considered cleaner is to set a proper LIBRARY_OUTPUT_DIRECTORY to hide the gory specifics of the internal build path by defining a proper distribution_dir :
`
set_target_properties(tangram
                      PROPERTIES
                      LIBRARY_OUTPUT_DIRECTORY
"${distribution_DIR}/lib/${ANDROID_ABI}")
`

I did not include that in the present PR as it's a little more invasive (would require change in android.cmake).
Let me know what you think.


matteblair:
Hi Guillaume, thanks for the suggestion! I'll give this a try over the next few days and see if we can integrate it.

matteblair:
I believe that with the latest version of Android Studio one should be able to set and break on breakpoints in native code in 'libtangram' using the current master branch without any changes, at least that's the case when I try it - let me know if that's not true for you. 

After looking closely at the NDK sample project you've linked, I think it is meant as an example for using native libraries from native code in your app itself. We're currently focused on the Java interface for our Android library and currently there aren't any plans for supporting a native interface on Android, so I don't think the recommendations in that sample apply for our exact situation.

Guillaume227:
It would also make more sense to me if Android Studio was able to debug seamlessly accross module boundaries but I am currently using both Android Studio 2.2 and 2.3 Beta 1 and that PR is the only way I managed to break in c++ code when I have some native code in a library module (e.g. tangram) and not directly in the app module (e.g. demo).

- Within the context of a single module bundling the app and jni/cpp code I have observed that breaking in cpp code just works.
- In the dual module setup I have since observed that outputting the libtangram.so in <module with native code>**src/main/jniLibs** makes breakpoint work without having to change anything in the build.gradle of dependant module, so it's nicer.

Anyway, feel free to ignore that PR.
As a side note, I have been pushing some of what's currently in the android java layer to the native side so I can leverage the c++ nature of tangram-es and maximise the amount of code reuse accross iOS and Android.
I have been using djinni (https://github.com/dropbox/djinni) to handle the JNI boiler plate and it's working quite nicely.
I understand it's not your focus right now but let me know if at some point that becomes PR material.

PR #1201, opened 2016-12-27
Fix marker functions

### LINKED ISSUE #1094: Marker with text styling
With the new Marker things (I'm specifically looking in Android), Is it possible to add a point marker with a text style?

```
...

Marker marker = mapController.addMarker();
marker.setPoint(coords);
marker.setStyling("{ style: 'text', text_source: function() { return 'Test'; }, font: { size: 14px } }");

...
```

Currently Tangram crashes on the setStyling line, Taking out the function works but I can't think of a work around to put text in there without a function? 

(On Android at least) any invalid yaml in setStyling kills the app with just which made playing with the new markers a little more tricky:

```
A/libc: Fatal signal 6 (SIGABRT), code -6 in tid 31076 (id.tangramstest)                                                             
[17 13:50:05.517  3029: 3029 W/         ] debuggerd: handling request: pid=31076 uid=10224 gid=10224 tid=31076
```

tallytalwar:
2 main fixes here:
- Style Context need to increment JS function counter when adding new JS functions
- Since each marker has its own set of rules, the js evaluations for these must be unique to each marker. Previously these evaluations were saved in MarkerManager::m_ruleSet::m_evaluated, which was being overwritten by subsequent marker JS function evaluations.
Fixes #1094

### DIFF
--- a/core/src/marker/marker.cpp
+++ b/core/src/marker/marker.cpp
@@ -11,6 +11,7 @@
 namespace Tangram {
 
 Marker::Marker(MarkerID id) : m_id(id) {
+    m_ruleSet.reset(new DrawRuleMergeSet());
 }
 
 Marker::~Marker() {
@@ -27,6 +28,10 @@ void Marker::setFeature(std::unique_ptr<Feature> feature) {
 
 void Marker::setStylingString(std::string stylingString) {
     m_stylingString = stylingString;
+}
+
+bool Marker::evaluateRuleForContext(StyleContext& ctx) {
+    return m_ruleSet->evaluateRuleForContext(*m_drawRule, ctx);
 }
 
 void Marker::setDrawRule(std::unique_ptr<DrawRuleData> drawRuleData) {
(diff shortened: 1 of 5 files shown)


matteblair:
Makes sense. Is there a simple use case to test the fix?

tallytalwar:
I tested this by added 2 markers each having a different js function for the drawrule.

matteblair:
After applying the function counter fix in https://github.com/tangrams/tangram-es/pull/1201/commits/cde97547ad82a2d7186f675f3d21a67e266dfc56 I can't seem to reproduce the issue. I've attached the diff I'm using to test. Are there additional conditions to trigger the bug?
[two-marker-functions-diff.txt](https://github.com/tangrams/tangram-es/files/685654/two-marker-functions-diff.txt)

tallytalwar:
Try this without the second commit: https://github.com/tangrams/tangram-es/commit/08b03278241b3502c77a084288fbbf7c6f95dd44

matteblair:
Yep, that does it.

PR #1205, opened 2017-01-05
Preparation for curved labels

hjanetzek:
- rename WorldTransform -> *Label::Coordinates
  the coordinates are given in tile/model space, not world space
- test out-of-screen (+buffer) in updateScreenTransform
  check against maximum label bbox combining all anchors
- use shared buffer for label screen-transforms (ScreenTransform::Buffer)
- add wrappers to interpret buffer for the respective label-type
- use shared buffer for obbs in Labels and LabelCollider
  this change allows for multiple obbs per label
- simplify Label::canOcclude()

### DIFF
--- a/core/src/labels/label.cpp
+++ b/core/src/labels/label.cpp
@@ -1,9 +1,6 @@
 #include "label.h"
 
 #include "util/geom.h"
-#include "glm/gtx/rotate_vector.hpp"
-#include "glm/gtx/norm.hpp"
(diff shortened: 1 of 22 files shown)


karimnaaji:
Just pushed some minor changes on the branch, feel free to revert/squash the one that you want to keep. I need to do some more explicit testing on the features that this affects tomorrow!

PR #1208, opened 2017-01-06
Update scene file to match more closely with JS

tallytalwar:
- has so much more info now.
- also one MAJOR change is the use of the v1 tile urls

Few differences from current JS scene file:
1. Use of `px2` is commented out as its not supported in ES yet!
2. Label/Poi rendering uses transitions in ES
3. Road labels in ES show off repeat distance and repeat group (Probably JS should also use this)
4. some rearrangement of `visible` parameter is done in ES. (moved visible to draw group instead of layer property).


scenes/scene.yaml
```diff
@@ -6,10 +6,10 @@ fonts:
         url: https://fonts.gstatic.com/s/montserrat/v7/zhcz-_WihjSQC0oHJ9TCYL3hpw3pgy2gAi-Ip7WPMi0.woff
     Open Sans:
         - weight: 300
-          url: fonts/open sans-300normal.ttf
+          url: https://github.com/tangrams/bubble-wrap/raw/gh-pages/fonts/OpenSans-Regular.woff
```
matteblair: Is there a more stable URL we can use for these? Google Fonts maybe?

nvkelso: For Mapzen house styles we usually add a comment with the remote URL and follow that with an uncommented local resources:

``` yaml
#url: https://fonts.gstatic.com/s/montserrat.woff
url: fonts/montserrat.woff
```

nvkelso: (suggest including a local copy of the resource in your repo, too)

tallytalwar: We have been trying to not include fonts in tangram es repo..

We can probably point this to some proper url, but this is a just a copy of what tangram JS has in its demo scene.

scenes/scene.yaml
```diff
@@ -1,19 +1,19 @@
...
 
 fonts:
     Montserrat:
         url: https://fonts.gstatic.com/s/montserrat/v7/zhcz-_WihjSQC0oHJ9TCYL3hpw3pgy2gAi-Ip7WPMi0.woff
     Open Sans:
-        - weight: 300
-          url: fonts/open sans-300normal.ttf
-        - weight: 300
+        - weight: 400
+          url: http://fonts.gstatic.com/s/opensans/v13/IgZJs4-7SA1XX_edsoXWog.ttf
```
matteblair: Are there woff files available? Wouldn't truetype files be much larger?

karimnaaji: Those are quite simple fonts with only a few hundreds kilobytes in it (at most 200k), there's not much gain we could have. Finding other formats that `woff2` is quite tedious since they don't propose it anymore on their website https://fonts.google.com/. I had to use an API key and download a full list of their fonts with TTF formats from https://www.googleapis.com/webfonts/v1/webfonts:

[GoogleFonts.txt](https://github.com/tangrams/tangram-es/files/694178/GoogleFonts.txt)

PR #1211, opened 2017-01-09
Label placement 2.0

### LINKED ISSUE #1179: Directional arrows on road
Hi

I was wondering if there is a way to show directional arrows on road. e.g.
google

### LINKED ISSUE #1183: Point-on-line placement strategies
Tangram JS recently added more sophisticated ways of placing point drawings along lines. The new options are described in the release notes for v0.11: https://github.com/tangrams/tangram/releases/tag/v0.11.0, with related PRs https://github.com/tangrams/tangram/pull/409 and https://github.com/tangrams/tangram/pull/441

tallytalwar:
Implements label placement strategies to match JS. Refer https://github.com/tangrams/tangram/pull/409 and JS release notes: https://github.com/tangrams/tangram/releases/tag/v0.11.0

- Fixes incorrect non-occlusion of parent label when child is not placed because of repeat distance.
- Implements `vertex`, `spaced` and `midpoint` placement strategies
- Implements `auto` angle calculations and application.

TODO:
- [x] Implement `placement_min_length_ratio` parameter.
- [x] ~~Implement `repeat_distance` and `repeat_group` for point based styles.~~ (will handle this in another PR: #1228)
- [x] Fix stop interpolations for flat point labels
- [x] Fix angle update for flat point labels
- [ ] ~~More stop interpolation updates to allow 2d stops for point labels, by making extrude vec2 instead of float~~ (Will be opening a subsequent PR for this)

Fixes #1183 
Fixes #1179 

cc. @bcamper We will need to set `flat: true` for arrows.


core/src/labels/spriteLabel.cpp
```diff
@@ -48,14 +48,14 @@ bool SpriteLabel::updateScreenTransform(const glm::mat4& _mvp, const ViewState&
                 if (m_vertexAttrib.extrudeScale != 1.f) {
                     scale *= pow(2, _viewState.fractZoom) * m_vertexAttrib.extrudeScale;
                 }
-                glm::vec2 dim = m_dim * scale;
 
-                positions[0] = p0 - dim;
```
tallytalwar: I wonder where I was getting that offset in all my arrows :). Nice catch

core/src/style/pointStyleBuilder.cpp
```diff
@@ -153,11 +153,21 @@ auto PointStyleBuilder::applyRule(const DrawRule& _rule, const Properties& _prop
...
+            // size for next style zoom for interpolation
+            float higherSize = sizeParam.stops->evalSize(m_styleZoom + 1).get<float>();
+            p.extrudeScale = (higherSize - lowerSize);
+            p.size = glm::vec2(lowerSize);
+        } else if (sizeParam.value.is<glm::vec2>()) {
+            p.size = sizeParam.stops->evalExpVec2(m_styleZoom);
+            // NB: this assumes that the width/height ratio is
+            // constant for all stops
+            glm::vec2 higherSize = sizeParam.stops->evalExpVec2(m_styleZoom + 1);
+            p.extrudeScale = (higherSize.x - p.size.x);
```
tallytalwar: I was considering doing this in another PR, to accomodate `extrudeScale` as a vec2.

core/src/style/pointStyleBuilder.cpp
```diff
@@ -358,44 +323,28 @@ void PointStyleBuilder::labelPointsPlacing(const Line& _line, const glm::vec4& u
             }
             break;
         case LabelProperty::Placement::spaced: {
-            Line interpolatedLine;
-            std::vector<float> allAngles;
+            LineSampler<std::vector<Point>> sampler;
```
tallytalwar: nitpick ... you can make this `LineSampler<Line>`.

core/src/util/lineSampler.h
```diff
@@ -0,0 +1,195 @@
+#pragma once
+
+#include "glm/glm.hpp"
+
+namespace Tangram {
+
+template<typename Points>
+struct LineSampler {
```
tallytalwar: Nice One.

core/src/style/pointStyleBuilder.cpp
```diff
@@ -358,44 +323,28 @@ void PointStyleBuilder::labelPointsPlacing(const Line& _line, const glm::vec4& u
...
-
-            for (size_t i = 0; i < interpolatedLine.size(); i++) {
-                auto p = interpolatedLine[i];
-                if (params.keepTileEdges || !isOutsideTile(p)) {
-                    if (params.autoAngle && !allAngles.empty()) {
-                        params.labelOptions.angle = allAngles[i];
-                    }
-                    addLabel(p, uvsQuad, params, _rule);
+            glm::vec2 p, r;
+            sampler.advance(distance, p, r);
```
tallytalwar: `advance` should also take `minLength` into consideration.

core/src/style/pointStyleBuilder.cpp
hjanetzek: This template type is for the internal storage of LineSampler, the z value is used for length. So I would tend to use `std::vector<glm::vec3>` here.

core/src/style/pointStyleBuilder.cpp
hjanetzek: yeah. just a quick fix. without checking style param value type it broke testing with bubble-wrap

core/src/style/pointStyleBuilder.cpp
tallytalwar: yup make sense, I saw the use of `z` in LineSampler after writing this comment :D.

core/src/style/pointStyleBuilder.cpp
tallytalwar: Ohh it would crash!. Cool. But it does still make sense to get `extrudeScale` in vec2 right?

hjanetzek: Technically - yes. Practically only useful for super weird styles I guess :)

core/src/style/pointStyleBuilder.cpp
```diff
@@ -255,6 +274,95 @@ bool PointStyleBuilder::getUVQuad(PointStyle::Parameters& _params, glm::vec4& _q
...
+                if (params.keepTileEdges || !isOutsideTile(p)) {
+                    if (params.autoAngle) {
+                        params.labelOptions.angle = angleBetween(p, q);
+                    }
+                    addLabel(p, uvsQuad, params, _rule);
+                }
+            }
+
+            // Place label on endpoint
+            auto &p = *(_line.rbegin() + 1);
```
karimnaaji: I don't think we actually need to explicitly add this iteration. The last iteration on the for loop already treats that last case.

hjanetzek: Ah, there's a out-of-bounds read in the above loop

tallytalwar: 2 things:
1.  @karimnaaji we are adding the label correspeonding to the point at the end and not end-1 as in the last iteration of the loop.
2. @hjanetzek where is the out of bound?

karimnaaji:
```
(_line.rbegin() + 1) === _line[_line.size() - 2];
(_line.back()) === _line[_line.size() - 1];
```
How is that different from the last iteration?

karimnaaji: @hjanetzek I also don't see any out-of-bounds read in the loop?

hjanetzek: I don't see it anymore either :) 

@karim the last iteration of the loop adds the last point - the loop goes over all segments, adding the startpoint of the segment.

Another way to write this:

```
            for (size_t i = 0; i < _line.size() - 1; i++) {
                auto& p = _line[i];
                auto& q = _line[i+1];
                if (params.keepTileEdges || !isOutsideTile(p)) {
                    if (params.autoAngle) {
                        params.labelOptions.angle = angleBetween(p, q);
                    }
                    addLabel(p, uvsQuad, params, _rule);
                    if (i == _line.size() - 2) {
                        addLabel(q, uvsQuad, params, _rule);
                    }
                }
            }
```

karimnaaji: @hjanetzek thanks we just spoke about having that exact code IRL with @tallytalwar :)

tallytalwar: Yup, already did this change, to make it more readable.

core/src/style/pointStyleBuilder.h
```diff
@@ -50,12 +55,20 @@ struct PointStyleBuilder : public StyleBuilder {
     bool addFeature(const Feature& _feat, const DrawRule& _rule) override;
 
 private:
+
+
+    Point interpolateSegment(const Line& _line, int i, int j, float distance);
```
karimnaaji: Those definitions can be removed, I think it's from the previous implementation before the line sampler was added.

tallytalwar: Yup will clean this up.

Issue #1212, opened 2017-01-09
Reorganize project structure


matteblair:
Tangram-es has grown and changed over its life. Now that we have spent time working with, in, and around it, several of us on the tangram-es team have ideas of how to make the project structure more clear and useful.

Any contributors to tangram-es are free to voice ideas here for discussion!

matteblair:
A suggestion for revising the directory structure:
```
tangram-es
  - bench
    - benchmark
    - src
    - CMakeLists.txt
  - core
    - deps (external projects used in 'core')
      - alfons
      - benchmark
      - ...
    - include (all public headers for 'core')
      - tangram
        - map.h
        - platform.h
        - ...
    - src (sources and private headers for 'core')
      - (all current folders in /core/src)
      - shaders
      - map.cpp
    - CMakeLists.txt (for building libcore)
  - docs (new folder for documentation and guides)
  - platforms
    - android
      - (all current folders in /android)
      - android.cmake
    - common
      - platform_gl.h/.cpp
      - glfw
    - ios
      - (all current folders in /ios)
      - ios.cmake
      - ios.framework.cmake
      - ios.toolchain.cmake
    - linux
      - include
      - src
      - linux.cmake
    - osx
      - include
      - src
      - darwin.cmake (rename to osx.cmake?)
    - rpi
      - include
      - src
      - raspberrypi.cmake
    - tizen
      - include
      - src
      - tizen.cmake
      - tizen.toolchain.cmake
  - scenes
  - scripts
  - tests
  - CMakeLists.txt (delegates to core and platform cmake scripts)
  - README, Makefile, etc.
```
If anything about this is not obvious, let me know and I'll add detail

karimnaaji:
I think it's in pretty good shape already! The one part that I would suggest modifying in this structure is:

```
  - core
    - 3rdparty/deps (external projects used in 'core')
      - alfons
      - benchmark
      - ...
      + Header only libraries (if not already assumed to be there)
    - include (all public headers for 'core')
      - map.h
    - src (sources and private headers for 'core')
      - (all current folders in /core/src)
      - shaders
      - map.cpp
```
I think it's possible to only provide a single header `map.h` that would make it much easier to distribute the library as a single shared/static file + this single header.

I like the name deps but I think that `3rdparty` is a good alternative too.

[...]

PR #1213, opened 2017-01-09
Changed CMake Variable to access through "add_subdirectory"

MyZeD:
Tested and works through add_subdirectory and native with cmake


karimnaaji:
Hey @MyZeD , thanks for your addition! We can considerate your contribution and cherry-pick your commit when addressing this issue https://github.com/tangrams/tangram-es/issues/1212 where we will make sure the build system works nicely when used a submodule.

MyZeD:
Okay, thanks

Issue #1147, opened 2016-12-06
Can't create iOS Archives when using Tangram-es binary framework because of bitcode weirdness


msmollin:
Went through testing Xcode's archive generation (which is how most independent developers generate their IPAs to distribute) and I get this error:

<img width="855" alt="screen shot 2016-12-06 at 10 29 41 am" src="https://cloud.githubusercontent.com/assets/1163880/20931411/f6dc5120-bb9e-11e6-9f39-98ff86feaf8e.png">

This seems to be related to an error I ran across before, describe here: (https://github.com/Alamofire/Alamofire/issues/835#issuecomment-145404118) but wasn't able to reliably reproduce it. It seems that in Xcode 8, bitcode generation for regular builds versus archives is a different process.

When I attempt to reproduce this by building the ios-sdk I get a slightly different error:
```
2016-12-06 15:06:31 +0000 [MT] Presenting: Error Domain=IDEFoundationErrorDomain Code=1 "Failed to verify bitcode in TangramMap.framework/TangramMap:
error: Bundle only contains bitcode-marker /var/folders/ty/20jr0dzn5_z9s6vw3ttczj400000gp/T/XcodeDistPipeline.epy/Root/Payload/ios-sdk.app/…" UserInfo={NSLocalizedDescription=Failed to verify bitcode in TangramMap.framework/TangramMap:
error: Bundle only contains bitcode-marker /var/folders/ty/20jr0dzn5_z9s6vw3ttczj400000gp/T/XcodeDistPipeline.epy/Root/Payload/ios-sdk.app/…}
```

I believe this is the actual error, and the test project I'm using is obfuscating that.

To Reproduce the first error:
1) Download the Tangram-es binary framework from https://github.com/tangrams/ios-framework
2) Start a new iOS project
3) Add the binary framework and make sure to embed the binary in xcode like so:
<img width="702" alt="screen shot 2016-12-06 at 10 37 27 am" src="https://cloud.githubusercontent.com/assets/1163880/20931732/0c1b3e1a-bba0-11e6-8257-ce2da1593a57.png">
4) Attempt to archive the build (Product Menu -> Archive)

Or you could just use this project I made (you'll have to change the codesigning credentials for this one):
[test-tangram-archive.zip](https://github.com/tangrams/tangram-es/files/634255/test-tangram-archive.zip)

[...]

PR #1217, opened 2017-01-10
Pass custom CXX flag to all dependencies

karimnaaji:
Add custom compile options to all of our external dependencies; for iOS it enables bitcode and gives custom flag to be propagated when compiling all our libraries.

Fixes https://github.com/tangrams/tangram-es/issues/1147


karimnaaji:
@msmollin it's quite diverging from the original branch since I finally found the right command to enable sharing a flag between all of our dependencies, so it needs to be tested thoroughly within the final archiving process!

PR #1221, opened 2017-01-12
Handle empty and zero-area rings in centroid()

matteblair:
https://github.com/tangrams/tangram-es/pull/1144#issuecomment-272017394

~If a polygon passed to `centroid()` has no rings or the first ring is empty, it returns `{0, 0}`. If the signed area of the first ring is zero, it returns the first point in the first ring.~

`centroid()` now takes a single ring as input, to clarify the functionality. Zero-area rings produce a centroid of { NAN, NAN }.

### DIFF
--- a/core/src/style/pointStyleBuilder.cpp
+++ b/core/src/style/pointStyleBuilder.cpp
@@ -411,8 +411,11 @@ bool PointStyleBuilder::addPolygon(const Polygon& _polygon, const Properties& _p
             labelPointsPlacing(line, uvsQuad, p, _rule);
         }
     } else {
-        glm::vec2 c = centroid(_polygon);
-        addLabel(Point{c,0}, uvsQuad, p, _rule);
+        glm::vec3 c;
+        if (!_polygon.empty()) {
+            c = centroid(_polygon.front().begin(), _polygon.front().end());
+        }
+        addLabel(c, uvsQuad, p, _rule);
     }
 
     return true;
--- a/core/src/util/geom.cpp
+++ b/core/src/util/geom.cpp
@@ -101,30 +101,6 @@ glm::vec2 worldToScreenSpace(const glm::mat4& _mvp, const glm::vec4& _worldPosit
     return clipToScreenSpace(clipCoords, _screenSize);
 }
 
-glm::vec2 centroid(const std::vector<std::vector<glm::vec3>>& _polygon) {
-
-    if (_polygon.empty()) {
-        return glm::vec2();
-    }
-
-    // We will consider only the outer ring of the polygon when calculating
-    // the centroid - holes will be ignored. It would be possible to transform
-    // a polygon with holes into a polygon that covers the same set of points
-    // without any holes, but that seems overly complex for our needs.
-    const auto& ring = _polygon.front();
-    glm::vec2 centroid;
-    float area = 0.f;
-
-    for (auto curr = ring.begin(), prev = ring.end() - 1; curr != ring.end(); prev = curr, ++curr) {
-        float a = (prev->x * curr->y - curr->x * prev->y);
-        centroid.x += (prev->x + curr->x) * a;
-        centroid.y += (prev->y + curr->y) * a;
-        area += a;
-    }
-
-    return centroid / (3.f * area);
-}
-
 // square distance from a point <_p> to a segment <_p1,_p2>
 // http://stackoverflow.com/questions/849211/shortest-distance-between-a-point-and-a-line-segment
 //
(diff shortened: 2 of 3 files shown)


hjanetzek:
@matteblair why not return vec2(nan) when it cannot determine the centroid? This way the caller can easily determine whether the result is valid (or useful in this case)

tallytalwar:
I agree with @hjanetzek, returning a valid pair of number here is kinda incorrect "centroid" for that zero area polygon.

Also I think by default it will return a {nan, nan} if area evaluated to `0`.

matteblair:
Hmmm yeah that does make sense - returning { 0, 0 } could be misleading in those cases. The case that I wanted to fix here is when a polygon's vertices are all the same point. For that case it makes the most sense to return that single point, but these changes impact more cases than that.

I'll think about this more and either revise or close the PR.

matteblair:
Well I completely changed this diff haha

The end result should be the same as on `master`, but now I think the usage of `centroid` is much clearer and more practical.

westnordost:
So the centroid of a polygon with no area is `{NaN, NaN}`, that's fine - but what about a polygon like this:
`[{-5,5},{5,5},{-5,-5},{5,-5}]`
Wouldn't it be wrongfully `{NaN, NaN}` as well with the implementation? 

I mean, you mentioned this yourself there https://github.com/tangrams/tangram-es/pull/1144#issuecomment-272031048

matteblair:
@westnordost I believe your example would be a self-intersecting polygon, which we don't aim to support since it has very little practical use for map data.

nvkelso:
(famous last words)

westnordost:
Ah right. Yes, this makes sense.

core/src/style/pointStyleBuilder.cpp
```diff
@@ -304,9 +304,11 @@ bool PointStyleBuilder::addPolygon(const Polygon& _polygon, const Properties& _p
...
         }
     } else {
-        glm::vec2 c = centroid(_polygon);
-
-        addLabel(Point{c,0}, uvsQuad, p, _rule);
+        glm::vec3 c;
+        if (!_polygon.empty()) {
+            c = centroid(_polygon.front().begin(), _polygon.front().end());
+        }
+        addLabel(c, uvsQuad, p, _rule);
```
tallytalwar: Probably make sense to initialize `c` here to `{0, 0}`.

matteblair: `glm::vec` and `glm::mat` types all have default constructors that initialize the members to zero.

westnordost:
Sorry to bring this centroid topic up yet again.

I noticed that the computation produces wrong results for small polygons. Small being about 100m² and below, typical sizes for detached houses, swimming pools etc.
(In my test, a centroid of a small terraced house (5x10m) turned up 20m away from the house!)

The algorithm is not wrong, it is just a systematic floating point rounding error which is introduced [here](https://github.com/tangrams/tangram-es/commit/5a098f6ab47a556e9b944e46d6db1f08fca5959a#diff-240257c8d0870f7efd7f7e89b50c522eR95)
What will solve the problem is to calculate only with the offsets from a reference point, i.e. the first point in the polygon.

The issue was discussed before in context of Leaflet (as far as I can see, they never fixed it) [here](https://translate.google.com/translate?hl=en&sl=de&tl=en&u=https%3A%2F%2Fplus.google.com%2Fu%2F0%2F%2BThomasRupprechtximex%2Fposts%2FbeV4PQYM58x). Follow the link to see the algorithm in code.

matteblair:
Interesting, does the change you suggested fix your use case?

You are correct that a `float` can't represent Lng-Lat coordinates with sufficient precision, but this calculation is done in `float` precision because it operates on "tile-local" coordinates rather than Lng-Lat coordinates. Roughly speaking, a `float` should be able to precisely represent values down to 2^-24 times the size of a tile in this context.

We could make it use the type of the members of `value_type`, i.e.
```
using Scalar = decltype(offset.x);
```
But we would still be operating on `float`s in all of our current use cases.

westnordost:
I use the same centroid algorithm for my project and it was there where I noticed and fixed the error. The code above is just something I wrote up quickly in the comment box. 
However, if tangram calculates with tile-local coordinates anyway, I guess that it is not a problem for tangram at all, after all, since it is already something like calculating with an offset instead of the whole of latlon.

So I'd say "nevermind then", cause I did not actually notice the error in tangram, i.e. a label being displayed at a wrong position. I just felt I should inform you about this since it is the same algorithm that caused the problem for me.

PR #1224, opened 2017-01-12
Move tangram initialization for iOS earlier into the VC lifecycle

### LINKED ISSUE #1220: Move Tangram::Map() creation earlier in TGMapViewController lifecycle
Decoupling the map creation from the view controller's view lifecycle will make this easier to test some of the functions that are called by the iOS SDK.

msmollin:
Fixes #1220


msmollin:
@matteblair / @karimnaaji figured it out. I needed to also override the initializer that Storyboards use. Crash is now resolved.

PR #1225, opened 2017-01-12
Simple shader cache

hjanetzek:


tallytalwar:
this is good, we can handle resuse of linked shader programs in a subsequent pr.

core/src/gl/renderState.h
```diff
@@ -112,6 +113,9 @@ class RenderState {
 
     JobQueue jobQueue;
 
+    std::map<std::string, GLuint> fragmentShaders;
+    std::map<std::string, GLuint> vertexShaders;
```
matteblair: How about `unordered_map` for these? We don't need a specific ordering and it seems like hash tables would require fewer string comparisons.

PR #1228, opened 2017-01-13
Implement repeat group logic for point based styles

tallytalwar:
- fix: repeat group should only be applied when repeat distance is > 0
- alter repeat group application for text based styles to accomodate inheritance logic from parent
point style
- implement repeat group and repeat distance parsing for point based styles
- repeat distance is 0 by default for point based styles (unlike for text based styles, which has
256 as default repeat distance)

Refer other details of the concept: https://github.com/tangrams/tangram/pull/424

This PR branch is based on #1211

### DIFF
--- a/core/src/style/textStyleBuilder.cpp
+++ b/core/src/style/textStyleBuilder.cpp
@@ -311,6 +311,10 @@ TextStyle::Parameters TextStyleBuilder::applyRule(const DrawRule& _rule,
     _rule.get(StyleParamKey::transition_show_time, p.labelOptions.showTransition.time);
 
     uint32_t priority;
+    size_t repeatGroupHash = 0;
+    std::string repeatGroup;
+    StyleParam::Width repeatDistance;
+
     if (_iconText) {
 
         if (_rule.get(StyleParamKey::text_priority, priority)) {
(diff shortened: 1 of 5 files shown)


core/src/style/textStyleBuilder.cpp
```diff
@@ -330,6 +334,24 @@ TextStyle::Parameters TextStyleBuilder::applyRule(const DrawRule& _rule,
...
+        }
+
+        if (p.labelOptions.repeatDistance > 0.f) {
+            if (_rule.get(StyleParamKey::text_repeat_group, repeatGroup)) {
+                hash_combine(repeatGroupHash, repeatGroup);
+            } else if (_rule.get(StyleParamKey::repeat_group, repeatGroup)) { //inherit from parent point
+                hash_combine(repeatGroupHash, repeatGroup);
+            } else {
+                repeatGroupHash = _rule.getParamSetHash();
+            }
```
karimnaaji: Should add 
```
p.labelOptions.repeatDistance *= m_style.pixelScale();
```
around here.

karimnaaji: Oh no this is done later on.

Issue #1065, opened 2016-11-04
Thumb size for feature selection


karimnaaji:
We need to allow the user of the tangram API to set up a thumb size in logical pixel size for our mobile platform. This size would be convenient for feature selection queries where high density display makes it hard to select features without reading the framebuffer on a wider range.

blair1618:
@karimnaaji I didn't see that you had assigned yourself, I was working on this today and I should have noted it! Oh well.

I've made a proof-of-concept implementation for on approach to this: https://github.com/tangrams/tangram-es/commit/2cec9e5496a57dab624286e56011b40bf0bee932

This samples a square area of the framebuffer with a configurable size and returns the most common (non-zero) color in that area. The sampled area could be further refined into a circle or ellipse. This works alright, and does allow selecting some features when your mouse is just a little bit off. However, it has problems. Mainly, small selectable features surrounded by larger selectable features can actually be harder to select! In an extreme case, a feature whose area on screen is less than half of the area sampled could be _impossible_ to select. 

The next approach I want to try is to sample a configurable area of the framebuffer and return the first non-zero color that is closest to the center of the selection area. This will still cover some "near-miss" scenarios but will also allow selecting arbitrarily small features.

Guillaume227:
"The next approach" you describe gets my vote !

PR #1230, opened 2017-01-14
Feature picking within a radius

matteblair:
Since fingers are not great at selecting a single pixel from a touchscreen, it can be useful to check over an area when picking features from the map.

This change add a configurable "pick radius", measured in logical pixels, used when selecting features, labels, or markers from the map. Instead of sampling a single pixel from the selection buffer, we now sample a block of pixels around the selected point in screen space and find the nearest non-zero selection color within the pick radius in that block. 

The default pick radius is 0.5 logical pixels, which usually results in sampling one pixel, as before.

After finding a well-tuned pick radius, touch screen apps should find that picking small features is much less frustrating now. However, it should be noted that there is a small added cost to sampling an area of pixels which rises quadratically with the radius. I have not yet profiled this cost!

Resolves: https://github.com/tangrams/tangram-es/issues/1065

### DIFF
--- a/core/src/gl/framebuffer.cpp
+++ b/core/src/gl/framebuffer.cpp
@@ -74,10 +74,10 @@ GLuint FrameBuffer::readAt(float _normalizedX, float _normalizedY) const {
 FrameBuffer::PixelRect FrameBuffer::readRect(float _normalizedX, float _normalizedY, float _normalizedW, float _normalizedH) const {
 
     PixelRect rect;
-    rect.left = floorf(_normalizedX * m_width);
-    rect.bottom = floorf(_normalizedY * m_height);
-    rect.width = ceilf(_normalizedW * m_width);
-    rect.height = ceilf(_normalizedH * m_height);
+    rect.left = fminf(fmaxf(floorf(_normalizedX * m_width), 0.f), m_width);
+    rect.bottom = fminf(fmaxf(floorf(_normalizedY * m_height), 0.f), m_height);
+    rect.width = fminf(fmaxf(ceilf(_normalizedW * m_width), 0.f), m_width - rect.left);
+    rect.height = fminf(fmaxf(ceilf(_normalizedH * m_height), 0.f), m_height - rect.bottom);
 
     rect.pixels.resize(rect.width * rect.height);
 
--- a/core/src/gl/framebuffer.cpp
+++ b/core/src/gl/framebuffer.cpp
@@ -71,6 +71,21 @@ GLuint FrameBuffer::readAt(float _normalizedX, float _normalizedY) const {
     return pixel;
 }
 
+FrameBuffer::PixelRect FrameBuffer::readRect(float _normalizedX, float _normalizedY, float _normalizedW, float _normalizedH) const {
+
+    PixelRect rect;
+    rect.left = _normalizedX * m_width;
+    rect.bottom =  _normalizedY * m_height;
+    rect.width = _normalizedW * m_width;
+    rect.height = _normalizedH * m_height;
+
+    rect.pixels.resize(rect.width * rect.height);
+
+    GL::readPixels(rect.left, rect.bottom, rect.width, rect.height, GL_RGBA, GL_UNSIGNED_BYTE, rect.pixels.data());
+
+    return rect;
+}
+
 void FrameBuffer::init(RenderState& _rs) {
 
     if (!Hardware::supportsGLRGBA8OES && m_colorRenderBuffer) {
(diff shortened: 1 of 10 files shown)


core/src/selection/selectionQuery.cpp
```diff
@@ -23,17 +23,33 @@ QueryType SelectionQuery::type() const {
...
     if (it == _colorCache.end()) {
-        color = _framebuffer.readAt(windowCoordinates.x, windowCoordinates.y);
-        _colorCache.push_back({color, m_position});
+        // Find the first non-zero color nearest to the position and within the selection radius.
+        auto rect = _framebuffer.readRect(windowCoordinates.x, windowCoordinates.y, windowSize.x, windowSize.y);
+        float minDistance = std::fmin(rect.width, rect.height);
+        float hw = static_cast<float>(rect.width) / 2.f, hh = static_cast<float>(rect.height) / 2.f;
+        for (int32_t row = 0; row < rect.height; row++) {
+            for (int32_t col = 0; col < rect.width; col++) {
+                uint32_t sample = rect.pixels[row * rect.width + col];
```
vicb: move `row * rect.width` out of the inner loop ?

vicb: also we should break the loop if `dist = 1`

Also searching from the center might be more efficient more larger rect ?
ie with such a pattern (where number correspond to loop iteration):

```
3333333
3222223
3211123
3210123
3211123
3222223
3333333
```

The algo would need to skip location outside the rect,
The loop can stop when the min distance on the next iteration (=`n+1` above) is more than the hit min distance.

matteblair: Hey thanks for the review! I'll do some profiling next week and see if any performance tuning is needed.

core/src/selection/selectionQuery.cpp
```diff
@@ -23,17 +23,33 @@ QueryType SelectionQuery::type() const {
...
-    glm::vec2 windowCoordinates = _view.normalizedWindowCoordinates(m_position.x, m_position.y);
+    float radius = m_radius * _view.pixelScale();
+    glm::vec2 windowCoordinates = _view.normalizedWindowCoordinates(m_position.x - radius, m_position.y + radius);
+    glm::vec2 windowSize = _view.normalizedWindowCoordinates(m_position.x + radius, m_position.y - radius) - windowCoordinates;
 
     GLuint color = 0;
 
     auto it = std::find_if(_colorCache.begin(), _colorCache.end(), [=](const auto& _colorRead) {
-        return m_position == _colorRead.position;
+        return (m_position == _colorRead.position) && (m_radius == _colorRead.radius);
```
karimnaaji: 👍

core/src/gl/framebuffer.cpp
```diff
@@ -72,6 +72,21 @@ GLuint FrameBuffer::readAt(float _normalizedX, float _normalizedY) const {
...
+
+    PixelRect rect;
+    rect.left = floorf(_normalizedX * m_width);
+    rect.bottom = floorf(_normalizedY * m_height);
+    rect.width = ceilf(_normalizedW * m_width);
+    rect.height = ceilf(_normalizedH * m_height);
+
+    rect.pixels.resize(rect.width * rect.height);
+
+    GL::readPixels(rect.left, rect.bottom, rect.width, rect.height, GL_RGBA, GL_UNSIGNED_BYTE, rect.pixels.data());
```
karimnaaji: Since we have a radius now it is possible that we make viewport read out of the window bounds. Do you know the behavior in that case? Definitely not an error by reading the docs, but it would be worth checking whether this is a potential memory corruption.

matteblair: Gooooood point, looks like undefined behavior! I'll add clamps.

core/src/selection/selectionQuery.cpp
matteblair: Based on the results of my profiling, the time spent iterating over the selection pixels is negligible compared to the time spent in glReadPixels, so I don't think there's much value in optimizing this any further.

core/src/gl/framebuffer.cpp
matteblair: Resolved now.

karimnaaji: Checking the gl docs in a little bit more depth:
>with lower left corner at (x+i,y+j) for 0<=i<width and 0<=j<height.

`width` and `height` are under a lower strict; we should add `-1` to the `minf` to each of the bottom/left clamping equation since it seems that those conditions would still not be satisfied. Can you confirm that?

matteblair: From this: http://docs.gl/es2/glReadPixels it sounds like those ranges describe the positions of the returned pixels, where `x`, `y`, `width`, and `height` are the arguments to `glReadPixels`, not the dimensions of the framebuffer itself, so I think we're ok here.

In the cases where `rect.left == m_width` or `rect.bottom == m_height` we would end up sampling with 0 width or height, so there shouldn't be any undefined results.

PR #1234, opened 2017-01-17
Gl context loss

### LINKED ISSUE #1206: Crash after screen orientation change on Nexus 5X
Tangram is consistently crashing for me on my nexus 5x when changing the device orientation (portrait <-> landscape). I managed to reproduce it in a way that hopefully will work for you:

Fire up the simplemap demo with a Nexus 5X emulator.
Zoom in somewhere (you have to move the camera).
Rotate the screen and you should crash:

01-05 16:51:01.228 3803-4095/com.mapzen.tangram.demos.simplemap A/libc: Invalid address 0xdf611000 passed to free: value not allocated
01-05 16:51:01.228 3803-4095/com.mapzen.tangram.demos.simplemap A/libc: Fatal signal 6 (SIGABRT), code -6 in tid 4095 (GLThread 196)

I used the x86_64 Lollipop and Marshmallow images with Google APIs and was able to reproduce it.  I was *not* able to reproduce it using a Nexus 6 or 5 configuration, which makes me think it is something unique with the 5x. This happens on my actual device as well (Nougat), so I don't think it's an emulator issue.  I've also seen less predictable crashes occasionally during onDestroy going in and out of the activity; I'm not sure if it's related.

karimnaaji:
Revisit a little bit our context loss on Android:
- Remove index generation that was being used to track whenever a new context is created
- gl objects no longer retain the memory
- Reload entirely the scene `onSurfaceCreated` whenever the map view has not been destroyed

Fixes #1206

### DIFF
--- a/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
+++ b/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
@@ -172,6 +172,7 @@ public class MapController implements Renderer {
         mapView = view;
         view.setRenderer(this);
         view.setRenderMode(GLSurfaceView.RENDERMODE_WHEN_DIRTY);
+        view.setPreserveEGLContextOnPause(true);
 
         // Set a default HTTPHandler
         httpHandler = new HttpHandler();
--- a/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
+++ b/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
@@ -251,7 +251,6 @@ public class MapController implements Renderer {
      * @param sceneUpdates List of {@code SceneUpdate}
      */
     public void loadSceneFile(String path, List<SceneUpdate> sceneUpdates) {
-
         String[] updateStrings = bundleSceneUpdates(sceneUpdates);
         scenePath = path;
         checkPointer(mapPointer);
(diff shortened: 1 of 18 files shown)


android/tangram/src/main/java/com/mapzen/tangram/MapController.java
```diff
@@ -811,6 +812,8 @@ public void queueSceneUpdate(SceneUpdate sceneUpdate) {
      * @param sceneUpdates List of {@code SceneUpdate}
      */
     public void queueSceneUpdate(List<SceneUpdate> sceneUpdates) {
+        this.sceneUpdates.addAll(sceneUpdates);
```
matteblair: I think we should also add updates from the method above this one to the list.

android/tangram/src/main/java/com/mapzen/tangram/MapController.java
```diff
@@ -1073,7 +1078,13 @@ public void onSurfaceChanged(GL10 gl, int width, int height) {
 
     @Override
     public void onSurfaceCreated(GL10 gl, EGLConfig config) {
+        if (surfaceCreated) {
```
matteblair: I'm not sure I follow the logic of this. Can you elaborate or add comments?

karimnaaji:
Ok I will do some testings tomorrow and see how this goes!

PR #1238, opened 2017-01-18
Fix #1237

### LINKED ISSUE #1237: TGMapViewDelegate causing crash #iOS
In TGMapViewDelegate all the methods are 'optional' but the implementation to call the delegate method really make it 'mandatory' as it will crash if delegate doesn't implement these method. In TGMapViewController.mm the method ` - (void)pickFeatureAt:(CGPoint)screenPosition`  contains 
`if (!self.mapViewDelegate && ![self.mapViewDelegate respondsToSelector:@selector(mapView:didSelectFeature:atScreenPosition:)]) {
            return;
        }`
This will not execute if delegate is set but optional method `@selector(mapView:didSelectMarker:atScreenPosition:)` is not implemented and it makes the code below this point execute which includes `[self.mapViewDelegate mapView:self didSelectFeature:dictionary atScreenPosition:position];` and it will crash as this is not implemented in delegate. 
**-[ViewController mapView:didSelectLabel:atScreenPosition:]: unrecognized selector sent to instance 0x14a838a00**

karimnaaji:


ios/src/TGMapViewController.mm
```diff
@@ -178,7 +178,7 @@ - (void)pickFeatureAt:(CGPoint)screenPosition
     screenPosition.y *= self.contentScaleFactor;
 
     self.map->pickFeatureAt(screenPosition.x, screenPosition.y, [screenPosition, self](const Tangram::FeaturePickResult* featureResult) {
-        if (!self.mapViewDelegate && ![self.mapViewDelegate respondsToSelector:@selector(mapView:didSelectFeature:atScreenPosition:)]) {
+        if (!self.mapViewDelegate || (self.mapViewDelegate && ![self.mapViewDelegate respondsToSelector:@selector(mapView:didSelectFeature:atScreenPosition:)])) {
```
msmollin: You can probably just switch these checks to be straight OR checks, right? Because if either one of these is nil (or both), you want to return. Since It's evaluated left to right, when `!self.mapViewDelegate` is nil, it'll set the if to true and return, without evaluating the secondary condition.

Basically:
`if (!self.mapViewDelegate || ![self.mapViewDelegate respondsToSelector:@selector(mapView:didSelectFeature:atScreenPosition:)])`

karimnaaji: Yep exact same.

PR #1242, opened 2017-01-20
Allow marker styling to use draw rules from other explicit scene layers

### LINKED ISSUE #1239: Proposal: styling flexibility for markers to use draw rules from layers defined in a scene file
In a discussion with @ecgreb and @msmollin, it would be better for a developer to reference draw /styling rules for a marker from a layer defined in the scene file:

Example: 
```
mz_route_line:
        data: { source: mz_route_line }
        draw:
            ux-route-line-overlay:
                color: '#06a6d4'
                order: 500
                width: [[0,3.5px],[5,5px],[9,7px],[10,6px],[11,6px],[13,8px],[14,9px],[15,10px],[16,11px],[17,12px],[18,10px]]
```

Referring `mz_route_line` for a marker should grab draw rules defined in the layer for drawing of this marker
```
map.markerSetStyling(markerID, "layer: mz-route-line");
```

In the above API, if a `layer` is defined, one can grab draw rules for that layer and use it to draw this marker. If no `layer` is defined and draw rules are defined explicitly, then these explicit draw rules are used to draw the marker.

One question I have with this "grabbing draw rules from a layer for a marker styling" is what happens when a layer being references is a sub-layer?
- should the syntax allow full layer name qualification, example: "uberLayerName:subLayerName"
  In this case merged set of draw rules will be applied to the marker.
- we can also not allow sub-layer assigned for markers!! Won't be good though :/

tallytalwar:
This allows styling a Marker by specifying a path to a _draw rule group_ from the current scene file.

@ecgreb @msmollin Once this is done, you will be able to add a current location gem marker by styling it with:

```cpp
map.markerSetStylingByPath(markerID, "layers.mz_current_location_gem.draw.ux-location-gem-overlay");
```

where `ux-location-gem-overlay` is the draw rule group defined under the `mz_current_location_gem` layer of our house styles. 

To style something like an inactive search pin one would do:
```cpp
map.markerSetStylingByPath(markerID, "layers.mz_search_result.inactive.draw.ux-icons-overlay");
```

Resolves #1239

### DIFF
--- a/core/src/marker/markerManager.cpp
+++ b/core/src/marker/markerManager.cpp
@@ -318,29 +318,53 @@ bool MarkerManager::buildStyling(Marker& marker) {
 
     // If the Marker styling is a path, find the layers it specifies.
     if (markerStyling.isPath) {
-        auto& path = markerStyling.string;
-
-        //check for layers
-        const std::string LAYERS = "layers";
-        if (path.substr(0, LAYERS.length()) != LAYERS) { return false; }
-
-        //check for draw
-        const std::string DRAW = LAYER_DELIMITER + "draw" + LAYER_DELIMITER;
-        auto n = path.find(DRAW);
-        if (n == std::string::npos) { return false; }
-
-        auto layerPath = path.substr(0, n);
-
-        LOG("layerPath: %s" , layerPath.c_str());
-        // Replace the rule delimiter with scene's DELIMITER to extract sublayer
-        std::replace(layerPath.begin(), layerPath.end(), LAYER_DELIMITER[0], DELIMITER[0]);
-
-        std::vector<const SceneLayer*> layers = m_scene->getLayerHierarchy(layerPath);
-
-        return marker.setDrawRule(layers);
+        auto path = markerStyling.string;
+        // The DELIMITER used by layers is currently ":", but Marker paths use "." (scene.h).
+        std::replace(path.begin(), path.end(), '.', DELIMITER[0]);
+        // Start iterating over the delimited path components.
+        size_t start = 0, end = 0;
+        end = path.find(DELIMITER[0], start);
+        if (path.compare(start, end - start, "layers") != 0) {
+            // If the path doesn't begin with 'layers' it isn't a layer heirarchy.
+            return false;
+        }
+        // Find the DataLayer named in our path.
+        const SceneLayer* currentLayer = nullptr;
+        size_t layerStart = end + 1;
+        start = end + 1;
+        end = path.find(DELIMITER[0], start);
+        for (const auto& layer : m_scene->layers()) {
+            if (path.compare(layerStart, end - layerStart, layer.name()) == 0) {
+                currentLayer = &layer;
+                marker.mergeRules(layer);
+                break;
+            }
+        }
+        // Search sublayers recursively until we can't find another token or layer.
+        while (end != std::string::npos && currentLayer != nullptr) {
+            start = end + 1;
+            end = path.find(DELIMITER[0], start);
+            const auto& layers = currentLayer->sublayers();
+            currentLayer = nullptr;
+            for (const auto& layer : layers) {
+                if (path.compare(layerStart, end - layerStart, layer.name()) == 0) {
+                    currentLayer = &layer;
+                    marker.mergeRules(layer);
+                    break;
+                }
+            }
+        }
+        // The last token found should have been "draw".
+        if (path.compare(start, end - start, "draw") != 0) {
+            return false;
+        }
+        // The draw group name should come next.
+        start = end + 1;
+        end = path.find(DELIMITER[0], start);
+        // Find the rule in the merged set whose name matches the final token.
+        return marker.finalizeRuleMergingForName(path.substr(start, end - start));
     }
-
-    // If the styling is not a path, try to load is as a string of YAML.
+    // If the styling is not a path, try to load it as a string of YAML.
     try {
         YAML::Node node = YAML::Load(markerStyling.string);
         // Parse style parameters from the YAML node.
@@ -356,7 +380,9 @@ bool MarkerManager::buildStyling(Marker& marker) {
     }
     m_jsFnIndex = sceneJsFnList.size();
 
-    return marker.setDrawRule(std::make_unique<DrawRuleData>("", 0, std::move(params)));
+    marker.setDrawRuleData(std::make_unique<DrawRuleData>("", 0, std::move(params)));
+
+    return true;
 }
 
 bool MarkerManager::buildGeometry(Marker& marker, int zoom) {
--- a/core/src/marker/markerManager.cpp
+++ b/core/src/marker/markerManager.cpp
@@ -4,10 +4,13 @@
 #include "gl/texture.h"
 #include "marker/marker.h"
 #include "scene/sceneLoader.h"
+#include "scene/dataLayer.h"
 #include "style/style.h"
 #include "labels/labelSet.h"
 #include "log.h"
 #include "selection/featureSelection.h"
+
+#include <algorithm>
 
 namespace Tangram {
 
@@ -51,11 +54,24 @@ bool MarkerManager::remove(MarkerID markerID) {
     return false;
 }
 
-bool MarkerManager::setStyling(MarkerID markerID, const char* styling) {
+bool MarkerManager::setStylingFromString(MarkerID markerID, const char* styling) {
     Marker* marker = getMarkerOrNull(markerID);
     if (!marker) { return false; }
 
-    marker->setStylingString(std::string(styling));
+    marker->setStyling(std::string(styling), false);
+
+    // Create a draw rule from the styling string.
+    if (!buildStyling(*marker)) { return false; }
+
+    // Build the feature mesh for the marker's current geometry.
+    return buildGeometry(*marker, m_zoom);
+}
+
+bool MarkerManager::setStylingFromPath(MarkerID markerID, const char* path) {
+    Marker* marker = getMarkerOrNull(markerID);
+    if (!marker) { return false; }
+
+    marker->setStyling(std::string(path), true);
 
     // Create a draw rule from the styling string.
     if (!buildStyling(*marker)) { return false; }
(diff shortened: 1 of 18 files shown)


ecgreb:
How will this change affect layers with draw rules based on state for example search pins can be active or inactive?

https://github.com/tangrams/bubble-wrap/blob/gh-pages/bubble-wrap-style.yaml#L1179-L1193

tallytalwar:
Ahh!! That falls into the category of sublayers.. will think more on a potential workaround.

core/src/marker/markerManager.cpp
```diff
@@ -297,6 +298,27 @@ bool MarkerManager::buildStyling(Marker& marker) {
     try {
         // Update the draw rule for the marker.
         YAML::Node node = YAML::Load(marker.stylingString());
+
+        // Check if marker styling defines a layer in the yaml scene
+        if (Node layer = node["layer"]) {
```
matteblair: There really isn't a need to use YAML when specifying a layer. I would suggest creating a separate function along the lines of:
```
bool markerSetLayer(MarkerID _marker, const char* layerName)
```
Then there's no need to read or write the layer name in YAML syntax.

core/src/marker/markerManager.cpp
```diff
@@ -297,6 +298,27 @@ bool MarkerManager::buildStyling(Marker& marker) {
...
+            if (!layer.IsScalar()) {
+                LOGE("layer for marker styling needs to be a scalar value.");
+                return false;
+            }
+
+            const DataLayer* sceneLayer = m_scene->findLayer(layer.Scalar());
```
matteblair: Restricting ourselves to top-level scene layers here is pretty limiting. We can borrow from our 'scene update' syntax for a way to specify fully-qualified layer names, like `roads.highway.link`. 

When styling the marker with a layer, we would then need to split the fully-qualified name on `.` and follow the layer tree recursively (either here or in `Scene::findLayer`).

core/src/marker/markerManager.cpp
```diff
@@ -297,6 +298,27 @@ bool MarkerManager::buildStyling(Marker& marker) {
...
+            if (sceneLayer && !sceneLayer->rules().empty()) {
+                // select first drawRule from this layer's rules
+                auto rule = sceneLayer->rules().front();
```
matteblair: Using only the first draw rule for a layer is also quite limiting. If at all possible, we should use the same rule merging and application process that we use for non-marker geometry, to minimize potential confusion.

core/src/marker/markerManager.cpp
tallytalwar: I agree on the "not" using of yaml here. Based on the initial requirements from @ecgreb it seems they wanted to use the same method for their purposes. (Refering initial discussion on slack).

I had suggested the exact same function method.

core/src/marker/markerManager.cpp
tallytalwar: My aim initially with this implementation was to have minimal new code and max reuse. Allowing to specify fully qualified layername is doable, but complicate the merging process. Will see if I can get it simplified.

core/src/marker/markerManager.cpp
tallytalwar: 👍  Will see if I can reuse the rule merging aptly to get this to a less "limiting" zone.

With respect to allowing more than one draw groups to draw a marker, we will have to modify the marker architecture to have meshes for more than one "styles". Can be done, but I would prefer to make that another task and not mingle it with this one.

core/src/marker/markerManager.cpp
tallytalwar: https://mapzen.slack.com/archives/pragma/p1484758215000054

@ecgreb and @msmollin on the API usage here.

core/src/marker/markerManager.cpp
tallytalwar: I am going to still keep it to draw only 1 style for now. Adding markers to be drawn with multiple styles seems to be too involved, which should land in its own PR.

tallytalwar:
Just finished adding support for styling marker by specifying an absolute path to a scene layer (delimited by ":").

@matteblair wanna give this a look and see if you have any other suggestions to implement this. I could not think of any other way to implement this without convoluting the marker styling pipeline.

nvkelso:
This seems related to my JS feature request: 

- https://github.com/tangrams/tangram/issues/489

matteblair:
@nvkelso Yes! If we allowed Style definitions to specify default Draw parameters then we wouldn't need to specify Draw parameters separately (either in-line or via a named layer). I don't think we'll wait on that feature, but it could be a nice way to simplify Marker styling in the future.

tallytalwar:
Totally https://github.com/tangrams/tangram/issues/489 will definitely simplify marker styling.

matteblair:
Ok new thought: Instead of styling a Marker by identifying a layer in the scene, we can allow identifying a specific draw rule within a layer. Example:
```
layers:
  touch:
    data: { source: touch }
    point:
      filter: { type: point }
      draw:
        icons:
          sprite: sunburst
          collide: false
          transition: { show: { time: 0s }, hide: { time: 0s } }
```
Styling a Marker would look like:
```
map->markerSetDrawRule(marker, "layers.touch.point.draw.icons");
```
This has a few advantages:
 - Disambiguates the desired draw rule, if a layer has several.
 - Doesn't merge rules, since we only care about a specific one.
 - Uses the same scene hierarchy syntax as SceneUpdate.

matteblair:
Also we don't necessarily need to change this in the current PR but the interface for styling Markers doesn't make the semantics apparent at all. We have multiple functions that can abstractly set the "styling" of a Marker but the interface doesn't indicate that they are mutually exclusive in their application. What seems better to me is something like:
```
class MarkerStyling {
  virtual DrawRule* getDrawRule() = 0;
};
bool Marker::setStyling(MarkerStyling* styling);
```
This makes it clear that a Marker can only have one "styling" set at a time.

tallytalwar:
Ok I refactored the code to incorporate the above 2 suggestions.

1. The interface to style the marker uses a drawRuleGroup specified under the apt layer in the scene file. Example:
    ```yaml
    layers:
        layer1:
            draw:
                grp1:
                    rulesGrp1... 
            layer2:
                draw:
                     grp1:
                         moreRulesGrp1...
                     grp2:
                         rulesGrp2...
    ```
    - `map->markerSetLayerDrawGroup(marker, "layer1:layer2:grp2")` will style the marker using `rulesGrp2` draw rules.
    - `map->markerSetLayerDrawGroup(marker, "layer1:layer2:grp1")` will style the marker using merged `rulesGrp1` and `moreRulesGrp1` draw rules.
    - `map->markerSetLayerDrawGroup(marker, "layer1:grp2")` will style the marker using `rulesGrp1` draw rules.

2. Restructured  the internal code to make sure the 2 ways of styling a marker (by specifying a yaml with draw rules, or by specifying a drawgroup from the loaded scene) are mutually exclusive. This also allows the user to switch styling of a marker from one way to another. @matteblair I think this is better design usage than what you suggested.

tallytalwar:
Updated description for the usage for the platform sdks.

core/src/scene/drawRule.cpp
```diff
@@ -157,53 +157,59 @@ bool DrawRuleMergeSet::match(const Feature& _feature, const SceneLayer& _layer,
...
+    }
+}
 
-void DrawRuleMergeSet::mergeRules(const SceneLayer& _layer) {
+void DrawRuleMergeSet::mergeRules(const SceneLayer& _layer, const std::string& drawGroupName) {
 
     size_t pos, end = m_matchedRules.size();
 
     for (const auto& rule : _layer.rules()) {
+        if (!drawGroupName.empty() && rule.name != drawGroupName) { continue; }
```
matteblair: Rather than adding this logic to DrawRuleMergeSet, it seems more appropriate to me to instead search the set of merged rules for the desired name in the Marker styling code.

core/src/scene/scene.cpp
```diff
@@ -83,6 +83,49 @@ const Style* Scene::findStyle(const std::string& _name) const {
 
 }
 
+std::vector<const SceneLayer*> Scene::getLayerHierarchy(const std::string& _name) const {
+
+    std::vector<const SceneLayer*> layers;
+
+    const std::string delimiter = ":";
```
matteblair: Defining a delimiter in several places like this is fragile, I'd suggest defining the delimiter once in a public variable associated with a relevant class.

core/src/scene/drawRule.cpp
tallytalwar: I did that before, but thought this would be more optimal to not call merge drawGroups which are not required.

matteblair: Checking names in DrawRuleMergeSet is slightly less work for Marker styling, but slightly more work for styling every non-marker feature. Since there are vastly more non-marker features than marker features, we should optimize for that case.

Efficiency aside, I think the code becomes confusing when responsibilities are mixed like this.

tallytalwar:
Addressed all comments here: 6a08f56

core/src/marker/markerManager.cpp
```diff
@@ -297,13 +298,36 @@ bool MarkerManager::buildStyling(Marker& marker) {
     if (!m_scene) { return false; }
 
     std::vector<StyleParam> params;
+
+    const auto& markerStyling = marker.styling();
+    const char* delimiter = ":";
```
tallytalwar: Since its the layers we are using for marker styling here, I think its redundant to add `layers`. So I have not modified this behavior, but included the use of `.` as a delimiter instead of `:`.

core/src/scene/drawRule.cpp
tallytalwar: Addressed the logic.

core/src/scene/scene.cpp
tallytalwar: Addressed.

platforms/osx/src/main.mm
```diff
@@ -13,11 +13,9 @@
 
 std::string sceneFile = "scene.yaml";
 
-std::string markerStyling = "{ style: 'points', interactive: true, color: 'white', size: [25px, 25px], order: 100, collide: false }";
+std::string markerStyling = "touch.point.icons";
```
matteblair: I commented on this earlier, but it was part of another comment so might not have been clear. This "path" is probably not going to make sense to a user. Even I, knowing the implementation, wouldn't find it intuitive.

The full path to the element we are trying to identify (i.e. the path we would use for a scene update) is `layers.touch.point.draw.icons`. If we are calling this a "path" then there's no reason to expect that certain parts should just be omitted.

tallytalwar: I did reply to your earlier comment. Its implicit in the naming of the API that it is a `layer` path, this is much different than the usage of the scene update interface where the user must identity with the complete yaml path.

That said I have addressed this in a270ba951bd4a38e3ff605ff01461811a14aa563. Thanks.

matteblair: That commit addresses part of my comment, but `draw` is still omitted from the path without explanation.

tallytalwar: Hmm will update. I would like to get some opinion on the user of our API to make sure I am not the only one who feel this is redundant information being asked from the user.

cc. @ecgreb 

But update coming.

matteblair: I can see how one might guess from a function name how 'layer' is implied, but guessing is not a good developer experience. Also, the comments above the function seem to directly contradict that implication, describing the argument as `the absolute path to a draw group in the scene`.

tallytalwar: Addressed: 80d0afc1160675f5b7c124dd10d31eede9af4201

ecgreb: Sometimes its better to be consistent that to be correct ;) Or in other words I'm also in favor of explicitly including all path elements (as with scene updates) to avoid any potential confusion for the developer.

platforms/ios/src/TangramMap/TGMarker.mm
```diff
@@ -57,13 +57,24 @@ - (instancetype)initWithMapView:(TGMapViewController *)mapView
     return self;
 }
 
-- (void)setStyling:(NSString *)styling
+- (void)setStylingString:(NSString *)styling
```
tallytalwar: We should have api name consistency between iOS and Android:
Android: `setStylingFromString` and `setStylingFromPath`
iOS: `setStylingString` and `setStylingPath`

matteblair: Hmm I see what you mean. The reason for these names on iOS is that these are setters for a "property" (an obj-c thing), so an application wouldn't actually call "setStylingString", it would just call `marker.stylingString = aStylingString`. Do you think it would be better as `marker.stylingFromString = aStylingString`?

tallytalwar: Ahh I see. I did not look at the `@properties` at all.

Anyways I think it would make sense to have them named like what you suggested above.

matteblair: @tallytalwar Which suggested names? :P

tallytalwar: `marker.stylingFromString = string1`
`marker.stylingFromPath = string2`

matteblair: Mmmmm that just doesn't feel quite right to me, linguistically. I think the slight difference in wording is ok since it's a different syntax anyway. If we really dislike it later we can change it.

tallytalwar:
Sure not too strong on it. Go ahead merge it.  Thanks

PR #1248, opened 2017-01-24
setPixelScale does nothing if new value same as old

matteblair:
When Android resumes a GLSurfaceView it calls resize, which prompts tangram to set the display size and pixel density again. Calling setPixelSize would then clear the tile sets and caches because a change in pixel density requires re-styling the tile data. This resulted in a noticeable delay to map usability when the view was resumed and all tiles on screen were re-loaded. 

This fix just checks whether the call to setPixelScale is redundant, and if so does nothing, making the view resume from pause effectively instantaneously.

### DIFF
--- a/core/src/tangram.cpp
+++ b/core/src/tangram.cpp
@@ -672,6 +672,12 @@ void Map::setPixelScale(float _pixelsPerPoint) {
 
 void Map::Impl::setPixelScale(float _pixelsPerPoint) {
 
+    // If the pixel scale changes we need to re-build all the tiles.
+    // This is expensive, so first check whether the new value is different.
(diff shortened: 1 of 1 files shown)


matteblair:
Before: | After:
--------|------
resume-slow | resume-fast

tallytalwar:
Ahh I only tested this on android. Just tried master on osx (having retina display) ... this breaks something..

tallytalwar:
Ok I got it now ... `setPixelScale` is called in the beginning before the scene is loaded (hence before styles are generated). So pixel scale is set for everything other than the styles. These styles need to get the apt pixel scale when these are generated.
Will refactor this a bit and submit another PR.

PR #1249, opened 2017-01-25
Allow float values for dash parameter

### LINKED ISSUE #951: Allow fractional dashed line patterns
Follow up for https://github.com/tangrams/tangram/pull/404.

tallytalwar:
- Parse float values
- Scale dash values with a fixed scale factor "20" (In accordance with JS) before filling the dash
texture
- Make sure to adjust texture coordinate to account for the texture data
- fixes #951

### DIFF
--- a/core/src/style/polylineStyle.cpp
+++ b/core/src/style/polylineStyle.cpp
@@ -123,7 +123,7 @@ void PolylineStyle::constructShaderProgram() {
             m_shaderProgram->addSourceBlock("defines", "#define TANGRAM_LINE_BACKGROUND_COLOR vec3(" +
                 std::to_string(m_dashBackgroundColor.r) + ", " +
                 std::to_string(m_dashBackgroundColor.g) + ", " +
-                std::to_string(m_dashBackgroundColor.b) + ")");
+                std::to_string(m_dashBackgroundColor.b) + ")\n");
         }
     }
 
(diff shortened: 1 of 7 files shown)


core/src/style/polylineStyle.cpp
```diff
@@ -128,6 +130,13 @@ void PolylineStyle::constructShaderProgram() {
     if (m_dashArray.size() > 0 || m_texture) {
         m_shaderProgram->addSourceBlock("defines", "#define TANGRAM_LINE_TEXTURE\n", false);
         m_shaderProgram->addSourceBlock("defines", "#define TANGRAM_ALPHA_TEST 0.25\n", false);
+        if (m_dashArray.size() > 0) {
+            m_shaderProgram->addSourceBlock("defines", "#define TANGRAM_DASHLINE_TEX_SCALE " +
+                                                std::to_string(dash_scale) + "\n", false);
+        } else {
+            m_shaderProgram->addSourceBlock("defines", "#define TANGRAM_DASHLINE_TEX_SCALE 1",
```
matteblair: Have you tested this case? The way this is used in the shader I think this could cause a float-int mismatch error.

matteblair: ^ Not outdated, despite what github says

karimnaaji: Fixed!

tallytalwar: I had not!. Good one. Thanks @karimnaaji .

PR #1252, opened 2017-01-25
Fix pixel scale to set for styles post loading/updating of scene

tallytalwar:
Just moving around the check to make sure new styles get apt pixel scale once these get loaded from a scene load/update.

### DIFF
--- a/core/src/tangram.cpp
+++ b/core/src/tangram.cpp
@@ -672,6 +672,8 @@ void Map::setPixelScale(float _pixelsPerPoint) {
 
 void Map::Impl::setPixelScale(float _pixelsPerPoint) {
 
+    auto oldPixelScale = scene->pixelScale();
+
     view.setPixelScale(_pixelsPerPoint);
     scene->setPixelScale(_pixelsPerPoint);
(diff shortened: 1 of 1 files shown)


core/src/tangram.cpp
```diff
@@ -672,17 +672,18 @@ void Map::setPixelScale(float _pixelsPerPoint) {
 
 void Map::Impl::setPixelScale(float _pixelsPerPoint) {
 
+    view.setPixelScale(_pixelsPerPoint);
```
matteblair: This guarantees that `_pixelsPerPoint == view.pixelScale()` below is always true, so the tile clearing steps are unreachable.

tallytalwar:
Ohh, that was a bad one! I did not look at the check properly. Anyways I have updated.

Regarding what the issue is, following screenshots should show you the difference:

With current master:
screen shot 2017-01-26 at 9 22 09 am

With this fix:
screen shot 2017-01-26 at 9 23 18 am

tallytalwar:
Closing in favor of #1257

PR #1256, opened 2017-01-26
Curved labels pt2

hjanetzek:
Follow-up PR's
- [ ] Handle line offset for curved labels.

### DIFF
--- a/core/src/labels/curvedLabel.cpp
+++ b/core/src/labels/curvedLabel.cpp
@@ -72,6 +72,7 @@ bool CurvedLabel::updateScreenTransform(const glm::mat4& _mvp, const ViewState& 
 
     // Set center for repeatGroup distance calculations
     m_screenCenter = glm::vec2(center);
+    m_screenAnchorPoint = m_anchorPoint;
 
     // Chord length for minimal ~120 degree inner angles (squared)
     // sin(60)*2
@@ -80,7 +81,7 @@ bool CurvedLabel::updateScreenTransform(const glm::mat4& _mvp, const ViewState& 
     const float sampleWindow = 20;
 
     float width = m_dim.x;
-    float start = sampler.point(m_screenAnchorPoint).z - width * 0.5f;
+    float start = sampler.point(m_anchorPoint).z - width * 0.5f;
 
     if (start < 0 || start + width > sampler.sumLength()) {
         sampler.clearPoints();
@@ -88,13 +89,33 @@ bool CurvedLabel::updateScreenTransform(const glm::mat4& _mvp, const ViewState& 
     }
 
     // Sets current segment to the segment on which the text starts.
-    glm::vec2 p, r; // unused
-    sampler.sample(start, p, r);
+    glm::vec2 startPos, r;
+    sampler.sample(start, startPos, r);
     size_t startSegment = sampler.curSegment();
+
+    glm::vec2 dir = sampler.segmentDirection(startSegment);
+
+    // Cannot reverse the direction when some glyphs must be
+    // placed in forward direction and vice versa.
+    const float flipTolerance = sin(DEG_TO_RAD * 45);
+    bool mustForward = dir.x > flipTolerance;
+    bool mustReverse = dir.x < -flipTolerance;
 
     for (size_t i = startSegment + 1; i < _transform.size(); i++) {
         float currLength = sampler.point(i).z;
-        glm::vec2 dir = sampler.segmentDirection(i);
+        dir = sampler.segmentDirection(i);
+
+        if (dir.x > flipTolerance) {
+            mustForward = true;
+        }
+        if (dir.x < -flipTolerance) {
+            mustReverse = true;
+        }
+
+        if (mustForward && mustReverse) {
+            sampler.clearPoints();
+            return false;
+        }
 
         // Go back within window to check for hard direction changes
         for (int k = i - 1; k >= int(startSegment); k--) {
@@ -109,6 +130,20 @@ bool CurvedLabel::updateScreenTransform(const glm::mat4& _mvp, const ViewState& 
         if (currLength > start + width) {
             break;
         }
+    }
+
+    if (!mustReverse) {
+        // TODO use better heuristic to decide flipping
+        glm::vec2 endPos;
+        sampler.sample(start + width, endPos, r);
+        if (startPos.x > endPos.x) {
+            mustReverse = true;
+        }
+    }
+
+    if (mustReverse) {
+        sampler.reversePoints();
+        m_screenAnchorPoint = _transform.size() - m_anchorPoint - 1;
     }
 
     return true;
@@ -186,18 +221,6 @@ void CurvedLabel::addVerticesToMesh(ScreenTransform& _transform, const glm::vec2
     if (sampler.sumLength() < width) { return; }
 
     float center = sampler.point(m_screenAnchorPoint).z;
-
-    glm::vec2 p1, p2;
-    sampler.sample(center + it->quad[0].pos.x * TextVertex::position_inv_scale, p1, rotation);
-    // Check based on first charater whether labels needs to be flipped
-    // sampler.sample(center + it->quad[2].pos.x, p2, rotation);
-    sampler.sample(center + (end-1)->quad[2].pos.x * TextVertex::position_inv_scale, p2, rotation);
-
-
-    if (p1.x > p2.x) {
-        sampler.reversePoints();
-        center = sampler.sumLength() - center;
-    }
 
     std::array<glm::i16vec2, 4> vertexPosition;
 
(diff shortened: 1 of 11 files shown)


hjanetzek:
@karimnaaji Ready for testing and initial review. Please ignore the commented out debugging logs :)

karimnaaji:
Starting to review this tomorrow!

core/src/style/textStyleBuilder.cpp
```diff
@@ -209,59 +207,240 @@ bool TextStyleBuilder::addFeature(const Feature& _feat, const DrawRule& _rule) {
...
-                    if (d > tolerance) { break; }
+        for (size_t j = i + 1; j < _line.size()-1; j++) {
+            glm::vec2 dir2 = sampler.segmentDirection(j);
 
-                    // Skip merged segment
-                    merged = true;
-                    i += 1;
+            float angle = crossProduct(dir1, dir2);
+
+            if (std::abs(angle) > flipTolerance) {
```
karimnaaji: Should we also have a similar rule on flip tolerance at screen space time? Such that we ensure text is in a readable direction. See _Place Denfert-Rochereau_ on the screenshot:

screen shot 2017-02-02 at 4 04 37 pm

core/src/labels/curvedLabel.cpp
```diff
@@ -0,0 +1,264 @@
...
+
+        float currLength = sampler.point(i).z;
+        float segmentLength = currLength - prevLength;
+        count++;
+
+        if (start + width > currLength) {
+            p2 = glm::vec2(sampler.point(i));
+
+            rotation = sampler.segmentDirection(i-1);
+            _obbs.append({(p1 + p2) * 0.5f, rotation, segmentLength, dim.y});
```
karimnaaji: I think that at this point `p1` and `p2` are missing the label offset, what do you think of:

```
glm::vec2 obbPosition = (p1 + p2) * 0.5f;
glm::vec2 dir = sampler.segmentDirection(i);
glm::vec2 normal = glm::normalize(glm::vec2(-dir.y, dir.x));
obbPosition += normal * m_options.offset;
_obbs.append({obbPosition, rotation, segmentLength, dim.y});
```

To move the obbs along the normal defined by the segment they are currently attached to.

core/src/labels/curvedLabel.h
```diff
@@ -0,0 +1,72 @@
...
+        return m_prio;
+    }
+
+    glm::vec2 modelCenter() const override {
+        return m_worldTransform[m_anchorPoint];
+    }
+
+protected:
+
+    const std::vector<glm::vec2> m_worldTransform;
```
karimnaaji: Should we name that `m_modelTransform` since we moved away from calling it world in previous PR?

core/src/style/textStyleBuilder.cpp
nvkelso: Yes, please :)

hjanetzek: @karimnaaji the case above could be avoided in screen space placement by not allowing quads being placed from right-to-left (with some threshold). Though to me the above label is still within in readable threshold.

`flipTolerance` is checked to count the direction changes of the curve.

core/src/labels/curvedLabel.h
hjanetzek: yup

core/src/labels/curvedLabel.cpp
hjanetzek: Label offset for curved labels needs the be applied to the polyline in updateScreenTransform. I'll do this in the next PR.

karimnaaji: Ok sounds good to be in a separate PR.

core/src/labels/curvedLabel.cpp
```diff
@@ -0,0 +1,266 @@
...
+#include "log.h"
+#include "style/textStyle.h"
+#include "text/fontContext.h"
+#include "textLabels.h"
+#include "util/geom.h"
+#include "util/lineSampler.h"
+#include "view/view.h"
+
+#include "alfons/path/splinePath.h"
+#include "alfons/path/ASPC.h"
```
karimnaaji: Both of those include seem unused

core/src/style/textStyleBuilder.cpp
```diff
@@ -209,59 +207,240 @@ bool TextStyleBuilder::addFeature(const Feature& _feat, const DrawRule& _rule) {
...
 
-                    float d = sqPointSegmentDistance(pp, p1, p);
-                    if (d > tolerance) { break; }
+        for (size_t j = i + 1; j < _line.size()-1; j++) {
+            glm::vec2 dir2 = sampler.segmentDirection(j);
 
-                    // Skip merged segment
-                    merged = true;
-                    i += 1;
+            float angle = crossProduct(dir1, dir2);
```
karimnaaji: Can we call this area instead? I think it's misleading to call it an angle.

core/src/style/textStyleBuilder.cpp
karimnaaji: Yep the rule for 'squiggly' labels is definitely important. The one up there seems to be _almost_ readable, but we can still run into cases where the letters can be upside down. It would be safe to add a rule to ensure that no quads gets upside down in screen space. I will give a look to implement that today.

Cases like this one are quite hard to read:
screen shot 2017-02-02 at 3 01 19 pm

hjanetzek: Yeah, that one should be handled. You can probably check in `updateScreenTransform` that within the label range `sampler.segmentDirection(i).x > threshold`. (After having determined the overall label direction)

core/src/style/textStyleBuilder.cpp
hjanetzek: I've updated the code and docs - are you ok with angle?

https://www.physicsforums.com/threads/using-cross-product-to-find-angle-between-two-vectors.510814/

core/src/labels/curvedLabel.cpp
hjanetzek: fixed

core/src/style/textStyleBuilder.cpp
karimnaaji: Yeah if dir1 and dir2 are normalized that's fine.

karimnaaji:
Merging this after https://github.com/tangrams/tangram-es/pull/1263 so we do a patch release for iOS.

tallytalwar:
Can we check all the places being tested for tangram JS, refer: https://github.com/tangrams/tangram/pull/484

hjanetzek:
@tallytalwar  yes, we should ask our cartography team for final acceptance test after the next PR :)

core/src/style/textStyleBuilder.cpp
```diff
@@ -322,9 +501,7 @@ TextStyle::Parameters TextStyleBuilder::applyRule(const DrawRule& _rule,
             p.labelOptions.priority = (float)priority;
         }
         _rule.get(StyleParamKey::text_collide, p.labelOptions.collide);
-        if (!_rule.get(StyleParamKey::text_interactive, p.interactive)) {
-            _rule.get(StyleParamKey::interactive, p.interactive);
-        }
+        _rule.get(StyleParamKey::text_interactive, p.interactive);
```
tallytalwar: `interactive` parameter is inherited from the parent `point` style, if `text` style itself is not having one defined.

core/src/style/textStyleBuilder.cpp
```diff
@@ -210,59 +208,240 @@ bool TextStyleBuilder::addFeature(const Feature& _feat, const DrawRule& _rule) {
...
-                    addLabel(_params, Label::Type::line, { a, a+b }, _rule);
-                    a += b;
+            if (!splitLine) {
+                // Go back within window to check for hard direction changes
+                for (int k = j - 1; k >= int(i); k--){
+                    if (glm::length2(sampler.segmentDirection(k) + dir2) < sqDirLimit) {
+                        //LOG("back break %s, %d %d", _params.text.c_str(), j, k);
+                        splitLine = true;
+                    }
+                    if (sampler.point(k).z < sampler.point(j).z - sampleWindow) {
```
tallytalwar: @hjanetzek can you explain the use of `sampleWindow` here? I am not sure I get it or maybe rename it to something which explains the context in which its used.

hjanetzek: The idea is to break the line also on sharp turns that are made of multiple segments. E.g. a 90° turn could be made of 3 30° 1px long segments. To detect these turns the current segment direction is compared to the previous segment directions within sampleWindow range.

core/src/style/textStyleBuilder.cpp
hjanetzek: Oh yes - fixed in pt3

PR #1257, opened 2017-01-26
Scene sets the pixel scale of its Styles

matteblair:
The Scene 'owns' its Styles, so when setting a pixel scale for the Scene, it should be applied to Styles in the Scene too.

This fixes a previous inconsistency: the correct pixel scale was set on the Scene in `Map::Impl::setScene`, but the Styles in the Scene would still have their default pixel scale.

This solves the same problem as https://github.com/tangrams/tangram-es/pull/1252, but I think this refactoring could spare us from similar problems in the future.


tallytalwar:
Cool, this is more robust of a fix. 👍

PR #1259, opened 2017-01-26
Scale SDF radius in FontContext with pixel scale

matteblair:
This should drastically reduce the frequency of the familiar log message: `NOTIFY textStyleBuilder.cpp:506: stroke_width too large: 12.000000 / 2.000000`. 

In FontContext, the width of a text stroke in pixels is capped by the radius used when generating SDFs, which is initialized to a value of 6 pixels. This is large enough for most reasonable strokes, but previously it _did not_ scale with pixel density! This meant that on a 2x display our maximum stroke was 3 pixels, on 3x it was 2 pixels, etc. 

The change here applies the pixel scale to the SDF radius in FontContext, allowing for a much more reasonable range of stroke widths.

### DIFF
--- a/core/src/scene/scene.cpp
+++ b/core/src/scene/scene.cpp
@@ -146,6 +146,7 @@ void Scene::setPixelScale(float _scale) {
     for (auto& style : m_styles) {
         style->setPixelScale(_scale);
     }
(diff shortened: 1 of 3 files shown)


matteblair:
@hjanetzek does this look alright to you?

PR #1260, opened 2017-01-27
Initial iOS release - 0.4.0

msmollin:
Updates CMake and the Info.Plist entry manually (for now) to set the appropriate alpha version release number.


msmollin:
@matteblair / @karimnaaji how do you folks normally handle / do you handle tagging releases on the main ES codebase for Android?

ios/framework/Info.plist
```diff
@@ -15,7 +15,7 @@
 	<key>CFBundlePackageType</key>
 	<string>FMWK</string>
 	<key>CFBundleShortVersionString</key>
-	<string>1.0</string>
+	<string>0.4.0</string>
```
ecgreb: Is it an issue that there is a trailing zero in the version number here but not in the framework cmake file?

msmollin: Not really :) We could fix it to be completely clear but its not a technical issue.

ecgreb: Fine with me then just making sure

tallytalwar:
We do tag releases for android. Example last android release is tagged as `0.4.8-android`.
We can follow same semantics for tagging ios releases. Example this one will be: `0.4.0-ios`. We also make sure the version number in the tag matches the one in [`android/tangram/gradle.properties`](https://github.com/tangrams/tangram-es/blob/833d615d08a6d954149e9cc7e543dd5c2a8aea41/android/tangram/gradle.properties#L2)

msmollin:
Ok I'll update the cmake file for 0.4.0 then so we're properly semantic.

toolchains/iOS.framework.cmake
```diff
@@ -3,7 +3,7 @@ include(${CMAKE_SOURCE_DIR}/toolchains/iOS.toolchain.cmake)
 message(STATUS "Build for iOS archs " ${CMAKE_OSX_ARCHITECTURES})
 
 set(FRAMEWORK_NAME TangramMap)
-set(FRAMEWORK_VERSION "1.0")
```
tallytalwar: Can we keep this consistent with major, minor and patch numbers here?

This will keep it consistent with what we have for android in https://github.com/tangrams/tangram-es/blob/master/android/tangram/gradle.properties#L2.

Also how are we planning on differentiating `SNAPSHOT`s with `RELEASE` versions?

tallytalwar: Ahha ... you updated with the trailing `0`. But `SNAPSHOT` question please :D.

msmollin: There is no general concept of `SNAPSHOT` releases in iOS land. If you wanted to do something like that, you could use `ALPHA` or `BETA` tags.

ecgreb: I don't believe Cocoapods has the same concept of SNAPSHOT artifacts as Maven/Gradle. Seems like the closest analog would be pointing the podspec at a specific commit hash?

msmollin: Yeah something like `0.4.1-alpha1` is generally the nomenclature you'd see.

tallytalwar: OK, something like a `-beta`, `-alpha` or maybe `-dev` will probably be meaningful.

msmollin: Are you suggesting this for this release or in the future?

tallytalwar: Post this release.

karimnaaji:
To cleanup this a bit we can safely remove all the cmake related versioning, I added those as a placeholders when adding binary framework through cmake (reference [here](https://cmake.org/cmake/help/v3.5/prop_tgt/FRAMEWORK.html)) but 
reading the docs more closely it seems that this particular option [FRAMEWORK_VERSION](https://cmake.org/cmake/help/v3.5/prop_tgt/FRAMEWORK_VERSION.html) only affects OSX.

msmollin:
@karimnaaji removed.

toolchains/iOS.framework.cmake
msmollin: So, just to follow up - this is going to be updated to use `-dev` when we're in development, and `-alpha`/`-beta` tags to line up with appropriate level of release status.

PR #1263, opened 2017-01-27
Add platform instance 

### LINKED ISSUE #1243: Multiple map instances causes strange rendering issues on iOS
The Mapzen iOS SDK sample app utilizes a UITabBarController which has 3 view controllers connected. All 3 VCs utilizes a TGMapViewController to display either route lines, search pins, or the device's current location.

Two major things seem to happen when, after launching, we select one of the other tabs (which loads a second map):
1) The original map (currently the Maps tab) appears to become completely unresponsive to gesture inputs. However, watching the logging output with LLDB attached appears that the code is still getting gesture inputs, resulting in the next bullet.

2) Tiles stop rendering correctly, resulting in screens that look like this:
screen shot 2017-01-20 at 4 54 00 pm

So there appears to be something getting paused in the rendering loop. Thoughts?

karimnaaji:
For Tangram to be fully instantiable on some platforms, we need the platform code to be tied to the map instance whenever the instance depends on the some of platform state (http handler/pixel scale/continuous rendering...). Whenever the platform code share common states, we have seen some noticeable issues on iOS #1243. 

 - [x] Update tests
 - [x] Update android, rpi, ~~tizen~~, osx, linux

The relationship between platform and map instance is a 1:1. For now, it is held as a `shared_ptr` in the map instance, mostly for safety reason within lambda functions and threads (we discussed usage of `unique_ptr`, but the lifetime of the platform instance would involve more sync within the threads to ensure the platform remains whenever accessed). To reduce `shared_ptr` cost, another solution is the usage of a raw pointer that has the process lifetime.

Tested on iOS with multiple view controllers and fixes #1243.


android/tangram/src/main/cpp/platform_android.cpp
```diff
@@ -423,7 +398,7 @@ void markerPickCallback(jobject listener, const Tangram::MarkerPickResult* marke
 
     jniEnv->CallVoidMethod(listener, onMarkerPickMID, markerPickResultObject, position[0], position[1]);
     jniEnv->DeleteGlobalRef(listener);
-
+    jniEnv->DeleteGlobalRef(tangramInstance);
```
matteblair: Why delete a global ref for this object? Did we create a global ref to it somewhere?

karimnaaji: No it's good to have feedback on it already! 
Good call, I added the `GlobalRef` since `markerPickCallback` is called from [a lambda](https://github.com/tangrams/tangram-es/pull/1263/files/d08218364accf2d6cc30afef9adb569cc1e13a5f#diff-ca9632f59860c7cc13ebd3493d69312dR246) earlier, but the platform instance already holds a global ref on it until the disposal of the map. It can be removed but it seems safe to leave it as it is.

karimnaaji:
@hjanetzek can you make sure this runs fine on linux platform?

hjanetzek:
@karimnaaji  testing now. 

What is the reason that you use 'const Platform'?  If it is intended to divide which functions may be called I think it would be better to use interfaces for this like:

`class Platform : DownloadQueue, PlatformOptions, ... {} `

and only pass the interfaces down to the using classes should be available in, say, TileWorker, etc.

karimnaaji:
I only wanted to fix issues relating to multiple map instances that were causing problems on #1243 without adding to much complexity, dividing which functions may be called from different part of the codebase is another issue that we can discuss more after this PR, it seems fine to leave as a unique interface to address that issue in a first step.

core/src/scene/sceneLoader.cpp
```diff
@@ -156,14 +156,14 @@ void SceneLoader::applyGlobals(Node root, Scene& scene) {
...
-bool SceneLoader::applyConfig(const std::shared_ptr<Scene>& _scene) {
+bool SceneLoader::applyConfig(const std::shared_ptr<Platform>& _platform, const std::shared_ptr<Scene>& _scene) {
 
     Node& config = _scene->config();
 
     // Instantiate built-in styles
     _scene->styles().emplace_back(new PolygonStyle("polygons"));
     _scene->styles().emplace_back(new PolylineStyle("lines"));
-    _scene->styles().emplace_back(new DebugTextStyle("debugtext", true));
+    _scene->styles().emplace_back(new DebugTextStyle("debugtext", std::make_shared<FontContext>(_platform), true));
```
tallytalwar: Why not use the `Scene::m_fontContext` here also like dont for other TextStyle?

core/src/tangram.cpp
```diff
@@ -243,6 +250,10 @@ void Map::queueSceneUpdate(const std::vector<SceneUpdate>& sceneUpdates) {
     impl->sceneUpdates.insert(impl->sceneUpdates.end(), sceneUpdates.begin(), sceneUpdates.end());
 }
 
+std::shared_ptr<Platform> Map::getPlatform() const {
```
tallytalwar: nitpick, lets call this `platform` for consistency with rest of the code base.

core/src/tangram.h
```diff
@@ -296,11 +299,17 @@ class Map {
     // Run this task asynchronously to Tangram's main update loop.
     void runAsyncTask(std::function<void()> _task);
 
+    std::shared_ptr<Platform> getPlatform() const;
```
tallytalwar: maybe should return a const reference!

core/src/tile/tileManager.cpp
```diff
@@ -15,16 +15,16 @@
...
 
-TileManager::TileManager(TileTaskQueue& _tileWorker) :
+TileManager::TileManager(std::shared_ptr<Platform> platform, TileTaskQueue& _tileWorker) :
     m_workers(_tileWorker) {
 
     m_tileCache = std::unique_ptr<TileCache>(new TileCache(DEFAULT_CACHE_SIZE));
 
     // Callback to pass task from Download-Thread to Worker-Queue
-    m_dataCallback = TileTaskCb{[this](std::shared_ptr<TileTask> task) {
+    m_dataCallback = TileTaskCb{[this, &platform](std::shared_ptr<TileTask> task) {
```
tallytalwar: how much difference will it be to pass an explicit reference of a shared_ptr to a lambda vs passing a copy and creating a internal shared reference of `platform`?

core/src/tile/tileWorker.h
```diff
@@ -32,6 +34,8 @@ class TileWorker : public TileTaskQueue {
 
     void setScene(std::shared_ptr<Scene>& _scene);
 
+    std::shared_ptr<const Platform> getPlatform() { return m_platform; }
```
tallytalwar: are we using this, probably do not need this as off now!

rpi/src/platform_rpi.h
```diff
@@ -3,3 +3,24 @@
 #include "platform.h"
 
 void processNetworkQueue();
+
+#define NUM_WORKERS 3
+
+class RPiPlatform : public Platform {
```
tallytalwar: Since RPi and Linux platforms are very similar, probably make sense for RPiPlatform to inherit from LinuxPlatform and override relevant bits from there. Will result in more code reuse.

lygstate: +1

lygstate: Also the Tizen.

core/src/tangram.cpp
karimnaaji: We have to fix ourselves some coding rules concerning this within the codebase. I stayed consistent with the rest of the public API here (`getPixelScale`, `getViewportHeight`, `getTilt`...).

core/src/scene/sceneLoader.cpp
karimnaaji: The behavior stays the same here, see https://github.com/tangrams/tangram-es/pull/1263/files/7ebd47ea2232dab4febaf26c21363126738046b3#diff-7822ef08be0ac1769ea05762681bbf15L22.

The debug text style would have no overhead synchronizing on regular text rendering; I also didn't want to introduce changes in this behavior on this PR.

android/tangram/src/main/cpp/jniExports.cpp
```diff
@@ -237,12 +239,13 @@ extern "C" {
...
     }
 
-    JNIEXPORT void JNICALL Java_com_mapzen_tangram_MapController_nativePickMarker(JNIEnv* jniEnv, jobject obj, jlong mapPtr, jfloat posX, jfloat posY, jobject listener) {
+    JNIEXPORT void JNICALL Java_com_mapzen_tangram_MapController_nativePickMarker(JNIEnv* jniEnv, jobject obj, jobject tangramInstance, jlong mapPtr, jfloat posX, jfloat posY, jobject listener) {
         assert(mapPtr > 0);
         auto map = reinterpret_cast<Tangram::Map*>(mapPtr);
         auto object = jniEnv->NewGlobalRef(listener);
-        map->pickMarkerAt(posX, posY, [object](auto pickMarkerResult) {
-            markerPickCallback(object, pickMarkerResult);
+        auto instance = jniEnv->NewGlobalRef(tangramInstance);
```
karimnaaji: Yep

core/src/tile/tileManager.cpp
karimnaaji: Difference concerning access time to the pointer? I would say that it's fairly similar, increase usage count is done only once if passed by copy, no increment should be done on usage count if passed by ref.

It's good you pointed that out, it's a leftover from a previous implementation where a reference was used, and it shouldn't be safe to pass it by reference here considering the scope of `platform`. Will update.

core/src/tile/tileWorker.h
karimnaaji: It's used in the platform code in general, but I would prefer not to expose that :/
Let me know if you have ideas on how we could get rid of this.

core/src/data/clientGeoJsonSource.cpp
```diff
@@ -44,15 +44,15 @@ ClientGeoJsonSource::ClientGeoJsonSource(const std::string& _name, const std::st
         std::regex r("^(http|https):/");
         std::smatch match;
         if (std::regex_search(_url, match, r)) {
-            startUrlRequest(_url,
+            m_platform->startUrlRequest(_url,
```
karimnaaji: It is already included here https://github.com/tangrams/tangram-es/pull/1263/files/7ebd47ea2232dab4febaf26c21363126738046b3#diff-3cdb39a640f4b049c126ddf4f998bffaR6.

core/src/tangram.h
karimnaaji: Updated.

core/src/tile/tileManager.cpp
karimnaaji: Updated.

tallytalwar: Yup, I was not sure whether a reference is apt here, a copy seems more safe and logical. 👍

android/tangram/src/main/cpp/jniExports.cpp
```diff
@@ -367,7 +370,7 @@ extern "C" {
         assert(mapPtr > 0);
         auto map = reinterpret_cast<Tangram::Map*>(mapPtr);
         auto sourceName = stringFromJString(jniEnv, name);
-        auto source = std::shared_ptr<Tangram::TileSource>(new Tangram::ClientGeoJsonSource(sourceName, ""));
+        auto source = std::shared_ptr<Tangram::TileSource>(new Tangram::ClientGeoJsonSource(map->getPlatform(), sourceName, ""));
```
tallytalwar: Ok, I see the use of `getPlatform` here (In reference to my previous comment at the declaration of `getPlatform`).

Probably it make sense to abstract out these methods to apt `Map` functions.

Example: 
`map->createClientSource(sourceName, "");`

tallytalwar: Ahh.. but I am not sure if [this](https://github.com/tangrams/tangram-es/blob/168903c9feb5cdd22e40c09e79123320707e23d5/android/tangram/src/main/cpp/jniExports.cpp#L111) can be abstracted out within a `Map` function.

core/src/data/clientGeoJsonSource.cpp
tallytalwar: Ohh I missed it somehow. 👍

rpi/src/platform_rpi.h
karimnaaji: I only made Raspberry Pi instantiate a `LinuxPlatform` directly.

Updated in: https://github.com/tangrams/tangram-es/pull/1263/commits/63e5a18585b28b8cce9debfa56880cb1713cadc3

Issue #1265, opened 2017-01-30
Demo app fails to load Refill style (Android)


ecgreb:
Steps to reproduce:

1. Clone https://github.com/tangrams/refill-style-more-labels into `[PROJECT_ROOT]/scenes` folder.
2. Update `MainActivity.java` line 57 to load Refill style.

```java
view.getMapAsync(this, "refill-style-more-labels/refill-style-more-labels.yaml");
```

3. Launch demo app and notice no map is loaded (black screen). Setting other house styles in this fashion successfully loads a map (tested with Bubble Wrap and Cinnibar).

matteblair:
Verified - this is not specific to Android, it's a bug in a core library dependency. The commit of yaml-cpp that we use currently has a bug that causes a byte-order mark in a YAML file to cause a parsing error. 

We've fixed this temporarily in tangram-es by just skipping a BOM before we pass the scene file to the parser.

ecgreb:
👍

PR #1267, opened 2017-01-30
Eat UTF-8 BOM from scene file

matteblair:
Resolves https://github.com/tangrams/tangram-es/issues/1265

This will be handled within yaml-cpp later.


nvkelso:
Is this something I should fix in the style?

matteblair:
@nvkelso No. You _could_ modify the file to fix this, but it's a difference that's not even visible in a text editor so it's unreasonable to try to maintain.

PR #1272, opened 2017-02-02
Curved labels part III

hjanetzek:
- Disable curved layout for complex shaped scripts
- use ICU for better text-transforms (i.e. support non-ascii chars)


karimnaaji:
@hjanetzek will start to review this when rebased.

tallytalwar:
> One experiment that we could try on this would be to lower down breakage to words instead of label when the shaping is complex. It could show up more labels on those areas without breaking too much readability, I think.

I agree. Breaking on word boundaries will be good. Unless for some language that also breaks the context (I do not think so, based on the list here: http://r12a.github.io/scripts/featurelist)

hjanetzek:
Breaking on word boundaries seems to be quite a bit of work I would rather move to another PR.

One thing we could also try: Instead of completely disabling curved placement for complex scripts we could use a stricter limit for angle between segments.

PR #1277, opened 2017-02-06
With polygon geometry text style should be drawn at centroid

tallytalwar:
Based on the [documentation](https://mapzen.com/documentation/tangram/Styles-Overview/#text) for a polygon geometry text should be drawn at the centroid of the polygon. 

Also tested this in play for tangram js behavior.

### DIFF
--- a/core/src/style/textStyleBuilder.cpp
+++ b/core/src/style/textStyleBuilder.cpp
@@ -189,11 +189,12 @@ bool TextStyleBuilder::addFeature(const Feature& _feat, const DrawRule& _rule) {
 
     } else if (_feat.geometryType == GeometryType::polygons) {
 
-        for (auto& polygon : _feat.polygons) {
-            for (auto& line : polygon) {
-                for (auto& point : line) {
-                    addLabel(params, Label::Type::point, { point }, _rule);
-                }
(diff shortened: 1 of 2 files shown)


core/src/style/textStyleBuilder.cpp
```diff
@@ -189,12 +189,13 @@ bool TextStyleBuilder::addFeature(const Feature& _feat, const DrawRule& _rule) {
...
-                for (auto& point : line) {
-                    addLabel(params, Label::Type::point, { point }, _rule);
-                }
+        glm::vec3 c;
+        const auto& polygons = _feat.polygons;
+        for (const auto& polygon : polygons) {
+            if (!polygon.empty()) {
+                c = centroid(polygon.front().begin(), polygon.front().end());
             }
+            addLabel(params, Label::Type::point, { c }, _rule);
```
karimnaaji: Why declare `c` out of the loop? We risk to use it initialized to `(0,0,0)`.

We should probably put this statement in the condition.

tallytalwar: Updated.

tallytalwar: Actually updated now!

PR #1279, opened 2017-02-08
Style shader cache

hjanetzek:
This PR modifies `Style` to use the same ShaderProgram for styles with identical sources 

- refactor: build `Light` shader blocks only once
- refactor ShaderProgram, move template engine to ShaderSource.

### DIFF
--- a/core/src/gl/shaderSource.h
+++ b/core/src/gl/shaderSource.h
@@ -1,0 +1,67 @@
+#pragma once
+
+#include <map>
+#include <string>
+#include <vector>
+
+#define SHADER_SOURCE(NAME) ShaderSource::shaderSourceBlock(NAME ## _data, NAME ## _size)
+
+namespace Tangram {
+
+class ShaderSource {
+public:
+
+    // Set the vertex and fragment shader GLSL source to the given strings/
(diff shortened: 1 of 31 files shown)


core/src/gl/shaderSource.h
```diff
@@ -0,0 +1,67 @@
...
+    // Add a block of GLSL to be injected at "#pragma tangram: [_tagName]" in the shader sources.
+    void addSourceBlock(const std::string& _tagName, const std::string& _glslSource,
+                        bool _allowDuplicate = true);
+
+    void addExtensionDeclaration(const std::string& _extension);
+
+    const auto& getSourceBlocks() const { return m_sourceBlocks; }
+
+    // Build vertex shader source
+    std::string vertexSource() const {
```
matteblair: I think most of the core code uses the convention that `someMember()` (with no verb) is an accessor and `getSomeMember()` or `verbSomeMember()` is doing nontrivial work. Do you think `getVertexSource()` or `composeVertexSource()` would be more clear?

hjanetzek:
Updated with all suggestions included - thanks

PR #1280, opened 2017-02-08
Add release buildType for tangram lib

tallytalwar:
- to disable `-g` debug flag being set by the ndk default android cmake toolchain
- Refer: https://github.com/android-ndk/ndk/issues/243

TODO: Keep an eye for future versions of ndk/android-studio to add a fix for this


matteblair:
This reduces the `fullRelease` AAR from 62.5 MB to 12.8 MB. :+1: :+1: :+1:

PR #1287, opened 2017-02-10
Take marker visible into account for selection and zoom rebuilds

### LINKED ISSUE #1276: Marker Visibility
markerSetVisible is not working as expected. 
1. when visible is set to false it hides the marker but when tap on the area where marker was shown it throws the callback for marker selection.
2. when visible is set to false it hides the marker but on panning or pinching as the marker rebuilt itself it start showing.

tallytalwar:
Fixes #1276


karimnaaji:
@tallytalwar Added a fix to set visibility back on context loss, can you confirm https://github.com/tangrams/tangram-es/pull/1287/commits/8b527a64ad8527467e257199545fc3d2d9a97589?

tallytalwar:
@karimnaaji looks good. Will update and merge this.

PR #1290, opened 2017-02-10
Add iOS documentation

### LINKED ISSUE #1266: Add tangram ios documentation
And publish it aptly under https://mapzen.com/documentation/tangram/

cc. @karimnaaji @msmollin

karimnaaji:
The iOS framework had no documentation (moved from 0% to 97% so far), this PR adds documentation with a jazzy integration to the build system to be able to generate them to host them later on.

Install jazzy with:

```sh
$ gem install jazzy
```

Generate the docs with:
```sh
$ make ios-docs
```

The docs should be created under `build/ios-docs/`:

screen shot 2017-02-10 at 5 11 07 pm

fixes #1266


ios/src/TangramMap/TGLabelPickResult.h
```diff
@@ -0,0 +1,43 @@
...
+};
+
+/**
+ Data structure holding the result of a label selection that occured on the map view.
+
+ See `-[TGMapViewController pickLabelAt:]` and `[TGMapViewDelegate mapView:didSelectLabel:atScreenPosition:]`.
+ */
+@interface TGLabelPickResult : NSObject
+
+- (instancetype) initWithCoordinates:(TGGeoPoint)coordinates type:(TGLabelType)type properties:(NSDictionary *)properties;
```
msmollin: Need docs for initializer.

ios/src/TangramMap/TGMapViewController.h
```diff
@@ -0,0 +1,731 @@
+//
+//  TGMapViewController.h
+//  TangramiOS
```
msmollin: TangramMap?

ios/src/TangramMap/TGMapViewController.h
```diff
@@ -0,0 +1,731 @@
...
+    /// Type for a <a href="https://mapzen.com/documentation/tangram/Cameras-Overview/#perspective-camera">perspective camera</a>
+    TGCameraTypePerspective = 0,
+    /// Type for an <a href="https://mapzen.com/documentation/tangram/Cameras-Overview/#isometric-camera">isometric camera</a>
+    TGCameraTypeIsometric,
+    /// Type for a <a href="https://mapzen.com/documentation/tangram/Cameras-Overview/#flat-camera">flat camera</a>
+    TGCameraTypeFlat,
+};
+
+/**
+ These defines describe an ease type function to be used for camera or other transition animation.
```
msmollin: These *enumerations* describe

ios/src/TangramMap/TGMapViewController.h
```diff
@@ -0,0 +1,731 @@
...
+    TGDebugFlagTangramStats,
+    /// Draw feature selection framebuffer
+    TGDebugFlagSelectionBuffer,
+};
+
+
+NS_ASSUME_NONNULL_BEGIN
+
+/**
+ A gesture recognizer delegate that can be implemented to receive gesture events from the map view.
```
msmollin: Just a thought - much of the documentation below refers to a "logical pixel" - it might make sense up here to define just what that is because most developers won't understand there's a fair amount of translation from OpenGL -> UIKit pixel space.

ios/src/TangramMap/TGMapViewController.h
```diff
@@ -0,0 +1,731 @@
...
+
+    - If the `TGRecognizerDelegate` doesn't respond to `shouldRecognize*`, the map view will proceed with
+    calling default gesture handling code.
+
+ Finally, the map view checks whether the `TGRecognizerDelegate` implements `didRecognize*` and call this
+ method if so.
+
+ @note Those methods are all **optional**.
+ */
+@protocol TGRecognizerDelegate <NSObject>
```
msmollin: Code organization thought (my OCD kicking in) for these methods:

1) Put all the *shouldRecognize*  functions together, and all the *didRecognize* functions together beneath them, since that's the order they happen in.

OR

2) Put the *shouldRecognizeX* function and the corresponding *didRecognizeX* function together in blocks, to indicate the pairing. I can't remember without looking at the code but I'm pretty sure this is how the implementation file is organized.

ios/src/TangramMap/TGMapViewController.h
```diff
@@ -0,0 +1,731 @@
...
+ This view provides default gesture handlers for tap, double tap, long press, pan,
+ pinch, rotate and shove.
+
+ The public interface provides dynamic map marker placement, change of camera view
+ settings, and map description changes through scene updates.
+
+ This view uses scenes descibed by the
+  <a href="https://mapzen.com/documentation/tangram/">Tangram YAML scene syntax</a>
+ allowing you to fully customize your map description using your own data.
+ Some already made basemap styles can be found
```
msmollin: Nitpick: Some *pre-made* basemap styles

ios/src/TangramMap/TGMapViewController.h
```diff
@@ -0,0 +1,731 @@
...
+ <a href="https://mapzen.com/documentation/cartography/styles/">here</a> using Mapzen
+ data sources.
+
+ To use basemap styles you can <a href="https://mapzen.com/developers/sign_in">sign up for
+ an API key</a> and load it through your application:
+
+ ```swift
+ let sceneURL = "https://mapzen.com/carto/walkabout-style-more-labels/walkabout-style-more-labels.yaml";
+ view.loadSceneFile(sceneURL, sceneUpdates: [ TGSceneUpdate(path: "sources.mapzen.url_params", value: "{ api_key: \(YOUR_API_KEY) }") ]);
```
msmollin: What, no objective-c example? :D

ios/src/TangramMap/TGMapViewController.h
```diff
@@ -0,0 +1,731 @@
...
+ @param mapView a pointer to the map view
+ */
+- (void)mapViewDidCompleteLoading:(nonnull TGMapViewController *)mapView;
+@end
+
+NS_ASSUME_NONNULL_BEGIN
+
+/**
+ `TGMapViewController` is a flexible and customizable map view, allowing you to embed
+ it inside a <a href="https://developer.apple.com/reference/glkit/glkview">GLKView</a> component.
```
msmollin: Mmmm TGMapViewController is a `GLKitViewController` which manages the lifecycle of a `GLKView` so this might need a little re-wording? In practice, you don't embed anything inside the `GLKView`, unless I'm misunderstanding something here?

ios/src/TangramMap/TGMapViewController.h
```diff
@@ -0,0 +1,731 @@
...
+ ```swift
+ let sceneURL = "https://mapzen.com/carto/walkabout-style-more-labels/walkabout-style-more-labels.yaml";
+ view.loadSceneFile(sceneURL, sceneUpdates: [ TGSceneUpdate(path: "sources.mapzen.url_params", value: "{ api_key: \(YOUR_API_KEY) }") ]);
+ ```
+ */
+@interface TGMapViewController : GLKViewController <UIGestureRecognizerDelegate>
+
+/**
+ If continuous rendering is set to `true`, the map view will render continuously, otherwise,
+ the map will render only when necessary.
```
msmollin: "When necessary" is slightly ambiguous. Maybe elaborate on what might trigger a render?

ios/src/TangramMap/TGMapViewController.h
```diff
@@ -0,0 +1,731 @@
...
+ Some styles can be set to be `animated` and this default value will be set appropriately,
+ see the <a href="https://mapzen.com/documentation/tangram/scene/#animated">
+ style parameter</a> for more details.
+
+ @note Any changes to this value will override the default induced value from the `animated`
+ style parameter.
+ Before making changes to this value, make sure battery usage is not something critical
```
msmollin: I'd rewrite this to be more explicit: "Setting this parameter to `true` will negatively impact battery life if left set  for extended periods of time. Before setting, make sure battery usage is not something critical to your application."

ios/src/TangramMap/TGMapViewController.h
```diff
@@ -0,0 +1,731 @@
...
+ any network related configuration (cache path and network cache size, both memory and disk,
+ or be notified when a network request completes).
+ More informations on the default provided configuration can be found on the description
+ of `TGHttpHandler`.
+ */
+@property (strong, nonatomic) TGHttpHandler* httpHandler;
+
+#pragma mark Debug toggles
+
+/**
```
msmollin: Outside of scope - it might be nice to have a way to set an array of debug flags some day.

ios/src/TangramMap/TGMapViewController.h
```diff
@@ -0,0 +1,731 @@
...
+ */
+- (void)markerRemoveAll;
+
+/**
+ Creates a marker identifier which can be used to dynamically add points and polylines
+ to the map.
+
+ Example on how to add a point to the map for a geographic coordinate.
+
+ ```objc
```
msmollin: What, no swift example? :D

ios/src/TangramMap/TGMapViewController.h
```diff
@@ -0,0 +1,731 @@
...
+ [view markerSetStyling:identifier styling:@"{ style: 'points', color: 'white', size: [25px, 25px], order:500 }"];
+ [view markerSetPoint:identifier coordinates:coordinate];
+ ```
+
+ @return a marker identifier
+ @note The marker should be appropriately styled using `-markerSetStyling:styling:`
+ and a type must be set (point, polyline, polygon) through `markerSetPoint*`, `markerSetPolyline` or `markerSetPolygon`
+ for it to be visible.
+
+ @note A marker identifier of 0 is non-valid.
```
msmollin: Might want to be a little more explanatory and indicate that 0 is returned when adding a marker to the map failed.

ios/src/TangramMap/TGMapViewController.h
```diff
@@ -0,0 +1,731 @@
...
+ Sets an image loaded with a <a href="https://developer.apple.com/reference/uikit/uiimage">
+ UIImage</a> to a marker styled with a <a href="https://mapzen.com/documentation/tangram/Styles-Overview/#points">
+ points style</a>.
+
+ ```obj
+ TGMapMarkerId mid = [vc markerAdd];
+ [mapView markerSetStyling:mid styling:@"{ style: 'points', color: 'white', size: [25px, 25px], order:500 }"];
+ [mapView markerSetPoint:mid coordinates:coordinate];
+ [mapView markerSetImage:mid image:[UIImage imageNamed:@"marker-image.png"]];
+ ```
```
msmollin: No swift love :(

ios/src/TangramMap/TGMapViewController.h
```diff
@@ -0,0 +1,731 @@
...
+
+#pragma mark Scene loading - updates interface
+
+/**
+ Loads a scene file synchronously.
+
+ If the scene file is set as a resource in your application bundle, make sure to resolve
+ the path for this URL relative to your bundle (for example `Resources/scene.yaml`).
+
+ If your scene is hosted remotely (any path starting with `https://` is considered a remote scene file)
```
msmollin: Do we support http? if not, should be clear about that.

ios/src/TangramMap/TGMapViewController.h
```diff
@@ -0,0 +1,731 @@
...
+
+ @param componentPath the path of the scene component to update
+ @value the value to assign to the YAML component selected with `componentPath`
+
+ @note Scene updates must be applied with `-applySceneUpdates:`
+ */
+- (void)queueSceneUpdate:(NSString*)componentPath withValue:(NSString*)value;
+
+/**
+ Queue a list of scene updates.
```
msmollin: Needs @param value for sceneUpdates

ios/src/TangramMap/TGMapViewController.h
```diff
@@ -0,0 +1,731 @@
...
+ @return the longitude and latitude, `(nil, nil)` if the point is not visible on the screen
+ */
+- (TGGeoPoint)screenPositionToLngLat:(CGPoint)screenPosition;
+
+#pragma mark Map View animations - Position interface
+
+/**
+ Animate the map view to another geographic coordinate.
+
+@param position the map view geographic coordinate
```
msmollin: Should be indented here?

ios/src/TangramMap/TGMapViewController.mm
```diff
@@ -712,7 +706,8 @@ - (void)respondToShoveGesture:(UIPanGestureRecognizer *)shoveRecognizer
 }
 #pragma mark Standard Initializer
 
-- (instancetype)initWithNibName:(NSString *)nibNameOrNil bundle:(NSBundle *)nibBundleOrNil {
+- (instancetype)initWithNibName:(NSString *)nibNameOrNil bundle:(NSBundle *)nibBundleOrNil
+{
```
msmollin: Scandalous haha. But seriously do you prefer { on its own line? It's something the iOS community has moved away from but used to be the way to go.  I don't really care but we should standardize on it either way.

ios/src/TangramMap/TGGeoPolygon.h
```diff
@@ -0,0 +1,88 @@
...
+ */
+- (TGGeoPoint*)coordinates;
+
+/**
+ The list of polygon `rings` contained in this geometry.
+ The sum of the values in `-[TGGeoPolygon rings:]` sum up to `-[TGeoPolygon count:]`.
+
+ @return a pointer to the list of polygon rings.
+ */
+- (int*)rings;
```
karimnaaji: The user shouldn't have to use this, it should be fine to leave this one as it is since I use it only to directly interface with the C++ API of Tangram. There could be another helper accessor using `NSArray` for convenience if we find the need to.

This method is only used internally, it is private/internal for the framework but still exposed publicly for simplicity. I'm planing to add a `TGGeoPolygon+Internal.h` as described [here](http://stackoverflow.com/questions/19095716/hiding-properties-from-public-framework-headers-but-leaving-available-internall) that has those methods used only in the scope of the framework. Let me know your thoughts on this!

ios/src/TangramMap/TGMapViewController.h
karimnaaji: Considering the numerous cases when a render needs to be requested https://github.com/tangrams/tangram-es/search?utf8=%E2%9C%93&q=requestRender it's a little hard to describe, but I'll reword that bit.

ios/src/TangramMap/TGMapViewController.mm
karimnaaji: I don't have any preference here, just tried to be consistent with initial styling, but we can switch based on what is more standard.

ios/src/TangramMap/TGHttpHandler.h
```diff
@@ -0,0 +1,78 @@
...
+ completed an asynchronous request.
+*/
+typedef void(^TGDownloadCompletionHandler)(NSData*, NSURLResponse*, NSError*);
+
+/**
+ A configurable http handler used in a `TGMapViewController`.
+
+ `TGMapViewController` provides a default http handler with the following configuration:
+
+    - cache location: `/tangram_cache`
```
karimnaaji: This interfaces directly with `NSURLCache`, and it seems to be a subdirectory within the bundle application on iOS:

>In iOS, path is the name of a subdirectory of the application’s default cache directory in which to store the on-disk cache (the subdirectory is created if it does not exist).

c.f: https://developer.apple.com/reference/foundation/urlcache/1415637-init

karimnaaji: We set the default to `tangram_cache` [here](https://github.com/tangrams/tangram-es/blob/master/ios/src/TGMapViewController.mm#L748).

ios/src/TangramMap/TGMapViewController.h
karimnaaji: [Yes we do](https://github.com/tangrams/tangram-es/blob/master/core/src/util/url.cpp#L391_L392), I'll update the comment here.

ios/src/TangramMap/TGLabelPickResult.h
karimnaaji: I've set `skip_undocumented` to `YES` in the jazzy yaml description so we don't expose things that the user shouldn't use, here it is an initializer that should only be used by `TGMapViewController` to create the results of a selection query but I can still document it and put a note about that?

ios/src/TangramMap/TGMapViewController.h
karimnaaji: Updated in https://github.com/tangrams/tangram-es/pull/1290/commits/9a0f38447fc50b6a0585571cb969b83198050358

ios/src/TangramMap/TGLabelPickResult.h
msmollin: Yeah I'd probably still make a note about it.

ios/src/TangramMap/TGMapViewController.h
karimnaaji: I would update to only objective-c since I think that both are a little too much to give the user any additional information on the API usage, but let me know your preference, only swift, only objective-c or both?

ios/src/TangramMap/TGLabelPickResult.h
karimnaaji: Ok then, _nothing to see here_.

ios/src/TangramMap/TGMapViewController.h
karimnaaji: Maybe swift is better since it has color highlighting in the generated docs.

msmollin:
Look good to me!

ios/src/TangramMap/TGMapViewController.h
```diff
@@ -0,0 +1,765 @@
...
+ */
+- (TGMapMarkerId)markerAdd;
+
+/**
+ Sets the styling for a marker identifier.
+
+ See the more detailed scene <a href="https://mapzen.com/documentation/tangram/Styles-Overview/">documentation</a>
+ to get more styling informations.
+
+ @param identifier the marker identifier created with `-markerAdd`
```
tallytalwar: missing second parameter here?

Issue #1264, opened 2017-01-28
Using Marker Animation causing run away view Re-rendering.


ac87:
See this sample app and in particular this [commit ](https://github.com/ac87/tangram_android/commit/dfcfb974b85667fecc450ddef22979c6457aa35d)

If `myLocationMarker.setPointEased` is used, the `MapController.ViewCompleteListener()` registered in `MainActivity ` gets called constantly ("View complete" appearing in the LogCat). This would be expected for 200ms but not constantly after.

Stopping the location updates doesn't stop it either.

matteblair:
Hi again, thanks for the report. From the code here and your description this does sound like a bug - the view should only re-render continuously for the duration of the easing. 

I will try to replicate this behavior and see if I can find a work-around for you while we fix it.

matteblair:
Haha whoops - it looks like Marker easing is currently interpreting the time parameter as _seconds_ when it should interpret it as _milliseconds_, so the Markers are actually moving continuously - just very very slowly. That's our bad :\

We'll get this fixed in the next release but until then you can just divide your Marker easing times by 1000.

PR #1298, opened 2017-02-14
Use milliseconds for duration in Marker setPointEased

matteblair:
Resolves https://github.com/tangrams/tangram-es/issues/1264


matteblair:
@tallytalwar The iOS interface appears to be consistent here, it uses seconds for all durations. Apple's UIKit animations use seconds as well, so I think this is the right choice.

matteblair:
I will, however, change the name of the 'duration' parameters in TGMapViewController, to clarify the units.

karimnaaji:
@tallytalwar updated the param name to `seconds` in #1290 .

tallytalwar:
Need to rebase and resolve conflicts, good to go when done.

PR #1299, opened 2017-02-14
Flip image markers on iOS

karimnaaji:


ios/src/TGMapViewController.mm
```diff
@@ -361,6 +361,8 @@ - (BOOL)markerSetImage:(TGMapMarkerId)identifier image:(UIImage *)image
 
     CGColorSpaceRef colorSpace = CGImageGetColorSpace(cgImage);
     CGContextRef cgContext = CGBitmapContextCreate(bitmap.data(), w, h, 8, w * 4, colorSpace, kCGImageAlphaPremultipliedLast);
+    CGAffineTransform affineTransform = CGAffineTransformMake(1, 0, 0, -1, 0, h);
```
tallytalwar: just a small nitpick, probably make sense to rename this as flipTransform or flipAffineTransform.

Issue #1204, opened 2017-01-04
'make ios' should not require xcpretty


matteblair:
As shown in https://github.com/tangrams/tangram-es/issues/1200, `make ios` and related tasks will fail in obscure ways if the `xcpretty` Ruby gem is not installed. 

Our `Makefile` tries to optionally use `xcpretty` for Xcode output formatting:
```
XCPRETTY = `command -v xcpretty || echo 'xargs echo'`
...
ios: ${IOS_BUILD_DIR}/${IOS_XCODE_PROJ}
	xcodebuild -target ${IOS_TARGET} ARCHS='i386 x86_64' \
		ONLY_ACTIVE_ARCH=NO \
		-sdk iphonesimulator \
		-project ${IOS_BUILD_DIR}/${IOS_XCODE_PROJ} \
		-configuration ${CONFIG} | ${XCPRETTY}
```
https://github.com/tangrams/tangram-es/blob/master/Makefile#L58
https://github.com/tangrams/tangram-es/blob/master/Makefile#L237-L242

matteblair:
This came up in the past and was resolved by simply not trying to format the Xcode output. I added the optional formatting again recently because the raw Xcode output overruns the logs in CircleCI.

PR #1300, opened 2017-02-16
Do not require xcpretty for ios builds

karimnaaji:
Fixes https://github.com/tangrams/tangram-es/issues/1204

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -57,8 +57,7 @@ OSX_XCODE_PROJ = tangram.xcodeproj
 IOS_XCODE_PROJ = tangram.xcodeproj
 IOS_FRAMEWORK_XCODE_PROJ = tangram.xcodeproj
 
-XCPRETTY = `command -v xcpretty`
-JAZZY = $(shell command -v jazzys 2> /dev/null)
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -222,9 +221,16 @@ osx: ${OSX_BUILD_DIR}/Makefile
 
 ${OSX_BUILD_DIR}/Makefile: cmake-osx
 
-xcode: ${OSX_XCODE_BUILD_DIR}/${OSX_XCODE_PROJ}
+${OSX_BUILD} = \
```
matteblair: Should just be `OSX_BUILD`, no braces

ios/src/TangramMap/TGMapViewController.h
```diff
@@ -420,6 +420,8 @@ NS_ASSUME_NONNULL_BEGIN
  to get more styling informations.
 
  @param identifier the marker identifier created with `-markerAdd`
+ @param styling the styling string in YAML syntax to set to the marker
```
karimnaaji: @tallytalwar updated the missing params you mentioned in the merged PR here.

PR #1301, opened 2017-02-16
Use built-in SQLite3 on iOS instead of building our own

matteblair:
Should reduce the framework size a little.

### DIFF
--- a/core/deps/CMakeLists.txt
+++ b/core/deps/CMakeLists.txt
@@ -84,5 +84,9 @@ add_subdirectory(${CMAKE_CURRENT_SOURCE_DIR}/alfons)
 ###############
 set(SQLITECPP_RUN_CPPLINT OFF CACHE BOOL "")
 set(SQLITECPP_RUN_CPPCHECK OFF CACHE BOOL "")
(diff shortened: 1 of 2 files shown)


core/deps/CMakeLists.txt
```diff
@@ -84,5 +84,9 @@ add_subdirectory(${CMAKE_CURRENT_SOURCE_DIR}/alfons)
 ###############
 set(SQLITECPP_RUN_CPPLINT OFF CACHE BOOL "")
 set(SQLITECPP_RUN_CPPCHECK OFF CACHE BOOL "")
+set(SQLITE_ENABLE_COLUMN_METADATA OFF CACHE BOOL "")
 
 add_subdirectory(SQLiteCpp)
+
+# Extensions aren't needed for MBTiles and aren't available in older versions of sqlite3.
+target_compile_definitions(SQLiteCpp PRIVATE SQLITE_OMIT_LOAD_EXTENSION)
```
tallytalwar: We can probably omit more here, but I could not figure out what might be best set here from giving a quick look at https://www.sqlite.org/compile.html.

Probably can also use `SQLITE_OMIT_DEPRECATED`

PR #1302, opened 2017-02-17
Create UrlClient class and use it in Linux platform

matteblair:
- Simpler implementation of async curl
 - Allows cancelling requests during transfer
 - Allows configurable number of worker threads
 - Can be shared by all platforms where libcurl is available


matteblair:
@hjanetzek this is similar to the UrlWorker implementation for Tizen that I believe you wrote - can you give this a look?

hjanetzek:
@matteblair sure, reviewing. Did you have a 2 space indentation day?

matteblair:
@hjanetzek Ah oops, I think my editor at home defaults to 2 spaces haha. I'll re-indent when I rebase.

PR #1304, opened 2017-02-20
Update hb-icu-ft

hjanetzek:


hjanetzek:
@karimnaaji ok. it's updated to hb-icu-ft master

karimnaaji:
👍 merging after CI.

PR #1306, opened 2017-02-21
Update geojson-vt

hjanetzek:
I've included geometry.hpp and gejson.hpp as git subtrees in our geojson-vt branch to avoid further submodules (and not having to fork them to include some tweaks for building with gcc 4.9)

### DIFF
--- a/core/src/data/clientGeoJsonSource.cpp
+++ b/core/src/data/clientGeoJsonSource.cpp
@@ -10,6 +10,11 @@
 
 #include "mapbox/geojsonvt.hpp"
 
+// RapidJson parser
+#include "mapbox/geojson.hpp"
+#include <mapbox/geojson_impl.hpp>
+
+
 #include <regex>
 
 namespace Tangram {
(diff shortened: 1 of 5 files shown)


matteblair:
Nice! Would it be alright if we cherry-picked the rapidjson update onto master, just to make this change set more manageable?

core/src/data/clientGeoJsonSource.cpp
```diff
@@ -1,44 +1,54 @@
...
+#include <regex>
 
 namespace Tangram {
 
-const double extent = 4096;
 const uint32_t indexMaxPoints = 100000;
-double tolerance = 1E-8;
+const double tolerance = 1E-8;
+using namespace mapbox;
+using vt_feature = mapbox::geojsonvt::detail::vt_feature;
```
tallytalwar: nitpic, no need to mention `mapbox::` when `using` is mentioned above.

tallytalwar:
Rebase, and ready to go.

PR #1308, opened 2017-02-21
Add Javadoc for com.mapzen.tangram.goemetry

matteblair:
The 'geometry' package is now used in the public interface, so the Javadoc should be treated accordingly.


karimnaaji:
Needs rebase then we can merge this.

PR #1311, opened 2017-02-22
Remove GLM submodule

matteblair:
We don't make changes to GLM, so there's not much value to maintaining it as a submodule -  copying in the files for the latest release is easier.

This also removes the GLM docs, which are over 10MB by themselves!


hjanetzek:
Why not use git subtree to ease updating?

lygstate:
He may means docs are hudge? but still prefre subtree.

hjanetzek:
@lygstate yeah we also considered that - still undecided

tallytalwar:
Looks like circle just broke on this. Will see if creating a new PR with the same commits fixes it.

Issue #692, opened 2016-04-20
Add CONTRIBUTING.md


tallytalwar:
Split some of the tangram-es description from Readme and move to contributing.md.

tallytalwar:
Could be as simple as the one in mapzen/android project: https://github.com/mapzen/android/blob/add-contributing/CONTRIBUTING.md

matteblair:
Done :)

PR #1312, opened 2017-02-23
Split README up by platform

matteblair:
The main README file was getting unwieldy, this should make it easier to use. 

I mostly just split up the existing markdown for each platform, but in some cases I added some information that was notably missing - so please do take a look at the new READMEs.


platforms/android/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+To build for Android you'll need [Android Studio](https://developer.android.com/studio/index.html) version 2.2 or newer on Mac OS X, Ubuntu, or Windows 10. Using the Android Studio SDK Manager, install or update the 'CMake', 'LLDB', and 'NDK' packages from the 'SDK Tools' tab.
+
+## Build ##
+
+After installing dependencies in Android Studio, you can execute Android builds from either the command line or the Android Studio interface.
+
+### Command Line ###
+
+The Android project is executed with Gradle commands from the 'platforms/android' folder. To build the demo application for the ARMv7 architecture (covers most Android devices), run:
```
tallytalwar: Probably make sense to mention about ndk and sdk path settings in relevant file if one is straight away trying from command line.

platforms/ios/README.md
```diff
@@ -0,0 +1,63 @@
...
+You can optionally install xcpretty to produce much prettier and more legible output during the build process:
+
+```bash
+gem install xcpretty
+```
+
+## Build ##
+
+Building the iOS demo application requires Xcode 8.0 or newer. First, run:
+
```
tallytalwar: Probably make sense to be explicit over here and mention to run these commands from the root of the project.

platforms/ios/README.md
```diff
@@ -0,0 +1,63 @@
...
+This project uses CMake (minimum version 3.0), you can download it [here](http://www.cmake.org/download/) or use your favorite installation package tool like [homebrew](http://brew.sh/).
+
+```bash
+brew install cmake
+```
+
+Make sure to update git submodules before you build:
+
+```bash
+git submodule update --init
```
tallytalwar: We still need recursive? (I might have missed some pr in last couple of days from Hannes or karim which removes recursive submodules)

platforms/osx/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+To build a runnable OS X application bundle, run:
+
+```bash
+make osx
+```
+And open the application with:
+
+```bash
+open build/osx/bin/tangram.app
```
tallytalwar: Can we also mention about the use of `-f` command line argument to specify a scene file (remote)?

README.md
```diff
@@ -16,201 +16,18 @@ This repository contains both the core rendering library and sample applications
...
-
-Android Studio supports debugging both the Java and C++ parts of tangram-es on a connected device or emulator. Choose one of the 'debug' build variants, set your desired breakpoints, and press the 'Debug' button (^D).
+Currently we are targeting five platforms:
 
-### Raspberry Pi ###
-To build on Rasberry Pi you will need a C++ toolchain with support for C++14. GCC 4.9 (or higher) is known to work (refer [here](https://community.thinger.io/t/starting-with-the-raspberry-pi/36) for instructions on getting GCC 4.9).
+ - [Android](platforms/android)
+ - [iOS](platforms/ios)
+ - [Mac OS X](platforms/osx)
+ - [Ubuntu Linux](platforms/linux)
```
tallytalwar: Any Linux flavor will work (arch fedora...), the only difference in the instructions be on how dependent libraries like curl etc are installed on different flavors.

platforms/rpi/README.md
```diff
@@ -0,0 +1,58 @@
...
+Tangram will be rendered directly to the screen without a window manager. To show a mouse cursor, run the application with the `-m`:
+
+```
+cd build/rpi/bin
+./tangram -m
+```
+
+You can also move the map with `w`, `a`, `s`, and `z`, zoom in and out with `-` and `=`, and quit with `q`.
+
+To build in `RELEASE` or `DEBUG` mode run the following:
```
tallytalwar: Probably move this to the common README?

README.md
```diff
@@ -16,201 +16,18 @@ This repository contains both the core rendering library and sample applications
...
+ - [Raspberry Pi](platforms/rpi)
 
-First, install CMake and libcurl:
+## Code Style ##
```
karimnaaji: We could add that in a `Contributing.md` as mentioned in https://github.com/tangrams/tangram-es/issues/692.

tallytalwar:
I accidently hit `x` yesterday, meant to approve 👍.

platforms/android/README.md
matteblair: It's really inconvenient to do the first-time build without running Android studio, you can see in the Travis script how many steps it takes (https://github.com/tangrams/tangram-es/blob/master/scripts/travis/before_install.sh) so I don't want to recommend that process to users.

platforms/ios/README.md
matteblair: Not anymore! Now that we use a subtree for harfbuzz there are no required recursive submodules.

platforms/ios/README.md
matteblair: Sounds good.

platforms/osx/README.md
matteblair: Good idea, adding it.

README.md
matteblair: Probably true, and I'm glad that it works on those distributions, but I don't want to imply that we'll provide help or fix bugs on distributions other than Ubuntu (it wouldn't be a good use of our time).

README.md
matteblair: Yeah, I can do that next :+1:

platforms/rpi/README.md
matteblair: I've added notes about choosing build type to the other relevant platforms.

PR #1315, opened 2017-02-23
Add Marker helper class to iOS SDK

karimnaaji:
- Add `TGMarker` class and update interface to not use marker identifiers and simplify interface.
- Add missing `drawOrder` from interface.
- Add private/internal interface with framework only visibility for `TGMapViewController`.
- Add `markers` accessor to retrieve all marker objects added to a `TGMapViewController`.

### DIFF
--- a/platforms/ios/src/TangramMap/TGGeoPolygon.h
+++ b/platforms/ios/src/TangramMap/TGGeoPolygon.h
@@ -12,7 +12,7 @@
 NS_ASSUME_NONNULL_BEGIN
 
 /**
- Helper class to contain a polygon geometry for use in `-[TGMapViewController markerSetPolygon:polygon:]`.
+ Helper class to contain a polygon geometry for use in `-[TGMarker polygon]`.
 
  The polygon winding order and internal polygons must be set according to the <a href="http://geojson.org/geojson-spec.html#polygon">
  GeoJSON specification</a>.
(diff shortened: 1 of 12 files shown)


platforms/ios/src/TangramMap/TGGeoPolygon.h
```diff
@@ -83,6 +83,8 @@ NS_ASSUME_NONNULL_BEGIN
  */
 - (void)removeAll;
 
+- (instancetype)userData;
```
karimnaaji: This is not used, will remove.

platforms/ios/src/TangramMap/TGMapViewController+Internal.h
```diff
@@ -0,0 +1,28 @@
...
+
+NS_ASSUME_NONNULL_BEGIN
+
+- (void)addMarker:(TGMarker *)marker withIdentifier:(Tangram::MarkerID)identifier;
+
+- (void)removeMarker:(Tangram::MarkerID)identifier;
+
+NS_ASSUME_NONNULL_END
+
+- (TGMarker *)getMarker:(Tangram::MarkerID)identifier;
```
karimnaaji: Might be a leftover, will remove.

tallytalwar:
@msmollin wanna give this a quick look to check for ios specific things I might be missing. Thanks

tallytalwar:
Circle is actually good, but not updating the status on github. Merging.

PR #1318, opened 2017-02-24
Combined desktop demo

matteblair:
- Moves `Platform` and related types into the `Tangram` namespace (we should avoid populating the global namespace in our API)
 - Merges the GLFW demo application code into a single set of shared functions that runs on both OS X and Linux (and eventually Windows, hopefully). This could be further refactored into a class, for now it still uses global state.


lygstate:
Yeap, that's would be a lost easier for Windows:)

lygstate:
+1 for me

platforms/common/glfwApp.h
```diff
@@ -0,0 +1,27 @@
...
+
+namespace GlfwApp {
+
+    void create(std::shared_ptr<Platform> platform, std::string sceneFile, int width, int height);
+    void destroy();
+
+    void run();
+    void stop(int);
+
+    void mouseButtonCallback(GLFWwindow* window, int button, int action, int mods);
```
karimnaaji: We could remove those from the header definition, so we would include `<GLFW/glfw3.h>` from implementation file only. Not that harmful though.

platforms/linux/src/main.cpp
```diff
@@ -1,423 +1,38 @@
...
+    auto platform = std::make_shared<LinuxPlatform>();
 
+    std::string sceneFile = "scene.yaml";
+    // Load file from command line, if given.
     int argi = 0;
     while (++argi < argc) {
         if (strcmp(argv[argi - 1], "-f") == 0) {
             sceneFile = std::string(argv[argi]);
-            LOG("File from command line: %s", argv[argi]);
+            logMsg("File from command line: %s\n", argv[argi]);
```
karimnaaji: We can remove the `log.h` inclusion if we switch to this one.

platforms/common/glfwApp.h
matteblair: :+1:

platforms/linux/src/main.cpp
matteblair: :+1:

PR #1322, opened 2017-02-27
Rpi platform

patriciogonzalezvivo:
Cleaning some leftovers and fixing Raspberry Pi compilation

### DIFF
--- a/platforms/rpi/src/main.cpp
+++ b/platforms/rpi/src/main.cpp
@@ -58,7 +58,7 @@ int main(int argc, char **argv){
     while (bUpdate) {
         updateGL();
 
-        platform->processNetworkQueue();
+        // platform->processNetworkQueue();
         if (getRenderRequest() || platform->isContinuousRendering() ) {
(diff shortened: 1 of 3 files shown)


platforms/rpi/src/main.cpp
```diff
@@ -40,7 +40,12 @@ void newFrame();
 
 int main(int argc, char **argv){
 
-    platform = std::make_shared<LinuxPlatform>();
+    UrlClient::Environment urlClientEnvironment;
+
+    UrlClient::Options urlClientOptions;
+    urlClientOptions.numberOfThreads = 4;
```
tallytalwar: Will 4 threads be too much for a pi? Fine if you are fine :D

patriciogonzalezvivo: hehehe... it can take it. I'm assuming a RPi2 or higher (that have 4 cores)

tallytalwar: cool cool :D

PR #1328, opened 2017-02-28
Compilation warnings

Guillaume227:
Fixed compilation warnings (seen in Android Studio) due to:
- returning const bool
- unused named function params

### DIFF
--- a/core/include/tangram/data/properties.h
+++ b/core/include/tangram/data/properties.h
@@ -40,7 +40,7 @@ struct Properties {
 
     std::string getAsString(const std::string& key) const;
 
-    const bool getAsString(const std::string& key, std::string& value) const;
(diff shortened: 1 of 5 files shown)


tallytalwar:
Just to confirm this is also on a windows machine, right?

Guillaume227:
Yes it is. Windows 10, Android studio 2.3 RC1

hjanetzek:
The change for ignored-qualifiers warnings is certainly good. 

For unused parameters I would prefer `-Wno-unused-parameter` instead. We should discuss it in the team.

matteblair:
Changes partly incorporated into https://github.com/tangrams/tangram-es/pull/1335 and merged. 

Still haven't addressed the unused parameter warning.

PR #1333, opened 2017-03-01
iOS cleanup public interface

karimnaaji:
- Always use strongly typed templates to unwrap types more easily in Swift.

### DIFF
--- a/platforms/ios/src/TangramMap/TGMapData.h
+++ b/platforms/ios/src/TangramMap/TGMapData.h
@@ -14,7 +14,9 @@
 /**
  Dictionary of feature properties keyed by their property name
  */
-typedef NSDictionary<NSString *, NSString *> FeatureProperties;
+typedef NSDictionary<NSString *, NSString *> TGFeatureProperties;
+
+@class TGMapViewController;
 
 /**
  A `TGMapData` is a convenience class to display point, polygons or polylines from a dynamic data layer.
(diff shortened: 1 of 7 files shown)


platforms/ios/src/TangramMap/TGMapData.h
```diff
@@ -14,7 +14,9 @@
 /**
  Dictionary of feature properties keyed by their property name
  */
-typedef NSDictionary<NSString *, NSString *> FeatureProperties;
+typedef NSDictionary<NSString *, NSString *> TGFeatureProperties;
+
+@class TGMapViewController;
```
tallytalwar: How was it working before without the forward declaration?

msmollin: Clang is smart enough to work around some of the forward declaration annoyances which can then result in things occasionally not working either. I think the rule is: If wherever `TGMapData` gets used for example also was importing `TGMapViewController.h` then clang will resolve the dependency that way.

karimnaaji: We didn't need that before. Now `TGLabelPickResult.h` uses `TGMapData.h` for the typedef  `TGFeatureProperties`, but `TGLabelPickResult.h` is also included by `TGMapViewController.h` which, without the forward declaration, would include `TGMapData.h` including `TGMapViewController.h` including `TGLabelPickResult.h` including `TGMapData.h`... Classic inclusion cycle annoyance.

PR #1335, opened 2017-03-02
More compiler warnings

hjanetzek:
Adding gcc warnings. I'll see if the warning flag work on clang


tallytalwar:
@hjanetzek added couple of commits to add same warning for android and osx (both use clang)

PR #1340, opened 2017-03-03
View complete only when all visible tiles are generated

### LINKED ISSUE #1324: `Tangram::update()` returns `true` before the scene finish loading
I'm using last version of Tangram-Es Map at paparazzi, here https://github.com/tangrams/paparazzi/blob/rearch/worker/src/paparazzi.cpp#L152 and returns true way before finishing... producing incomplete scenes screenshots : /

tallytalwar:
Fixes #1324

### DIFF
--- a/core/src/tile/tileManager.cpp
+++ b/core/src/tile/tileManager.cpp
@@ -121,6 +121,7 @@ void TileManager::updateTileSets(const ViewState& _view,
                                  const std::set<TileID>& _visibleTiles) {
     m_tiles.clear();
     m_tilesInProgress = 0;
+    m_tilesInGeneration = 0;
     m_tileSetChanged = false;
 
     for (auto& tileSet : m_tileSets) {
@@ -237,6 +238,9 @@ void TileManager::updateTileSet(TileSet& _tileSet, const ViewState& _view,
                 entry.task = _tileSet.source->createTask(visTileId);
                 enqueueTask(_tileSet, visTileId, _view);
                 m_tilesInProgress++;
+            } else if (!entry.needsLoading() && !entry.isReady()) {
+                // tile not generated yet!
+                m_tilesInGeneration++;
             }
 
             if (newTiles && entry.isLoading()) {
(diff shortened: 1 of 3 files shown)


tallytalwar:
cc. @patriciogonzalezvivo fyi.

patriciogonzalezvivo:
confirmed... this work on Paparazzi

core/src/tile/tileManager.cpp
```diff
@@ -237,6 +238,9 @@ void TileManager::updateTileSet(TileSet& _tileSet, const ViewState& _view,
                 entry.task = _tileSet.source->createTask(visTileId);
                 enqueueTask(_tileSet, visTileId, _view);
                 m_tilesInProgress++;
+            } else if (!entry.needsLoading() && !entry.isReady()) {
+                // tile not generated yet!
+                m_tilesInGeneration++;
             }
```
hjanetzek: Do we need to differentiate `m_tilesInProgress` and `m_tilesInGeneration`? Otherwise I would count this case as 'in progress'

tallytalwar: I wanted to explicitly differentiate it. Makes understanding tilemanager code better.

hjanetzek: Ok then `m_tilesInProgress` should be renamed to `m_tilesAwaitDownload` I think

tallytalwar: That's a better name will update.

hjanetzek: I'm not yet convinced that we should have both. How about `m_tilesInProgress` counter for all tiles that are `!entry.isReady() || (entry.tile->sourceGeneration() < generation)` ?

tallytalwar: ^ Basically merge `m_tilesGenerating` in `m_tilesInProgress`. I am fine with that, can add some code comments along with it to mention what we mean by `InProgress`.

tallytalwar: Updated! Addressed!

core/src/tile/tileManager.h
```diff
@@ -7,6 +7,7 @@
 #include "tile/tileTask.h"
 #include "tile/tileWorker.h"
 #include "util/fastmap.h"
+#include "log.h"
```
hjanetzek: small cleanup. Then LGTM

tallytalwar: addressed!

Issue #1313, opened 2017-02-23
Directional arrows


sheikh-iffi:
Hi guys

I am facing some issues regarding directional arrows. 

**1:** At some places, arrows are drawn at wrong places e.g. in the below image arrow is not drawn on road
![screenshot from 2017-02-23 12-46-28](https://cloud.githubusercontent.com/assets/12157369/23256760/b695fdaa-f9e2-11e6-9305-cb7f77f4ab20.png)

**2:** At some places arrows direction is wrong. e.g. in the below image you can see E Houston St to have arrows in both direction
![screenshot from 2017-02-23 12-55-15](https://cloud.githubusercontent.com/assets/12157369/23256858/0fd24536-f9e3-11e6-90b1-acdcd35ca029.png)

**3:** There is no consistency between placing distance. e.g. in the below image one road is filled with arrows and other with little or no arrows.
![screenshot from 2017-02-23 16-06-31](https://cloud.githubusercontent.com/assets/12157369/23256951/6275dbc2-f9e3-11e6-94f1-c03f897be809.png)

Thanks

tallytalwar:
Can you share the scene file with us? And are you using our vector tile source?

Arrows are drawn based on the underlying geometry. My initial hunch is that the underlying polyline data could be incorrect. But can not be sure without looking at the scene file and referring the underlying vector tile data.

sheikh-iffi:
I am using your scene file [scene.yaml](https://github.com/tangrams/tangram-es/blob/master/scenes/scene.yaml) without any changes.

tallytalwar:
K will give this a look and update. Thanks

tallytalwar:
So looked some more into this:

1. **Arrow not on road**
  It looks like we do have some issues with arrow placement and overzooming.
  At zoom < 18:
  ![screen shot 2017-03-02 at 10 23 57 am](https://cloud.githubusercontent.com/assets/360641/23513763/30d9bab6-ff33-11e6-8a61-e2d0fc92823c.png)
  At zoom > 18:
  ![screen shot 2017-03-02 at 10 24 06 am](https://cloud.githubusercontent.com/assets/360641/23513786/45e22e7a-ff33-11e6-9891-5fa196ccb504.png)
  This further needs investigation.

2. **arrows direction is wrong**
   I think this is due to underlying data. I checked on JS and it looks the same:
[...]

PR #1342, opened 2017-03-03
fix: placement_min_length_ratio handling for spaced point sprites

hjanetzek:
https://github.com/tangrams/tangram-es/issues/1313

tested with:
```
        arrows:
            filter: { oneway: yes, shield_text: false, $zoom: { min: 8 }, name: "Eastern Parkway Service Rd." }
            draw:
                icons:
                    flat: true
                    priority: 0
                    sprite: arrow
                    size: [[17, 18px], [18, 20px], [20, 32px]]
                    placement: spaced
                    #placement_spacing: [[17, 70px], [20, 175px]]
                    placement_spacing: 10px
                    angle: auto
                points:
                    priority: 0
                    collide: false
                    color: red
                    size: 5px
                    placement: vertex
```

### DIFF
--- a/core/src/util/lineSampler.h
+++ b/core/src/util/lineSampler.h
@@ -125,17 +125,6 @@ struct LineSampler {
                 // length from cur to next point
                 float segmentLength = next.z - curr.z;
 
-                if (segmentLength <= m_minSampleLength) {
-                    if (m_curPoint >= m_points.size() - 2) {
-                        _point = glm::vec2(next);
-                        _rotation = (glm::vec2(next) - glm::vec2(curr)) / segmentLength;
-                        m_curAdvance = sumLength();
-                        return false;
-                    }
-                    m_curPoint += 1;
-                    continue;
-                }
-
                 if (length <= segmentLength) {
                     float f = length / segmentLength;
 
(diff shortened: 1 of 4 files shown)


core/src/util/lineSampler.h
```diff
@@ -202,19 +180,25 @@ struct LineSampler {
...
+    }
+    float lengthToPrevSegment() {
+        if (m_curPoint >= m_points.size()) { return 0; }
+        return m_curAdvance - m_points[m_curPoint].z;
     }
 
     LineSampler(Points _points) : m_points(_points) { }
     LineSampler() { }
+
+    auto& points() { return m_points; }
```
tallytalwar: Doesn't seemed to be used in the current implementation, can we remove this?

hjanetzek:
The dots are placed on each vertex to indicate line segment start and end. Looks right to me :)

core/src/util/lineSampler.h
hjanetzek: hmm yeah can be removed

tallytalwar:
Ahaa.. thats right. I was reading it incorrectly. Thanks for the clarification.

Issue #1061, opened 2016-11-03
Add completion callback for applying scene updates


msmollin:
It would be nice to be able to get a callback when scene updates complete (after calling `applySceneUpdate`) in the event you wish to update/add/remove certain UI elements based upon if the map is changing or not.

tallytalwar:
This is not `ios` specific I guess. Should we remove that tag?

msmollin:
Sure! I just noticed it on iOS so I didn't know if it was specific or not :)

PR #1347, opened 2017-03-07
Scene update status

karimnaaji:
- Extend our scene update API to return on scene update complete with errors information: 
   - path not found
   - yaml syntax error on path
   - yaml syntax error on update value

Fixes https://github.com/tangrams/tangram-es/issues/1061


platforms/android/tangram/src/main/cpp/platform_android.cpp
```diff
@@ -350,6 +373,30 @@ void setCurrentThreadPriority(int priority) {
...
+        jstring jUpdateStatusPath = jniEnv->NewStringUTF(updateStatus.update.path.c_str());
+        jstring jUpdateStatusValue = jniEnv->NewStringUTF(updateStatus.update.value.c_str());
+        jint jError = (jint)updateStatus.error;
+        jobject jUpdateStatus = jniEnv->NewObject(sceneUpdateStatusClass, sceneUpdateStatusInitMID,
+                                                  jUpdateStatusPath, jUpdateStatusValue, jError);
+
+        jniEnv->CallBooleanMethod(arrayList, arrayListAddMID, jUpdateStatus);
+    }
+
+    jniEnv->CallVoidMethod(updateStatusCallbackRef, onSceneUpdateErrorMID, arrayList);
```
tallytalwar: You could have created a method on which takes in 2 strings and a int and created the array list on java itself... but maybe what you currently have has less handshakes with jni interface.. Just a thought.

platforms/ios/demo/src/MapViewController.m
```diff
@@ -18,6 +18,15 @@ - (void)addAlert:(NSString *)message withTitle:(NSString *)title;
 
 @implementation MapViewControllerDelegate
 
+- (void)mapView:(TGMapViewController *)mapView didCompleteSceneUpdatesWithErrors:(NSArray<TGSceneUpdateStatus *> *)updateStatuses
+{
+    for (TGSceneUpdateStatus* updateStatus in updateStatuses) {
+        NSLog(@"Scene update error for update with path %@ and value %@",
```
tallytalwar: As I mentioned for android interface, instead of logging, does it make sense to send a NSError here? cc. @msmollin

platforms/ios/src/TangramMap/TGMapViewController.h
```diff
@@ -260,6 +261,15 @@ NS_ASSUME_NONNULL_END
  @param mapView a pointer to the map view
  */
 - (void)mapViewDidCompleteLoading:(nonnull TGMapViewController *)mapView;
+
+/**
+ Called whenever scene updates have been applied to the scene file.
+ The list of scene update statuses will be emtpy if all updates have been applied successfully.
```
tallytalwar: nitpick .. `empty`.

platforms/ios/demo/src/MapViewController.m
msmollin: It would be neat to know what failed and what didn't, yep. And NSError would be the way to go.

platforms/ios/src/TangramMap/TGMapViewController.h
karimnaaji: will cleanup

platforms/ios/demo/src/MapViewController.m
karimnaaji: >As I mentioned for android interface, instead of logging

Agreed on NSError, but this is the demo app where it is fine to log to show example usage.

>It would be neat to know what failed and what didn't, yep. 

Please look at the interface in the public framework and usage of `TGSceneUpdateStatus`, here it is the demo app. Also, please mention how `NSError` would be best used for your needs.

core/include/tangram/tangram.h
```diff
@@ -63,6 +63,21 @@ struct SceneUpdate {
     SceneUpdate(std::string p, std::string v) : path(p), value(v) {}
 };
 
+enum SceneUpdateError {
```
karimnaaji: Yes sounds good, I can change that for a `TangramError` which where we could add more errors incoming from tangram API use.

platforms/android/demo/src/main/java/com/mapzen/tangram/android/MainActivity.java
```diff
@@ -237,5 +241,14 @@ public void run() {
             }
         });
     }
+
+    @Override
+    public void onSceneUpdateError(List<SceneUpdateStatus> updateStatuses) {
+        for (SceneUpdateStatus status : updateStatuses) {
+            Log.d("Tangram", "Scene update errors "
```
karimnaaji: If we have something on those lines:
```java
try {
    map.loadScene("scene.yaml", sceneUpdates);
} catch(SceneUpdateException) {

} 
```
Around anything that applies the scene updates. 

Some problem with that pattern is that some of our API calls that applies the scene updates are not synchronous, so if we catch an error for a scene update and throw an exception, the `try` block would be already out of scope and the exception thrown on a different context (basically our `update` loop).

Another solution, is that we could surround our `update` loop by a `try` `catch` for those exceptions.

core/src/scene/sceneLoader.cpp
```diff
@@ -74,24 +75,37 @@ bool SceneLoader::loadScene(const std::shared_ptr<Platform>& _platform, std::sha
...
         } catch (YAML::ParserException e) {
             LOGE("Parsing scene update string failed. '%s'", e.what());
+            updateStatus.push_back({update, SceneUpdateError::value_yaml_syntax_error});
         }
-        if (value) {
-            try {
-                auto node = YamlPath(update.path).get(root);
+        try {
+            // Dummy node to trigger YAML exception on YAML syntax errors
+            auto parse = YAML::Load(update.path);
```
karimnaaji: I'm not sure to see, where exactly do you mention to use it further?

tallytalwar: Ahh.. I misread `YamlPath` below to `Yaml::Load`. All good here.

platforms/ios/demo/src/MapViewController.m
tallytalwar: Yup, fine to have logs in the demo app. But my question was more on the usage of NSError :D. Will be good to have usage of the NSError from @msmollin though. Thanks

platforms/ios/src/TangramMap/TGSceneUpdateStatus.h
```diff
@@ -0,0 +1,38 @@
...
+/**
+ Holds information relative to a scene update error status
+ */
+@interface TGSceneUpdateStatus : NSObject
+
+/// A scene update holding the path and the scene update value
+@property (readonly, nonatomic) TGSceneUpdate* sceneUpdate;
+
+/// The type of error the scene update triggered when applied to the scene
+@property (readonly, nonatomic) TGSceneUpdateError error;
```
msmollin: This is a good start. Where we could go with this would be to use TGSceneUpdateError types as the error codes for NSError.

I'd imagine, from an implementing developer side, a nullable property (since NSError is an object) that you'd check to see if its nil or not, and then check the code.

So it build an NSError at bare minimum you need an Error domain (a string constant) and an error code. Then in the implementation it's as simple as constructing it and passing it here as a property instead of the enum.

msmollin: See here: https://developer.apple.com/reference/foundation/nserror/1417063-initwithdomain?language=objc

platforms/ios/demo/src/MapViewController.m
msmollin: Sorry yeah was doing the entire review still - just wanted to comment here first :) Placed a more in-depth comment on TGStatusUpdate

platforms/ios/src/TangramMap/TGMapViewController.h
```diff
@@ -260,6 +261,15 @@ NS_ASSUME_NONNULL_END
...
+
+ @param mapView a pointer to the map view
+ @param updateStatuses the list of scene update error status, empty if all updates were successfull
+ */
+- (void)mapView:(nonnull TGMapViewController *)mapView didCompleteSceneUpdatesWithErrors:(NSArray<TGSceneUpdateStatus *> *)updateStatuses;
```
msmollin: 👍 indicating internal types for collection objects makes swift happy :)

platforms/ios/src/TangramMap/TGMapViewController.mm
```diff
@@ -135,7 +136,13 @@ - (void)loadSceneFile:(NSString *)path sceneUpdates:(NSArray<TGSceneUpdate *> *)
     }
 
     self.scenePath = path;
-    self.map->loadScene([path UTF8String], false, updates);
+
+    auto updateCallbackStatus = [self](auto sceneUpdateStatuses) {
+        if (!self.mapViewDelegate || ![self.mapViewDelegate respondsToSelector:@selector(mapView:didCompleteSceneUpdatesWithErrors:)]) { return; }
+        [self.mapViewDelegate mapView:self didCompleteSceneUpdatesWithErrors:[TGSceneUpdateStatus sceneUpdateStatuses:sceneUpdateStatuses]];
```
msmollin: Hmmm now that I see this in usage, would this be better named `didCompleteSceneUpdatesWithStatus` or is this only going to return when there are errors? Just getting tripped up on the name of things since typical Apple objc syntax is to return something named an error when you say `withErrors` - not a status.

platforms/ios/src/TangramMap/TGSceneUpdateStatus.mm
```diff
@@ -0,0 +1,55 @@
...
+        self.sceneUpdate = sceneUpdate;
+        self.error = error;
+    }
+
+    return self;
+}
+
++ (NSArray<TGSceneUpdateStatus *> *)sceneUpdateStatuses:(const std::vector<Tangram::SceneUpdateStatus>&)updateStatuses
+{
+    NSMutableArray* nsUpdateStatuses = [[NSMutableArray alloc] initWithCapacity:(NSUInteger)updateStatuses.size()];
```
msmollin: Nitpick - probably want to do a check here on `updateStatuses` to make sure it's not nil or otherwise this will crash. Objective-c will let you do the message send using something like `performSelector` so a little defensive programming here might be wise.

platforms/ios/src/TangramMap/TGMapViewController.mm
```diff
@@ -135,7 +136,13 @@ - (void)loadSceneFile:(NSString *)path sceneUpdates:(NSArray<TGSceneUpdate *> *)
     }
 
     self.scenePath = path;
-    self.map->loadScene([path UTF8String], false, updates);
+
+    auto updateCallbackStatus = [self](auto sceneUpdateStatuses) {
```
msmollin: Hmmmm this looks like a closure called from C++ yeah? If so, you'll want to do the weakSelf/strongSelf dance to avoid a retain cycle with self. You can see a discussion on the topic here - https://github.com/objc-zen/objc-zen-book#retain-cycles-on-self

I tend to err on the side of caution here - there are times when you don't have to do this dance because everything is behaving and not holding onto long-term references to the block or the code using the block. However it's a safeguard that I use a lot.

platforms/ios/demo/src/MapViewController.m
karimnaaji: Sounds good, I understand more the use cases of `NSError` and how it'll be used, I'll look how this can fit nicely with this interface.

platforms/ios/src/TangramMap/TGSceneUpdateStatus.mm
karimnaaji: `updateStatuses` is a reference though, which should theoretically be initialized and not `null`, not even testable against it (unless you're doing something strange and dereference something already `null`), we're safe here.

platforms/ios/src/TangramMap/TGMapViewController.mm
karimnaaji: I see, I will look at that in details and see what would be safer on this. The capture on `self` is made by reference so this might be a good call, if we use a copy we might be able to retain it automatically.

core/include/tangram/tangram.h
```diff
@@ -61,16 +61,32 @@ struct SceneUpdate {
     std::string path;
     std::string value;
     SceneUpdate(std::string p, std::string v) : path(p), value(v) {}
+    SceneUpdate() {}
 };
 
+enum Error {
+    scene_update_path_not_found,
+    scene_update_path_yaml_syntax_error,
+    scene_update_value_yaml_syntax_error,
```
tallytalwar: does it make sense to add a no_error here?

core/include/tangram/tangram.h
```diff
@@ -61,16 +61,32 @@ struct SceneUpdate {
...
+};
+
+struct SceneUpdateError {
+    SceneUpdate update;
+    Error error;
+};
+
+using SceneUpdateErrorCallback = std::function<void(const SceneUpdateError&)>;
```
tallytalwar: nitpick. I think in other instances of platform callbacks functionality we use Cb for brevity. So maybe instead of `SceneUpdateErrorCallback` call this `SceneUpdateErrorCb`

platforms/ios/src/TangramMap/TGMapViewController.mm
```diff
@@ -253,7 +270,7 @@ - (void)pickFeatureAt:(CGPoint)screenPosition
     screenPosition.x *= self.contentScaleFactor;
     screenPosition.y *= self.contentScaleFactor;
 
-    self.map->pickFeatureAt(screenPosition.x, screenPosition.y, [screenPosition, self](const Tangram::FeaturePickResult* featureResult) {
+    self.map->pickFeatureAt(screenPosition.x, screenPosition.y, [=](const Tangram::FeaturePickResult* featureResult) {
```
tallytalwar: Why switching the closure to copy all? Will a reference to `self` not suffice? Or `self` is asyncly updated?

karimnaaji: This does not copy _all_, only what's referenced from the lambda body, so `self`, to increment its counter in ARC context. 

Refer https://github.com/tangrams/tangram-es/pull/1347#discussion_r107471131.

core/include/tangram/tangram.h
karimnaaji: I usually prefer explicit names, `Cb` doesn't seem that clear to me at first sight, especially in the public interface.
But I can switch back if you're strong on it.

core/include/tangram/tangram.h
karimnaaji: For now `no_error` wouldn't be used, but we can add it in the future if there is anything that needs to return a `no_error` signal.

tallytalwar: Yup, i meant for future only. Works for me.

core/include/tangram/tangram.h
tallytalwar: No I am not too strong on it. Works for me.

Also just confirmed, we are using `Cb` in our internal code, but all public facing code is using `Callback`. So we are good here.
Thanks

Issue #1309, opened 2017-02-22
Android cache


ivangorelow:
How can I use cache on android?

matteblair:
@ivangorelow Hello again! With the Android library you can use HTTP caching by creating an `HttpHandler` object with a file cache and then giving that handler to the `MapController` with `setHttpHandler`. An example of this is in our demo application: https://github.com/tangrams/tangram-es/blob/master/platforms/android/demo/src/main/java/com/mapzen/tangram/android/MainActivity.java#L84-L114

Does this answer your question?

ivangorelow:
and is it possible to load map localy?

tallytalwar:
Yes we added support for `mbtiles` in #1015. You will need to have the tile set in a [`mbtiles`](https://github.com/mapbox/mbtiles-spec) format.

Setting the scene file using the following syntax, you should be able to load the map using your `mbtiles` file.

```
sources:
    _savedTiles: #This is an example MBTiles data source.
        type: TopoJSON #We expect TopoJSON data from the source.
        url: /path/to/my/file.mbtiles #Only absolute file paths supported for now.
        max_zoom: 16 #Other optional parameters are applied as usual.
```

Let us know if this works for you.

ivangorelow:
no, it doesn't workt

I have error `ERROR mbtilesDataSource.cpp:246: Unable to open SQLite database:`

tallytalwar:
Thanks @ivangorelow 

Can you share the MBTiles file you are trying to open?

Thanks

ivangorelow:
@tallytalwar  https://openmaptiles.os.zhdk.cloud.switch.ch/v3.3/extracts/moscow_russia.mbtiles

tallytalwar:
Ok looks like we need to do resolve android asset path for sqlite database construction, which is not being done at present.

@ivangorelow what path are you putting the mbtiles file in? Are you bundling with the asset?

JinYongwHwa:
I try offline tangram map by mbtiles

this is my scene file
```
sources:
    pinonmap:
        type: TopoJSON #GeoJSON, MVT, TopoJSON
        url: file:///storage/emulated/0/test/seoul.mbtiles
        max_zoom: 16
```

this is my test mbtiles file (1-10 zoom level seoul )
[seoul.zip](https://github.com/tangrams/tangram-es/files/815984/seoul.zip)

i have error too

[...]

PR #1348, opened 2017-03-08
Mbtiles platform asset paths

tallytalwar:
Addressed #1309 

Android platform:
- if the mbtiles path specified in the scene file is an apk asset path, then this file is copied at runtime to app's externalCacheDir (no special permission is required for this).
- if the mbtiles specified in the scene file is an absolute path on the device, `READ_EXTERNAL_STORAGE` permission is required both in the `AndroidManifest.xml` as well as during runtime using the following (post android M):
   ```
  if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
           requestPermissions(new String[]{Manifest.permission.READ_EXTERNAL_STORAGE}, 1);
  }
   ```
   NOTE: from @ecgreb 
   > The permission READ_EXTERNAL_STORAGE should not be required if the mbtiles bundle is stored in the app's designated files folder accessed via Context.getExternalFilesDir(...).

iOS platform:
- if the mbtiles path specified in the scene file is app bundle path, then this file's path is resolved with the app's Bundle path which is then used to open the sqlite db.
- absolute path to an mbtiles works out of the box.

All other platforms:
The path to the mbtiles are resolved (relative to base scene yaml, or absolute path) automatically and works out of the box.

### DIFF
--- a/core/include/tangram/platform.h
+++ b/core/include/tangram/platform.h
@@ -53,7 +53,8 @@ public:
 
     // returns the path of an asset given the name of the asset
     // ios: returns absolute path of the asset in the asset bundle
-    // android: returns a tmp path where the asset file is extracted from the apk
+    // android: returns a tmp path where the asset file is extracted from the apk, given the path of
+    // the asset, OR absolute path of an external resource
     // other: returns the absolute path of the asset (should be by default)
     virtual std::string getAssetPath(const Tangram::Url& _path) const;
 
--- a/core/include/tangram/platform.h
+++ b/core/include/tangram/platform.h
@@ -14,9 +14,7 @@ using UrlCallback = std::function<void(std::vector<char>&&)>;
 
 using FontSourceLoader = std::function<std::vector<char>()>;
 
-namespace Tangram {
-    class Url;
-}
+class Url;
 
 struct FontSourceHandle {
     FontSourceHandle(std::string _path) : path(_path) {}
--- a/core/include/tangram/platform.h
+++ b/core/include/tangram/platform.h
@@ -14,6 +14,10 @@ using UrlCallback = std::function<void(std::vector<char>&&)>;
 
 using FontSourceLoader = std::function<std::vector<char>()>;
 
+namespace Tangram {
+    class Url;
+}
+
 struct FontSourceHandle {
     FontSourceHandle(std::string _path) : path(_path) {}
     FontSourceHandle(FontSourceLoader _loader) : load(_loader) {}
@@ -46,6 +50,12 @@ public:
     virtual void setContinuousRendering(bool _isContinuous);
 
     virtual bool isContinuousRendering() const;
+
+    // returns the path of an asset given the name of the asset
+    // ios: returns absolute path of the asset in the asset bundle
+    // android: returns a tmp path where the asset file is extracted from the apk
+    // other: returns the absolute path of the asset (should be by default)
+    virtual std::string getAssetPath(const Tangram::Url& _path) const;
 
     // Read a file as a string
     // Opens the file at the _path and returns a string with its contents.
--- a/core/src/data/mbtilesDataSource.cpp
+++ b/core/src/data/mbtilesDataSource.cpp
@@ -241,6 +241,11 @@ void MBTilesDataSource::openMBTiles(const std::shared_ptr<Platform>& _platform) 
         }
 
         auto mbTilesPath = _platform->getAssetPath(Url(m_path));
+        if (mbTilesPath.empty()) {
+            LOGW("Empty mbtiles asset path set.");
+            m_db.reset();
+            return;
+        }
         m_db = std::make_unique<SQLite::Database>(mbTilesPath, mode);
         LOG("SQLite database opened: %s", m_path.c_str());
 
--- a/core/src/data/mbtilesDataSource.cpp
+++ b/core/src/data/mbtilesDataSource.cpp
@@ -1,6 +1,7 @@
 #include "data/mbtilesDataSource.h"
 
 #include "util/asyncWorker.h"
+#include "util/url.h"
 #include "util/zlibHelper.h"
 #include "log.h"
 #include "platform.h"
@@ -125,7 +126,7 @@ MBTilesDataSource::MBTilesDataSource(std::shared_ptr<Platform> _platform, std::s
 
     m_worker = std::make_unique<AsyncWorker>();
 
-    openMBTiles();
+    openMBTiles(_platform);
 }
 
 MBTilesDataSource::~MBTilesDataSource() {
@@ -229,7 +230,7 @@ bool MBTilesDataSource::loadNextSource(std::shared_ptr<TileTask> _task, TileTask
     return next->loadTileData(_task, cb);
 }
 
-void MBTilesDataSource::openMBTiles() {
+void MBTilesDataSource::openMBTiles(const std::shared_ptr<Platform>& _platform) {
 
     try {
         auto mode = SQLite::OPEN_READONLY;
@@ -239,7 +240,8 @@ void MBTilesDataSource::openMBTiles() {
             mode = SQLite::OPEN_READWRITE | SQLite::OPEN_CREATE;
         }
 
-        m_db = std::make_unique<SQLite::Database>(m_path, mode);
+        auto mbTilesPath = _platform->getAssetPath(Url(m_path));
+        m_db = std::make_unique<SQLite::Database>(mbTilesPath, mode);
         LOG("SQLite database opened: %s", m_path.c_str());
 
     } catch (std::exception& e) {
--- a/platforms/android/tangram/src/main/cpp/platform_android.cpp
+++ b/platforms/android/tangram/src/main/cpp/platform_android.cpp
@@ -45,6 +45,7 @@ static jmethodID setRenderModeMethodID = 0;
 static jmethodID startUrlRequestMID = 0;
 static jmethodID cancelUrlRequestMID = 0;
 static jmethodID getFontFilePath = 0;
+static jmethodID copyAssetToTmpPath = 0;
 static jmethodID getFontFallbackFilePath = 0;
 static jmethodID onFeaturePickMID = 0;
 static jmethodID onLabelPickMID = 0;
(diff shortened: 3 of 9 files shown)


ecgreb:
The permission `READ_EXTERNAL_STORAGE` should not be required if the `mbtiles` bundle is stored in the app's designated files folder accessed via `Context.getExternalFilesDir(...)`.

https://developer.android.com/reference/android/content/Context.html#getExternalFilesDir(java.lang.String)

> Starting in KITKAT, no permissions are required to read or write to the returned path; it's always accessible to the calling app. This only applies to paths generated for package name of the calling application. To access paths belonging to other packages, WRITE_EXTERNAL_STORAGE and/or READ_EXTERNAL_STORAGE are required.

core/include/tangram/platform.h
```diff
@@ -45,6 +49,12 @@ class Platform {
 
     virtual bool isContinuousRendering() const;
 
+    // returns the path of an asset given the name of the asset
```
karimnaaji: This description line can probably be removed since it's covered under the other specific cases. We could also  replace "given the name of the asset" -> "given the path of the asset within the application bundle" for iOS and Android.

platforms/android/tangram/src/main/cpp/platform_android.cpp
```diff
@@ -231,6 +233,21 @@ void AndroidPlatform::setContinuousRendering(bool _isContinuous) {
     jniEnv->CallVoidMethod(m_tangramInstance, setRenderModeMethodID, _isContinuous ? 1 : 0);
 }
 
+std::string AndroidPlatform::getAssetPath(const Tangram::Url& _path) const {
+
+    if (_path.scheme() == "asset") {
+        JniThreadBinding jniEnv(jvm);
+        auto assetPath = _path.path()[0] == '/' ? _path.path().substr(1) : _path.path();
```
karimnaaji: Probably makes sense to use `isAbsolute()` here?
```c++
auto assetPath = _path.isAbsolute() ? _path.path().substr(1) : _path.path()
```

platforms/ios/src/TangramMap/platform_ios.mm
```diff
@@ -94,6 +95,15 @@ void initGLExtensions() {
     return data;
 }
 
+std::string iOSPlatform::getAssetPath(const Tangram::Url& _path) const {
+    if (_path.isAbsolute()) {
```
karimnaaji: What covers this use case? iOS client application read scope is only within the bundle.

platforms/android/tangram/src/main/cpp/platform_android.cpp
tallytalwar: I think `isAbsolute` check for the presence of a `scheme`, in this case it has a `scheme` => `asset`.

platforms/ios/src/TangramMap/platform_ios.mm
tallytalwar: Yeah I was not sure about this, can a developer provide any absolute path in an ios app? If its not possible we can get rid of this!

core/include/tangram/platform.h
tallytalwar: Updated description here: 1306421

platforms/ios/src/TangramMap/platform_ios.mm
tallytalwar: Updated here: d12d919

platforms/android/tangram/src/main/cpp/platform_android.cpp
karimnaaji: How the presence of a scheme makes it non-absolute? I think it should work.

matteblair: A _path_ is absolute if it starts with `/`, a _URL_ is absolute if it begins with a scheme. `Url::isAbsolute()` operates in this second sense. https://tools.ietf.org/html/rfc3986#section-4.3

tallytalwar: What @matteblair said. And I did not look at the spec, but just the code we have in our `Url` handling: https://github.com/tangrams/tangram-es/blob/master/core/src/util/url.cpp#L350-L361

platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
```diff
@@ -1134,4 +1138,30 @@ String getFontFallbackFilePath(int importance, int weightHint) {
         return fontFileParser.getFontFallback(importance, weightHint);
     }
 
+    // Tmp Asset Path
+    // ==============
+    String copyAssetToTmpPath(String path) throws IOException {
```
karimnaaji: We shouldn't throw the exception here this is called from JNI code otherwise we would crash if the application fails to find this mbtile file.

platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
```diff
@@ -1134,4 +1138,30 @@ String getFontFallbackFilePath(int importance, int weightHint) {
...
+        String tmpDir = mapView.getContext().getExternalCacheDir().getAbsolutePath();
+        InputStream in = assetManager.open(path);
```
karimnaaji: We should add a `try`, `catch` block here with the right exception check (`IOException`), return either an empty string or null string and do the appropriate error check within tangram core in that case.

tallytalwar: True on it.

platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
```diff
@@ -1134,4 +1138,37 @@ String getFontFallbackFilePath(int importance, int weightHint) {
...
+    // ==============
+    String copyAssetToTmpPath(String path) {
+
+        String tmpDir = mapView.getContext().getExternalCacheDir().getAbsolutePath();
+
+        try {
+            InputStream in = assetManager.open(path);
+            File outFile = new File(tmpDir, path);
+
+            if (!outFile.exists()) {
```
matteblair: So if a file at this path already exists then we just return the path. But we don't know for sure that this file is the asset we're trying to access (the application using tangram also has access to this folder and can put files there). 

It's an unlikely scenario, but it's unfortunately kind of tricky to solve. We could:
 - Compare the contents of the asset with the contents of the file, and create a new file at a different path if they differ.
 - Ask the client application to provide a directory reserved for tangram's use.

platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
```diff
@@ -1134,4 +1138,37 @@ String getFontFallbackFilePath(int importance, int weightHint) {
...
+                OutputStream out = new FileOutputStream(outFile);
+
+                // Copy File
+                byte[] buffer = new byte[1024];
+                int read;
+                while ((read = in.read(buffer)) != -1) {
+                    out.write(buffer, 0, read);
+                }
+
+                in.close();
```
matteblair: The input buffer is never closed if the output file already exists.

platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
tallytalwar: I agree its a tricky thing to solve but I think it will be a very rare case.

The path returned from `getExternalCacheDir` is: `/storage/emulated/0/Android/data/com.mapzen.tangram.android/cache`, so the asset's path comes out to be `/storage/emulated/0/Android/data/com.mapzen.tangram.android/cache/new-york_new-york.mbtiles` for example. Its the same path http cache is saved.

platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
tallytalwar: oops. fixing.

tallytalwar:
Yup I found the [sqlite-ndk]((https://github.com/krystianbigaj/sqlite-ndk) for this purpose only. But I wanted to take an incremental approach since this was a much simpler solution (considering I wanted to get a fix out for this soon). 

I would like to push this one out with the next release (tomorrow), and get sqlite-ndk work done post that.

platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
tallytalwar: Updated here: ceebc9b

core/src/data/mbtilesDataSource.cpp
```diff
@@ -239,7 +240,8 @@ void MBTilesDataSource::openMBTiles() {
             mode = SQLite::OPEN_READWRITE | SQLite::OPEN_CREATE;
         }
 
-        m_db = std::make_unique<SQLite::Database>(m_path, mode);
+        auto mbTilesPath = _platform->getAssetPath(Url(m_path));
+        m_db = std::make_unique<SQLite::Database>(mbTilesPath, mode);
```
karimnaaji: @tallytalwar Do you know how sqlite behaves when given an empty string as a database path? If it's unclear it would be worth checking above whether the asset path is empty string and not create the db in that case.
Either way we should at least add a warning log for that error case.

tallytalwar: 7c5e4b0

platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
```diff
@@ -1134,4 +1138,39 @@ String getFontFallbackFilePath(int importance, int weightHint) {
...
+                out.flush();
+                out.close();
+            } else {
+                in.close();
+            }
+
+            return outFile.getAbsolutePath();
+        } catch (IOException e) {
```
karimnaaji: It's more of a Java thing, but streams should be close in a `finally` block in general.

matteblair:
Proof-of-concept using the 'ndk-asset' sqlite VFS: https://github.com/tangrams/tangram-es/commit/cb85ca83a868031f1d2570560a5d6a2a673e72b3

platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
```diff
@@ -1134,4 +1138,45 @@ String getFontFallbackFilePath(int importance, int weightHint) {
...
+                    out.write(buffer, 0, read);
+                }
+
+                out.flush();
+            }
+
+            return outFile.getAbsolutePath();
+        } catch (IOException e) {
+            e.printStackTrace();
+            return "";
```
karimnaaji: I would put that at the end of the function block.

tallytalwar:
Closing in favor of #1360

Issue #1339, opened 2017-03-02
ES equivalent for JS's Oriented lines feature


tallytalwar:
Refer the following PR and issue for details:

JS PR: https://github.com/tangrams/tangram/pull/506
JS Issue: https://github.com/tangrams/tangram/issues/471

PR #1349, opened 2017-03-08
[wip] Boundary Labels - text_source:left/right

hjanetzek:
Implement https://github.com/tangrams/tangram-es/issues/1339

### DIFF
--- a/core/src/scene/styleParam.h
+++ b/core/src/scene/styleParam.h
@@ -62,6 +62,8 @@ enum class StyleParamKey : uint8_t {
     sprite_default,
     style,
     text_source,
+    text_source_left,
+    text_source_right,
     text_wrap,
     tile_edges,
     text_transform,
--- a/core/src/scene/styleParam.h
+++ b/core/src/scene/styleParam.h
@@ -141,8 +141,20 @@ struct StyleParam {
         }
     };
 
-    using Value = variant<Undefined, none_type, bool, float, uint32_t, std::string, glm::vec2,
-                            Width, LabelProperty::Placement, LabelProperty::Anchors>;
+
+    struct TextSource {
+        std::vector<std::string> keys;
+        bool operator==(const TextSource& _other) const {
+            if (keys.size() != _other.keys.size()) { return false; }
+            for (size_t i = 0; i < keys.size(); i++) {
+                if (!(keys[i] == _other.keys[i])) { return false;}
+            }
+            return true;
+        }
+    };
+
+    using Value = variant<none_type, Undefined, bool, float, uint32_t, std::string, glm::vec2, Width,
+                          LabelProperty::Placement, LabelProperty::Anchors, TextSource>;
 
     StyleParam() :
         key(StyleParamKey::none),
--- a/core/src/scene/styleParam.h
+++ b/core/src/scene/styleParam.h
@@ -147,11 +147,7 @@ struct StyleParam {
     struct TextSource {
         std::vector<std::string> keys;
         bool operator==(const TextSource& _other) const {
-            if (keys.size() != _other.keys.size()) { return false; }
-            for (size_t i = 0; i < keys.size(); i++) {
-                if (!(keys[i] == _other.keys[i])) { return false;}
-            }
-            return true;
+            return keys == _other.keys;
         }
     };
 
--- a/core/src/style/textStyleBuilder.cpp
+++ b/core/src/style/textStyleBuilder.cpp
@@ -464,14 +464,21 @@ TextStyle::Parameters TextStyleBuilder::applyRule(const DrawRule& _rule,
 
     TextStyle::Parameters p;
 
-    _rule.get(StyleParamKey::text_source, p.text);
-    if (!_rule.isJSFunction(StyleParamKey::text_source)) {
-        if (p.text.empty()) {
-            p.text = _props.getString(key_name);
-        } else {
-            p.text = resolveTextSource(p.text, _props);
+    auto& textSource = _rule.findParameter(StyleParamKey::text_source);
+
+    if (textSource.value.is<StyleParam::TextSource>()) {
+        for (auto& key : textSource.value.get<StyleParam::TextSource>().keys) {
+            p.text = _props.getString(key);
+            if (!p.text.empty()) { break; }
         }
+    } else if (textSource.value.is<std::string>()) {
+        // From function evaluation
+        p.text = textSource.value.get<std::string>();
+    } else {
+        // Use default key
+        p.text = _props.getString(key_name);
     }
+
     if (p.text.empty()) { return p; }
 
     auto fontFamily = _rule.get<std::string>(StyleParamKey::text_font_family);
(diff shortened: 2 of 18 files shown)


bcamper:
Hey @hjanetzek @dmvaldman @nvkelso can we sync up on how we want anchor/offset to work here?

What I think @dmvaldman's [JS branch](https://github.com/tangrams/tangram/pull/516) is currently doing is *not* offsetting the left/right labels by default, but automatically applying the `offset` in the "correct" direction for each side of the label, e.g. a positive offset would push the right label to the right of the line, and the left label to the left of the line. This does require setting an explicit `offset` to render these labels:
```
draw:
  text:
    text_source:
      left: 'name:left'
      right: ‘name:right'
    offset : [0, 10px]
```

I infer from f68020025fc4c9fa8928999da2522fcb646d5cca that the ES branch is automatically applying an offset to move the left/right labels outside the line (according to their own text height anyway...? may not match the thickness of the line rendered below the label). Is that right?

Which of these do we want?

nvkelso:
I prefer sensible defaults, and in this case ES automatically applying an
offset so the left and right text both shows by default makes more sense to
me (and not also having to set the offset – without setting it on JS I
couldn’t see both the labels, they were mashed up on the same line).

I have the explicate offset set in a branch of Bubble Wrap to account for
the different line thickness. I’d either assume a 1px line weight + the
text sizes, or see the parent style’s property value and adjust that
dynamically.

But David should chime in – he was explaining to me how the offset
direction is more complicated with inner / outer calculation of the offset
for closed shapes.

bcamper:
Since the line label and line itself are independent `draw` groups, they don't have direct knowledge of each other, so I'd keep any complexity of trying to auto-detect out of this (currently anyway).

So the default would be an `offset` equal to the text height, and then any *additional*/explicit offset provided in the `draw` block would push the labels out further (accounting for left/right direction, as Dave's branch currently does)?

Meaning in this example, the total offset would be *height of text plus 4 pixels*:
```
draw:
  text:
    text_source:
      left: 'name:left'
      right: ‘name:right'
    offset : [0, 4px]
```

bcamper:
The above is consistent with point anchoring (auto offset + explicit offset) so it works for me.

nvkelso:
I’m down with it :)

dmvaldman:
Cool. I'll make the necessary changes on the JS side.

hjanetzek:
@bcamper yes, using the text height as baseline offset for boundary labels sounds good to me. 

I've currently added the minimal offset that is required for left/right labels to not occlude each other in collision detection.  I'm still thinking about the best way to implement that 'sibling labels' do not occlude each other so that one could reduce the minimal offset.

dmvaldman:
So one problem on the JS side is that there is no `canvas` API to measure the height of text (only its width). To approximate the height, we use the text's pixel size and add a fudge factor. Unfortunately, actual text height is not only pixel size dependent, but font dependent, so for many fonts we cannot accurately measure the height of text.

How should we proceed as a result of this?

bcamper:
@dmvaldman it's not a great answer, but if you go back to separating them by the texture (rather than collision) box, they shouldn't collide, right? (though the labels will also end up a little further apart than they would in ES)

dmvaldman:
Yes, but the main problem is that it won't be consistent across fonts

bcamper:
> Yes, but the main problem is that it won't be consistent across fonts

But don't we have a fixed height that's consistent based on whatever is in the `size` of the `font` block?

Either way can we take a look to see how it feels :)

dmvaldman:
A 12px font will have a different height depending on the font (and characters)

bcamper:
@dmvaldman right, but *our* expected height will be the same (right?) -- since Canvas doesn't tell us the real height, we guess. So there's some visual variation depending on true font size, but this is already true for us, whether it's being offset or not. Seems like if there were artifacts (font higher than we expect and got cut off, for example), that would already be an issue for us and wouldn't be affected by this change?

core/src/scene/styleParam.h
```diff
@@ -141,8 +143,20 @@ struct StyleParam {
         }
     };
 
-    using Value = variant<Undefined, none_type, bool, float, uint32_t, std::string, glm::vec2,
-                            Width, LabelProperty::Placement, LabelProperty::Anchors>;
+
+    struct TextSource {
+        std::vector<std::string> keys;
+        bool operator==(const TextSource& _other) const {
+            if (keys.size() != _other.keys.size()) { return false; }
```
karimnaaji: I think that `keys == _other.keys` should be sufficient here and more succinct. 
See operator the doc here http://en.cppreference.com/w/cpp/container/vector/operator_cmp, the size is the first element compared before the content.

core/src/style/textStyleBuilder.cpp
```diff
@@ -268,7 +353,8 @@ bool TextStyleBuilder::addStraightTextLabels(const Line& _line, const TextStyle:
             glm::vec2 b = glm::vec2(p1 - p0) / float(run);
 
             for (int r = 0; r < run; r++) {
-                addLabel(_params, Label::Type::line, {{ a, a+b }}, _rule);
+                _cb(a, a+b);
```
karimnaaji: What about calling it `_onLabelAdd` instead of `_cb`?

core/src/style/textStyleBuilder.cpp
```diff
@@ -462,14 +554,12 @@ TextStyle::Parameters TextStyleBuilder::applyRule(const DrawRule& _rule,
 
     TextStyle::Parameters p;
 
-    _rule.get(StyleParamKey::text_source, p.text);
-    if (!_rule.isJSFunction(StyleParamKey::text_source)) {
```
karimnaaji: How is this case handled now?

karimnaaji:
Found a little bug while testing this more, when given a two dimensional offset, the boundary labels would go in opposite directions:

screen shot 2017-03-29 at 12 04 44 pm

hjanetzek:
@karimnaaji  yeah, I think we should disable horizontal offset for line labels for now. This is also not handled in finding anchor positions.

core/src/style/textStyleBuilder.cpp
hjanetzek: hmm ok

core/src/style/textStyleBuilder.cpp
hjanetzek: If the text is from a function the value of `text_source` is a string otherwise it's `StyleParam::TextSource`.

https://github.com/tangrams/tangram-es/pull/1349/files/110e1d9264bea34e2e73cae59bd3b7d5d62c45e7#diff-3f0d16e06da67d94777462bfd1445c27R171

karimnaaji:
It seems that JS supports horizontal offsets, by translating them in the same direction.

core/src/scene/styleParam.h
hjanetzek: :+1:

hjanetzek:
Do they also check that a label does not go over the line segment? Say the anchor is at the center of a 100px long line segment, then adding 100px offset would move the label completely off the line.

Also, since one cannot define the anchor position on a line the option seems generally not that useful :)

karimnaaji:
>Do they also check that a label does not go over the line segment? Say the anchor is at the center of a 100px long line segment, then adding 100px offset would move the label completely off the line.

Yes, I get your point, I don't think horizontal offset has good use case anyway, but we should still be consistent with JS. Having a 1 dimension offset like `offset: 10px` instead of `offset: [0, 10px]`  would start offsetting in opposite directions, results would become quite different on both platforms.

karimnaaji:
From the docs https://mapzen.com/documentation/tangram/draw/#offset it seems that we can't use offsets as a single value so we're good here if we simply deactivate horizontal offsets like you proposed.

Issue #935, opened 2016-08-26
Implement text/point style buffer option


karimnaaji:
https://github.com/tangrams/tangram-docs/issues/141

PR #1350, opened 2017-03-08
Parse buffer option

karimnaaji:
Fixes https://github.com/tangrams/tangram-es/issues/935


core/src/style/textStyleBuilder.cpp
```diff
@@ -537,11 +537,13 @@ TextStyle::Parameters TextStyleBuilder::applyRule(const DrawRule& _rule,
             }
         }
 
-
         _rule.get(StyleParamKey::text_transition_hide_time, p.labelOptions.hideTransition.time);
         _rule.get(StyleParamKey::text_transition_selected_time, p.labelOptions.selectTransition.time);
         _rule.get(StyleParamKey::text_transition_show_time, p.labelOptions.showTransition.time);
 
+        if (!_rule.get(StyleParamKey::text_buffer, p.labelOptions.buffer)) {
```
tallytalwar: Does the `buffer` value get inherited from parent `point` style if not specified? I think it is not in JS, which is what we have, but maybe good to confirm.

core/src/scene/styleParam.cpp
```diff
@@ -180,6 +184,17 @@ StyleParam::Value StyleParam::parseString(StyleParamKey key, const std::string&
         }
         return vec.value;
     }
+    case StyleParamKey::text_buffer:
```
tallytalwar: Can `buffer` param take stops as a value? I could not find it in the documentation, but probably good to confirm from JS implementation.

core/src/style/textStyleBuilder.cpp
karimnaaji: It's unclear, have you tried and seen that there was no inheritance? I don't see any mention of that in https://github.com/tangrams/tangram/pull/306.

core/src/scene/styleParam.cpp
karimnaaji: I don't think so, not mentioned in reference issue https://github.com/tangrams/tangram-docs/issues/141.

core/src/style/textStyleBuilder.cpp
tallytalwar: Yeah I couldn't find anywhere in the documentation for both the things I mentioned. Will have to give it a shot in play to confirm.

tallytalwar: Yup doesn't look like this gets inherited, so we are good here.

core/src/scene/styleParam.cpp
tallytalwar: Yup, stops seems to not work with JS either. So we are all good here.

PR #1354, opened 2017-03-09
Fix view complete on tile cancel

### LINKED ISSUE #1353: Tiles loading is always returning true on iOS and Android
The tiles loading boolean that we use to know whether the whole set of visible tiles are loaded for the current screen view is always returning true on both Android and iOS platforms, which makes our on map ready event broken. This is not reproducible on OSX and Linux where this works as expected.

tallytalwar:
- was happening on platforms because both mobile platform demos, use clientdatasource which
initially has no geometry and hence have their respective tiles cancelled for loading.
- on desktop demos we are not using clientdatasource so this use case was not being hit.

fixes #1353

### DIFF
--- a/core/src/tile/tileManager.cpp
+++ b/core/src/tile/tileManager.cpp
@@ -223,7 +223,7 @@ void TileManager::updateTileSet(TileSet& _tileSet, const ViewState& _view,
             if (entry.isReady()) {
                 m_tiles.push_back(entry.tile);
 
-                if (!entry.isLoading() &&
+                if (!entry.isInProgress() &&
                     (sourceGeneration < generation)) {
                     // Tile needs update - enqueue for loading
                     entry.task = _tileSet.source->createTask(visTileId);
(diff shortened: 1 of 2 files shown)


hjanetzek:
I wonder ` if (entry.isLoading()) { m_tilesInProgress++;  }` could be sufficient

tallytalwar:
yup that should be fine. Will give that a try. Thanks

tallytalwar:
Updated!

hjanetzek:
hmm I think the `needsLoading()` is needs a better name as it means needsToBeEnqueuedForDownloadAgain. An entry can be both `needsLoading()` and `isLoading()` at the same time.
@tallytalwar do you have a case in mind where `(!entry.isCanceled() && !entry.isReady())` is not covered by `entry.isLoading()` ?

tallytalwar:
`entry.isLoading` does the cover the following cases:
1. when a tile's data is downloaded so its ready to be build, but its not ready to be rendered, and will only be done when `tile` has been build, => `entry.isReady()` is true.
2. when a clientdatasource task is created (when clientdatasource has no geometry data in it), the `TileTask::process` cancels this task, and hence in next update cycle `m_tilesInProcess` should not be incremented for this task.

I agree on the naming of `isLoading` and `needsLoading`. How about naming `needsLoading` to `entry.enqueueAgain()`?

hjanetzek:
Or the other way round: clarifying `isLoading()` to `inProgress()` :)  I can elaborate more tomorrow

tallytalwar:
yup, either works.

tallytalwar:
Yup, make sense. A `tileEntry`'s task is `reset` when its tile is ready. Which makes its `isLoading` (now `isInProgress`) return false. Updated here: af8e321

hjanetzek:
LGTM!

Issue #857, opened 2016-07-22
Implement Screenshot API


msmollin:

lygstate:
This could be able to implemented with framebufferobject

PR #1359, opened 2017-03-10
iOS screenshot API

karimnaaji:
Fixes https://github.com/tangrams/tangram-es/issues/857

### DIFF
--- a/platforms/ios/demo/src/MapViewController.m
+++ b/platforms/ios/demo/src/MapViewController.m
@@ -18,9 +18,17 @@
 
 @implementation MapViewControllerDelegate
 
+- (void)mapView:(TGMapViewController *)view didCaptureScreenshot:(UIImage *)screenshot
+{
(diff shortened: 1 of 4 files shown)


platforms/ios/demo/src/MapViewController.m
```diff
@@ -18,9 +18,17 @@ - (void)addAlert:(NSString *)message withTitle:(NSString *)title;
 
 @implementation MapViewControllerDelegate
 
+- (void)mapView:(TGMapViewController *)view didCaptureScreenshot:(UIImage *)screenshot
+{
+    NSArray* paths = NSSearchPathForDirectoriesInDomains(NSDocumentDirectory, NSUserDomainMask, YES);
```
karimnaaji: We actually don't really need that, it was used only for testing in the demo app, I will remove it and rebase.

tallytalwar:
Also can you add some documentation for this? Thanks

PR #1360, opened 2017-03-10
Sqlite assets

tallytalwar:
Addressed #1309 

Android platform:
- if the mbtiles path specified in the scene file is an apk asset path, then ndk-sqlite vfs is used to read the mbtiles file into sqlite db.
- if the mbtiles specified in the scene file is an absolute path on the device, `READ_EXTERNAL_STORAGE` permission is required both in the `AndroidManifest.xml` as well as during runtime using the following (post android M):
   ```
  if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
           requestPermissions(new String[]{Manifest.permission.READ_EXTERNAL_STORAGE}, 1);
  }
   ```
   NOTE: from @ecgreb 
   > The permission READ_EXTERNAL_STORAGE should not be required if the mbtiles bundle is stored in the app's designated files folder accessed via Context.getExternalFilesDir(...).

iOS platform:
- if the mbtiles path specified in the scene file is app bundle path, then this file's path is resolved with the app's Bundle path which is then used to open the sqlite db.

All other platforms:
The path to the mbtiles (or for that matter all sources) are resolved (relative to base scene yaml, or absolute path) automatically and works out of the box.

### DIFF
--- a/core/include/tangram/platform.h
+++ b/core/include/tangram/platform.h
@@ -47,6 +47,8 @@ public:
 
     virtual bool isContinuousRendering() const;
 
+    virtual std::string resolveAssetPath(const std::string& path) const;
+
     // Read a file as a string
     // Opens the file at the _path and returns a string with its contents.
     // If the file cannot be found or read, the returned string is empty.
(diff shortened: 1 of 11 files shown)


matteblair:
Oh also we should probably squash the two commits in the middle.

tallytalwar:
Ohh I did, forgot to push, now updated.

PR #1367, opened 2017-03-13
fix linux debug build / initialize curl

hjanetzek:


matteblair:
This class is definitely a "spooky action at a distance" kind of thing, which isn't intuitive to use, but I couldn't think of a better way to implement the global setup or libcurl with the recommended constraints: https://curl.haxx.se/libcurl/c/curl_global_init.html

hjanetzek:
Yeah we could put that static global also in urlClient I guess

>   In a C++ module, it is common to deal with the global constant situation by defining a special class that represents the global constant environment of the module. A program always has exactly one object of the class, in static storage. That way, the program automatically calls the constructor of the object as the program starts up and the destructor as it terminates. As the author of this libcurl-using module, you can make the constructor call curl_global_init and the destructor call curl_global_cleanup and satisfy libcurl's requirements without your user having to think about it.

hjanetzek:
@matteblair curl counts the calls to global_init/shutdown. so it should be save to have our static global Environment. Only minor issue could be that the init flags are set by who comes first.

https://github.com/curl/curl/blob/0afbcfd800c45e766e225e4ce273b128ee6a8c25/lib/easy.c#L200-L274

PR #1368, opened 2017-03-13
fix: PointStyle spacing for overzoomed tiles

hjanetzek:


tallytalwar:
Travis is not picking up this, but since this is a small change it should be just fine. Merging.

PR #1376, opened 2017-03-15
update datasource urls

meetar:


nvkelso:
These probably all need API keys specified in their URL params (if they don't have it requests will be blocked from the server).

meetar:
Yep - I see a few which appear to have an api_key nearby, but as I'm not sure how keys are handled in the other cases (or which keys are to be used) I'll let the maintainers modify this as needed before merging.

matteblair:
Looks like these have API keys in all the necessary places - the only thing breaking is our URL unit tests (they expected the old tile URL as input). I'll update that to prevent future confusion.

PR #1378, opened 2017-03-15
fix: Skip YAML source node if null at merging time

karimnaaji:
At style import time, don't merge any map field where the YAML source node is null.


matteblair:
Superseded by #1387

PR #1381, opened 2017-03-16
Fix label collider repeat group

hjanetzek:
Also tiles were scaled up twice by tileScale so repeat groups were not properly filtered for the zoom range in which the tile is visible. 

Filtering out more labels that cannot become visible should also help to have slightly better highway arrow placement :)

### DIFF
--- a/core/src/labels/labelCollider.cpp
+++ b/core/src/labels/labelCollider.cpp
@@ -19,35 +19,41 @@ void LabelCollider::addLabels(std::vector<std::unique_ptr<Label>>& _labels) {
     }
 }
 
-void LabelCollider::handleRepeatGroup(size_t startPos) {
+size_t LabelCollider::filterRepeatGroups(size_t startPos, size_t curPos) {
 
-    float threshold2 = pow(m_labels[startPos].label->options().repeatDistance, 2);
-    size_t repeatGroup = m_labels[startPos].label->options().repeatGroup;
+    size_t endGroup = m_labels[curPos].label->options().repeatGroup;
+    size_t endPos = curPos;
 
-    // Get the range
-    size_t endPos = startPos;
-    for (;startPos > 0; startPos--) {
-        if (m_labels[startPos-1].label->options().repeatGroup != repeatGroup) { break; }
-    }
-    for (;endPos < m_labels.size()-1; endPos++) {
-        if (m_labels[endPos+1].label->options().repeatGroup != repeatGroup) { break; }
-    }
+    for (size_t pos = startPos; pos <= curPos; pos = endPos + 1) {
+        size_t repeatGroup = m_labels[pos].label->options().repeatGroup;
+        float  threshold2 = pow(m_labels[pos].label->options().repeatDistance, 2);
 
+        // Find end of the current repeatGroup
+        endPos = pos;
 
-    for (size_t i = startPos; i < endPos; i++) {
-        Label* l1 = m_labels[i].label;
-        if (l1->isOccluded()) { continue; }
-
-        for (size_t j = i+1; j <= endPos; j++) {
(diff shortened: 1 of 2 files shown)


hjanetzek:
Hmm, maybe this filters out too many labels - doing the tile placement at 2x overzoom would keep more POI labels.

shot-2017-03-16_13-20-23

shot-2017-03-16_13-20-05

core/src/labels/labelCollider.cpp
```diff
@@ -20,35 +20,41 @@ void LabelCollider::addLabels(std::vector<std::unique_ptr<Label>>& _labels) {
...
-    for (;endPos < m_labels.size()-1; endPos++) {
-        if (m_labels[endPos+1].label->options().repeatGroup != repeatGroup) { break; }
-    }
+    for (size_t pos = startPos; pos <= curPos; pos = endPos) {
 
+        size_t repeatGroup = m_labels[pos].label->options().repeatGroup;
+        float  threshold2 = pow(m_labels[pos].label->options().repeatDistance, 2);
+
+        // Find end of the current repeatGroup
+        size_t endPos = pos;
```
karimnaaji: This variable `endPos` is shadowing the previously defined one line 26.

core/src/labels/labelCollider.cpp
```diff
@@ -190,7 +204,7 @@ void LabelCollider::process(TileID _tileID, float _tileInverseScale, float _tile
             repeatGroup = l1->options().repeatGroup;
 
             if (repeatGroup != 0) {
-                handleRepeatGroup(pair.first);
+                repeatPos = filterRepeatGroups(repeatPos, pair.first);
```
karimnaaji: `repeatPos` is a little ambiguous (it seems to be where we last stopped? `lastFilteredIndex` or something on those lines seems more appropriate).

core/src/labels/labelCollider.cpp
hjanetzek: oh right - this should 'only' have caused some labels being checked twice though.

core/src/labels/labelCollider.cpp
hjanetzek: `lastFilteredRepeatGroupLabelIndex` ?

karimnaaji: Haha, well, your call.

core/src/labels/labelCollider.cpp
```diff
@@ -251,6 +265,8 @@ void LabelCollider::process(TileID _tileID, float _tileInverseScale, float _tile
         }
     }
 
+    filterRepeatGroups(repeatPos, m_labels.size()-1);
```
karimnaaji: This checks for the last range of groups and in general it converges quite quickly, but have you considered doing all the filtering at once at the end?

hjanetzek: The idea is to filter out repeated labels early so that they don't block lower priority labels.

hold on - updating the PR to address some comments

platforms/ios/src/TangramMap/TGMarker.h
```diff
@@ -179,6 +179,8 @@ NS_ASSUME_NONNULL_BEGIN
 @property (readonly, nonatomic) NSInteger drawOrder;
 /// Access the marker icon (readonly)
 @property (readonly, nonatomic) UIImage* icon;
+/// A custom user data
+@property (assign, nonatomic) void* userData;
```
karimnaaji: Oups, how did that get in there?

hjanetzek: hmm yeah.. fixing

hjanetzek: It was `git rebase --interactive` on the wrong commit hash

PR #1383, opened 2017-03-16
Resize the map when an accurate bounds is available

### LINKED ISSUE #1377: Improve Marker Selection
It is currently very hard to select markers. Pressing on top of them is not triggering the selection delegate to be called with a `TGMarkerPickResult`. Increasing the pick radius does not seem to improve marker selection when tapping directing on the marker (it does increase the area around the marker that is selectable though).

To reproduce:
1. Clone the iOS demo app repo:
```
git clone git@github.com:mapzen/ios.git
```
2. Install app to phone
3. Press "Toggle Marker" button to make sample marker visible
4. Press on marker
5. Observe that alert is not shown to indicate selection

Here is a video showing what happens when the pick radius is set to 2:
https://www.dropbox.com/s/21auvjk3c51zh5z/20170309_161650.mp4?dl=0

cc/ @karimnaaji @matteblair

sarahlensing:
This fixes a bug which set the map's size before an accurate bounds was available. Previously the map was resized in `viewDidLoad` but now it is set in ~~`viewDidLayoutSubviews`~~ `viewWillLayoutSubviews`. When the `TGMapViewController` was a view controller within a `UITabBarViewController` without a transparent tab bar, the map's view would not be properly sized.

This [branch](https://github.com/tangrams/tangram-es/tree/pick-test) shows that bug manifesting. Clicking on the screen draws a marker above where you tap (which makes it appear unselectable because you have to tap below where the marker is drawn to trigger the selection delegate).

Closes #1377

### DIFF
--- a/platforms/ios/demo/src/MapViewController.m
+++ b/platforms/ios/demo/src/MapViewController.m
@@ -120,7 +120,7 @@
     {
         if (!vc.markerPolygon) {
             vc.markerPolygon = [[TGMarker alloc] init];
-            vc.markerPolygon.styling = @"{ style: 'polygons', color: 'blue', order: 500 }";
+            vc.markerPolygon.stylingString = @"{ style: 'polygons', color: 'blue', order: 500 }";
 
             // Add the marker to the current view
             vc.markerPolygon.map = view;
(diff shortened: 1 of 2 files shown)


sarahlensing:
Updated to 4 space indentation and also to a slightly earlier lifecycle method: `viewWillLayoutSubviews`. I tested it still works but you might want to too!

karimnaaji:
Ok I'll give it another try with this one before merging.

PR #1385, opened 2017-03-17
Misc. render optimizations

karimnaaji:
- Use vaos for first batch of dynamic buffer mesh (looking at the trace, vertex bytes offset is different only for specific use cases; when a mesh is too big, or when a texture is specific to a mesh through a bitmap marker. Since the byte vertex offset cannot be captured by the vertex array object, only use it when this offset is 0, which makes it being used most of the time)
- Optimize costly shaders (Shader program 60 on the screenshot 3.59ms -> 1.55ms): 
   - Offload some of costly shader work from fragment to vertex shader
   - Pass part of it through glsl optimizer and adapt to something still human readable
- Configure `GLKView` to disable stencil and use no msaa on iOS

For the same area on walkabout the frame rate time changed from ~6.3ms to ~3.8ms on iPhone 6S Plus:

screen shot 2017-03-17 at 11 52 20 am
screen shot 2017-03-17 at 11 52 37 am

### DIFF
--- a/core/shaders/point.vs
+++ b/core/shaders/point.vs
@@ -74,35 +74,26 @@ void main() {
     v_sdf_scale = a_scale / 64.0;
 
     if (u_pass == 0) {
-        // fill
         v_sdf_threshold = 0.5;
-        //v_alpha = 0.0;
-    } else if (a_stroke.a > 0.0) {
-        // stroke
-        // (0.5 / 3.0) <= sdf change by pixel distance to outline == 0.083
-        float sdf_pixel = 0.5/u_max_stroke_width;
-
-        // de-normalize [0..1] -> [0..max_stroke_width]
-        float stroke_width = a_stroke.a * u_max_stroke_width;
-
-        // scale to sdf pixel
-        stroke_width *= sdf_pixel;
-
-        // scale sdf (texture is scaled depeding on font size)
-        stroke_width /= v_sdf_scale;
-
-        v_sdf_threshold = max(0.5 - stroke_width, 0.0);
-
-        v_color.rgb = a_stroke.rgb;
     } else {
-        v_alpha = 0.0;
+        if ((a_stroke.a > 0.0)) {
+            float stroke_width_2;
+            stroke_width_2 = ((a_stroke.a * u_max_stroke_width) * (0.5 / u_max_stroke_width));
+            stroke_width_2 = (stroke_width_2 / v_sdf_scale);
+            v_sdf_threshold = max ((0.5 - stroke_width_2), 0.0);
+            v_color.xyz = a_stroke.xyz;
+        } else {
+            v_alpha = 0.0;
+        }
     }
 
-    vec4 position = vec4(vertex_pos, 0.0, 1.0);
+    vec4 position;
+    position.zw = vec2(0.0, 1.0);
+    position.xy = vertex_pos;
 
     #pragma tangram: position
 
-    gl_Position = u_ortho * position;
+    gl_Position = (u_ortho * position);
 
 #else
     v_texcoords = a_uv;
--- a/core/shaders/point.vs
+++ b/core/shaders/point.vs
@@ -44,7 +44,7 @@ varying vec4 v_color;
 varying vec2 v_texcoords;
 #ifdef TANGRAM_TEXT
 varying float v_sdf_threshold;
-varying float v_sdf_scale;
+varying float v_sdf_pixel;
 #endif
 varying float v_alpha;
 
@@ -71,16 +71,17 @@ void main() {
 #ifdef TANGRAM_TEXT
     vec2 vertex_pos = UNPACK_POSITION(a_position);
     v_texcoords = UNPACK_TEXTURE(a_uv);
-    v_sdf_scale = a_scale / 64.0;
+    float sdf_scale = a_scale / 64.0;
+    v_sdf_pixel = 0.5 / (u_max_stroke_width * sdf_scale);
 
     if (u_pass == 0) {
         v_sdf_threshold = 0.5;
     } else {
         if ((a_stroke.a > 0.0)) {
-            float stroke_width_2;
-            stroke_width_2 = ((a_stroke.a * u_max_stroke_width) * (0.5 / u_max_stroke_width));
-            stroke_width_2 = (stroke_width_2 / v_sdf_scale);
-            v_sdf_threshold = max ((0.5 - stroke_width_2), 0.0);
+            float stroke_width;
+            stroke_width = ((a_stroke.a * u_max_stroke_width) * (0.5 / u_max_stroke_width));
+            stroke_width = (stroke_width / sdf_scale);
+            v_sdf_threshold = max ((0.5 - stroke_width), 0.0);
             v_color.xyz = a_stroke.xyz;
         } else {
             v_alpha = 0.0;
--- a/core/shaders/sdf.fs
+++ b/core/shaders/sdf.fs
@@ -25,7 +25,7 @@ varying vec4 v_color;
 varying vec2 v_texcoords;
 varying float v_sdf_threshold;
 varying float v_alpha;
-varying float v_sdf_scale;
+varying float v_sdf_pixel;
 
 #pragma tangram: global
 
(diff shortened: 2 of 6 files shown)


hjanetzek:
@karimnaaji do you have a timing just without msaa compared to master? I'm just curious.

DynamicMesh Vaos  :+1:

karimnaaji:
@hjanetzek a timing with master with only msaa disabled in same environment:
screen shot 2017-03-17 at 3 15 52 pm

karimnaaji:
@hjanetzek here are some more ideas that we can use for later improvements and similar tweaks:
- Allow to increase the size of the glyph texture atlas based on the pixel density, right now they get filled up quite quickly on higher density screen display. The only drawback that I faced when changing that is that we need a way to invalidate all texture atlases that could have been created, since pixel ratio can change at runtime when moving the window between screens, and it messes up all UVs previously created. This would reduces the number of vertex buffer objects since they are tight to their texture atlas. I tested 512x512 which reduces the number of draw calls for icons/text by 2 on those devices.
- Passing more shader within the glsl-optimizer and manually tweak them while keeping them human readable (I have not looked at polygon shader so far).
- Use https://www.shadertoy.com/view/4ldSD2 on sdf instead of smoothstep, I have not yet looked at the visual difference with that so I haven't incorporated it, but the amount of instructions generated is effectively reduced as expected.

hjanetzek:
512x512 could be a sensible default for glyph textures. One reason to keep them smaller was that an atlas gets only discarded and reused when no more meshes reference them. So when one makes a world tour visiting all map regions with different scripts one may hold on to large atlases with very few referenced glyphs.

karimnaaji:
For comparison @nvkelso, @sensescape  here are some screenshots for anti-aliasing, we will still leave that as an option on the iOS framework but we might turn it off by default.

enabled:
img_0007
disabled:
img_0008

nvkelso:
Were these generated on an iPhone "+" sized phone?

karimnaaji:
Yes

matteblair:
This probably isn't related to the changes here, but I'm noticing that in the iOS demo app a screenshot is being captured whenever "view complete" occurs, even when the call to ` captureScreenshot` is removed. 

Based on the logic for capturing screenshots in `TGMapViewController` (https://github.com/tangrams/tangram-es/blob/master/platforms/ios/src/TangramMap/TGMapViewController.mm#L871-L881) it seems that if a delegate is present that responds to `didCaptureScreenshot` then a screenshot will be captured every time "view complete" occurs (whether a screenshot was requested or not). That doesn't seem right.

karimnaaji:
That effectively doesn't seem right. There is a missing `self->captureFrameWaitForViewComplete = false` to reset ths flag on that block.

matteblair:
Something here is causing consistent, immediate crashes in the android demo on our Galaxy S3. The call stack isn't consistent, sometimes it's in our code, sometimes it's in the GL driver, but it's always a bad deallocation for the address `deadbaad`. Example:
```
F/libc    ( 5878): @@@ ABORTING: INVALID HEAP ADDRESS IN dlfree
F/libc    ( 5878): Fatal signal 11 (SIGSEGV) at 0xdeadbaad (code=1)
I/DEBUG   ( 5063): *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
I/DEBUG   ( 5063): Build fingerprint: 'samsung/d2vl/d2can:4.0.4/IMM76D/I747MVLALE8:user/release-keys'
I/DEBUG   ( 5063): pid: 5878, tid: 5891  >>> com.mapzen.tangram.android <<<
I/DEBUG   ( 5063): signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr deadbaad
I/DEBUG   ( 5063):  r0 deadbaad  r1 00000001  r2 40000000  r3 00000000
I/DEBUG   ( 5063):  r4 00000000  r5 00000027  r6 400b8526  r7 400c7574
I/DEBUG   ( 5063):  r8 00c90b90  r9 00000009  10 00000008  fp 009d0ca8
I/DEBUG   ( 5063):  ip ffffffff  sp 5dccd4b0  lr 40099525  pc 400958ac  cpsr 60000030
I/DEBUG   ( 5063):  d0  6772612064696c75  d1  766e49203232206d
...
I/DEBUG   ( 5063):          #00  pc 000178ac  /system/lib/libc.so
I/DEBUG   ( 5063):          #01  pc 0001373a  /system/lib/libc.so
I/DEBUG   ( 5063):          #02  pc 00015a74  /system/lib/libc.so (dlfree)
I/DEBUG   ( 5063):          #03  pc 000160ec  /system/lib/libc.so (free)
I/DEBUG   ( 5063):          #04  pc 000477e0  /system/lib/egl/libGLESv2_adreno200.so (rb_vbo_free)
I/DEBUG   ( 5063):          #05  pc 0003e50c  /system/lib/egl/libGLESv2_adreno200.so (rb_delete_detach_vbo_list)
I/DEBUG   ( 5063):          #06  pc 0005c1de  /system/lib/egl/libGLESv2_adreno200.so (leia_perform_resolve)
I/DEBUG   ( 5063):          #07  pc 0003f25a  /system/lib/egl/libGLESv2_adreno200.so (rb_resolve)
I/DEBUG   ( 5063):          #08  pc 00041760  /system/lib/egl/libGLESv2_adreno200.so (rb_surface_swap)
I/DEBUG   ( 5063):          #09  pc 00030882  /system/lib/egl/libGLESv2_adreno200.so (gl2_surface_swap)
I/DEBUG   ( 5063):          #10  pc 00026bb4  /system/lib/egl/libGLESv2_adreno200.so (oglSwapBuffer)
I/DEBUG   ( 5063):          #11  pc 0000da5c  /system/lib/egl/libEGL_adreno200.so (qeglDrvAPI_eglSwapBuffers)
I/DEBUG   ( 5063):          #12  pc 000055f8  /system/lib/egl/libEGL_adreno200.so (eglSwapBuffers)
I/DEBUG   ( 5063):          #13  pc 0000cc80  /system/lib/libEGL.so (eglSwapBuffers)
I/DEBUG   ( 5063):          #14  pc 0004c2aa  /system/lib/libandroid_runtime.so
I/DEBUG   ( 5063):          #15  pc 0001fbb0  /system/lib/libdvm.so (dvmPlatformInvoke)
I/DEBUG   ( 5063):          #16  pc 0005a734  /system/lib/libdvm.so (_Z16dvmCallJNIMethodPKjP6JValuePK6MethodP6Thread)
```
On the master branch, the android demo is slow, but stable on the S3.

bcamper:
> a bad deallocation for the address `deadbaad`

:metal:💀 ⛔️

bcamper:
Or 💀 🐑 ?

karimnaaji:
Ah no... think we've come up with that in the past which is why we haven't used them to render text, I wonder if there is not a bug in the driver whenever a shared index buffer is used between different VAOs. That would be the only difference with usage of GL resources from master that I would see in this branch. Nothing comes to mind when I try to think what we could be going wrong.

matteblair:
More complete log from another occurrence of the crash: https://pastebin.com/0VL6Se5s

matteblair:
The logs indicate that tangram thinks VAOs are supported here. I wouldn't be shocked if we're hitting a bug in the driver, but we can at least narrow down the cause a bit first. 

I'll try this on other devices and see how isolated the crash is.

matteblair:
More data points:

 - On the Nexus 5 everything is A.O.K.
 - On the Galaxy Note 8.0, point features do not draw at all and logcat is full of: `E/libEGL  ( 8113): called unimplemented OpenGL ES API`. The log also indicates that tangram detected VAOs _not_ supported. 

Now that I'm looking more closely, it doesn't seem like the `useVao` parameter added to `DynamicQuadMesh::draw` is ever passed in - so would we be using VAOs on every device regardless of support?

karimnaaji:
>so would we be using VAOs on every device regardless of support?

Seems like a bad idea. Fixed in af4b1d0.

matteblair:
Yep, that restores point features on the Galaxy Note 8.0. Galaxy S3 still crashes in the same way.

karimnaaji:
Can you try something on this device, to always generate a new index buffer whenever we have a new VAO by commenting this test https://github.com/tangrams/tangram-es/blob/master/core/src/gl/renderState.cpp#L322 should be enough.

matteblair:
Tried your suggestion, same result (with `deadbaad` again, curiously).

matteblair:
This crash wasn't reproducible on another Galaxy S3 running a different OS version, so I'm willing to believe that this is just isolated to the buggy version of Android on our particular handset. If we get reports of similar issues from users then we can investigate more.

Issue #1384, opened 2017-03-17
Library 'libc++_shared.so' not found


fokin33:
On Garmin Monterra :
Message: Cannot load library: link_image[1936]:   203 could not load needed library 'libc++_shared.so' for 'libtangram.so' (load_library[1091]: Library 'libc++_shared.so' not found)
Stacktrace:
	java.lang.Runtime.loadLibrary(Runtime.java:370)
	java.lang.System.loadLibrary(System.java:538)
	com.mapzen.tangram.MapController.<clinit>(MapController.java:962)
	com.mapzen.tangram.MapView.getMapInstance(MapView.java:107)
	com.mapzen.tangram.MapView.getMapAsync(MapView.java:77)
	com.mapzen.android.graphics.MapInitializer.loadMap(MapInitializer.java:90)
	com.mapzen.android.graphics.MapInitializer.loadMap(MapInitializer.java:82)
	com.mapzen.android.graphics.MapInitializer.init(MapInitializer.java:48)
	com.mapzen.android.graphics.MapView.getMapAsync(MapView.java:119)

But that apk work on LG G2 for sample.
Use in gradle:
compile 'com.mapzen:mapzen-android-sdk:1.3.2'

matteblair:
The _Garmin Monterra_, now that's a device we've definitely never tested before. Can you tell me what version of Android this device has?

fokin33:
4.0.3

matteblair:
Looks like we were missing a call to `System.loadLibrary` that's needed for older versions of Android. Should be resolved by #1389

PR #1389, opened 2017-03-17
Fix shared library loading on Android versions < 4.3

matteblair:
Resolves https://github.com/tangrams/tangram-es/issues/1384


tallytalwar:
For me android studio native debugging is breaking with 3.3 gradle.

matteblair:
@tallytalwar I reverted the build tool changes, we can investigate that separately.

tallytalwar:
Thanks @matteblair, I still have to check what in the latest gradle messes up debugging for me on android studio.

Rebased, once CI is done will merge.

Issue #1396, opened 2017-03-24
System locale independent float parsing


hjanetzek:
The common string parsing and printing functions handle float values differently depending on LC_NUMERIC. This breaks our scene parsing and shader generation.

Found the problem testing Tangram on Tizen when system language is set to e.g. German.

google: 'printf locale independent' for 1000+ solutions and workarounds to this issue.
http://stackoverflow.com/questions/272479/how-to-get-equivalent-of-printf-l-on-linux
http://www.netlib.org/fp/

PR #1397, opened 2017-03-24
Locale independent float2string

hjanetzek:
https://github.com/tangrams/tangram-es/issues/1396


hjanetzek:
@matteblair could you give this a look? Not yet complete but allows to load bubble-wrap independent of LC_NUMERIC. Would be good if we could get something like this into the tizen update next week

matteblair:
As I understand it, the bugs we saw were caused by the behavior of the `to_string` functions, not the `stof`/`stod` functions. The fixes to `to_string` don't depend on the V8 conversion functions, right? What is the value of changing `stod` to use these new functions?

hjanetzek:
@matteblair unfortunately it's both. just to_string would be too simple

matteblair:
Oh wait, is the `stod` change necessary for parsing numbers correctly from the scene file? That is, for correctly interpreting `1.000` and `1,000`?

hjanetzek:
yeah..

matteblair:
Now that I've had time to get my head around this problem, I think it makes sense to move ahead with this. I think we can make a few improvements before merging:
 - Choose a more descriptive name for `fixfloat.h`, maybe `stringConversion.h`
 - Use the V8 conversion function for the `to_string` functions (They seem easy to use, and I would guess they're more efficient than iostreams).

platforms/linux/src/main.cpp
```diff
@@ -22,6 +22,10 @@ int main(int argc, char* argv[]) {
             break;
         }
     }
+    char *loc = setlocale(LC_ALL, "de_DE.utf-8");
+    logMsg("LOCALE %s\n", loc);
+    loc = setlocale(LC_ALL, NULL);
+    logMsg("LOCALE %s - %f\n", loc, 435145.342134);
```
matteblair: Ah we should also remove this before we forget about it and confuse ourselves later :)

PR #1400, opened 2017-03-27
Platform memory warnings

### LINKED ISSUE #1393: TGMapViewController shouldn't destroy its GLKView in low memory scenarios
https://github.com/tangrams/tangram-es/blob/master/platforms/ios/src/TangramMap/TGMapViewController.mm#L763

This view is handed over from the system, so destroying it creates a strange problem of how to re-create it effectively within the view hierarchy. It also causes issues with any kind of constraints that may have been attached as now suddenly they are orphaned within the system.

karimnaaji:
Handle platform memory warning signals:
- Delete tile cache
- Release font resources that can be easily reloaded

Note for iOS: Since iOS 6 the view can't be dropped anymore, so we can reliably consider that we will still have a view when the signal is sent to `TGMapViewController`.

Fixes #1393

### DIFF
--- a/core/src/tangram.cpp
+++ b/core/src/tangram.cpp
@@ -901,6 +901,10 @@ void Map::onMemoryWarning() {
     if (tileCache) {
         tileCache->clear();
     }
+
+    if (impl->scene && impl->scene->fontContext()) {
+        impl->scene->fontContext()->releaseFonts();
+    }
 }
 
 void setDebugFlag(DebugFlags _flag, bool _on) {
--- a/core/src/tangram.cpp
+++ b/core/src/tangram.cpp
@@ -895,6 +895,14 @@ void Map::runAsyncTask(std::function<void()> _task) {
     }
 }
 
+void Map::onMemoryWarning() {
+    auto& tileCache = impl->tileManager.getTileCache();
+
+    if (tileCache) {
+        tileCache->clear();
+    }
+}
+
 void setDebugFlag(DebugFlags _flag, bool _on) {
 
     g_flags.set(_flag, _on);
--- a/core/src/tangram.cpp
+++ b/core/src/tangram.cpp
@@ -902,6 +902,10 @@ void Map::onMemoryWarning() {
         tileCache->clear();
     }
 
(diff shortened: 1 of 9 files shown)


tallytalwar:
#1400 fixes #1400 :D. Can you update what issue number this fixes?

core/src/tangram.cpp
```diff
@@ -887,6 +887,18 @@ void Map::runAsyncTask(std::function<void()> _task) {
     }
 }
 
+void Map::onMemoryWarning() {
+    auto& tileCache = impl->tileManager.getTileCache();
```
tallytalwar: Probably data source cache can also be cleared for all loaded datasource?

So:
```c++
for (auto& tileSet : impl->tileManager.getTileSets()) {
    tileSet.source->clearData();
}
```

core/src/tangram.cpp
```diff
@@ -887,6 +887,18 @@ void Map::runAsyncTask(std::function<void()> _task) {
...
+
+    if (tileCache) {
+        tileCache->clear();
+    }
+
+    if (impl->scene && impl->scene->fontContext()) {
+        impl->scene->fontContext()->releaseFonts();
```
tallytalwar: I need to check this but probably you already did, when/how will the fontContext be loaded again post memory warning cleanup? As I see it requires a scene init again, is that how a client app will handle this post memory warning opps are done?

karimnaaji:
Haha yes, updated.

core/src/tangram.cpp
karimnaaji: It will reload/load the font whenever a URI or a font handle loader has been given to the font context, this happens whenever it is needed again from the font context to render a glyph (not if the data was given by a `char*` where the fonts couldn't be loaded again), so no scene init involved in that process. 

You can trigger a memory warning using F3 in this branch to see how it goes through the process.

core/src/tangram.cpp
karimnaaji: Yep, seems like a good call, will look at adding this.

PR #1401, opened 2017-03-27
Fix #1394, #1391, #1370

### LINKED ISSUE #1370: TGMarker error handling
We will need a way to handle errors when adding markers to the map. Maybe tangram can throw an exception when `TGMarker#map` is set for an invalid marker?

cc/ @karimnaaji @msmollin

### LINKED ISSUE #1391: TGMarker.mm setMap attempts to return a boolean false but is marked as VOID
https://github.com/tangrams/tangram-es/blob/master/platforms/ios/src/TangramMap/TGMarker.mm#L193

Seems to be a refactor artifact.

The question is - do we want to re-expose the boolean return? Because `setMap:` is getting synthesized into the `.map` property here https://github.com/tangrams/tangram-es/blob/master/platforms/ios/src/TangramMap/TGMarker.h#L137

Overriding property setters has a requirement of a `void` return, otherwise you need to declare the setter explicitly as a function in the header. Not sure of the correct approach here given we  would like better error codes for some of this stuff per https://github.com/tangrams/tangram-es/issues/1370

### LINKED ISSUE #1394: Marker icon doesnt show if set before map
See the test code below:

```objc
TGMarker* markerPoint = [[TGMarker alloc] init];
markerPoint.stylingString = @"{ style: 'points', color: 'white', size: [52px, 52px], collide: false }";
markerPoint.point = coordinates;
// when set before icon, marker icon shows
//markerPoint.map = view;
markerPoint.icon = [UIImage imageNamed:@"mapzen_logo"];
// when set after icon, marker icon doesnt show
//markerPoint.map = view;
```

karimnaaji:
- Change properties to functions for marker usage to be able to return error status
- Return `NSError*` from this new interface
- Leave placeholder for future enum error code that can be returned through marker usage

Fixes #1394 
Fixes #1391 
Fixes #1370

### DIFF
--- a/platforms/ios/src/TangramMap/TGMarker.h
+++ b/platforms/ios/src/TangramMap/TGMarker.h
@@ -52,23 +52,26 @@ NS_ASSUME_NONNULL_BEGIN
  @param coordinates the longitude and latitude where the marker will be placed
  @param seconds the animation duration given in seconds
  @param ease the ease function to be used between animation timestep
- @return `YES` if this operation was successful, `NO` otherwise
+ @param error an error status pointer that will be assigned if non-nil and an error occured
+ @return Whether the marker coordinates were set succesfully
 
  @note Markers can have their geometry set multiple time with possibly different geometry types.
  */
-- (NSError *)pointEased:(TGGeoPoint)coordinates seconds:(float)seconds easeType:(TGEaseType)ease;
+- (BOOL)pointEased:(TGGeoPoint)coordinates seconds:(float)seconds easeType:(TGEaseType)ease error:(NSError **)error;
 
 /**
  Sets the styling for a marker with a string of YAML defining a 'draw rule'.
 
  See the more detailed scene <a href="https://mapzen.com/documentation/tangram/Styles-Overview/">documentation</a>
  to get more styling informations.
+ 
+ @param styling the styling to set to this marker
+ @param error an error status pointer that will be assigned if non-nil and an error occured
+ @return Whether the marker styling string was set successfully
 
  @note Setting the stylingString will overwrite any previously set stylingString or stylingPath.
-
- @return An error if the marker can't be shown on the map
  */
-- (NSError *)stylingString:(NSString *)styling;
+- (BOOL)stylingString:(NSString *)styling error:(NSError **)error;
 
 /**
  Sets the styling for a marker with a path, delimited by '.' that specifies a 'draw rule' in the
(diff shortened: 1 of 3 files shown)


platforms/ios/src/TangramMap/TGMarker.mm
```diff
@@ -51,99 +62,148 @@ - (instancetype)initWithMapView:(TGMapViewController *)mapView
...
+    if (!tangramInstance->markerSetStylingFromPath(identifier, [path UTF8String])) {
+        return [self createNSError];
+    }
 
-    tangramInstance->markerSetStylingFromPath(identifier, [path UTF8String]);
+    return nil;
 }
 
-- (void)setPoint:(TGGeoPoint)coordinates
+- (NSError *)point:(TGGeoPoint)coordinates
```
sarahlensing: Why did you choose to return an error as opposed to taking it as a parameter and returning a true if the call succeeded? The reason I ask is that this format seems more idiomatic for iOS dev (and will import into swift as a call marked as something that throws)

tallytalwar: Based on the discussion from last week, I think that is what the consensus was, ie. Swift equivalent API should be marked to "throw" an exception.

karimnaaji: Yes I first went what you are suggesting:

```objc
NSError* error;
BOOL success = [TGMarker point:coord error:&error];
if (!success) {
    // Use error
}
````
vs
```objc
NSError* error = [TGMarker point:coord];
if (error) {
   // Use error
}
```
I'm unsure what is the most common pattern here for iOS dev, so it can be changed to what you are  suggesting.

tallytalwar:
So tested the NSError interface. It seems to just error with `domain` as TGMarker. But I think it does make sense to have it return the apt error. Or you were thinking of handling that in a subsequent PR, seeing that `TODO` marked in `createNSError` function.

platforms/ios/src/TangramMap/TGMarker.mm
```diff
@@ -51,99 +62,148 @@ - (instancetype)initWithMapView:(TGMapViewController *)mapView
...
+                           userInfo:userInfo];
+}
+
+- (NSError *)stylingString:(NSString *)styling
 {
     _stylingString = styling;
     _stylingPath = nil;
 
-    if (!tangramInstance || !identifier) { return; }
+    if (!tangramInstance || !identifier) { return nil; }
```
karimnaaji: Unsure, in a context where exception might be thrown out to the user this can become quite annoying, technically this is not an error _yet_ since you don't know whether the styling is valid or not here, but I was also wondering whether something should be thrown here, so I'm open to suggestions.

platforms/ios/src/TangramMap/TGMarker.mm
```diff
@@ -51,99 +62,148 @@ - (instancetype)initWithMapView:(TGMapViewController *)mapView
...
-- (void)setStylingString:(NSString *)styling
+- (NSError *)createNSError
+{
+    NSMutableDictionary* userInfo = [[NSMutableDictionary alloc] init];
+    // TODO: add enum for object error
+    // [userInfo setObject: forKey:@"TGMarker"];
+
+    return [NSError errorWithDomain:@"TGMarker"
+                               code:0
+                           userInfo:userInfo];
```
karimnaaji: Yes, it's just a placeholder for us to leave more information in the future.

platforms/ios/src/TangramMap/TGMarker.mm
tallytalwar: Yeah maybe an exception can be thrown for the user to the API to use it aptly, which is make sure a map instance is set before setting styling on a marker.

karimnaaji:
@tallytalwar it does make sense. We have to handle that TODO when we address https://github.com/tangrams/tangram-es/issues/1370.

platforms/ios/src/TangramMap/TGMarker.mm
sarahlensing: Ah I didn't realize that. What you first had is a more common pattern.

platforms/ios/src/TangramMap/TGMarker.mm
karimnaaji: Ok sounds good, updating for that.

platforms/ios/src/TangramMap/TGMarker.mm
karimnaaji: Updated in https://github.com/tangrams/tangram-es/pull/1401/commits/7ad7f11c9171d53b4dfd3c4d02078e43deed91d7

platforms/ios/src/TangramMap/TGMarker.mm
karimnaaji: Updating in https://github.com/tangrams/tangram-es/pull/1401/commits/7ad7f11c9171d53b4dfd3c4d02078e43deed91d7

platforms/ios/src/TangramMap/TGMarker.h
```diff
@@ -77,52 +80,66 @@ NS_ASSUME_NONNULL_BEGIN
...
 /**
  Sets a marker styled to be a polyline (described in a `TGGeoPolyline`).
 
- @note Markers can have their geometry set multiple time wwith possibly different geometry types.
+ @param polyline the polyline geometry to set to this marker
+ @param error an error status pointer that will be assigned if non-nil and an error occured
+ @return Whether the marker polyline was set successfully
 
- @return An error if the marker can't be shown on the map
+ @note Markers can have their geometry set multiple time wwith possibly different geometry types.
```
tallytalwar: nit pick, spelling `with`.

sarahlensing:
@tallytalwar @tallytalwar Changes LGTM 👍

PR #1402, opened 2017-03-27
Since house styles use api key now

tallytalwar:
We need to specify api key programmatically to load these scenes.


hjanetzek:
@tallytalwar tested with Android?

tallytalwar:
Oops yeah every platform demo apps will need to set the apikey.

Will add.

tallytalwar:
@hjanetzek updated android and rpi demo apps to include the update for api key. Should be good now.

PR #1407, opened 2017-03-28
fix: consistent sorting of label pairs

hjanetzek:
- also on tizen this crashed for optimized builds previously


karimnaaji:
@hjanetzek what was the crash on this code?

hjanetzek:
@karimnaaji I haven't stepped through all the data - I guess the gcc 4.9 sort implementation is picky for strict weak ordering http://stackoverflow.com/questions/16824667/doesnt-stl-sort-require-a-strict-weak-ordering-to-work. The partition subroutine went reproducably out of bounds

karimnaaji:
@hjanetzek Did you actually want a merge commit in this one?

hjanetzek:
@karimnaaji I would say squash'n'merge

Issue #1409, opened 2017-03-28
Add userData to TGMarker


sarahlensing:
https://developers.google.com/maps/documentation/ios-sdk/reference/interface_g_m_s_overlay.html#a900ff2be989386d501ed6c64ea1d7ab3

PR #1414, opened 2017-03-30
Add TGMaker custom user data

karimnaaji:
See https://github.com/mapzen/ios/pull/243#discussion_r108527159 for context.

Fixes https://github.com/tangrams/tangram-es/issues/1409


sarahlensing:
@tallytalwar @karimnaaji These changes look good and satisfy the requirements. 👍  to making it available on android as well!

PR #1416, opened 2017-03-30
Fix marker visible

### LINKED ISSUE #1390: Marker.setVisible(boolean) function does not work properly.
Marker.setVisible(boolean) function doesn't work properly.

I thought the marker would not always be visible or invisible when this function was called. But when i manipulate the map, it doesn't work properly.

When the Visibility false setting is initially set, it works normally. However, if I move the map to the point where there is no marker and then return to first position where the setVisible (false) marker exist again, the marker is shown on map.

And The odd thing is that when zooming again, it looks like it was not visible.

i want your help T_T

tallytalwar:
Fixes #1390

### DIFF
--- a/core/src/labels/labels.cpp
+++ b/core/src/labels/labels.cpp
@@ -50,7 +50,7 @@ void Labels::processLabelUpdate(const ViewState& viewState,
                       viewState.viewportSize.y);
 
     for (auto& label : labelMesh->getLabels()) {
-        if (!drawAll && label->state() == Label::State::dead) {
+        if (!drawAll && (label->state() == Label::State::dead || label->forceInvisible()) ) {
             continue;
         }
 
@@ -65,6 +65,7 @@ void Labels::processLabelUpdate(const ViewState& viewState,
         if (!label->update(mvp, viewState, &bounds, transform)) {
             continue;
         }
+
 
         if (onlyTransitions) {
             if (label->occludedLastFrame()) { label->occlude(); }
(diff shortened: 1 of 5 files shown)


tallytalwar:
I think this is an interesting way of fixing this issue, since how marker relies on the label pipeline. Will be good to have @hjanetzek and @karimnaaji thoughts on this.

Also renamed and documented some conceptual logic in labels update code.

core/src/labels/labels.cpp
```diff
@@ -58,15 +58,17 @@ void Labels::processLabelUpdate(const ViewState& viewState,
...
+        auto bounds = (onlyRender || !label->canOcclude())
             ? screenBounds
             : extendedBounds;
 
         if (!label->update(mvp, viewState, &bounds, transform)) {
             continue;
         }
 
-        if (onlyTransitions) {
+        if (label->forceInvisible()) { label->occlude(); }
```
karimnaaji: I wonder if this wouldn't start a fading event.

What about modifying the test line 53 to:
```
if (!drawAll && (label->state() == Label::State::dead || label->forceInvisible))
```
So the label wouldn't even be put in the collidable set.

tallytalwar: Hmm ... but I still have to mark the label/marker to be occluded right? (occluded by force!)

karimnaaji: Not really if you don't want it to be included in the visible set you wouldn't need to occlude it, what could also work is to simply to put it in the `dead` state when the marker is set to not visible.

tallytalwar: Yeah `dead` state is a real `DEAD` state. Going with your initial suggestion.

tallytalwar:
Updated.

PR #1417, opened 2017-03-30
Refactor tile source subclasses into function namespaces

matteblair:
No change in functionality, just simplifying and cleaning some code.

Instead of creating a different subclass of `TileSource` for each data format, a `TileSource` will now just store its data format and call the appropriate function(s) when it needs to parse tile data.


core/src/scene/sceneLoader.cpp
```diff
@@ -944,29 +941,16 @@ void SceneLoader::loadSource(const std::shared_ptr<Platform>& platform, const st
...
-        mime =  "application/vnd.mapbox-vector-tile";
-    } else if (type == "Raster") {
-        // TODO try to guess from url
-        mime = "image/png";
-    } else {
-        LOGW("Unrecognized data source type '%s', skipping", type.c_str());
-        return;
+    if (auto typeNode = source["type"]) {
+        type = typeNode.Scalar();
     }
```
matteblair: Good point, we effectively default to the `GeoJSON` format if the string isn't recognized, but instead we should print a warning and not add the data source to the scene.

Issue #1388, opened 2017-03-17
Migrate 'visible' layer parameter to 'enabled'


matteblair:
See: https://github.com/tangrams/tangram/pull/523

Tangram-es already has the newly specified behavior implemented. We need only to check the new parameter name, `enabled`, and fall back to checking `visible` if `enabled` is absent (for backwards compatibility).

PR #1420, opened 2017-03-31
Migrate 'visible' layer parameter to 'enabled'

matteblair:
'visible' will still be applied as before if 'enabled' is not present; in a future release 'visible' will be disregarded.

Resolves https://github.com/tangrams/tangram-es/issues/1388


matteblair:
Thanks for updating those tests Varun!

PR #1434, opened 2017-04-04
Add missing NSError creation on polygon marker error

karimnaaji:


platforms/ios/src/TangramMap/TGMarker.mm
```diff
@@ -195,6 +195,7 @@ - (BOOL)polygon:(TGGeoPolygon *)polygon error:(NSError **)error
     auto polygonCoords = reinterpret_cast<Tangram::LngLat*>([polygon coordinates]);
 
     if (!tangramInstance->markerSetPolygon(identifier, polygonCoords, [polygon rings], [polygon ringsCount])) {
+        [self createNSError:error];
```
tallytalwar: ahh nice catch. Missed that during review.

Issue #1435, opened 2017-04-04
Enable 512px tile sources


matteblair:
**Background**

We currently assume that a tile from tiled data sources is intended to be viewed at a scale of 256 logical pixels per side. As the resolution of mobile devices increases, we incur a corresponding increase in the number of URL requests needed to cover a screen-sized area. A data source can reduce this number of requests by offering tiles meant for viewing at 512 logical pixels per side. These 512px tiles would cover the area of 4 256px tiles and would therefore reduce the number of URL requests needed for a typical view by a factor of 4. This would also result in fetching somewhat more off-screen data than with 256px tiles, but this trade-off is often beneficial in practice.

**Implementation**

The logical size of a tile is an attribute of the data source that provided that tile, and this size cannot be reliably inferred from the tile data itself. Therefore the tile size of a data source is probably best configured in the scene file within a `sources` entry. The syntax for this could be simply:
```yaml
sources:
  _myTileSource:
    tile_size: 512 # Defaults to 256 if absent.
```
Or, recognizing that sizes will almost certainly be powers of 2, we could specify an exponent relative to 256 instead:
```yaml
sources:
  _myTileSource:
    tile_scale: 2 # Defaults to 1 if absent, 2 denotes 512px, 3 denotes 1024px, etc.
    # Or maybe 0 denotes 256px, 1 denotes 512px.
```

Within the engine, we'll need additional logic for determining the display size of a tile based on the size specified in its data source configuration. The semantics of a Tile Coordinate will remain the same - that is, a Tile Coordinate {x, y, z} will denote the same geographic area for tiles of any pixel size. This means that the pixel size specified for a tile is effectively a measure of the _level of detail_ present in the tile data. Therefore, for a view covering the Tile Coordinate {x, y, z} using a data source of 512px tiles, we would display the tile {x/2, y/2, z-1} so that we show the same level of detail as we would for 256px tiles.

nvkelso:
@bcamper Is there an issue tracking this on the JS side? I remember you have a branch for this, but don't see it pushed to the server?

nvkelso:
The new (undocumented) endpoint for 512px tiles is:

- `tile.mapzen.com/{user}/{tileset}/v{version}/{tilesize}/{layer}/{z}/{x}/{y}.{format}`

For example:

- `tile.mapzen.com/mapzen/vector/v1/512/all/{z}/{x}/{y}.mvt`

bcamper:
[...]

PR #1436, opened 2017-04-04
512/scaled source tiles

tallytalwar:
#1435


matteblair:
I'm still reviewing the code itself, but I'm noticing that proxy tiles don't seem to be working as intended when zooming out.

tallytalwar:
Hmm, maybe some logic needs to be added to `TileID::getChild` to take tile scale into account.

tallytalwar:
Closing in favor of https://github.com/tangrams/tangram-es/pull/1498

Issue #1365, opened 2017-03-12
Android: *PickListeners not called on main thread


westnordost:
The methods called in the LabelPickListener, MarkerPickListener and FeaturePickListener are not called on the main (UI) thread, but on some GLThread, which is not documented and in my opinion also not desirable behavor since usually a "pick" of a feature, label or marker should do something visual on the screen, just as the callback made on a tap of a button is made on the UI thread.

Notably, the callback is even made on the GLThread when  `mapController.pick*` was called from the UI thread, i.e. the standard case:
```
	@Override public boolean onSingleTapConfirmed(float x, float y)
	{
		// I am on the UI thread
		controller.pickLabel(x,y);
		return true;
	}

	@Override
	public void onLabelPick(LabelPickResult labelPickResult, float positionX, float positionY)
	{
		// I am not on UI thread. This will crash the application.
		Toast.makeText(getContext(), "Crash me", Toast.LENGTH_SHORT).show();
	}
```
Actually, the above example does not even produce any error log on the console like it would normally happen when trying to do something UI on a non-UI-thread. Instead, it simply crashes with no further comment. I guess this has something to do with JNI/native code that can't be helped.

My suggested solution is to post the callback on the main thread by default.

karimnaaji:
Hey @westnordost , 

Thanks, this is a fair suggestion. You're right that there should be consistency with the callee thread and callback thread. For now, the workaround for you would be to call `runOnUIThread` from the listener callback, but this is not convenient for end-user. 

I'll look into this shortly and make sure the UI thread is the one being called from the listener callbacks.

matteblair:
I've been aware of this situation for a while and have avoided changing the behavior because not all callbacks originate from the UI thread or require synchronization with the UI thread. In those cases, we're imposing a small, but unnecessary delay and synchronization cost. The most flexible option, as I see it, is to allow the user to perform any necessary synchronization.

There are some benefits to consistently running callbacks on the UI thread. It's more convenient in the case where the callback performs some View operation, as you mentioned. It also prevents a long callback operation from delaying the rendering loop, which is possible with the current implementation.

Just sharing my two cents here.

matteblair:
This is resolved by #1495 and will be available in the next release.

matteblair:
This is available now in 0.7.0

PR #1437, opened 2017-04-05
Android: add ability to perform UI thread tasks from native code

karimnaaji:
- Add ability to queue Android UI tasks from native code
- Update marker/label/feature pick events to queue their listener event callbacks

TODO:
- [x] Document behavior of pick events
- [x] Apply the same behavior to `sceneUpdateErrorCallback`

See https://github.com/tangrams/tangram-es/issues/1365 for more context.

Fixes https://github.com/tangrams/tangram-es/issues/1365.

### DIFF
--- a/platforms/android/tangram/src/main/cpp/platform_android.cpp
+++ b/platforms/android/tangram/src/main/cpp/platform_android.cpp
@@ -433,7 +433,7 @@ void sceneUpdateErrorCallback(jobject updateCallbackRef, const SceneUpdateError&
     jniEnv->DeleteGlobalRef(updateCallbackRef);
 }
 
-void labelPickCallback(AndroidPlatform& platform, jobject listenerRef, const Tangram::LabelPickResult* labelPickResult) {
+void labelPickCallback(AndroidPlatform& platform, jobject listenerRef, const LabelPickResult* labelPickResult) {
 
     JniThreadBinding jniEnv(jvm);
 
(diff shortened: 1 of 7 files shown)


matteblair:
This seems to work as intended, and I now see that implementing this feature is way more complicated than I originally imagined!

Another approach that I imagined while reviewing this: 
 - When async operation requests are made, they are represented as Java objects and put into a queue in MapController.
 - During MapController$onDrawFrame, those query objects are all sent to the native Map.
 - In the native update and render calls, the queries are processed and results are produced, populating result values in the Java request objects.
 - At the end of MapController$onDraw, after the requests are known to be completed, we post the results to the UI thread using the convenient runOnUiThread method.

This would also require substantial refactoring, but I think it could end up being a simpler implementation.

platforms/android/tangram/src/main/cpp/platform_android.cpp
```diff
@@ -433,12 +499,17 @@ void markerPickCallback(jobject listener, jobject tangramInstance, const Tangram
         }
     }
 
-    jniEnv->CallVoidMethod(listener, onMarkerPickMID, markerPickResultObject, position[0], position[1]);
-    jniEnv->DeleteGlobalRef(listener);
-    jniEnv->DeleteGlobalRef(tangramInstance);
+    jobject markerPickResultRef = jniEnv->NewGlobalRef(markerPickResultObject);
```
tallytalwar: This should also be inside the `if (markerPickResult)` scope.

karimnaaji:
@matteblair This seems like an alternative. I've made a more specialized implementation based on your idea here https://github.com/tangrams/tangram-es/commit/9bbf021d4ca8ded1262301edb206bdc0d03c35f3 where synchronization is made for selection queries on the Java side and then posted to the UI thread as expected. If the code looks good I'll make another PR with this commit that supersedes this one.
I would still leave the code of this PR in a branch since it seems possible to need a this kind of `queueUITask ` helper from native code for other purposes than selection queries in the future.

matteblair:
@karimnaaji Awesome, I'll take a look at that branch. I agree with your point about a more general UI-thread-task construct, we may find ourselves wanting something like that in the future.

matteblair:
@karimnaaji See how you like #1495 as well

karimnaaji:
Closing in favor of #1495.

PR #1439, opened 2017-04-06
Force iOS to use system sqlite3 instead of building it

matteblair:
Maybe we have to set the value in the CMake cache?


matteblair:
Yep, that seems to do the trick.

screen shot 2017-04-06 at 3 33 27 pm

Issue #1440, opened 2017-04-07
Don't create errors when no map set


sarahlensing:
I'm sorry that this is contrary to what i originally advocated for here: https://github.com/tangrams/tangram-es/pull/1401#discussion_r108501077

In practice it feels wrong to throw an error when the user sets the geometry or styling before the map. Can we update the implementation to just return NO if there is no map set? And then when the map is set, if the styling string or any other properties are invalid we can throw an error at that point?

msmollin:
@sarahlensing brought me up to speed on this a little. I took a look at the code in #1441 and I had a bit of a thought there.

Let's take for example `- (BOOL)stylingPath:(NSString *)path error:(NSError **)error`- but this applies to all these types of setters in `TGMarker`.

I feel like the BOOL return and the NSError are overlapping in their use case. The code is returning "false" in the event the map isn't set or the identifier is set to zero, which happens when the marker has either not been added or been removed. It can *also* return false if there is an actual error (like an invalid yaml string for styling). 

Going through the use cases in my head, I come up with three scenarios following the above logic:

1) That's intentional usage. The developer doesn't want the user to see anything change styling so they either don't add it before setting styling or they remove it to redo the style / reset the polyline.

2) The developer forgot to set the map, which is an implementation error.

3) The developer actually passed in something incorrect, which is a bug.

For number 1: returning a BOOL really doesn't matter. If they're not setting the map intentionally (or they removed it from the map), they're going to ignore the `false` return and just set it later (which is what #1441 is allowing them to do). 

For number 2: they missed a step in implementation. The code could return an error like `MapNotSet`, but per #1441 we don't consider it a blocking error. 

For number 3: they should be checking the `NSError` for actual errors. Its existence is inherently a boolean success or failure state, except it also contains additional information about what happened.

So my thought is we should just remove the `BOOL` returns altogether and use the `NSError` objects to indicate success or failure. That would simplify the API significantly.

I welcome comments though, because honestly there is probably a use case here I missed.

karimnaaji:
>So my thought is we should just remove the BOOL returns altogether and use the NSError objects to indicate success or failure. That would simplify the API significantly.

There might be a little bit of confusion in here. You shouldn't see the boolean return as a indicator of an _error_ state, but whether the marker has been successfully constructed in order to be _visible_ on the map. A marker is also perfectly valid to live without a map on it.

tallytalwar:
[...]

PR #1441, opened 2017-04-07
Fix 1440

karimnaaji:
Following a discussion with @sarahlensing 

See #1440 for more details and original thread here https://github.com/tangrams/tangram-es/pull/1401#discussion_r108501077


msmollin:
@karimnaaji I put some thoughts in 1440 for additional thoughts on the changes here. I think we can simplify this API further.

Issue #1442, opened 2017-04-07
[Discussion] Rethink TGMarker <-> TangramMap interaction model.


msmollin:
While writing my response to #1440 I started thinking about the interaction model between TGMarker and the map that displays it. Rather than pollute that ticket further I wanted to open a larger discussion which I think will lead to rearchitecting that dependency model a bit.

Markers really have very little to do with a map, when you think about it. Markers are just a set of metadata, location, and styling information about something on the map. That could be a device's current location, or a POI, or a tappable highway shield, etc. The map takes that information, and renders it for display (same as with vector tiles). 

Ultimately this means the coupling between our existing TGMarker and TangramMap does little more than complicate the interaction model between the two. We should mimic the interface off of the model "A map displays a marker".

There are 2 possible developer workflows currently for TGMarkers being added to a map:

1) Developer creates instance of `TGMarker`
2) Developer sets the map instance that owns the `TGMarker` on the marker itself
3) Developer sets the rest of the options which at some point in the future take effect.

**OR**

1) Developer creates instance of `TGMarker`
2) Developer sets the options they want.
3) Developers sets the map instance that owns the `TGMarker` with an implicit "opt-in" that by setting this instance, they would like the marker to be displayed on the map.

There's a logical separation here which didn't really showcase itself in the original PR, but has showed up on usage in the Mapzen iOS SDK. The API as it stands now is not explicit with the intent to display a marker. A developer sets the map instance (https://github.com/tangrams/tangram-es/blob/427b70422640343410c7c395708001814df9a465/platforms/ios/src/TangramMap/TGMarker.mm#L269) and through a series of "side effects" from setting that map instance, the marker may or may not get displayed.

What I suggest is we flip the interaction model such that rather than setting TGMapViewController instances on one or more TGMarkers, we tell TGMapViewController to display one or more TGMarkers. 

I think this has a number of benefits:
1) We condense down to one workflow which mimics the second original workflow in its steps, but instead of setting a map instance on the marker, the developer explicitly tells the map to render and display their marker(s). 
2) From a performance perspective, we're only going to render once the developer is ready to render that data. This avoids some unnecessary rendering churn that could occur from the original first workflow.
3) We can rather drastically simplify the internal implementation of TGMarker by centralizing the error management code inside TGMapViewController and just having the marker return those errors like currently implemented: only if it's been added to a map.

[...]

PR #1448, opened 2017-04-10
Prototype for #1442

karimnaaji:
@msmollin https://github.com/tangrams/tangram-es/issues/1442 description is quite confusing to me for most part, but here is another proposal, so let's discuss around a piece of code instead.

- We decided of returning `BOOL` instead of a `void` (based on initially referenced documentation https://developer.apple.com/library/content/documentation/Swift/Conceptual/BuildingCocoaApps/AdoptingCocoaDesignPatterns.html#//apple_ref/doc/uid/TP40014216-CH7-ID100) but I can see how it is redundant.
- Concerning the initial approach, we designed it to be closer to the Google's api https://developers.google.com/maps/documentation/ios-sdk/reference/interface_g_m_s_overlay#ae1a9e4e14105b4fc5c284865c9cbf5ee but here is a different approach.

### DIFF
--- a/platforms/ios/demo/src/MapViewController.m
+++ b/platforms/ios/demo/src/MapViewController.m
@@ -130,11 +130,8 @@
     // Add polygon marker
     {
         if (!vc.markerPolygon) {
-            vc.markerPolygon = [[TGMarker alloc] init];
+            vc.markerPolygon = [view markerAdd];
             [vc.markerPolygon stylingString:@"{ style: 'polygons', color: 'blue', order: 500 }" error:nil];
-
-            // Add the marker to the current view
-            [vc.markerPolygon map:view error:nil];
         }
         static TGGeoPolygon* polygon = nil;
         if (!polygon) { polygon = [[TGGeoPolygon alloc] init]; }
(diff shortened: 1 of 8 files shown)


msmollin:
Sorry for the confusion - I basically was advocating for an interface that was more declarative. I'll take a look at this in a bit.

tallytalwar:
>We decided of returning BOOL instead of a void (based on initially referenced documentation https://developer.apple.com/library/content/documentation/Swift/Conceptual/BuildingCocoaApps/AdoptingCocoaDesignPatterns.html#//apple_ref/doc/uid/TP40014216-CH7-ID100) but I can see how it is redundant.

I do not think this is redundant now. After reading what @sarahlensing posted in another issue (https://developer.apple.com/library/content/documentation/Cocoa/Conceptual/ProgrammingWithObjectiveC/ErrorHandling/ErrorHandling.html), it make sense to have the return value of a Boolean, 

>`When dealing with errors passed by reference, it’s important to test the return value of the method to see whether an error occurred, as shown above. Don’t just test to see whether the error pointer was set to point to an error.`

It looks like `Bool` return is meant to be redundant.

PR #1459, opened 2017-04-17
Adds support for providing subdomains for source urls

### LINKED ISSUE #1386: Tile sources should allow naming a set of url subdomains
See: https://github.com/tangrams/tangram/issues/445

Tiled data sources should allow specifying a set of _subdomains_ to use for requests by using `{s}` as a placeholder in the URL template and a new `url_subdomains` parameter:
```yaml
url: https://{s}.some-tile-server.com/{z}/{y}/{x}.pbf
url_subdomains: [a, b, c]
```
Behavior:
 - If `url_subdomains` contains a sequence of _N_ > 0 scalar values, then:
    - Request number _i_ made by the associated data source will replace `{s}` in the URL template with the (_i_ % _N_)'th value from `url_subdomains`.
 - Otherwise, `{s}` will not be replaced in the URL template.

Implementation:
 - It may behoove us to create a data structure representing a URL template, including subdomains. This structure could store the offsets of the placeholders to replace, rather than searching the string repeatedly.

tallytalwar:
- Tiles for the source will be fetched in a round robin manner to cover all subdomains

Closes #1386

### DIFF
--- a/core/src/data/networkDataSource.cpp
+++ b/core/src/data/networkDataSource.cpp
@@ -7,13 +7,14 @@
 
 namespace Tangram {
 
-NetworkDataSource::NetworkDataSource(std::shared_ptr<Platform> _platform, const std::string& _urlTemplate) :
+NetworkDataSource::NetworkDataSource(std::shared_ptr<Platform> _platform,
+        std::vector<std::string>&& _urlTemplates) :
     m_platform(_platform),
-    m_urlTemplate(_urlTemplate),
+    m_urlTemplates(std::move(_urlTemplates)),
     m_maxDownloads(MAX_DOWNLOADS) {}
 
-void NetworkDataSource::constructURL(const TileID& _tileCoord, std::string& _url) const {
-    _url.assign(m_urlTemplate);
+void NetworkDataSource::constructURL(const TileID& _tileCoord, std::string& _url, int32_t index) const {
+    _url.assign(m_urlTemplates[index]);
 
     try {
         size_t xpos = _url.find("{x}");
@@ -48,7 +49,8 @@ bool NetworkDataSource::loadTileData(std::shared_ptr<TileTask> _task, TileTaskCb
         m_pending.push_back(tileId);
     }
 
-    std::string url(constructURL(_task->tileId()));
+    m_urlIndex = (++m_urlIndex) % m_urlTemplates.size();
+    std::string url(constructURL(_task->tileId(), m_urlIndex));
 
     bool started = m_platform->startUrlRequest(url,
         [this, cb = _cb, task = _task](std::vector<char>&& _rawData) mutable {
@@ -89,7 +91,10 @@ void NetworkDataSource::removePending(const TileID& _tileId) {
 
 void NetworkDataSource::cancelLoadingTile(const TileID& _tileId) {
     removePending(_tileId);
-    m_platform->cancelUrlRequest(constructURL(_tileId));
+    // cancel all possible requests for this tile
+    for (size_t index = 0; index < m_urlTemplates.size(); index++) {
+        m_platform->cancelUrlRequest(constructURL(_tileId, index));
+    }
 }
 
 }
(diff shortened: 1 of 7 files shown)


core/src/scene/sceneLoader.cpp
```diff
@@ -990,6 +992,24 @@ void SceneLoader::loadSource(const std::shared_ptr<Platform>& platform, const st
...
+        urls.push_back(url);
+        urls.back().replace(loc, 3, subD);
+    };
+
+    if (Node subDomainNode = source["url_subdomains"]) {
+        auto loc = url.find("{s}");
+        if (loc != std::string::npos && subDomainNode.IsSequence()) {
+            for (const auto& domain : subDomainNode) {
+                if (domain.IsScalar()) {
+                    constructUrls(loc, domain.Scalar());
```
matteblair: This seems more complicated than storing the list of subdomains and substituting them when we build the request. I find it especially confusing that subdomain substitution is handled to differently from tile coordinate substitution even though they do almost exactly the same thing.

matteblair: Also I don't think this logic is correct when `url_subdomains` is specified but the URL contains no `{s}` (the URL should be used without substitution).

core/src/scene/sceneLoader.cpp
```diff
@@ -1011,9 +1031,9 @@ void SceneLoader::loadSource(const std::shared_ptr<Platform>& platform, const st
         // If we have MBTiles, we know the source is tiled.
         tiled = true;
         // Create an MBTiles data source from the file at the url and add it to the source chain.
-        rawSources->setNext(std::make_unique<MBTilesDataSource>(platform, name, url, ""));
+        rawSources->setNext(std::make_unique<MBTilesDataSource>(platform, name, std::move(urls), ""));
```
matteblair: `MBTilesDataSource` should not use URL subdomains, since it makes (at most) one network request. We can note this in the relevant documentation as well.

core/src/data/networkDataSource.cpp
```diff
@@ -89,7 +91,10 @@ void NetworkDataSource::removePending(const TileID& _tileId) {
 
 void NetworkDataSource::cancelLoadingTile(const TileID& _tileId) {
     removePending(_tileId);
-    m_platform->cancelUrlRequest(constructURL(_tileId));
+    // cancel all possible requests for this tile
```
matteblair: Good catch - this is one more limitation of the current URL request system and I'll be thrilled to update it.

core/src/scene/sceneLoader.cpp
matteblair: Ah nevermind my second comment, I think this works.

tallytalwar: Ahh .. I just pushed a fix for the above (second comment)

With respect to subdomain substitution being similar to tile id substitution, I did think of doing it in the same way, but subdomains set is a **constant** **small** set, which does not need to be substituted in every request (which is obviously not true for tile ids).

core/src/scene/sceneLoader.cpp
tallytalwar: Ohh Yes. I wanted to ask about mbtiles and document it.

tallytalwar: Reverted for mbtiles datasources.

core/src/scene/sceneLoader.cpp
matteblair: Oh good, I was questioning my eyesight for a second there haha

True, subdomains aren't exactly the same as tile coordinates. To me it still seems simpler to substitute at the time of a request if we have a list of subdomains - otherwise we're doing more work for all the requests that won't use subdomains (which is most).

tallytalwar: Its not doing too much work irrespective. It feels you are strong on this so will change it. Stay tuned.

tallytalwar: Updated!

tallytalwar:
Cool. Looks like I commented on an earlier commit hash.. so those comments are lost:).

PR #1466, opened 2017-04-20
Use properties on TGMarker

karimnaaji:
- Defer marker errors to another api call `getError` that consume the error on the marker
- Use of properties instead of functions for setters of TGMarker
- Don't allow invalid ids on markers except on marker removal (based on #1448)


platforms/ios/src/TangramMap/TGMarker.mm
```diff
@@ -7,244 +7,145 @@
...
+    self.point = {NAN, NAN};
 }
 
-- (void)createNSError:(NSError **)error
+- (void)createNSError
 {
-    if (!error) {
-        return;
-    }
+    // TODO: Check for all Tangram internal tangram error and
```
tallytalwar: I am not sure I understand the "Check for all Tangram internal error" here. Shouldn't `createNSError` receive some context of the error from the callee to be used to get this additional info?

Or if tangram core is maintaining a stack or error, does this grab the last error and use it to populate the additional info?

platforms/ios/src/TangramMap/TGTypes.mm
```diff
@@ -0,0 +1,11 @@
...
+//  TGTypes.mm
+//  TangramMap
+//
+//  Created by Karim Naaji on 4/20/17.
+//  Copyright (c) 2017 Mapzen. All rights reserved.
+//
+
+#import "TGTypes.h"
+
+NSString* const TGErrorDomain = @"TGErrorDomain";
```
tallytalwar: Can this "static" "const" string not be defined in the header?

karimnaaji: In general I like to leave those kind of global in a single compilation unit. Otherwise each includer would define its own instance of this variable.

platforms/ios/src/TangramMap/TGMarker.mm
karimnaaji: Since the informations given by the string itself and the marker API gives either success or failure, the error string here is quite simple. This string could be given more description details around the kind of error that actually happened if the core marker API evolves.

What is possible in this implementation is appending errors in the string as they come, to give more context and change the interface by adding the callee name:
```
[self createNSError:"setStylingString"];
```
But this wouldn't give you much more informations if you check errors after the API calls of interest.

tallytalwar: K make sense for the current marker interface implementation. I was mainly asking about when we have a generic tangram error handling interface.

tallytalwar:
Good to go when docs are added. @msmollin and @sarahlensing can you give this a look before we merge this, since you will be using this error handling interface soon :D.

platforms/ios/src/TangramMap/TGTypes.mm
msmollin: I think that problem can be resolved by an extern version of this in the header, and then defining the string constant inside the implementation. Thoughts?

msmollin: Oh duh I see. That's exactly what you did. Nevermind. Haha :)

msmollin:
Also good to go once docs are added. Can we get a test build version of this put into the iOS framework repo once it merges?

karimnaaji:
Yep, we will provide a test version on the `ios-framework` repo once this is merged!

PR #1467, opened 2017-04-20
Remove markers on scene reload

### LINKED ISSUE #1464: Marker crash after new scene is loaded
When we load a new scene, we rebuild all the markers against that new scene and we end up with two cases:
- a marker is attached to a style that exists in both the previously loaded and new scene, 
- the marker doesn't find a new style to be built against

Inconsistency can happen in both those cases; a style can be found with the same name but have  different resources, where we result in a bad access or, the marker wouldn't show if a style was available in the previous scene but not in the new one.

To resolve this inconsistency, the best solution seems to drop all markers whenever a scene gets reloaded, since those are in fact tightly coupled with the scene itself when they get created.

karimnaaji:
Fixes #1464 

Needs rebase after #1466 is merged to properly be used on iOS

### DIFF
--- a/core/src/marker/markerManager.cpp
+++ b/core/src/marker/markerManager.cpp
@@ -33,7 +33,7 @@ void MarkerManager::setScene(std::shared_ptr<Scene> scene) {
         m_styleBuilders[style->getName()] = style->createBuilder();
     }
 
-    rebuildAll();
+    removeAll();
 }
 
 MarkerID MarkerManager::add() {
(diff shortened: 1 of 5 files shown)


tallytalwar:
The demo app is crashing on adding a point marker. Debugging what is causing this.

tallytalwar:
Ok fixes the crash with  2d7e36d, but still one more issue here. 

Another issue here:
Not all markers are being removed on loading a new scene, however their references are partially deleted somewhere in the pipeline, causing a crash when the one tries to access old markers. More tomorrow.

tallytalwar:
@karimnaaji updated with a fix (for an issue of infinite `clearGeometry` called, currently present in master). Good to go.

PR #1468, opened 2017-04-20
Android scoped global ref

karimnaaji:
- Use ScopedGlobalRef to wrap creation and deletion of JNI global references

Needs rebase after #1437 is merged


karimnaaji:
Closing for now since we took a different approach to redirect UI tasks from the Android interface.

Issue #1460, opened 2017-04-18
Native crash at /system/vendor/lib/egl/libGLESv2_adreno.so


westnordost:
I have so many users now, StreetComplete is becoming a stress test for tangram-es ;-)

So, here is another crash report I get via the Android developer console. It happened on the following devices and Android versions so far:

Android 5.0, 5.1
Devices: OnePlus One, Galaxy Note3, Galaxy S4

One user added the following comment:
> "The app crashed after I turned off the smartphone with the power button and then turned it on again."

```
*** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
Build fingerprint: 'samsung/jfltexx/jflte:5.0.1/LRX22C/I9505XXUHPK2:user/release-keys'
Revision: '11'
ABI: 'arm'
pid: 18988, tid: 19025, name: RenderThread >>> de.westnordost.streetcomplete <<<
signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0xe
r0 00000000 r1 00004001 r2 00000000 r3 a0a181e8
r4 a0a181e8 r5 0000000e r6 00000078 r7 a0a41000
r8 00000000 r9 00000005 sl 9850c980 fp a0a404c0
ip ad5ddea8 sp a0cd3788 lr b6f65293 pc a950fa40 cpsr 600d0030

backtrace:
#00 pc 00080a40 /system/vendor/lib/egl/libGLESv2_adreno.so (cpumempool_malloc+67)
#01 pc 0008c1f1 /system/vendor/lib/egl/libGLESv2_adreno.so (rb_vbo_alloc+16)
#02 pc 00050e49 /system/vendor/lib/egl/libGLESv2_adreno.so (cache_temp_vbo_elements+12)
#03 pc 00054885 /system/vendor/lib/egl/libGLESv2_adreno.so (core_glDrawElementsInstancedXXX+440)
#04 pc 000549b1 /system/vendor/lib/egl/libGLESv2_adreno.so (core_glDrawElements+10)
#05 pc 00043ca3 /system/vendor/lib/egl/libGLESv2_adreno.so (glDrawElements+28)
#06 pc 000283d1 /system/lib/libhwui.so
#07 pc 000283fb /system/lib/libhwui.so
#08 pc 000285c1 /system/lib/libhwui.so
#09 pc 000322ff /system/lib/libhwui.so
#10 pc 0001caf9 /system/lib/libhwui.so
[...]

PR #1469, opened 2017-04-21
Disable VAOS for DynamicMesh

hjanetzek:
- some drivers crash when using vaos together with dynamic vbos
   https://github.com/westnordost/StreetComplete/issues/130

The flickering effect before crashing makes it likely that it happens with drivers that store the absolute glVertexPointer address in VAO instead of determining the address from vbo address + byteoffset. So most likely a driver bug..

Should also fix https://github.com/tangrams/tangram-es/issues/1460


matteblair:
This was merged into `master` as part of the 0.6.2 patch (c4b4bc5bb5c6a19ab5bae78e19712a27796fa3e3)

Issue #1452, opened 2017-04-11
Remove API keys and document Mapzen key requirements


matteblair:
The demo applications and scenes in this repository have historically used a single Mapzen API key designated for test use of vector tiles. Our intent with this key was that users could check out the code for the demo applications and run them without any modifications, then get their own key when/if they decided to continue developing an application. 

There are several reasons that it's important for developers to use their own API key:
 - Better quality of service! If one application starts abusing the vector tiles API using the demo key, we have to shut off service for that key. Any other innocent applications using that key will also have their service shut off without warning.
 - More information for developers! Using your own API key, you can track the usage of Mapzen APIs in your own application with the Mapzen Developer dashboard.
 - More information for Mapzen! We are always trying to figure out what our API users need more of, and with individual user keys this job is much easier.

However, we've observed that a large number of developers have ended up copy-pasting our demo API key into their applications, rather than obtaining and using their own key. Here are some incremental steps we can take to prevent the accidental or intentional usage of our demo API key in other applications:

1. Improve documentation of key requirements: We should document that the Mapzen services used in our demos require an API key, and link to the relevant Mapzen pages.
2. Refresh the demo API key: We'll replace the API key in our demo applications with a new key and shut off service to the old one. Any applications using the old demo key will then be forced to obtain a new key if they want to continue using the vector tiles service.
3. Remove API keys from code: We can remove all usage of API keys from the repository and rely instead on environment variables or other local configuration files to provide keys. This will prevent any future users from accidentally using an API key from the demo code.

**Progress:**
- [x] Remove API keys from Android demo
- [x] Remove API keys from iOS demo
- [x] Remove API keys from desktop demos

tallytalwar:
Some of the work being done here: https://github.com/tangrams/tangram-es/pull/1466 (cc. @karimnaaji)

matteblair:
cc @msmollin @karimnaaji What do we think is the easiest way to pull environment variables like API keys into an iOS project?

karimnaaji:
I think we can safely add it into the plist and query the value from the sample app code if that makes sense with you @msmollin .
By default we would leave the spot empty and mention that the user would need to fill it from the documentation.

matteblair:
But don't we track the plist in git? It might get annoying to add and remove a local API key from the plist when one wants to do git operations, especially since plist files can't be nicely diff'ed.

tallytalwar:
[...]

PR #1470, opened 2017-04-22
Update demo app to include style switcher/setter

### LINKED ISSUE #1362: Android demo app should allow easily loading custom scenes from GUI
This could be as simple as a text field to enter a url for an arbitrary scene file. A person with an Android phone should be able to download this demo from android.mapzen.com and put in their own scene file to quickly evaluate it.

@nvkelso @sensescape The goal of this is to make it easier for you to test scene files with tangram-es. Do you have other ideas of how we could facilitate this? We could probably do something similar with iOS, or distribute an OS X tangram.app that you could just drag-and-drop scene files onto.

tallytalwar:
- Choose between CDN bundled scene files
- Or specify a URL for a scene to be loaded

A custom ArrayAdapter was required to specify a custom filter to include a string outside of the initial filter list.

Closes #1362

### DIFF
--- a/platforms/android/demo/build.gradle
+++ b/platforms/android/demo/build.gradle
@@ -25,5 +25,5 @@ android {
 
 dependencies {
   compile project(path: ':tangram')
-    compile 'com.android.support:design:25.3.1'
+  compile 'com.android.support:design:25.3.1'
 }
--- a/platforms/android/demo/build.gradle
+++ b/platforms/android/demo/build.gradle
@@ -25,4 +25,5 @@ android {
 
 dependencies {
   compile project(path: ':tangram')
+    compile 'com.android.support:design:25.3.1'
 }
--- a/platforms/android/demo/src/main/java/com/mapzen/tangram/android/MainActivity.java
+++ b/platforms/android/demo/src/main/java/com/mapzen/tangram/android/MainActivity.java
@@ -8,6 +8,7 @@ import android.support.v7.widget.Toolbar;
 import android.util.Log;
 import android.view.View;
 import android.view.Window;
+import android.view.inputmethod.InputMethodManager;
 import android.widget.AdapterView;
 import android.widget.ArrayAdapter;
 import android.widget.Filter;
@@ -47,6 +48,9 @@ import java.util.Map;
 public class MainActivity extends AppCompatActivity implements OnMapReadyCallback, TapResponder,
         DoubleTapResponder, LongPressResponder, FeaturePickListener, LabelPickListener, MarkerPickListener, SceneUpdateErrorListener,
         AdapterView.OnItemSelectedListener, AdapterView.OnItemClickListener {
+
+    private final String apiKey = "vector-tiles-tyHL4AY";
+    private ArrayList<SceneUpdate> sceneUpdates = new ArrayList<>();
 
     /* Override ArrayAdaptor to include any text */
     private class CustomArrayAdapter extends ArrayAdapter {
@@ -119,7 +123,7 @@ public class MainActivity extends AppCompatActivity implements OnMapReadyCallbac
         "https://mapzen.com/carto/zinc-style/zinc-style.zip"
     };
 
-    AutoCompleteTextView sceneString;
+    AutoCompleteTextView autoCompleteView;
     CustomArrayAdapter customAdapter;
     Spinner spinner;
 
@@ -130,8 +134,6 @@ public class MainActivity extends AppCompatActivity implements OnMapReadyCallbac
 
     @Override
     public void onCreate(Bundle savedInstanceState) {
-        final ArrayList<SceneUpdate> sceneUpdates = new ArrayList<>(1);
-        final String apiKey = "vector-tiles-tyHL4AY";
         sceneUpdates.add(new SceneUpdate("global.sdk_mapzen_api_key", apiKey));
         super.onCreate(savedInstanceState);
 
@@ -140,12 +142,12 @@ public class MainActivity extends AppCompatActivity implements OnMapReadyCallbac
         setContentView(R.layout.main);
 
         /* setup autocompletetextview adapter */
-        sceneString = (AutoCompleteTextView)findViewById(R.id.autoCompleteTextView);
+        autoCompleteView = (AutoCompleteTextView)findViewById(R.id.autoCompleteTextView);
         List<String> lst = new ArrayList<String>(Arrays.asList(scenes));
         customAdapter = new CustomArrayAdapter(this, android.R.layout.simple_list_item_1, lst);
-        sceneString.setAdapter(customAdapter);
-        sceneString.setThreshold(1);
-        sceneString.setOnItemClickListener(this);
+        autoCompleteView.setAdapter(customAdapter);
+        autoCompleteView.setThreshold(1);
+        autoCompleteView.setOnItemClickListener(this);
 
         /* setup spinner style selecter */
         Toolbar toolbar = (Toolbar) findViewById(R.id.toolbar);
@@ -161,7 +163,7 @@ public class MainActivity extends AppCompatActivity implements OnMapReadyCallbac
         /* setup map view */
         view = (MapView)findViewById(R.id.map);
         view.onCreate(savedInstanceState);
-        sceneString.setText(scenes[0]);
+        autoCompleteView.setText(scenes[0]);
         view.getMapAsync(this, "scene.yaml", sceneUpdates);
     }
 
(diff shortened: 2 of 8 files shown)


platforms/android/demo/src/main/java/com/mapzen/tangram/android/MainActivity.java
```diff
@@ -88,6 +189,63 @@ public void onLowMemory() {
         view.onLowMemory();
     }
 
+    @Override public void onItemClick(AdapterView<?> parent, View view, int position, long id) {
+        final ArrayList<SceneUpdate> sceneUpdates = new ArrayList<>(1);
+        final String apiKey = "vector-tiles-tyHL4AY";
```
sarahlensing: remove key and replace with value that is read from gradle.properties?

1. add to `~/.gradle/gradle.properties`:
```
tangramDemoApiKey=the_api_key
```

2. add to demo app `build.gradle`:
```
def DEMO_API_KEY = '"' + tangramDemoApiKey + '"';

buildTypes {
    debug {
      debuggable true
      jniDebuggable true
      buildConfigField 'String', 'API_KEY', DEMO_API_KEY
    }
  }
```

3. update this line to: `final String apiKey = BuildConfig.API_KEY;`

platforms/android/demo/src/main/java/com/mapzen/tangram/android/MainActivity.java
```diff
@@ -88,6 +189,63 @@ public void onLowMemory() {
...
+    }
+
+    @Override public void onItemSelected(AdapterView<?> parent, View view, int position, long id) {
+        if (map == null) { return; }
+
+        final ArrayList<SceneUpdate> sceneUpdates = new ArrayList<>(1);
+        final String apiKey = "vector-tiles-tyHL4AY";
+        sceneUpdates.add(new SceneUpdate("global.sdk_mapzen_api_key", apiKey));
+
+        switch (position) {
```
sarahlensing: since you do the same thing for all index values < scenes.length, you could refactor this to something like:

```
if (position < scenes.length) {
   sceneString.setText(scenes[position]);
   map.loadSceneFile(scenes[position], sceneUpdates);
} else {
   sceneString.setText(null);
   sceneString.requestFocus();
   InputMethodManager in = (InputMethodManager) getSystemService(Context.INPUT_METHOD_SERVICE);
   in.showSoftInput(sceneString, 0);
}
```

the else clause adds text clearing and shows the keyboard so the user can enter the custom scene location (if you don't want this behavior feel free to ignore though :) )

platforms/android/demo/src/main/java/com/mapzen/tangram/android/MainActivity.java
```diff
@@ -88,6 +189,63 @@ public void onLowMemory() {
         view.onLowMemory();
     }
 
+    @Override public void onItemClick(AdapterView<?> parent, View view, int position, long id) {
+        final ArrayList<SceneUpdate> sceneUpdates = new ArrayList<>(1);
+        final String apiKey = "vector-tiles-tyHL4AY";
+        sceneUpdates.add(new SceneUpdate("global.sdk_mapzen_api_key", apiKey));
+        spinner.setSelection(7);
+        map.loadSceneFile(sceneString.getText().toString(),sceneUpdates);
```
sarahlensing: if you replace `onItemClick` with something like this:

```
int selectedIndex = 7; // custom
String sceneText = sceneString.getText().toString();
for (int i = 0; i < scenes.length; i++) {
    if (sceneText.equals(scenes[i])) {
        selectedIndex = i;
        break;
    }
}
spinner.setSelection(selectedIndex);
        
sceneString.clearFocus();
InputMethodManager in = (InputMethodManager) getSystemService(
Context.INPUT_METHOD_SERVICE);
in.hideSoftInputFromWindow(sceneString.getWindowToken(), 0);
```

it will add some polish by dismissing the keyboard on selection and detect the corresponding spinner item selected and update it to be the selected spinner item (right now if you click a style within the scenes list it wont update the spinner).

platforms/android/demo/src/main/java/com/mapzen/tangram/android/MainActivity.java
tallytalwar: This is good.

We have an issue to track the work for the same for all platform demos. https://github.com/tangrams/tangram-es/issues/1452

platforms/android/demo/src/main/java/com/mapzen/tangram/android/MainActivity.java
```diff
@@ -31,18 +39,90 @@
...
+        "scene.yaml",
+        "https://mapzen.com/carto/bubble-wrap-style/bubble-wrap-style.zip",
+        "https://mapzen.com/carto/refill-style/refill-style.zip",
+        "https://mapzen.com/carto/walkabout-style/walkabout-style.zip",
+        "https://mapzen.com/carto/tron-style/tron-style.zip",
+        "https://mapzen.com/carto/cinnabar-style/cinnabar-style.zip",
+        "https://mapzen.com/carto/zinc-style/zinc-style.zip"
+    };
+
+    AutoCompleteTextView sceneString;
```
tallytalwar: Addressed.

platforms/android/demo/src/main/java/com/mapzen/tangram/android/MainActivity.java
tallytalwar: Refactored a bit to at least not have same scene API setting 3 times in the code.

platforms/android/demo/src/main/java/com/mapzen/tangram/android/MainActivity.java
tallytalwar: Works great. Addressed.

platforms/android/demo/src/main/java/com/mapzen/tangram/android/MainActivity.java
tallytalwar: Works great. Addressed.

platforms/android/demo/src/main/java/com/mapzen/tangram/android/MainActivity.java
```diff
@@ -31,36 +40,130 @@
...
+                ArrayAdapter.createFromResource(this, R.array.style_array, R.layout.simple_spinner_item);
+        sceneSelector.setDropDownViewResource(R.layout.simple_spinner_dropdown_item);
+        spinner.setAdapter(sceneSelector);
+        spinner.setOnItemSelectedListener(this);
+
+        /* setup map view */
         view = (MapView)findViewById(R.id.map);
         view.onCreate(savedInstanceState);
+        autoCompleteView.setText(scenes[0]);
         view.getMapAsync(this, "scene.yaml", sceneUpdates);
```
matteblair: This call should use the current text in the `AutoCompleteView` instead of a constant.

platforms/android/demo/src/main/java/com/mapzen/tangram/android/MainActivity.java
```diff
@@ -88,6 +191,42 @@ public void onLowMemory() {
         view.onLowMemory();
     }
 
+    @Override public void onItemClick(AdapterView<?> parent, View view, int position, long id) {
+        int selectedIndex = 7; // custom
```
matteblair: This smells like trouble. `onItemClick` and `onItemSelected` seem overly complicated and dependent on data that's defined elsewhere. As in my earlier comment, I'd suggest removing the `Spinner` element entirely.

platforms/android/demo/src/main/java/com/mapzen/tangram/android/MainActivity.java
```diff
@@ -88,6 +191,42 @@ public void onLowMemory() {
...
+                break;
+            }
+        }
+        spinner.setSelection(selectedIndex);
+
+        autoCompleteView.clearFocus();
+        InputMethodManager in = (InputMethodManager) getSystemService(Context.INPUT_METHOD_SERVICE);
+        in.hideSoftInputFromWindow(autoCompleteView.getWindowToken(), 0);
+
+        map.loadSceneFile(autoCompleteView.getText().toString(),sceneUpdates);
```
matteblair: Loading a scene file synchronously from a network location on the UI thread is ... _no bueno_. Android could very well kill our app or produce the famous "Application Not Responding" message if this takes too long. However, there does seem to be a problem with calling `MapView$getMapAsync` more than once per `MapView` ... so either we live with an unresponsive UI for now or we fix `getMapAsync`.

platforms/android/demo/build.gradle
```diff
@@ -25,4 +25,5 @@ android {
 
 dependencies {
   compile project(path: ':tangram')
+    compile 'com.android.support:design:25.3.1'
```
tallytalwar: My Android Studio settings did this. Addressed.

platforms/android/demo/src/main/java/com/mapzen/tangram/android/MainActivity.java
```diff
@@ -31,36 +40,130 @@
...
-public class MainActivity extends Activity implements OnMapReadyCallback, TapResponder,
-        DoubleTapResponder, LongPressResponder, FeaturePickListener, LabelPickListener, MarkerPickListener, SceneUpdateErrorListener {
+public class MainActivity extends AppCompatActivity implements OnMapReadyCallback, TapResponder,
+        DoubleTapResponder, LongPressResponder, FeaturePickListener, LabelPickListener, MarkerPickListener, SceneUpdateErrorListener,
+        AdapterView.OnItemSelectedListener, AdapterView.OnItemClickListener {
+
+    private final String apiKey = "vector-tiles-tyHL4AY";
+    private ArrayList<SceneUpdate> sceneUpdates = new ArrayList<>();
+
+    /* Override ArrayAdaptor to include any text */
```
tallytalwar: Addressed.

platforms/android/demo/src/main/java/com/mapzen/tangram/android/MainActivity.java
```diff
@@ -31,36 +40,130 @@
...
+    private class CustomArrayAdapter extends ArrayAdapter {
```
tallytalwar: Make sense, now done in a separate class `DemoSceneManager`. The most apt name I could come for `CustomArrayAdapter` is `SceneAutoCompleteArrayAdapter`.

platforms/android/demo/src/main/java/com/mapzen/tangram/android/MainActivity.java
```diff
@@ -31,36 +40,130 @@
...
+                    clear();
+                    for (String scene: filterList) {
+                        add(scene);
+                        notifyDataSetChanged();
+                    }
+                }
+            }
+        };
+
+        CustomArrayAdapter(Context context, int resource, List<String> objects) {
```
tallytalwar: I think you can do that. I get an `java.lang.UnsupportedOperationException`. Checked a couple of stackoverflows and they say because one can not `clear` the contents as is being done in the custom filter here. Of course if I am missing something java(y) then I am not sure.

platforms/android/demo/src/main/java/com/mapzen/tangram/android/MainActivity.java
```diff
@@ -31,36 +40,130 @@
...
+        "https://mapzen.com/carto/refill-style/refill-style.zip",
+        "https://mapzen.com/carto/walkabout-style/walkabout-style.zip",
+        "https://mapzen.com/carto/tron-style/tron-style.zip",
+        "https://mapzen.com/carto/cinnabar-style/cinnabar-style.zip",
+        "https://mapzen.com/carto/zinc-style/zinc-style.zip"
+    };
+
+    AutoCompleteTextView autoCompleteView;
+    CustomArrayAdapter customAdapter;
+    Spinner spinner;
```
tallytalwar: Addressed. no spinner now. Make sense.

platforms/android/demo/src/main/java/com/mapzen/tangram/android/MainActivity.java
tallytalwar: Addressed.

platforms/android/demo/src/main/java/com/mapzen/tangram/android/MainActivity.java
tallytalwar: Implicitly handled by removing spinner.

platforms/android/demo/src/main/java/com/mapzen/tangram/android/MainActivity.java
tallytalwar: Agreed. I would suggest implementing `loadSceneFileAsync` call in our `MapController` to address this issue. But outside the scope of this PR and can be addressed in a separate PR.

tallytalwar:
@sarahlensing can you again give this a look, post the changes. Thanks

platforms/android/demo/src/main/java/com/mapzen/tangram/android/MainActivity.java
sarahlensing: Yep, thats right, when you use `String[]`, the adapter creates an immutable list

sarahlensing:
@tallytalwar LGTM

tallytalwar:
Updated the previous commit with couple of tiny improvements:
1. Switched to use `more-label` variants of house style from cdn
2. AutoCompleteTextView now highlights the whole text on getting focus. Makes it easy for the user to get rid of the entire text.

matteblair:
The last issue I'm trying to track down in this is that when the `MainActivity` is re-created due to rotating the screen, the text in the `AutoCompleteTextView` is preserved from the last instance, but the scene itself always reverts to the original "scene.yaml".

What's weird is that `MainActivity$onCreate` gets the scene URL from the current text of the `AutoCompleteTextView`, but the preserved text isn't restored in that view until _after_ `onCreate`!

matteblair:
The first quick "solution" that I thought of was just to make the `AutoCompleteTextView` _not_ preserve its text, but this behavior seems to be an irrevocable part of `EditText` and its subclasses (https://developer.android.com/reference/android/widget/TextView.html#setFreezesText(boolean)).

platforms/android/demo/src/main/java/com/mapzen/tangram/android/MainActivity.java
```diff
@@ -76,8 +76,16 @@ public void loadSceneCallback(String scene) {
 
         /* setup map view */
         view = (MapView)findViewById(R.id.map);
-        view.onCreate(savedInstanceState);
-        view.getMapAsync(this, sceneManager.getCurrentScene(), sceneUpdates);
+    }
+
+    @Override
+    public void onPostCreate(Bundle savedInstanceState) {
```
tallytalwar: 👍

PR #1471, opened 2017-04-24
Find fallback fonts on pre Lollipop Android versions

hjanetzek:
This PR adds sans-serif and no-name font family groups to the list of fallback fonts.

### DIFF
--- a/platforms/android/tangram/src/main/java/com/mapzen/tangram/FontFileParser.java
+++ b/platforms/android/tangram/src/main/java/com/mapzen/tangram/FontFileParser.java
@@ -23,8 +23,10 @@ class FontFileParser {
     private static String systemFontPath = "/system/fonts/";
     // Android version >= 5.0
     private static String fontXMLPath = "/system/etc/fonts.xml";
+
     // Android version < 5.0
-    private static String fontXMLFallbackPath = "/etc/system_fonts.xml";
+    private static String oldFontXMLPath = "/system/etc/system_fonts.xml";
+    private static String oldFontXMLFallbackPath = "/system/etc/fallback_fonts.xml";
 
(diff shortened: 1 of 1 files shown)


tallytalwar:
I will give one of the phones in the stock here a try.

tallytalwar:
The S3 we have running 4.0.4 has the following:

```
shell@android:/ $ ls -l /system/etc/fallback_fonts.xml                                                <
-rw-r--r-- root     root         4031 2012-05-23 10:23 fallback_fonts.xml
shell@android:/ $ ls -l /etc/fallback_fonts.xml
-rw-r--r-- root     root         4031 2012-05-23 10:23 fallback_fonts.xml

shell@android:/ $ ls -l /system/etc/system_fonts.xml
-rw-r--r-- root     root         2595 2012-05-23 10:23 system_fonts.xml
shell@android:/ $ ls -l /etc/system_fonts.xml
-rw-r--r-- root     root         2595 2012-05-23 10:23 system_fonts.xml
```

I see that you changed the paths here. Any reason for that?

tallytalwar:
Other than the above question .. renders look good wrt all fallback fonts.

hjanetzek:
@tallytalwar it's the path that is stated in '/system/etc/system_fonts.xml' comments. Also we use `/system/` prefix for post lollipop path. So just consistency.

@matteblair you can download older Android emulator images in AVD with a few clicks when creating a new device.

tallytalwar:
Thanks @hjanetzek, makes sense. Good to go.

matteblair:
This was merged into `master` in the 0.6.2 patch (d0a66e7037b7c9a5de663e188fcb5aaff83b03f1)

Issue #1473, opened 2017-04-24
assert crash with applySceneUpdates and sceneAssets


Guillaume227:
I am hitting the assert at sceneLoader.cpp:587 following a call to applySceneUpdates();
That's because it's looking for a texture url in scene->sceneAssets().
That texture asset was properly loaded during the loading of the scene.
However that line at he end of in SceneLoader::loadScene:
`    _scene->sceneAssets().clear();
`
clears the loaded assets.

I am trying to understand the logic of calling clear at that stage in relation to the applySceneUpdates calls that may follow.
Map displays fine if I do not call applySceneUpdates()

Any idea?

tallytalwar:
Yup, this seems to be a bug in the recently merged PR (https://github.com/tangrams/tangram-es/pull/1373) which does not handle scene assets loaded from initial scene load aptly when applying scene updated. 

Since this PR was merged after 0.6.1 release, this bug does not manifests in any of tangram-es official releases.

Guillaume227:
@tallytalwar : thank you !

PR #1476, opened 2017-04-24
Share scene assets from old scene when updating scene

tallytalwar:
Addresses #1473

### DIFF
--- a/core/src/scene/scene.cpp
+++ b/core/src/scene/scene.cpp
@@ -71,7 +71,7 @@ Scene::Scene(const Scene& _other)
     m_globalRefs = _other.m_globalRefs;
 
     m_mapProjection.reset(new MercatorProjection());
-    m_sceneAssets = _other.sceneAssets();
+    m_assets = _other.assets();
 }
 
 Scene::~Scene() {}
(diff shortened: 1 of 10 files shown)


core/src/scene/scene.cpp
```diff
@@ -70,6 +71,9 @@ Scene::Scene(const Scene& _other)
     m_globalRefs = _other.m_globalRefs;
 
     m_mapProjection.reset(new MercatorProjection());
+    for (auto& sceneAsset: _other.sceneAssets()) {
+        m_sceneAssets[sceneAsset.first.k] = sceneAsset.second;
+    }
 }
```
hjanetzek: Any reason for not using `m_sceneAssets = _other.sceneAssets()` ?

core/src/scene/scene.h
```diff
@@ -103,6 +105,9 @@ class Scene {
     const auto& featureSelection() const { return m_featureSelection; }
     const auto& sceneAssets() const { return m_sceneAssets; };
 
+    void createSceneAsset(const std::shared_ptr<Platform>& platform, const Url& resolvedUrl, const Url& relativeUrl,
+                          const Url& base);
+
```
hjanetzek: No need for this PR - I would rename these to`createAsset() / assets()` as it's clear that it's about parts of the scene

tallytalwar:
There is also another bug I found in this. Pushing soon.

tallytalwar:
Updated with a fix and scene urls are resolved post scene updates.

core/src/scene/scene.cpp
tallytalwar: Ohh I think I had a different structure for scene assets before.. hence this. But make sense with `fastmap`. Updated.

core/src/scene/scene.h
tallytalwar: Yeah make sense. Will push an explicit commit with this PR only. Thanks

PR #1479, opened 2017-04-25
Fix rpi build

hjanetzek:
- missing 'using namespace'
- curl setup/teardown is done by UrlClient


tschesnok:
I can confirm that this now builds and works on the RPI

PR #1480, opened 2017-04-25
Restrict font fallback families to be added

tallytalwar:


tallytalwar:
@hjanetzek Updated. Thanks

tallytalwar:
Note from @hjanetzek on why we are not using other styles for fallback fonts (in reference to 0f8a8d5):

```
the fallback font list is used to find glyphs that are not available otherwise. if that list contains e.g. a font with bold and then regular variant it would only always use the first since they provide the same glyphs

we could have separate fallback lists for different styles though making things more complicated (which I would not recommend right now though)
```

platforms/ios/src/TangramMap/platform_ios.mm
```diff
@@ -115,17 +115,36 @@ void initGLExtensions() {
     return data;
 }
 
+bool allowedFamily(NSString* familyName) {
+    const NSArray<NSString *> *allowedFamilyList = @[ @"Hebrew", @"Kohinoor", @"Gumurki", @"Thonburi", @"Tamil",
+                                                    @"Gurmukhi", @"Kailasa", @"Sangam", @"PingFang", @"Geeza",
+                                                    @"Mishafi", @"Farah", @"Hiragino" ];
+
+    for (NSString* allowedFamily in allowedFamilyList) {
```
msmollin: I think this makes sense since font family names can contain additional characters beyond just what's been listed above.

However, just for informational sake: You can simplify this if you want to check for content *equality*. `NSArray` has a method called `containsObject:` which you can pass your `familyName` value to. This subsequently calls `isEqual` on each string and does a content-based (not pointer-based) comparison on the strings.

tallytalwar: Word. Yeah I went with `containsString` because there are fallback fonts like 
"Kohinoor Devanagiri", "Kohinoor Telugu", etc.

platforms/ios/src/TangramMap/platform_ios.mm
```diff
@@ -115,17 +115,36 @@ void initGLExtensions() {
...
     for (id fallback in fallbacks) {
-
-        handles.emplace_back([fallback]() {
-            auto data = loadUIFont([UIFont fontWithName:fallback size:1.0]);
-            return data;
-        });
+    
+        if (!allowedFamily(fallback)) { continue; }
+
+        for (id fontName in [UIFont fontNamesForFamilyName:fallback]) {
```
msmollin: I generally try *really* hard to avoid using `id` for things if possible.

Looking at https://developer.apple.com/reference/uikit/uifont/1619023-fontnamesforfamilyname it's returning an array of NSStrings, so I'd rather see that type explicitly set if possible.

tallytalwar: Thanks @msmollin. Addressed e228edf.

Issue #1474, opened 2017-04-24
Error in libtangram.so (_ZN7Tangram6Labels18processLabelUpdateERKNS_9ViewStateEPNS_10StyledMeshEPNS_4TileERKN3glm7tmat4x4IfLNS8_9precisionE0EEEfbbb+763)


fokin33:
Build fingerprint: 'asus/WW_P01Y/P01Y:5.0.2/LRX22G/WW_ZenPad-12.26.4.66-20161201:user/release-keys'
Revision: '0'
ABI: 'x86'
pid: 11129, tid: 12640, name: GLThread 1471  >>> com.OsMoDroidV <<<
signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x68
    eax 5bb5ce20  ebx 5e7bdc70  ecx 00000000  edx 61ebf6d8
    esi 5face594  edi 5ddfaf80
    xcs 00000073  xds 0000007b  xes 0000007b  xfs 0000011f  xss 0000007b
    eip 5e3b7770  ebp 5d66fd00  esp 61ebf620  flags 00010202

backtrace:
    #00 pc 0024c770  /data/app/com.OsMoDroidV-2/lib/x86/libtangram.so (_ZN7Tangram11SpriteLabel17addVerticesToMeshERNS_15ScreenTransformERKN3glm5tvec2IfLNS3_9precisionE0EEE+160)
    #01 pc 0019818b  /data/app/com.OsMoDroidV-2/lib/x86/libtangram.so (_ZN7Tangram6Labels18processLabelUpdateERKNS_9ViewStateEPNS_10StyledMeshEPNS_4TileERKN3glm7tmat4x4IfLNS8_9precisionE0EEEfbbb+763)
    #02 pc 001985bc  /data/app/com.OsMoDroidV-2/lib/x86/libtangram.so (_ZN7Tangram6Labels12updateLabelsERKNS_9ViewStateEfRKNSt6__ndk16vectorINS4_10unique_ptrINS_5StyleENS4_14default_deleteIS7_EEEENS4_9allocatorISA_EEEERKNS5_INS4_10shared_ptrINS_4TileEEENSB_ISI_EEEERKNS5_INS6_INS_6MarkerENS8_ISN_EEEENSB_ISP_EEEEb+668)
[...]

PR #1482, opened 2017-04-26
MarkerManager scene switching

hjanetzek:
This branch fixes some cases where references to the old scene could still be used:
- StyleBuilder of the old scene were not removed for styles with different names than the new scene
- If the marker geometry was not rebuild due to missing rules or for other reasons the old mesh still remained. This could lead to the crash reported in https://github.com/tangrams/tangram-es/issues/1474

### DIFF
--- a/core/src/marker/markerManager.cpp
+++ b/core/src/marker/markerManager.cpp
@@ -314,6 +314,8 @@ const std::vector<std::unique_ptr<Marker>>& MarkerManager::markers() const {
 
 bool MarkerManager::buildStyling(Marker& marker) {
 
+    marker.setDrawRuleData(nullptr);
+
     if (!m_scene) { return false; }
 
     std::vector<StyleParam> params;
@@ -392,6 +394,9 @@ bool MarkerManager::buildStyling(Marker& marker) {
 }
 
 bool MarkerManager::buildGeometry(Marker& marker, int zoom) {
+
+    // Make sure old mesh gets invalidated
+    marker.setMesh(0, zoom, nullptr);
 
     auto feature = marker.feature();
     auto rule = marker.drawRule();
--- a/core/src/marker/markerManager.cpp
+++ b/core/src/marker/markerManager.cpp
@@ -51,6 +51,7 @@ MarkerID MarkerManager::add() {
 }
 
 bool MarkerManager::remove(MarkerID markerID) {
+
     for (auto it = m_markers.begin(), end = m_markers.end(); it != end; ++it) {
         if (it->get()->id() == markerID) {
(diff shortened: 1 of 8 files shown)


tallytalwar:
Some of this might also have been done @karimnaaji here: https://github.com/tangrams/tangram-es/pull/1467

core/src/marker/markerManager.cpp
```diff
@@ -14,15 +15,20 @@
...
     m_scene = scene;
     m_mapProjection = scene->mapProjection().get();
-    m_styleContext.initFunctions(*scene);
-    m_jsFnIndex = scene->functions().size();
 
+    m_styleContext = std::make_unique<StyleContext>();
+    m_styleContext->initFunctions(*scene);
 
     // Initialize StyleBuilders.
+    m_styleBuilders.clear();
```
tallytalwar: In accordance with the discussion I had with @karimnaaji for https://github.com/tangrams/tangram-es/pull/1467/files, it make sense to remove all markers on a scene reset.

Reason being, a marker could have been using some style in its styling string or a drawGroup which no longer exists in the new scene. So it becomes the responsibility of the user to re-add/re-style these markers.

tallytalwar: And when this gets included we can close https://github.com/tangrams/tangram-es/pull/1467 in favor of this PR.

hjanetzek: I'm not sure if it's really necessary to remove all markers on scene load. 
It's the question whether the api for loadScene is more friendly when it states that after scene loading all markers are invalidated and need to be recreated or that markers which use custom styles not defined in the new scene need to be updated with a new styling.

I think it's more common to use markers that use the default styles and it would be nice if we could keep that case simple.

matteblair:
This whole issue is more evidence to me that the current Marker system needs to be completely overhauled.

On the question of clearing all Markers when a new scene is loaded: It may be inconvenient in some cases, but I think the only rational choice for the interface is to clear all Markers whenever a new scene is loaded. Some of the Markers on a Map might be using built-in styles and therefore we can theoretically reproduce them perfectly in the new scene, but some of the Markers might not be reproducible in the new scene, and then the question is: how does a user's code know which is which? There's really no good way to communicate programmatically that a Marker needs to be re-styled or removed - it just stops being rendered and feels like a bug. And what happens to the Marker handle that the user's program is holding? Is it still valid? Do they need to dispose it? From the perspective of a user's program, it is much simpler to reason about Markers if you know for sure that they will all be invalidated and disposed when a new scene is loaded.

tallytalwar:
Closing in favor of #1497

PR #1490, opened 2017-05-05
Tidy up the Android demo app

matteblair:
- Separates the URL entry view logic from the concept of scene files
 - Removes the demo API key from the source code, using local environment variables instead (Travis CI now has a new environment variable for this)
 - Updates the Android `buildToolsVersion`


matteblair:
@tallytalwar Can you give this a try and see whether the local API key retrieval works for your setup too?

There's some nuance to the variables it uses: Android Studio on macOS doesn't use your shell environment variables, so for that situation you _must_ use a Gradle property. But if you build with command line Gradle, you should be able to use either.

matteblair:
Oh I see, good to know that there is a workaround for environment variables in Android Studio. I found it easier to set a `mapzenApiKey` value in `~/.gradle/gradle.properties`, so I left that option.

tallytalwar:
Yeah make sense to advertise the API key setting in gradle properties. For example I have one set in my global gradle properties.

Also can you update the android readme to document the use of API key here. Once there good to merge.

PR #1495, opened 2017-05-09
Yet Another Version of Android UI Thread Callbacks

matteblair:
Somehow we had overlooked this simpler option. We may be able to reduce the repetition here too, but I don't think it's too bad for now.

Question: Should this be applied to the `ViewCompleteListener` as well?

_Edit: The same change is now applied to ViewCompleteListener too._


tallytalwar:
This seems much concise and simpler approach. I think we can do this for `ViewCompleteListener` also.

PR #1496, opened 2017-05-10
Remove tile side buffers for client data source tiles

tallytalwar:
Remove tile paddings from client data by setting tile buffer on each side to zero for GeoJSONVT option.

Fixes https://github.com/mapzen/android/issues/364


hjanetzek:
lets see if this initializer list compiles on gcc 4.9

PR #1498, opened 2017-05-12
Scaled tiles

tallytalwar:
Ok this is much better now.

- visible tile ids (scale and max zoom adjusted) are stored per datasource 
- Does not unnecessarily bloat TileID struct
- getParent and getChild TileID takes into account a datasource' tileScale and maxZoom respectively.

Addressed #1435 
- label pipeline modified a bit to include the set of datasources to help find parent and child tiles.

### DIFF
--- a/core/include/tangram/tile/tileID.h
+++ b/core/include/tangram/tile/tileID.h
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <cassert>
 #include <cstdint>
 #include <string>
 
@@ -56,15 +57,17 @@ struct TileID {
         return TileID(x >> over, y >> over, _maxZoom, s, wrap);
     }
 
-    TileID scaled(int32_t _tileScale) const {
-        if (_tileScale <= 0) { return *this; }
+    TileID zoomBiasAdjusted(int32_t _zoomBias) const {
+        assert(_zoomBias >= 0);
 
-        auto scaledZ = std::max(0, z - _tileScale);
-        return TileID(x >> _tileScale, y >> _tileScale, scaledZ, z, wrap);
+        if (!_zoomBias) { return *this; }
+
+        auto scaledZ = std::max(0, z - _zoomBias);
+        return TileID(x >> _zoomBias, y >> _zoomBias, scaledZ, z, wrap);
     }
 
-    TileID getParent(int32_t _tileScale = 0) const {
-        if (s > (z + _tileScale)) {
+    TileID getParent(int32_t _zoomBias = 0) const {
+        if (s > (z + _zoomBias)) {
             // Over-zoomed, keep the same data coordinates
             return TileID(x, y, z, s - 1, wrap);
         }
--- a/core/include/tangram/tile/tileID.h
+++ b/core/include/tangram/tile/tileID.h
@@ -45,28 +45,33 @@ struct TileID {
         return isValid() && z <= _maxZoom;
     }
 
-    TileID withMaxSourceZoom(int32_t _max) const {
+    TileID withMaxSourceZoom(int32_t _maxZoom) const {
 
-        if (z <= _max) {
+        if (z <= _maxZoom) {
             return *this;
         }
 
-        int32_t over = z - _max;
+        int32_t over = z - _maxZoom;
 
-        return TileID(x >> over, y >> over, _max, z, wrap);
+        return TileID(x >> over, y >> over, _maxZoom, s, wrap);
     }
 
-    TileID getParent() const {
+    TileID scaled(int32_t _tileScale) const {
+        if (_tileScale <= 0) { return *this; }
 
-        if (s > z) {
+        auto scaledZ = std::max(0, z - _tileScale);
+        return TileID(x >> _tileScale, y >> _tileScale, scaledZ, z, wrap);
+    }
+
+    TileID getParent(int32_t _tileScale = 0) const {
+        if (s > (z + _tileScale)) {
             // Over-zoomed, keep the same data coordinates
             return TileID(x, y, z, s - 1, wrap);
         }
-        return TileID(x >> 1, y >> 1, z-1, z-1, wrap);
+        return TileID(x >> 1, y >> 1, z - 1, s - 1, wrap);
     }
 
-    TileID getChild(int32_t _index) const {
-
+    TileID getChild(int32_t _index, int32_t _maxZoom) const {
         if (_index > 3 || _index < 0) {
             return TileID(-1, -1, -1, -1, -1);
         }
@@ -78,15 +83,12 @@ struct TileID {
         // i:      0, 0, 1, 1
         // j:      0, 1, 0, 1
 
-        return TileID((x<<1)+i, (y<<1)+j, z+1, z+1, wrap);
-    }
-
-    TileID getChild(int32_t _index, int32_t _maxSourceZoom) const {
-        return getChild(_index).withMaxSourceZoom(_maxSourceZoom);
+        auto childID = TileID((x<<1)+i, (y<<1)+j, z + 1, s + 1, wrap);
+        return childID.withMaxSourceZoom(_maxZoom);
     }
 
     std::string toString() const {
-        return std::to_string(x) + "/" + std::to_string(y) + "/" + std::to_string(z);
+        return std::to_string(x) + "/" + std::to_string(y) + "/" + std::to_string(z) + "/" + std::to_string(s);
     }
 
 };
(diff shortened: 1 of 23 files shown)


hjanetzek:
yup, looks much better :) 

Added some changes to keep the TileSource specific handling in TileManager instead of View::updateTiles

@matteblair, @tallytalwar your turn

tallytalwar:
@hjanetzek your changes look good to me.

core/src/tile/tileManager.cpp
```diff
@@ -154,8 +154,11 @@ void TileManager::updateTileSets(const View& _view) {
 
     // Make m_tiles an unique list of tiles for rendering sorted from
     // high to low zoom-levels.
-    std::sort(m_tiles.begin(), m_tiles.end(), [](auto& a, auto& b){
-            return a->getID() < b->getID(); });
+    std::sort(m_tiles.begin(), m_tiles.end(), [](auto& a, auto& b) {
```
tallytalwar: 👍

core/src/tile/tileManager.cpp
```diff
@@ -65,17 +65,22 @@ void TileManager::setTileSources(const std::vector<std::shared_ptr<TileSource>>&
...
+                             });
 
-            m_tileSets.push_back({ source, false });
+            if (sIt == m_tileSets.end()) {
+                LOGN("Add source %s", source->name().c_str());
+                m_tileSets.push_back({ source, false });
+            } else {
+                LOGN("Reuse source from previous scene %s", source->name().c_str());
+                source = sIt->source;
+            }
```
hjanetzek: We could also remove this TileSource reusing logic completely instead Imo.

tallytalwar: Yeah, we were discussing this for https://github.com/tangrams/tangram-es/pull/1504.

tallytalwar: Will address it in that PR.

core/src/scene/sceneLoader.cpp
```diff
@@ -971,6 +971,7 @@ void SceneLoader::loadSource(const std::shared_ptr<Platform>& platform, const st
     int32_t minDisplayZoom = -1;
     int32_t maxDisplayZoom = -1;
     int32_t maxZoom = 18;
+    int32_t tileScale = 0; // Equivalent of tileSize of 256px
```
matteblair: More comment! 'A tileScale of 0 corresponds to a tile_size of 256, 1 corresponds to 512, 2 corresponds to 1024, etc.'

core/include/tangram/tile/tileID.h
```diff
@@ -45,28 +45,33 @@ struct TileID {
...
-        int32_t over = z - _max;
+        int32_t over = z - _maxZoom;
 
-        return TileID(x >> over, y >> over, _max, z, wrap);
+        return TileID(x >> over, y >> over, _maxZoom, s, wrap);
     }
 
-    TileID getParent() const {
+    TileID scaled(int32_t _tileScale) const {
+        if (_tileScale <= 0) { return *this; }
```
matteblair: Actually, maybe it makes more sense for these functions that require parameters from the tile source to be members of `TileSource` itself, like `TileID TileSource::getParent(TileID tile)`. Not sure that it helps substantially for this PR, but something to keep in mind perhaps.

core/src/view/view.h
```diff
@@ -9,6 +9,7 @@
 #include "glm/vec3.hpp"
 #include <set>
 #include <memory>
+#include <unordered_map>
```
matteblair: Unused also, I think. We can probably remove `<set>` now too.

core/src/view/view.h
```diff
@@ -133,7 +134,7 @@ class View {
     /* Get the current pitch angle in radians */
     float getPitch() const { return m_pitch; }
 
-    /* Updates the view and projection matrices if properties have changed */
+    /* Updates the view and projection matrices if properties have changed  */
```
matteblair: Sneaky space sneakin in

core/src/scene/sceneLoader.cpp
```diff
@@ -987,6 +988,15 @@ void SceneLoader::loadSource(const std::shared_ptr<Platform>& platform, const st
     if (auto maxZoomNode = source["max_zoom"]) {
         maxZoom = maxZoomNode.as<int32_t>(maxZoom);
     }
+    if (auto tileSizeNode = source["tile_size"]) {
+        const auto BaseTileSize = 256;
+        auto tileSize = tileSizeNode.as<int32_t>();
+        if (tileSize && !( tileSize & (tileSize - 1))) {
```
karimnaaji: We do have a routine for checking power of two used for textures defined in `geom.h`, which would make this more explicit to read.

core/include/tangram/tile/tileID.h
tallytalwar: Maybe, I won't consider it to be for this PR.

core/src/scene/sceneLoader.cpp
tallytalwar: @karimnaaji will use the one you mentioned. @matteblair LOG statement below in the same context, one can easily tell what this statement is checking.

tallytalwar:
K @matteblair addressed most of of your suggestions. Few things I did not address:
1. TileSource::Options: makes sense, but does not directly relate to this PR and adds a lot of diff, so make sense to have a separate PR
2. getChildren. not sure if this is any beneficial to us, we have to loop over the children anyhow. And in unit test we operate with a single child! That said, I am neutral on this, but again does not directly relates to this PR and will add a lot of diff, so should be done separately.

core/src/data/rasterSource.cpp
```diff
@@ -69,8 +69,8 @@ class RasterTileTask : public BinaryTileTask {
 
 RasterSource::RasterSource(const std::string& _name, std::unique_ptr<DataSource> _sources,
                            int32_t _minDisplayZoom, int32_t _maxDisplayZoom, int32_t _maxZoom,
-                           TextureOptions _options, bool _genMipmap)
-    : TileSource(_name, std::move(_sources), _minDisplayZoom, _maxDisplayZoom, _maxZoom),
+                           int32_t _, TextureOptions _options, bool _genMipmap)
+    : TileSource(_name, std::move(_sources), _minDisplayZoom, _maxDisplayZoom, _maxZoom, _),
       m_texOptions(_options),
```
hjanetzek: Parameter name could be more descriptive :)

tallytalwar: Ahh.. sed messed up. Will fix.

core/src/labels/labels.cpp
```diff
@@ -214,23 +216,18 @@ void Labels::skipTransitions(const std::vector<std::unique_ptr<Style>>& _styles,
         TileID tileID = tile->getID();
         std::shared_ptr<Tile> proxy;
 
+        const auto& source = _scene->getTileSource(tile->sourceID());
```
hjanetzek: Oh right, ClientDataSources are not added to the scene sources.

core/src/scene/scene.cpp
```diff
@@ -134,6 +134,15 @@ std::shared_ptr<Texture> Scene::getTexture(const std::string& textureName) const
     return texIt->second;
 }
 
+std::shared_ptr<TileSource> Scene::getTileSource(int32_t id) {
```
hjanetzek: If something returns a pointer one should always check for null case :)

core/src/data/rasterSource.cpp
tallytalwar: Updated.

core/src/labels/labels.cpp
tallytalwar: Yeah, this is fixed.

core/src/labels/labels.cpp
```diff
@@ -217,23 +221,19 @@ void Labels::skipTransitions(const std::vector<std::unique_ptr<Style>>& _styles,
...
+        if (!source) { continue; }
+
```
hjanetzek: no, ClientDataSources are not in Scene TileSources so this needs to be handled differently

tallytalwar:
No wait, @hjanetzek is correct. Labels skiptransition wont be checked for client datasources.

tallytalwar:
@hjanetzek can you check 382e479? I couldn't think of any other way to get client data sources accessed from within labels.

tallytalwar:
I took the scene file commit out from here, so this should be good to go.

PR #1502, opened 2017-05-15
0.5.3 patch

hjanetzek:


matteblair:
I think we can safely omit this patch release, the users who have reported SSL problems with 0.5.x seem to be content to work around it by using HTTP tile URLs.

tallytalwar:
We pushed 0.6.3 patched release yesterday with the apt bug fixes. `0.5.3` as such is not required.

PR #1504, opened 2017-05-15
Fix tilesource equality

tallytalwar:
Found this issue while debugging an issue on android sdk: https://github.com/mapzen/android/issues/363

We check for tileSource equality when adding/removing `TileSet` do utilize cache for previously loaded datasource(tileSource) from an older scene when the new scene adds the same datasource(tileSource).

The equality for these tileSources was broken.

Along with this it kinda make sense to remove all old tile sources on a new scene file load.

Also adds reading of API key from the environment variable for the desktop demo apps (refer issue: https://github.com/tangrams/tangram-es/issues/1452)


tallytalwar:
@sarahlensing tried this and included the SNAPSHOT repo in android sdk and it none of the styles work in style switcher, which is excepted and should be fixed by your work from https://github.com/mapzen/android/pull/371.

core/src/tile/tileManager.cpp
```diff
@@ -43,18 +43,14 @@ void TileManager::setTileSources(const std::vector<std::shared_ptr<TileSource>>&
...
         [&](auto& tileSet) {
             if (!tileSet.clientTileSource) {
-                auto sIt = std::find_if(_sources.begin(), _sources.end(),
-                                        [&](auto& source){ return source->equals(*tileSet.source); });
-
-                if (sIt == _sources.end() || !(*sIt)->generateGeometry()) {
-                    LOGD("remove source %s", tileSet.source->name().c_str());
-                    return true;
-                }
+                LOGD("remove source %s", tileSet.source->name().c_str());
```
tallytalwar: @hjanetzek This is what you were referring to in https://github.com/tangrams/tangram-es/pull/1498#discussion_r117051629.

Also though it is good to have the `TileSource::equals` functionality, with this change it is not used now. Does it make sense to remove it, I am fine either way, but probably can leave it.

toolchains/linux.cmake
```diff
@@ -34,6 +34,14 @@ set(CORE_COMPILE_DEFS PLATFORM_LINUX)
 # load core library
 add_subdirectory(${PROJECT_SOURCE_DIR}/core)
 
+if("${MAPZEN_API_KEY}" STREQUAL "")
```
karimnaaji: It would make send to add this as a CMake routine to reduce deduplication, it's used on almost every single platform now.

toolchains/darwin.cmake
```diff
@@ -14,6 +14,14 @@ set(CORE_COMPILE_DEFS PLATFORM_OSX)
 # Build core library with dependencies.
 add_subdirectory(${PROJECT_SOURCE_DIR}/core)
 
+if("${MAPZEN_API_KEY}" STREQUAL "")
```
matteblair: This syntax will check for a definition that was explicitly passed into CMake, which our `Makefile` currently does, but this doesn't automatically work if you're not using the `Makefile` to build (e.g. I use `ninja` almost exclusively on desktop).  We can check for environment variables directly in CMake using `$ENV{MAPZEN_API_KEY}` and then this will work without the `Makefile`.

platforms/rpi/src/main.cpp
```diff
@@ -83,7 +86,13 @@ void setup(int argc, char **argv) {
     double lat = 0.0f;
     double lon = 0.0f;
     std::string scene = "scene.yaml";
-    const std::string& apiKey = "vector-tiles-tyHL4AY";
+
+    const char* apiKey = std::getenv("MAPZEN_API_KEY");
```
matteblair: Since we already have access to the API key in CMake, we can just pass it in at compile-time as a `#define` rather than checking an environment variable at run-time. That is, in CMake we say `add_definitions(-DMAPZEN_API_KEY="$MAPZEN_API_KEY")` and in this file we just use the value of `MAPZEN_API_KEY` (I may not have the syntax exactly right).

core/include/tangram/data/tileSource.h
```diff
@@ -85,8 +85,6 @@ class TileSource : public std::enable_shared_from_this<TileSource> {
     virtual void clearRasters();
     virtual void clearRaster(const TileID& id);
 
-    bool equals(const TileSource& _other) const;
```
matteblair: I think we can also get rid of `DataSource::equals` now?

tallytalwar:
Updated.

Can I get some eyes on this? Need to get this in before #1498

toolchains/utils.cmake
```diff
@@ -22,6 +22,19 @@ function(check_unsupported_compiler_version)
 
 endfunction(check_unsupported_compiler_version)
 
+function(check_mapzen_api_key)
+
+    if("${MAPZEN_API_KEY}" STREQUAL "")
```
matteblair: This syntax will check for a definition that was explicitly passed into CMake, which our `Makefile` currently does, but this doesn't automatically work if you're not using the `Makefile` to build (e.g. I use `ninja` almost exclusively on desktop). We can check for environment variables directly in CMake using `$ENV{MAPZEN_API_KEY}` and then this will work without the `Makefile`. (Comment moved for clarity)

core/include/tangram/data/tileSource.h
tallytalwar: Too late @matteblair

toolchains/utils.cmake
tallytalwar: Yup make sense. 

P.S. you could have deleted the other comment :D.

platforms/common/glfwApp.cpp
```diff
@@ -397,8 +404,10 @@ void keyCallback(GLFWwindow* window, int key, int scancode, int action, int mods
 
 void dropCallback(GLFWwindow* window, int count, const char** paths) {
 
+    const char* apiKey = std::getenv("MAPZEN_API_KEY");
```
karimnaaji: This value is already accessed once at setup, we can change its scope to be global to this file.

platforms/osx/README.md
```diff
@@ -15,7 +15,17 @@ Make sure to update git submodules before you build:
...
 ```
 
-## Build ##
+The demo application uses the Mapzen vector tile service, so you will need a Mapzen API key to build and run the demo. 
+
+ 1. Visit https://mapzen.com/documentation/overview/#get-started-developing-with-mapzen to get an API key.
+
+ 2. Setup an environment variable (`MAPZEN_API_KEY`) to point to your API key. 
+ If you are using an IDE on osx, you need to do the following:
+    `launchctl setenv MAPZEN_API_KEY YOUR_API_KEY`
```
karimnaaji: Can we use ``` markdown syntax similarly to other shell commands to make this formatting more clear.

toolchains/utils.cmake
tallytalwar: Addressed.

platforms/common/glfwApp.cpp
tallytalwar: accessed the macro set by cmake now.

platforms/osx/README.md
tallytalwar: addressed.

platforms/osx/README.md
```diff
@@ -15,7 +15,21 @@ Make sure to update git submodules before you build:
...
+ If you are using an IDE on osx, you need to do the following:
+    ```bash
+    launchctl setenv MAPZEN_API_KEY YOUR_API_KEY
+    ```
+ If you are running the app from a terminal you need to do the following:
+    ```bash
+    export MAPZEN_API_KEY=YOUR_API_KEY
+    ```
+
+# #Build ##
```
karimnaaji: Extra space got in here.

toolchains/linux.cmake
```diff
@@ -34,6 +34,9 @@ set(CORE_COMPILE_DEFS PLATFORM_LINUX)
 # load core library
 add_subdirectory(${PROJECT_SOURCE_DIR}/core)
 
+check_mapzen_api_key()
+add_definitions( -DMAPZEN_API_KEY="$ENV{MAPZEN_API_KEY}" )
```
matteblair: I would suggest putting this within the `if(APPLICATION)` block below, since the API key isn't actually required unless you're building the demo (rather than just the library).

It would also be nice if `check_mapzen_api_key()` obtained the key for us (with an in-out variable, see https://cmake.org/pipermail/cmake/2008-November/025423.html)

matteblair: Oh I was too slow, I'll submit this separately.

PR #1508, opened 2017-05-17
Fix iOS framework version injection

karimnaaji:
At deployment time we only update the value of `CFBundleShortVersionString` explicitly to the framework version number and leave `CFBundleVersion` to `$(CURRENT_PROJECT_VERSION)`. Unfortunately this value was not set through `set_xcode_property` in CMake and was left blank.

This PR changes slightly the deployment process to update the framework value directly in CMake and then sets the value of the XCode property `CURRENT_PROJECT_VERSION` accordingly which is then set on both `CFBundleShortVersionString` and `CFBundleVersion`.


tallytalwar:
Rebased, merging when CI goes green.

PR #1516, opened 2017-05-19
CMake helper function to get a local Mapzen API key

matteblair:
- Makes it easier to change the name or location of the local API key from just one place.
 - Moves API key checks into demo application build steps, where applicable.

### DIFF
--- a/toolchains/iOS.cmake
+++ b/toolchains/iOS.cmake
@@ -18,7 +18,8 @@ foreach(_ext ${IOS_EXTENSIONS_FILES})
         ${PROJECT_SOURCE_DIR}/platforms/ios/demo/src/${_ext})
 endforeach()
 
-check_mapzen_api_key()
+get_mapzen_api_key(MAPZEN_API_KEY)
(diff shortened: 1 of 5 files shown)


toolchains/iOS.cmake
```diff
@@ -18,7 +18,8 @@ foreach(_ext ${IOS_EXTENSIONS_FILES})
         ${PROJECT_SOURCE_DIR}/platforms/ios/demo/src/${_ext})
 endforeach()
 
-check_mapzen_api_key()
+get_mapzen_api_key(MAPZEN_API_KEY)
+add_definitions(-DMAPZEN_API_KEY="${MAPZEN_API_KEY}")
```
tallytalwar: We do not need to have this added for iOS. The API key gets ingested in the plist file.

matteblair: I wasn't sure whether it would be replaced in the plist file without a define, it probably doesn't hurt

matteblair: Seems fine without the define. We could probably remove the step of writing/reading it from a plist file and just use the define in code, so I'll just leave it for now!

Issue #1500, opened 2017-05-14
Node images become invisible on certain zoom levels


westnordost:
This is not a duplicate of #1415, as that issue was solved by adding priority:0 to the draw rule in question. Now with tangram-es 0.6.2, some people are experiencing the same bug, but with the right priority already set in the draw rule. I cannot say of course if this behavior existed before 0.6.2, it also does not appear too often.

Note the red marker with the question mark. It vanishes at a certain zoom level, even though the geojson for that layer is not changed.

![gif](https://cloud.githubusercontent.com/assets/899988/26034134/213d5eb4-38b8-11e7-81dd-27dd9fa619ff.gif)
The video was made with a Android 6.0.1 MMB29M, Redmi 4 device, StreetComplete 0.10.

I reproduced it like this right now:
1. turn on "show all notes" in the StreetComplete options
2. go here: http://www.openstreetmap.org/#map=19/50.06217/19.93807 and scan for quests. A red marker should appear on the northern end of the long building
3. zoom in and out, see that the marker appears and disappears depending on the zoom.

It is  conspicuous that markers seem to vanish whenever a text (road name) is displayed at the same location.
But not always. You see this effect with other markers on that plaza, for example there is an orange clock marker on the small building on the western side of this plaza which also shows the same behavior but there is no street name close to there.

The style I use: https://github.com/westnordost/StreetComplete/blob/master/app/src/main/assets/streetcomplete.yaml

Related StreetComplete bug: https://github.com/westnordost/StreetComplete/issues/225

tallytalwar:
@westnordost Going to mark this resolved. Fix for this went in master with #1517 and I also tested this with `StreetComplete`'s `master` branch.

Issue #1499, opened 2017-05-13
Geometry added with GeoJson flicker depending on zoom level


westnordost:
I add line geometry which is directly derived from OSM data as a new layer to the map (`geometryLayer.addPolyline`), using an own style to render it half-transparently.

Some users are reporting the following effect when zooming in and out. See this video for an illustration:
https://files.natenom.com/nidx/nobackup/bug204.mp4

I myself can not reproduce it, but what I can reproduce is that sections of the way seem to be rendered duplicate, this might be related.
![img](https://cloud.githubusercontent.com/assets/13040148/25828146/fcb50654-344e-11e7-99a0-77fd90c15649.png)

Related StreetComplete issue: https://github.com/westnordost/StreetComplete/issues/204

Using Tangram-ES Version 0.6.2

westnordost:
The video was made with LineageOS 14.1, Android 7.1.2 on a Samsung Galaxy S5.

tallytalwar:
Hey @westnordost 

Thanks for keeping us on our toes :D. There are 2 issues pointed here and one of which ("sections of the way seem to be rendered duplicate") is resolved by https://github.com/tangrams/tangram-es/pull/1496, the other issue of disappearing of certain client data source geometry at certain zoom level is same as being discussed in https://github.com/tangrams/tangram-es/issues/1463. We are aware of this and will try to get it resolved soon. I am going to close this issue in favor of #1463.

Thanks

matteblair:
(The polylines will overlap less after the fix from #1496 is applied, but the round caps at the ends will still overlap - you can change those to 'butt' to keep them from extending past the end of the line https://mapzen.com/documentation/tangram/draw/#cap)

westnordost:
Right, cool. Sorry for the duplicate.

tallytalwar:
@westnordost The fix for this should be in master. Refer #1517

PR #1517, opened 2017-05-22
Fix GeoJsonClientSource multi geometries

hjanetzek:
Tested with the case in https://github.com/tangrams/tangram-es/issues/1499

http://geojson.io/#id=gist:anonymous/d00b5090e73256d284444cadb5b0f0bc&map=18/52.27367/20.84775


tallytalwar:
Testing this now, this might have also resolved #1500

tallytalwar:
Looks like this is also fixing #1500, just tried this with StreetComplete master.

hjanetzek:
@tallytalwar updated

Issue #1450, opened 2017-04-10
Custom placeholder when loading scene asynchronously


sarahlensing:
We recently updated the iOS SDK sample app to load scene files asynchronously. When making that change we noticed that the map background is black before any of the tiles load. This doesn't occur when loading the scene file synchronously (the background is gray). 

Could we update tangram to provide a gray background in all cases by default and also to allow users to designate a custom background color (or view) to display before tiles load?

[Here](https://github.com/tangrams/ios-framework/tree/tab-bar) is an updated example which highlights the issue by creating two `TGMapViewController` subclasses within a tab bar and sets the background to be red before the tiles finish loading.

nvkelso:
I vote for white by default.

matteblair:
That is an odd difference, I agree that it would be preferable to make the behavior more consistent. 

To clarify, we're talking about the application background while the _scene_ is loading, yes? There is already an option in the scene file to specify a background color where _tiles_ are loading, but of course we can't apply that option before the scene is parsed and loaded.

karimnaaji:
@sarahlensing have you tried changing the background color in the interface builder?

![screen shot 2017-04-11 at 10 57 18 am](https://cloud.githubusercontent.com/assets/7061573/24915699/de43478a-1ea5-11e7-9087-8208bee8bea8.png)

In tangram, setting the clear color in the first GL call that happens in like so:
```diff
diff --git a/core/src/tangram.cpp b/core/src/tangram.cpp
index a9e19f0..4ed3a50 100644
--- a/core/src/tangram.cpp
+++ b/core/src/tangram.cpp
@@ -864,6 +864,8 @@ void Map::setupGL() {
 
     LOG("setup GL");
 
+    impl->renderState.clearColor(1.0, 0.0, 0.0, 1.0);
+
     impl->renderState.invalidate();
 
     impl->tileManager.clearTileSets();
(END)

```
Doesn't seem to resolve the issue, even though it would still be good to add that diff to provide a default color before the scene is loaded.

sarahlensing:
"To clarify, we're talking about the application background while the scene is loading, yes?"
@matteblair we're talking about the `GLKView` background while the scene is loading

@karimnaaji I tried that but it didnt have any effect. I was able to change the color with this code:
```swift
override func glkView(_ view: GLKView, drawIn rect: CGRect) {
    super.glkView(view, drawIn: rect)
    if(!loaded) {
[...]

PR #1518, opened 2017-05-22
Background color placeholder

karimnaaji:
Fixes https://github.com/tangrams/tangram-es/issues/1450

### DIFF
--- a/core/src/gl/framebuffer.cpp
+++ b/core/src/gl/framebuffer.cpp
@@ -38,10 +38,12 @@ void FrameBuffer::apply(RenderState& _rs, GLuint _handle, glm::vec2 _viewport, g
     _rs.framebuffer(_handle);
     _rs.viewport(0, 0, _viewport.x, _viewport.y);
 
-    _rs.clearColor(_clearColor.r / 255.f,
-                   _clearColor.g / 255.f,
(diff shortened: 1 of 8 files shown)


core/src/gl/framebuffer.cpp
```diff
@@ -38,10 +38,12 @@ void FrameBuffer::apply(RenderState& _rs, GLuint _handle, glm::vec2 _viewport, g
     _rs.framebuffer(_handle);
     _rs.viewport(0, 0, _viewport.x, _viewport.y);
 
-    _rs.clearColor(_clearColor.r / 255.f,
-                   _clearColor.g / 255.f,
-                   _clearColor.b / 255.f,
-                   _clearColor.a / 255.f);
+    if (_clearColor == glm::vec4(0.0) && _rs.defaultOpaqueClearColor()) {
```
tallytalwar: looks like you check for the `defaultOpaqueClearColor` to be set in `clearDefaultOpaqueColor` so this second check in the if statement is redundant.

karimnaaji: If the user hasn't set a `defaultOpaqueColor` through `setDefaultBackgroundColor` in `tangram.h` we still want to clear with the default color.

PR #1519, opened 2017-05-22
Set debug level project-wide

hjanetzek:


matteblair:
@hjanetzek Feel free to merge after rebasing!

PR #1521, opened 2017-05-23
replaced CMAKE_SOURCE_DIR with PROJECT_SOURCE_DIR for Android toolchain

pbula:
Made a little change in cmake file for Android so the project could be included as part of a larger cmake based project. Alternatively, CMAKE_CURRENT_SOURCE_DIR works as well, but in this case I think PROJECT_SOURCE_DIR makes more sense, especially seeing how it's used in other cmake files.


matteblair:
Hi @pbula, thanks for the PR. The change here makes sense, our CI build for Android is failing here because of a recent change we made with the handling of API keys - it's not your fault. We're working on a solution now.

matteblair:
Applied in #1590

Issue #1477, opened 2017-04-24
Implement label centroid generation


matteblair:
A data source can specify a Boolean parameter `generate_label_centroids`. When this is `true`, the data source should generate a new point geometry for every polygon, as described here: https://mapzen.com/documentation/tangram/sources/#generate_label_centroids

karimnaaji:
Resolved by https://github.com/tangrams/tangram-es/pull/1522

PR #1522, opened 2017-05-23
Label centroid

tallytalwar:
Addressed #1477 

TODO:
- [x] Implement centroid calculation relative to first coordinate in the polygon ring

Note: Not an API breaking change, but adds on to the current API for MapData interface for ios and android.

### DIFF
--- a/core/include/tangram/data/tileSource.h
+++ b/core/include/tangram/data/tileSource.h
@@ -29,6 +29,8 @@ public:
             : minDisplayZoom(_minDisplayZoom), maxDisplayZoom(_maxDisplayZoom),
               maxZoom(_maxZoom), zoomBias(_zoomBias) {}
 
+        ZoomOptions() {}
+
         // Minimum zoom for which tiles will be displayed
         int32_t minDisplayZoom = -1;
         // Maximum zoom for which tiles will be displayed
@@ -81,7 +83,7 @@ public:
      * and zoom level of tiles to produce their URL.
      */
     TileSource(const std::string& _name, std::unique_ptr<DataSource> _sources,
-               ZoomOptions _zoomOptions);
+               ZoomOptions _zoomOptions = {});
 
     virtual ~TileSource();
 
--- a/core/include/tangram/data/tileSource.h
+++ b/core/include/tangram/data/tileSource.h
@@ -22,6 +22,26 @@ class TileSource : public std::enable_shared_from_this<TileSource> {
 
 public:
 
+    struct ZoomOptions {
+
+        ZoomOptions(int32_t _minDisplayZoom, int32_t _maxDisplayZoom,
+                      int32_t _maxZoom, int32_t _zoomBias)
+            : minDisplayZoom(_minDisplayZoom), maxDisplayZoom(_maxDisplayZoom),
+              maxZoom(_maxZoom), zoomBias(_zoomBias) {}
+
+        // Minimum zoom for which tiles will be displayed
+        int32_t minDisplayZoom = -1;
+        // Maximum zoom for which tiles will be displayed
+        int32_t maxDisplayZoom = -1;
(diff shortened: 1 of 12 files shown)


core/include/tangram/data/tileSource.h
```diff
@@ -22,6 +22,26 @@ class TileSource : public std::enable_shared_from_this<TileSource> {
 
 public:
 
+    struct SourceOptions {
+
```
hjanetzek: The 'Source' in TileSource::SourceOptions is a bit redundant.

tallytalwar:
@hjanetzek I am done here. You can finally review this. 👍

tallytalwar:
Wait this needs API change for platforms!!

tallytalwar:
Ok Finally all good now.

tallytalwar:
@karimnaaji can you also look at the ios API additions here?

Thanks

tallytalwar:
@karimnaaji updated the platform interface here to reflect what we discussed in the meeting. Can you give this a look?

core/src/data/clientGeoJsonSource.cpp
```diff
@@ -77,6 +78,83 @@ ClientGeoJsonSource::ClientGeoJsonSource(std::shared_ptr<Platform> _platform,
...
+            if (g.empty()){
+                continue;
+            }
+            auto area = signedArea(g.front().begin(), g.front().end());
+            if (area > largestArea) {
+                largestArea = area;
+                largestAreaIndex = index;
+            }
+        }
+        return (*this)(geom[largestAreaIndex]);
```
karimnaaji: nit: do you find this more explicit?
```c++
this->operator()(geom[largestAreaIndex]);
```

core/src/data/clientGeoJsonSource.cpp
```diff
@@ -77,6 +78,83 @@ ClientGeoJsonSource::ClientGeoJsonSource(std::shared_ptr<Platform> _platform,
 
 ClientGeoJsonSource::~ClientGeoJsonSource() {}
 
+struct add_centroid {
+
+    geometry::point<double>& pt;
```
karimnaaji: nit: we might be able to find a better name for `pt`, what about `centroidPoint`?

core/src/data/clientGeoJsonSource.cpp
```diff
@@ -77,6 +78,83 @@ ClientGeoJsonSource::ClientGeoJsonSource(std::shared_ptr<Platform> _platform,
...
+    bool operator()(const geometry::multi_polygon<double>& geom) {
+        float largestArea = 0.f;
+        size_t largestAreaIndex = 0;
+        auto size = geom.size();
+        for (size_t index = 0; index < geom.size(); index++) {
+            auto& g = geom[index];
+            if (g.empty()){
+                continue;
+            }
+            auto area = signedArea(g.front().begin(), g.front().end());
```
karimnaaji: We might want to check for `NaN` here and skip it in that case. Later on we could throw an assert if `operator->()` returns `false` in that case.

core/src/data/clientGeoJsonSource.cpp
tallytalwar: Sorry, missed these. Will address some of these in a followup.

Sure we can, its obvious though? Will change it to be explicit.

core/src/data/clientGeoJsonSource.cpp
tallytalwar: Ohh I only used this because that how it was being used with other piece of code here. But its one and the same thing :D.

core/src/data/clientGeoJsonSource.cpp
tallytalwar: Not sure this is needed, we can not have a `NaN` `signedArea`. So it should be fine here.

Issue #1525, opened 2017-05-30
Proper way to handle Markers on Scene Update


ac87:
I've updated to the latest version and now have the change to fix #1464, so if the scene changes all markers are invalidated/removed, subsequently my test app now crashes toggling raster/vector.

I see the comment from @tallytalwar "document the behavior such that the user handles readding of any marker in the asyncSceneLoad/Update callback."

Before I make my own events to update all markers, does this callback not yet exist (in Android)? or am I looking in the wrong place?

hjanetzek:
Hi @ac87, thanks for the report! There is indeed no good way to recreate markers on scene updates at the moment. I'll look into this now.

hjanetzek:
It probably works if you set a `ViewCompleteListener` after the call to `applySceneUpdates` like this
```
mapController.setViewCompleteListener(new ViewCompleteListener() {
   mapController.removeAllMarkers();
   ... recreate markers ...
});

**Comment by ac87** [comment]
The 0.8.0 update for this is working well for me thanks.

One subtle additional change (in case anyone else was using it) was where I was previously advised to use something like:
 
```
mapController.queueEvent(reloadRunnable);
 
private Runnable reloadRunnable = new Runnable() {
        @Override
        public void run() {
            mapController.loadSceneFile(getBaseScene(isVector), getSceneUpdates());
        }
    };
```

This queuing to the mapcontroller is no longer required and if left in causes concurrent modification exceptions, the scene can now be loaded directly without the queue.

PR #1526, opened 2017-05-30
Scene complete callback

hjanetzek:
https://github.com/tangrams/tangram-es/issues/1525

- this PR enables to recreate markers reliably in SceneLoaderLister.onReady()
  (after applySceneUpdates)

- fix unreleased GlobalRef in case SceneUpdate succeeded, otherwise
  MapController probably never got garbage collected
- update IOS/OSX

### DIFF
--- a/core/include/tangram/map.h
+++ b/core/include/tangram/map.h
@@ -65,20 +65,19 @@ struct SceneUpdate {
 };
 
 enum Error {
+    none,
     scene_update_path_not_found,
     scene_update_path_yaml_syntax_error,
     scene_update_value_yaml_syntax_error,
 };
 
-struct SceneUpdateError {
+struct SceneError {
     SceneUpdate update;
     Error error;
 };
 
-using SceneUpdateErrorCallback = std::function<void(const SceneUpdateError&)>;
-
-// Function type for a mapReady callback
-using MapReady = std::function<void(void*)>;
+// Function type for a sceneReady callback
+using SceneReadyCallback = std::function<void(bool success, const SceneError&)>;
 
 enum class EaseType : char {
     linear = 0,
@@ -99,17 +98,15 @@ public:
     // Any pending scene update will be cleared
     void loadSceneAsync(const char* _scenePath,
                         bool _useScenePosition = false,
-                        MapReady _onMapReady = nullptr,
-                        void *_onMapReadyUserData = nullptr,
-                        const std::vector<SceneUpdate>& sceneUpdates = {},
-                        SceneUpdateErrorCallback _onSceneUpdateError = nullptr);
+                        SceneReadyCallback _onSceneReady = nullptr,
+                        const std::vector<SceneUpdate>& sceneUpdates = {});
 
     // Load the scene at the given absolute file path synchronously
     // Any pending scene update will be cleared
     void loadScene(const char* _scenePath,
                    bool _useScenePosition = false,
-                   const std::vector<SceneUpdate>& sceneUpdates = {},
-                   SceneUpdateErrorCallback _onSceneUpdateError = nullptr);
+                   SceneReadyCallback _onSceneReady = nullptr,
+                   const std::vector<SceneUpdate>& sceneUpdates = {});
 
     // Request an update to the scene configuration; the path is a series of yaml keys
     // separated by a '.' and the value is a string of yaml to replace the current value
@@ -118,7 +115,7 @@ public:
     void queueSceneUpdate(const std::vector<SceneUpdate>& sceneUpdates);
 
     // Apply all previously requested scene updates
-    void applySceneUpdates(SceneUpdateErrorCallback _onSceneUpdateError = nullptr);
+    void applySceneUpdates(SceneReadyCallback _onSceneReady = nullptr);
 
     // Set an MBTiles SQLite database file for a DataSource in the scene.
     void setMBTiles(const char* _dataSourceName, const char* _mbtilesFilePath);
--- a/core/include/tangram/map.h
+++ b/core/include/tangram/map.h
@@ -98,14 +98,12 @@ public:
     // Any pending scene update will be cleared
     void loadSceneAsync(const char* _scenePath,
                         bool _useScenePosition = false,
-                        SceneReadyCallback _onSceneReady = nullptr,
                         const std::vector<SceneUpdate>& sceneUpdates = {});
 
     // Load the scene at the given absolute file path synchronously
     // Any pending scene update will be cleared
     void loadScene(const char* _scenePath,
                    bool _useScenePosition = false,
-                   SceneReadyCallback _onSceneReady = nullptr,
                    const std::vector<SceneUpdate>& sceneUpdates = {});
 
     // Request an update to the scene configuration; the path is a series of yaml keys
@@ -114,8 +112,10 @@ public:
     void queueSceneUpdate(const char* _path, const char* _value);
     void queueSceneUpdate(const std::vector<SceneUpdate>& sceneUpdates);
 
+    void setSceneReadyListener(SceneReadyCallback _onSceneReady);
+
     // Apply all previously requested scene updates
-    void applySceneUpdates(SceneReadyCallback _onSceneReady = nullptr);
+    void applySceneUpdates();
 
     // Set an MBTiles SQLite database file for a DataSource in the scene.
     void setMBTiles(const char* _dataSourceName, const char* _mbtilesFilePath);
--- a/core/include/tangram/map.h
+++ b/core/include/tangram/map.h
@@ -76,8 +76,10 @@ struct SceneError {
     Error error;
 };
 
+using SceneID = int32_t;
+
 // Function type for a sceneReady callback
-using SceneReadyCallback = std::function<void(bool success, const SceneError&)>;
+using SceneReadyCallback = std::function<void(SceneID id, bool success, const SceneError&)>;
 
 enum class EaseType : char {
     linear = 0,
@@ -96,15 +98,15 @@ public:
 
     // Load the scene at the given absolute file path asynchronously
     // Any pending scene update will be cleared
-    void loadSceneAsync(const char* _scenePath,
-                        bool _useScenePosition = false,
-                        const std::vector<SceneUpdate>& sceneUpdates = {});
+    SceneID loadSceneAsync(const char* _scenePath,
+                           bool _useScenePosition = false,
+                           const std::vector<SceneUpdate>& sceneUpdates = {});
 
     // Load the scene at the given absolute file path synchronously
     // Any pending scene update will be cleared
-    void loadScene(const char* _scenePath,
-                   bool _useScenePosition = false,
-                   const std::vector<SceneUpdate>& sceneUpdates = {});
+    SceneID loadScene(const char* _scenePath,
+                      bool _useScenePosition = false,
+                      const std::vector<SceneUpdate>& sceneUpdates = {});
 
     // Request an update to the scene configuration; the path is a series of yaml keys
     // separated by a '.' and the value is a string of yaml to replace the current value
@@ -115,7 +117,7 @@ public:
     void setSceneReadyListener(SceneReadyCallback _onSceneReady);
 
     // Apply all previously requested scene updates
-    void applySceneUpdates();
+    SceneID applySceneUpdates();
 
     // Set an MBTiles SQLite database file for a DataSource in the scene.
     void setMBTiles(const char* _dataSourceName, const char* _mbtilesFilePath);
--- a/core/include/tangram/map.h
+++ b/core/include/tangram/map.h
@@ -108,16 +108,12 @@ public:
                       bool _useScenePosition = false,
                       const std::vector<SceneUpdate>& sceneUpdates = {});
 
+    void setSceneReadyListener(SceneReadyCallback _onSceneReady);
+
     // Request an update to the scene configuration; the path is a series of yaml keys
     // separated by a '.' and the value is a string of yaml to replace the current value
     // at the given path in the scene
-    void queueSceneUpdate(const char* _path, const char* _value);
-    void queueSceneUpdate(const std::vector<SceneUpdate>& sceneUpdates);
-
-    void setSceneReadyListener(SceneReadyCallback _onSceneReady);
-
-    // Apply all previously requested scene updates
-    SceneID applySceneUpdates();
+    SceneID updateSceneAsync(const std::vector<SceneUpdate>& sceneUpdates);
 
     // Set an MBTiles SQLite database file for a DataSource in the scene.
     void setMBTiles(const char* _dataSourceName, const char* _mbtilesFilePath);
--- a/core/include/tangram/map.h
+++ b/core/include/tangram/map.h
@@ -69,6 +69,7 @@ enum Error {
     scene_update_path_not_found,
     scene_update_path_yaml_syntax_error,
     scene_update_value_yaml_syntax_error,
+    no_valid_scene,
 };
 
 struct SceneError {
@@ -79,7 +80,7 @@ struct SceneError {
 using SceneID = int32_t;
 
 // Function type for a sceneReady callback
-using SceneReadyCallback = std::function<void(SceneID id, bool success, const SceneError&)>;
+using SceneReadyCallback = std::function<void(SceneID id, const SceneError*)>;
 
 enum class EaseType : char {
     linear = 0,
--- a/core/include/tangram/map.h
+++ b/core/include/tangram/map.h
@@ -97,24 +97,27 @@ public:
     Map(std::shared_ptr<Platform> _platform);
     ~Map();
 
-    // Load the scene at the given absolute file path asynchronously
-    // Any pending scene update will be cleared
+    // Load the scene at the given absolute file path asynchronously.
     SceneID loadSceneAsync(const char* _scenePath,
                            bool _useScenePosition = false,
                            const std::vector<SceneUpdate>& sceneUpdates = {});
 
     // Load the scene at the given absolute file path synchronously
-    // Any pending scene update will be cleared
     SceneID loadScene(const char* _scenePath,
                       bool _useScenePosition = false,
                       const std::vector<SceneUpdate>& sceneUpdates = {});
 
-    void setSceneReadyListener(SceneReadyCallback _onSceneReady);
-
-    // Request an update to the scene configuration; the path is a series of yaml keys
-    // separated by a '.' and the value is a string of yaml to replace the current value
-    // at the given path in the scene
+    // Request updates to the current scene configuration. This reloads the
+    // scene with the updated configuration.
+    // The SceneUpdate path is a series of yaml keys separated by a '.' and the
+    // value is a string of yaml to replace the current value at the given path
+    // in the scene.
     SceneID updateSceneAsync(const std::vector<SceneUpdate>& sceneUpdates);
+
+    // Set listener for scene load events. The callback receives the SceneID
+    // of the loaded scene and SceneError in case loading was not successful.
+    // The callback may be be called from the main or worker thread.
+    void setSceneReadyListener(SceneReadyCallback _onSceneReady);
 
     // Set an MBTiles SQLite database file for a DataSource in the scene.
     void setMBTiles(const char* _dataSourceName, const char* _mbtilesFilePath);
--- a/core/src/map.cpp
+++ b/core/src/map.cpp
@@ -90,6 +90,8 @@ public:
     float pickRadius = .5f;
 
     std::vector<SelectionQuery> selectionQueries;
+
+    SceneReadyCallback onSceneReady = nullptr;
 };
 
 void Map::Impl::setEase(EaseField _f, Ease _e) {
@@ -181,7 +183,6 @@ void Map::Impl::setScene(std::shared_ptr<Scene>& _scene) {
 // NB: Not thread-safe. Must be called on the main/render thread!
 // (Or externally synchronized with main/render thread)
 void Map::loadScene(const char* _scenePath, bool _useScenePosition,
-                    SceneReadyCallback _onSceneReady,
                     const std::vector<SceneUpdate>& _sceneUpdates) {
 
     LOG("Loading scene file: %s", _scenePath);
@@ -200,17 +201,16 @@ void Map::loadScene(const char* _scenePath, bool _useScenePosition,
 
     }
 
-    if (_onSceneReady) {
+    if (impl->onSceneReady) {
         if (scene->errors.empty()) {
-            _onSceneReady(true, {});
+            impl->onSceneReady(true, {});
         } else {
-            _onSceneReady(false, scene->errors.front());
+            impl->onSceneReady(false, scene->errors.front());
         }
     }
 }
 
 void Map::loadSceneAsync(const char* _scenePath, bool _useScenePosition,
-                         SceneReadyCallback _onSceneReady,
                          const std::vector<SceneUpdate>& _sceneUpdates) {
 
     LOG("Loading scene file (async): %s", _scenePath);
@@ -224,18 +224,18 @@ void Map::loadSceneAsync(const char* _scenePath, bool _useScenePosition,
         nextScene = impl->nextScene;
     }
 
-    runAsyncTask([nextScene, _onSceneReady, _sceneUpdates, this](){
+    runAsyncTask([nextScene, _sceneUpdates, this](){
 
             bool newSceneLoaded = SceneLoader::loadScene(platform, nextScene, _sceneUpdates);
 
-            impl->jobQueue.add([nextScene, newSceneLoaded, _onSceneReady, this]() {
+            impl->jobQueue.add([nextScene, newSceneLoaded, this]() {
                     {
                         std::lock_guard<std::mutex> lock(impl->sceneMutex);
                         if (nextScene == impl->nextScene) {
                             impl->nextScene.reset();
                         } else {
                             // loadScene[Async] was called in the meantime.
-                            if (_onSceneReady) { _onSceneReady(false, {}); }
+                            if (impl->onSceneReady) { impl->onSceneReady(false, {}); }
                             return;
                         }
                     }
@@ -243,10 +243,10 @@ void Map::loadSceneAsync(const char* _scenePath, bool _useScenePosition,
                     if (newSceneLoaded) {
                         auto s = nextScene;
                         impl->setScene(s);
-                        applySceneUpdates(_onSceneReady);
+                        applySceneUpdates();
                     }
 
-                    if (_onSceneReady) { _onSceneReady(newSceneLoaded, {}); }
+                    if (impl->onSceneReady) { impl->onSceneReady(newSceneLoaded, {}); }
                 });
             platform->requestRender();
         });
@@ -262,11 +262,15 @@ void Map::queueSceneUpdate(const std::vector<SceneUpdate>& sceneUpdates) {
     impl->sceneUpdates.insert(impl->sceneUpdates.end(), sceneUpdates.begin(), sceneUpdates.end());
 }
 
+void Map::setSceneReadyListener(SceneReadyCallback _onSceneReady) {
+    impl->onSceneReady = _onSceneReady;
+}
+
 std::shared_ptr<Platform>& Map::getPlatform() {
     return platform;
 }
 
-void Map::applySceneUpdates(SceneReadyCallback _onSceneReady) {
+void Map::applySceneUpdates() {
 
     std::shared_ptr<Scene> nextScene;
     std::vector<SceneUpdate> updates;
@@ -289,38 +293,38 @@ void Map::applySceneUpdates(SceneReadyCallback _onSceneReady) {
         nextScene = impl->nextScene;
     }
 
-    runAsyncTask([nextScene, updates = std::move(updates), _onSceneReady, this](){
+    runAsyncTask([nextScene, updates = std::move(updates), this](){
 
             if (!SceneLoader::applyUpdates(platform, *nextScene, updates)) {
                 LOGW("Scene updates not applied to current scene");
 
-                if (_onSceneReady) {
+                if (impl->onSceneReady) {
                     SceneError err;
                     if (!nextScene->errors.empty()) { err = nextScene->errors.front(); }
-                    _onSceneReady(false, err);
+                    impl->onSceneReady(false, err);
                 }
                 return;
             }
 
             bool configApplied = SceneLoader::applyConfig(platform, nextScene);
 
-            impl->jobQueue.add([nextScene, configApplied, _onSceneReady, this]() {
+            impl->jobQueue.add([nextScene, configApplied, this]() {
                     {
                         std::lock_guard<std::mutex> lock(impl->sceneMutex);
                         if (nextScene == impl->nextScene) {
                             impl->nextScene.reset();
                         } else {
                             // loadScene[Async] was called in the meantime.
-                            if (_onSceneReady) { _onSceneReady(false, {}); }
+                            if (impl->onSceneReady) { impl->onSceneReady(false, {}); }
                             return;
                         }
                     }
                     if (configApplied) {
                         auto s = nextScene;
                         impl->setScene(s);
-                        applySceneUpdates(_onSceneReady);
+                        applySceneUpdates();
                     }
-                    if (_onSceneReady) { _onSceneReady(true, {}); }
+                    if (impl->onSceneReady) { impl->onSceneReady(true, {}); }
                 });
             platform->requestRender();
         });
--- a/core/src/map.cpp
+++ b/core/src/map.cpp
@@ -181,8 +181,8 @@ void Map::Impl::setScene(std::shared_ptr<Scene>& _scene) {
 // NB: Not thread-safe. Must be called on the main/render thread!
 // (Or externally synchronized with main/render thread)
 void Map::loadScene(const char* _scenePath, bool _useScenePosition,
-                    const std::vector<SceneUpdate>& _sceneUpdates,
-                    SceneUpdateErrorCallback _onSceneUpdateError) {
+                    SceneReadyCallback _onSceneReady,
+                    const std::vector<SceneUpdate>& _sceneUpdates) {
 
     LOG("Loading scene file: %s", _scenePath);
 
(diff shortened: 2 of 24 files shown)


core/include/tangram/map.h
```diff
@@ -99,17 +98,15 @@ class Map {
     // Any pending scene update will be cleared
     void loadSceneAsync(const char* _scenePath,
                         bool _useScenePosition = false,
-                        MapReady _onMapReady = nullptr,
-                        void *_onMapReadyUserData = nullptr,
-                        const std::vector<SceneUpdate>& sceneUpdates = {},
-                        SceneUpdateErrorCallback _onSceneUpdateError = nullptr);
+                        SceneReadyCallback _onSceneReady = nullptr,
```
tallytalwar: Note: When these changes get merged and released we have to make sure to update tizen plugin with these updates.

core/src/map.cpp
```diff
@@ -216,30 +224,28 @@ void Map::loadSceneAsync(const char* _scenePath, bool _useScenePosition,
...
     }
 
-    runAsyncTask([nextScene, _onMapReady, _onMapReadyUserData, _sceneUpdates, _onSceneUpdateError, this](){
+    runAsyncTask([nextScene, _onSceneReady, _sceneUpdates, this](){
 
-            bool newSceneLoaded = SceneLoader::loadScene(platform, nextScene, _sceneUpdates, _onSceneUpdateError);
+            bool newSceneLoaded = SceneLoader::loadScene(platform, nextScene, _sceneUpdates);
 
-            impl->jobQueue.add([nextScene, newSceneLoaded, _onMapReady, _onMapReadyUserData, _onSceneUpdateError, this]() {
+            impl->jobQueue.add([nextScene, newSceneLoaded, _onSceneReady, this]() {
```
tallytalwar: Userdata is required in cases like the tizen plugin where we need to pass the maps view handle (`maps_view_h`) to be used in the sceneReady callback.

core/src/map.cpp
```diff
@@ -216,30 +224,28 @@ void Map::loadSceneAsync(const char* _scenePath, bool _useScenePosition,
...
                     {
                         std::lock_guard<std::mutex> lock(impl->sceneMutex);
                         if (nextScene == impl->nextScene) {
                             impl->nextScene.reset();
                         } else {
                             // loadScene[Async] was called in the meantime.
+                            if (_onSceneReady) { _onSceneReady(false, {}); }
```
tallytalwar: Do we need to honor this sceneReady callback, when another/new `loadScene` has been called?

core/src/map.cpp
```diff
@@ -216,30 +224,28 @@ void Map::loadSceneAsync(const char* _scenePath, bool _useScenePosition,
...
                             return;
                         }
                     }
 
                     if (newSceneLoaded) {
                         auto s = nextScene;
                         impl->setScene(s);
-                        applySceneUpdates(_onSceneUpdateError);
```
tallytalwar: Possible user queued certain scene updates post doing a async loadScene, I think this applySceneUpdates was taking care of this scenario.

core/src/map.cpp
```diff
@@ -216,30 +224,28 @@ void Map::loadSceneAsync(const char* _scenePath, bool _useScenePosition,
...
-
-                        if (_onMapReady) {
-                            _onMapReady(_onMapReadyUserData);
-                        }
                     }
+
+                    if (_onSceneReady) { _onSceneReady(newSceneLoaded, {}); }
```
tallytalwar: Why to honor the callback when `newScene` was not loaded?

tallytalwar: Or the aim is for the user to use the `status` of the loadScene/sceneUpdate and do accordingly.

core/src/map.cpp
```diff
@@ -282,31 +288,37 @@ void Map::applySceneUpdates(SceneUpdateErrorCallback _onSceneUpdateError) {
...
 
-            impl->jobQueue.add([nextScene, configApplied, _onSceneUpdateError, this]() {
+            impl->jobQueue.add([nextScene, configApplied, _onSceneReady, this]() {
                     {
                         std::lock_guard<std::mutex> lock(impl->sceneMutex);
                         if (nextScene == impl->nextScene) {
                             impl->nextScene.reset();
                         } else {
                             // loadScene[Async] was called in the meantime.
+                            if (_onSceneReady) { _onSceneReady(false, {}); }
```
tallytalwar: Again why to honor, when another/new loadScene has been called?

core/src/map.cpp
```diff
@@ -282,31 +288,37 @@ void Map::applySceneUpdates(SceneUpdateErrorCallback _onSceneUpdateError) {
...
                             return;
                         }
                     }
-
                     if (configApplied) {
                         auto s = nextScene;
                         impl->setScene(s);
-                        applySceneUpdates(_onSceneUpdateError);
```
tallytalwar: Similar to loadSceneAsync, possible certain SceneUpdates are queued after a `applySceneUpdate` is called.

core/src/map.cpp
```diff
@@ -282,31 +288,37 @@ void Map::applySceneUpdates(SceneUpdateErrorCallback _onSceneUpdateError) {
...
                     }
+                    if (_onSceneReady) { _onSceneReady(true, {}); }
```
tallytalwar: here `success` is always `true` irrespective of the value of `configApplied`

core/include/tangram/map.h
```diff
@@ -65,20 +65,19 @@ struct SceneUpdate {
...
 
 enum Error {
+    none,
     scene_update_path_not_found,
     scene_update_path_yaml_syntax_error,
     scene_update_value_yaml_syntax_error,
 };
 
-struct SceneUpdateError {
+struct SceneError {
```
tallytalwar: Does it make sense to include this in scene.h?

core/src/scene/scene.h
```diff
@@ -17,6 +17,7 @@
 #include "yaml-cpp/yaml.h"
 #include "util/yamlHelper.h"
 
+#include "tangram.h" // SceneError
```
tallytalwar: `map.h`!

core/src/scene/sceneLoader.cpp
```diff
@@ -78,47 +79,37 @@ bool SceneLoader::loadScene(const std::shared_ptr<Platform>& _platform, std::sha
...
         Node value;
 
         try {
             value = YAML::Load(update.value);
-        } catch (YAML::ParserException e) {
+        } catch (const YAML::ParserException& e) {
             LOGE("Parsing scene update string failed. '%s'", e.what());
-            updateError = {update, Error::scene_update_value_yaml_syntax_error};
-            hasError = true;
+            scene.errors.push_back({update, Error::scene_update_value_yaml_syntax_error});
```
tallytalwar: nice, error container should belong to scene 👍

platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
```diff
@@ -141,6 +143,13 @@
         void onSceneUpdateError(SceneUpdateError sceneUpdateError);
     }
 
+    public interface SceneLoadListener {
```
tallytalwar: 👍

core/src/map.cpp
tallytalwar: Ok figured the use of `SceneLoadListener` interface. Makes sense!

platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
```diff
@@ -1267,6 +1312,28 @@ public void onResponse(Call call, Response response) throws IOException {
         return true;
     }
 
+    // Called from JNI on worker or render-thread.
+    void sceneReadyCallback(final boolean success, final SceneUpdateError error) {
+        if (!success && sceneUpdateErrorListener != null) {
```
tallytalwar: Add comment/notes to mark this depricated and to remove when we remove `SceneUpdateErrorListener` in a major releases.

core/src/map.cpp
hjanetzek: Now that I've removed the use of global refs for the updateErrorCallback it's not really needed anymore. Previously calling scene-update from Android created a global ref that had to be released in the SceneErrorCallback function. There was a bug that the ref was not released when the scene-update succeeded.  https://github.com/tangrams/tangram-es/pull/1526/files#diff-3d0c7fc4582b39c4afec2835dcd73f43L134

core/src/map.cpp
hjanetzek: Yea right! There is one problem with reusing the `_onSceneUpdateError` callback function though: When one passes different functions in subsequent calls to `applySceneUpdates` the function that was passed in first is also used for the second call to    `applySceneUpdates`

core/src/map.cpp
hjanetzek: The maps view handle can be bound to the std::function SceneReadyCallback

core/src/map.cpp
tallytalwar: Hmm yeah, thats an issue.

tallytalwar:
Assigned to myself to finish work here for ios.

tallytalwar:
I added the necessary APIs for ios, to match the current behavior.

However I feel might run into issues with a single `SceneReadyCallback` per instance, specially if we have multiple map update calls being triggered and the user wants to perform different set of operations with the `SceneReadyCallback`. I think it might still make sense to pass about a `Callback` with the `loadScene` and `applyUpdates` API calls to take care of this use case.

hjanetzek:
Another way to handle this would be to return the SceneID from loadScene and applyUpdates, then pass the ID as parameter to SceneReadyCallback functions. @tallytalwar does this work for your case?

core/include/tangram/map.h
matteblair: I'd say no - otherwise we'd have to #include the Scene header and put it in the public include folder.

core/src/scene/scene.h
```diff
@@ -140,6 +141,8 @@ class Scene {
     std::atomic_ushort pendingTextures{0};
     std::atomic_ushort pendingFonts{0};
 
+    std::vector<SceneError> errors;
```
matteblair: It seems like we only put a maximum of 1 in the list of errors. Maybe instead of a list it can just be a single error state with a default value of "OK".

matteblair:
@tallytalwar talked this change over with @msmollin today and we're going to take the approach of using a single listener for scene change events and use a unique scene identifier to identify the scene change in each listener invocation, as @hjanetzek suggested. 

@hjanetzek can you implement this change for the core and android interfaces? I also have more comments on the changes so far, I'll add those inline here.

tallytalwar:
Will update the ios interface accordingly.

core/include/tangram/map.h
tallytalwar: Maybe we should do that.. this `map.h` is getting a lot of stuff now. But maybe this work belongs in another refactor PR.

platforms/android/tangram/src/main/cpp/androidPlatform.h
```diff
@@ -11,21 +11,25 @@ void setupJniEnv(JNIEnv* _jniEnv);
...
+struct SceneUpdate;
+struct SceneError;
+using SceneID = int32_t;
 
 void featurePickCallback(jobject listener, const Tangram::FeaturePickResult* featurePickResult);
 void markerPickCallback(jobject listener, jobject tangramInstance, const Tangram::MarkerPickResult* markerPickResult);
 void labelPickCallback(jobject listener, const Tangram::LabelPickResult* labelPickResult);
-void sceneUpdateErrorCallback(jobject updateStatusCallbackRef, const SceneUpdateError& sceneUpdateErrorStatus);
+
+void loadScene(Map& map, const char* cPath, const std::vector<SceneUpdate>& updates);
```
tallytalwar: I think these 2 functions are not used now.

platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
```diff
@@ -128,6 +130,7 @@
     }
 
     /**
+     * @deprecated
```
tallytalwar: We can get rid of these, as will be pushing a newer interface for 0.8.0.

platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
```diff
@@ -732,71 +759,96 @@ public void setPickRadius(float radius) {
...
+    /**
+     * Set a listener for scene update error statuses
+     * @param listener The {@link SceneLoadListener} to call after scene has loaded
+     */
+    public void setSceneLoadListener(final SceneLoadListener listener) {
+        sceneLoadListener = listener;
     }
 
     /**
+     * @deprecated use setSceneLoadListener instead
```
tallytalwar: Can be deleted now.

platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
```diff
@@ -1270,6 +1299,24 @@ public void onResponse(Call call, Response response) throws IOException {
         return true;
     }
 
+    // Called from JNI on worker or render-thread.
+    void sceneReadyCallback(final int sceneId, final boolean success, final SceneUpdateError error) {
```
tallytalwar: Should be setting a `SceneID` for a MapController and check for it to equal to the one set in the callback? To make sure the callback being called is for the apt sceneLoad/sceneUpdate?

platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
```diff
@@ -1270,6 +1299,24 @@ public void onResponse(Call call, Response response) throws IOException {
...
+
```
tallytalwar: Should we be setting a `SceneID` for the `MapController` and check for it to equal to the one set in the callback? To make sure the callback being called is for the apt sceneLoad/sceneUpdate?

tallytalwar:
ios sdk need a small refactor here. So hold on reviewing for ios.

hjanetzek:
@tallytalwar There were some cases where a SceneLoadListener call gets triggered with wrong SceneIDs  

- When a sceneLoad/Update is already in progress applyUpdates did not return the correct new SceneID
- SceneUpdates that were added while a scene is loading (or updates are applied) were automatically applied once the scene finished loading (independent of whether applyUpdate was called)

There is at least one remaining issue when sceneLoad gets called while applyUpdates or loadSceneAsync tasks are in progress.

core/src/map.cpp
```diff
@@ -220,34 +221,29 @@ SceneID Map::loadSceneAsync(const char* _scenePath, bool _useScenePosition,
...
-                        std::lock_guard<std::mutex> lock(impl->sceneMutex);
-                        if (nextScene == impl->nextScene) {
-                            impl->nextScene.reset();
-                        } else {
-                            // loadScene[Async] was called in the meantime.
-                            if (impl->onSceneReady) { impl->onSceneReady(nextScene->id, false, {}); }
-                            return;
-                        }
-                    }
+            // NB: Need to set the scene on the worker thread so that waiting
```
tallytalwar: Glad you have this note here. 👍

matteblair:
Yeah this is starting to look a lot better. A few things remaining:

 1. As Varun mentioned, we should be able to consolidate `queueSceneUpdate` and `applySceneUpdates` into a single function that takes a vector of updates. Since the updates happen async we could call this `updateSceneAsync`. This should allow us to simplify some logic and remove the `sceneUpdates` member of `Map::Impl`.
 2. We may be able to simplify the arguments to `SceneReadyCallback`. Instead of both `bool success` and `const SceneError&` we could provide just a `const SceneError*`, which is `nullptr` when no error occurred.
 3. There are several ways we could resolve the concurrent scene-loading issue:
  A. When `loadScene` is called, mark all async loading tasks as cancelled and run callbacks for them with a "cancelled" error.
  B. Only allow one scene change to run at a time. If a scene load/update is requested before the previous one finishes, log an error and return an invalid scene ID.
  C. Remove the async load/update calls from the core library. Let platform-specific code handle running tasks in the background as needed.

There may be more to address in the platform code, I'm mostly just looking at the core library for now.

tallytalwar:
>A. When loadScene is called, mark all async loading tasks as cancelled and run callbacks for them with a "cancelled" error.
Sounds good to me.

hjanetzek:
1. Sounds good
2. Makes sense - I would go for SceneError::none for success
3. If the issue is that pending scene-load/update calls are not canceled when new scene-loads come in, I think we should postpone handling this case.

core/include/tangram/map.h
```diff
@@ -97,28 +98,22 @@ class Map {
...
-                   bool _useScenePosition = false,
-                   const std::vector<SceneUpdate>& sceneUpdates = {},
-                   SceneUpdateErrorCallback _onSceneUpdateError = nullptr);
+    SceneID loadScene(const char* _scenePath,
+                      bool _useScenePosition = false,
+                      const std::vector<SceneUpdate>& sceneUpdates = {});
+
+    void setSceneReadyListener(SceneReadyCallback _onSceneReady);
 
     // Request an update to the scene configuration; the path is a series of yaml keys
```
tallytalwar: Comments need to be updated here.

platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
```diff
@@ -732,71 +767,73 @@ public void setPickRadius(float radius) {
...
-                    @Override
-                    public void run() {
-                        listener.onFeaturePick(properties, positionX, positionY);
-                    }
-                });
-            }
-        };
+        if (listener == null) {
+            featurePickListener = null;
+        } else {
```
matteblair: Can we do an early return here instead of an `else` to reduce indentation?

platforms/android/tangram/src/main/cpp/jniExports.cpp
```diff
@@ -515,19 +512,12 @@ extern "C" {
             jniEnv->DeleteLocalRef(value);
         }
 
-        if (sceneUpdates.empty()) { return; }
+        // Already checked on the java side
+        if (sceneUpdates.empty()) { return -1; }
```
tallytalwar: It looks like we already handle this in the java side (also from your comment), so probably no use to have this check here.

platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
```diff
@@ -128,17 +130,24 @@
...
+        void onSceneReady(int sceneId);
+
         /**
-         * Receive error status when a scene update failed
+         * Receive error status when a scene load failed
+         * @param sceneId returned by {@link #updateScene(List<SceneUpdate>)} or {@link #loadSceneFile(String, List<SceneUpdate>)}
          * @param sceneUpdateError The  {@link SceneUpdateError} holding error informations
          */
-        void onSceneUpdateError(SceneUpdateError sceneUpdateError);
+        void onSceneError(int sceneId, SceneUpdateError sceneUpdateError);
```
tallytalwar: Correct me if I misunderstood, I think we decided (last week?) to club these 2 methods in a single one and let user handle the functionality based on the error state.

platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
hjanetzek: @tallytalwar setting SceneID where? What's the case you have in mind?

core/src/scene/scene.h
hjanetzek: Yeah, I thought it would be good to collect all errors and warnings from scene loading. For now we can change it to hold a single SceneError

platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
tallytalwar: Sorry I thinking of something else when I mentioned this. Not needed now. Thanks

tallytalwar:
K Updated the ios sdks. Good to go from my testing.

platforms/ios/demo/src/MapViewController.m
```diff
@@ -29,10 +29,9 @@ - (void)mapViewDidCompleteLoading:(TGMapViewController *)mapView
     // [mapView captureScreenshot:YES];
 }
 
-- (void)mapView:(TGMapViewController *)mapView didSceneReady:(int)sceneID Success:(bool)success WithError:(nonnull NSError *)sceneError
+- (void)mapView:(TGMapViewController *)mapView didSceneReady:(int)sceneID WithError:(nonnull NSError *)sceneError
```
karimnaaji: `WithError` doesn't match the code style of the public interface, can you update for `withError`?

`didLoadScene` might sound more natural regarding to objective-c guidelines.

platforms/ios/src/TangramMap/TGMapViewController.h
```diff
@@ -492,8 +491,10 @@ NS_ASSUME_NONNULL_BEGIN
  Tangram will automatically load that remote scene file.
 
  @param path the scene path URL
+
+ @return the integer (SceneID) associated with this scene load.
  */
-- (void)loadSceneFile:(NSString *)path;
+- (int)loadSceneFile:(NSString *)path;
```
karimnaaji: We should mention that these return `-1` on failure.

platforms/ios/src/TangramMap/TGMapViewController.h
```diff
@@ -266,15 +259,19 @@ NS_ASSUME_NONNULL_END
  */
 - (void)mapView:(nonnull TGMapViewController *)view didCaptureScreenshot:(nonnull UIImage *)screenshot;
 
+// TODO: Update Documentation
```
karimnaaji: Can remove the `TODO`.

tallytalwar:
@karimnaaji addressed these changes. Thanks

platforms/ios/src/TangramMap/TGMapViewController.h
```diff
@@ -267,14 +260,17 @@ NS_ASSUME_NONNULL_END
...
+ `-[TGMapViewController loadSceneFileAsync:sceneUpdates:]` or
+ `-[TGMapViewController applySceneUpdate]` is completed.
 
  @param mapView a pointer to the map view
- @param sceneUpdateError a NSError containing information about the scene update that failed
+ @param sceneID sceneID corresponding to which this callback will be called 
+ @param sceneError any errors during scene load or update
  */
-- (void)mapView:(nonnull TGMapViewController *)mapView didFailSceneUpdateWithError:(nonnull NSError *)sceneUpdateError;
+- (void)mapView:(nonnull TGMapViewController *)mapView didLoadScene:(int)sceneID withError:(nonnull NSError *)sceneError;
```
matteblair: Why `nonnull` on the `NSError*`? It seems conventional to yield a `nil` error in these situations (see the async calls in `NSURLSession` for instance: https://developer.apple.com/documentation/foundation/nsurlsession/1411598-downloadtaskwithresumedata?language=objc)

matteblair: In fact it looks like we currently _do_ yield a `nil` when a scene loads without error, which seems contrary to the `nonnull` notation.

platforms/ios/src/TangramMap/TGHelpers.mm
```diff
@@ -36,10 +36,14 @@ + (TGError)convertTGErrorTypeFrom:(Tangram::Error)error
...
             return TGErrorSceneUpdateValueYAMLSyntaxError;
+        case Tangram::Error::no_valid_scene:
+            return TGErrorNoValidScene;
+        case Tangram::Error::none:
+            return TGErrorNone;
     }
 }
 
-+ (NSError *)errorFromSceneUpdateError:(Tangram::SceneUpdateError)updateError
++ (NSError *)errorFromSceneError:(Tangram::SceneError)updateError
```
matteblair: The `NSError` produced here could be a lot more useful. We should do our best to provide values for the conventional `userInfo` keys like `NSLocalizedDescriptionKey`, `NSLocalizedFailureReasonErrorKey`, and `NSLocalizedRecoverySuggestionErrorKey` (from http://nshipster.com/nserror/).

platforms/ios/src/TangramMap/TGMapViewController.h
```diff
@@ -522,36 +524,21 @@ NS_ASSUME_NONNULL_BEGIN
...
-- (void)queueSceneUpdates:(NSArray<TGSceneUpdate *> *)sceneUpdates;
-
-/**
- Apply scene updates queued with `-queueSceneUpdate*` methods.
  If a scene update error happens, scene udpates won't be applied.
+
+ @return the integer (SceneID) associated with this scene update or -1 if scene can not be updated.
  */
-- (void)applySceneUpdates;
+- (int)updateScene:(NSArray<TGSceneUpdate *> *)sceneUpdates;
```
matteblair: This should probably be called `updateSceneAsync` to remain consistent with `loadSceneFileAsync`.

tallytalwar: I had it called that, but modified it in line with what we have for [android](https://github.com/tangrams/tangram-es/blob/scene-complete-callback/platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java#L308).

platforms/ios/src/TangramMap/TGHelpers.mm
tallytalwar: Will read on this. I did not modify any of this code, to keep it in strict to the current task in hand. But will update.

platforms/ios/src/TangramMap/TGMapViewController.h
tallytalwar: Make sense. Will update.

platforms/ios/src/TangramMap/TGHelpers.mm
matteblair: Yeah that's fine, we don't have to fix it in this PR but I wanted to note the issue.

platforms/ios/src/TangramMap/TGMapViewController.h
matteblair: Yeah I'm not sure the naming on `MapController` methods is perfect either, but in that context there's no distinction between sync and async methods, here there is.

tallytalwar: Naming is hard!

I am not sure, accordingly to me in both the context it will always be async. I would be good keeping both the platform API consistent. Will update both `ios` and `android` to `updateSceneAsync`. Will also document that these are async tasks, I do not see it documented in doc comments for both platforms.

platforms/android/tangram/src/main/java/com/mapzen/tangram/MapView.java
```diff
@@ -18,93 +18,35 @@
...
-                callback.onMapReady(mapController);
-            }
-
-            @Override
-            protected void onCancelled(Boolean ok) {
-                mapInstance.dispose();
-            }
-
-        }.execute();
+    public MapController getMap(MapController.SceneLoadListener listener) {
```
matteblair: It feels weird that calling this method requires a `SceneLoadListener` but does not trigger a scene load - i.e. there's no need for the `MapController` to have a `SceneLoadListener` set at this point.

matteblair: I'm not sure that it's bad - passing in `null` is acceptable so it's not limiting in any way - it just feels weird.

platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
```diff
@@ -262,31 +266,57 @@ public void run() {
...
+     * Load a new scene file asynchronously.
      * If scene updates triggers an error, they won't be applied.
+     * Use {@link #setSceneLoadListener(SceneLoadListener)} for notification when the new scene is
+     * ready.
      * @param path Location of the YAML scene file within the application assets
      * @param sceneUpdates List of {@code SceneUpdate}
+     * @return Scene ID
      */
-    public void loadSceneFile(String path, List<SceneUpdate> sceneUpdates) {
+    public int loadSceneFile(String path, List<SceneUpdate> sceneUpdates) {
```
matteblair: `loadSceneFileAsync` seems like a clearer name - at a minimum it will reduce confusion for people upgrading from older Tangram versions.

platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
```diff
@@ -262,31 +266,57 @@ public void run() {
...
+
+    /**
+     * Apply SceneUpdates to the current scene.
+     * If a updates trigger an error, scene updates won't be applied.
+     * Use {@link #setSceneLoadListener(SceneLoadListener)} for notification when the new scene is
+     * ready.
+     * @param sceneUpdates List of {@code SceneUpdate}
+     * @return new scene ID
+     */
+    public int updateScene(List<SceneUpdate> sceneUpdates) {
```
matteblair: `updateSceneAsync` makes sense to me as well.

platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
```diff
@@ -262,31 +266,57 @@ public void run() {
         });
     }
 
-    static MapController getInstance(GLSurfaceView view) {
-        return new MapController(view);
-    }
 
     /**
-     * Load a new scene file
+     * Load a new scene file.
```
matteblair: += "asynchronously."

platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
```diff
@@ -262,31 +266,57 @@ public void run() {
...
+     * Use {@link #setSceneLoadListener(SceneLoadListener)} for notification when the new scene is
+     * ready.
      * @param path Location of the YAML scene file within the application assets
+     * @return Scene ID
```
matteblair: "An identifier for the scene being loaded, the same value will be passed to `SceneLoadListener.onSceneReady` when loading is complete." Does that sound right?

platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
tallytalwar: I agree it does, but at the same time if we only offer async scene loading or scene updating don't you think `Async` is not needed in the name? Just saying, I am open to either name.

matteblair: We will thank ourselves later if we add synchronous versions!

platforms/ios/src/TangramMap/TGMapViewController.h
tallytalwar: Updated name aptly.

platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
tallytalwar: Added synchronous scene loading methods for android platform. Updated names aptly and added apt documentation.

platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
tallytalwar: Updated!

platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
tallytalwar: Updated!

platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
tallytalwar: Yup, I have something along these lines on ios. Will add this for android also.

platforms/ios/src/TangramMap/TGMapViewController.h
tallytalwar: Made `sceneError` here `nullable`! Totally made sense.

platforms/ios/src/TangramMap/TGHelpers.mm
tallytalwar: Updated here: d5a7d16.

I think what error info I added make sense for current context. Will need to update this when we handle more error states in Tangram. Feel free to update the language if need be.

platforms/ios/src/TangramMap/TGHelpers.mm
```diff
@@ -49,8 +49,13 @@ + (NSError *)errorFromSceneError:(Tangram::SceneError)updateError
     NSString* value = [NSString stringWithUTF8String:updateError.update.value.c_str()];
     TGSceneUpdate* udpate = [[TGSceneUpdate alloc] initWithPath:path value:value];
```
karimnaaji: This is unused now, I think it would still be useful to somehow leave it in the `userInfo` error to know which update failed.

platforms/ios/src/TangramMap/TGMapViewController.mm
```diff
@@ -153,29 +153,41 @@ - (void)loadSceneFile:(NSString *)path sceneUpdates:(NSArray<TGSceneUpdate *> *)
...
-        if (!self.mapViewDelegate || ![self.mapViewDelegate respondsToSelector:@selector(mapView:didFailSceneUpdateWithError:)]) { return; }
-        [self.mapViewDelegate mapView:self didFailSceneUpdateWithError:[TGHelpers errorFromSceneUpdateError:sceneUpdateError]];
+    auto updateCallbackStatus = [=](int sceneID, auto sceneError) {
+        if (!self.mapViewDelegate || ![self.mapViewDelegate respondsToSelector:@selector(mapView:didLoadScene:withError:)]) { return; }
+
+        NSError* error = nil;
+        if (sceneError) {
+            error = [TGHelpers errorFromSceneError:*sceneError];
+        }
+        [self.mapViewDelegate mapView:self didLoadScene:sceneID withError:error];
```
karimnaaji: Missing removal of markers and render request here.

platforms/ios/src/TangramMap/TGMapViewController.mm
```diff
@@ -187,49 +199,41 @@ - (void)loadSceneFileAsync:(NSString *)path sceneUpdates:(NSArray<TGSceneUpdate
         }
     }
 
-    auto updateCallbackStatus = [=](auto sceneUpdateError) {
-        if (!self.mapViewDelegate || ![self.mapViewDelegate respondsToSelector:@selector(mapView:didFailSceneUpdateWithError:)]) { return; }
-        [self.mapViewDelegate mapView:self didFailSceneUpdateWithError:[TGHelpers errorFromSceneUpdateError:sceneUpdateError]];
-    };
-
-    self.map->loadSceneAsync([path UTF8String], false, onReadyCallback, nullptr, updates, updateCallbackStatus);
+    self.map->setSceneReadyListener(onReadyCallback);
```
karimnaaji: We could define this for convenience:
```
self.map->setSceneReadyListener([self sceneReadyListener]);
```
With a method defined as such
```
- (Tangram::SceneReadyCallback) sceneReadyListener {
    auto listener = [=](int sceneID, auto sceneError) {
        // ...
    }
    return listener;
}
```
And use it on line 235, and 166 too so we don't miss marker removal and render request from the listener.

platforms/ios/src/TangramMap/TGHelpers.mm
tallytalwar: Ahh, right, I should use that in `NSLocalizedDescriptionKey`.

platforms/ios/src/TangramMap/TGMapViewController.mm
tallytalwar: @karimnaaji can you check these updates now.

karimnaaji: Looks good!

PR #1527, opened 2017-05-30
Deploy iOS framework from CI

matteblair:
This is a major change to our iOS release process. Instead of building framework bundles locally and pushing them to GitHub for hosting, a CircleCI build server will run the framework builds and push them to an S3 bucket for hosting. This will make our [framework hosting repo](https://github.com/tangrams/ios-framework) unnecessary. The new process is described in the updated [iOS release checklist](https://github.com/tangrams/tangram-es/blob/deploy-ios-framework/platforms/ios/release-checklist.md).

TODO:
- [x] Combine iOS and Android release checklists, since we only do combined releases now.
- [x] Use better naming for release archives.
- [x] For release archives, produce debug and release variants to store on S3.


circle.yml
```diff
@@ -19,3 +22,18 @@ test:
       zip -r ${CIRCLE_ARTIFACTS}/framework.zip TangramMap.framework; fi
     - if [ $CIRCLE_BRANCH = 'master' ]; then cd build/ &&
       zip -r ${CIRCLE_ARTIFACTS}/docs.zip ios-docs; fi
+deployment:
+  snapshots:
+    branch: [master, deploy-ios-framework]
+    commands:
+      # Create a zip archive of files that will serve as the Cocoapod source.
+      - zip -r ${CIRCLE_ARTIFACTS}/pod.zip build/ios-framework-universal/Release/TangramMap.framework LICENSE
+      - aws s3 cp ${CIRCLE_ARTIFACTS}/pod.zip s3://ios.mapzen.com/tangram-es-snapshots/${CIRCLE_BUILD_NUM}.zip
```
msmollin: Are we worried about there being a ton of these snapshots sitting in the bucket?

matteblair: A little bit, but for Android we push snapshots at the same frequency and it hasn't been an issue. We actually used to do this for Android on *every* branch, which created a bonkers number of snapshots and I never heard complaints from our AWS bill.

msmollin: Cool. Looks good to me otherwise.

scripts/circle/build_cocoapods_archive.sh
```diff
@@ -0,0 +1,16 @@
...
+set -e
+set -o pipefail
+
+# Create a zip archive of files that will serve as the Cocoapods source.
+
+# Build the universal binary framework.
+make ios
+
+# First add the license file.
+zip ${CIRCLE_ARTIFACTS}/pod.zip LICENSE
```
karimnaaji: The pod will also need a README (the one that will be displayed on https://cocoapods.org/pods/Tangram-es). We can pick the one that was previously used in the ios-framework repo, since the one we have on the iOS platform folder is not really relevant for direct user of the pod.

platforms/ios/release-checklist.md
```diff
@@ -5,19 +5,40 @@
...
-4. Once #5's PR merges to master, tag the release with the same version number as #2.
-5. Update the Tangram-es.podspec file version number with that same tag, and run `pod spec lint` to
-make sure everything is happy. Fix issues if not happy (eventually we should document known possible
-issues). Submit to PR once lint is clean.
-6. Once #6 PR merges, push the updated pod spec to trunk: `pod trunk push Tangram-es.podspec`
-7. Submit a PR to increment the version number for #1's plist file. We add `-dev` to the end of the
-version numbers while in development.
-8. Before publishing the Tangram release, add the zipped version of both Debug and Release flavors
-to the Download attachments of the release.
+Steps to release the Tangram ES iOS Framework to Cocoapods:
```
karimnaaji: Should we mention the steps for the documentation updates?

scripts/circle/build_cocoapods_archive.sh
matteblair: Oh good catch, I'll add that.

platforms/ios/release-checklist.md
matteblair: We don't have that for Android either.. maybe we can add instructions for both platforms in another PR.

matteblair:
Ok, I think this is ready for another round of review @karimnaaji. We'll see for sure whether everything is working when it's merged to master.

circle.yml
```diff
@@ -5,17 +5,42 @@ machine:
...
+      # Upload the Cocoapods archive to S3.
+      - aws s3 cp ${CIRCLE_ARTIFACTS}/pod.zip s3://ios.mapzen.com/tangram-snapshots/tangram-snapshot-${CIRCLE_BUILD_NUM}.zip
+      - aws s3 cp ${CIRCLE_ARTIFACTS}/pod.zip s3://ios.mapzen.com/tangram-latest.zip
+  releases:
+    # For any tag of the form 1, 1.2, 1.2.3, etc. we will deploy a release build.
+    tag: /[0-9]+(\.[0-9]+)*/
+    owner: tangrams
+    commands:
+      # Build the framework in release mode and package it into pod.zip
+      - make ios RELEASE=1
```
karimnaaji: We might hit an annoying case here, if we have an error within the sample app this will fail and we wouldn't have catch it from the `test` step. What we could do is change that for `make ios-framework-universal RELEASE=1` or also build the demo app on the `test` step.

I think that was the reason we had 
```
if [ $CIRCLE_BRANCH = 'master' ]; then make ios; else make ios TANGRAM_IOS_FRAMEWORK_SLIM=1; fi
```
above; so that only master builds the demo app while having faster CI builds on iOS for branches.

karimnaaji: Just double-checked the steps with the option `TANGRAM_IOS_FRAMEWORK_SLIM`, and it builds the demo app, so all good here.

PR #1532, opened 2017-06-02
Enable curving for some complex scripts

tallytalwar:
- remaining are Arabic, Mongolian and Burmese (burmese has other known issues so not enabling it)


tallytalwar:
Still playing around with angle thresholds to get a good value for these complex scripts to avoid overlapping glyphs.

tallytalwar:
Reviving this.

I need some validation for the remaining scripts before I enable these for curving. Played around with a bunch of indic scripts and they pass very gracefully in all my test cases present in OSM as of right now. But I am in process of adding hindi at more critical points for testing this.

@hjanetzek I just commented out the language set you had right now, can you give this a quick look and merge?

Issue #1540, opened 2017-06-16
libtangram.so error


JunsuLime:
#00 pc 0000000000224764  /data/app/com.test.map-2/lib/arm64/libtangram.so (_ZN7Tangram15DynamicQuadMeshINS_12SpriteVertexEE4drawERNS_11RenderStateERNS_13ShaderProgramEib+148)
#01 pc 00000000002246a0  /data/app/com.test.map-2/lib/arm64/libtangram.so (_ZN7Tangram10PointStyle16onBeginDrawFrameERNS_11RenderStateERKNS_4ViewERNS_5SceneE+120)
#02 pc 00000000001756f4  /data/app/com.test.map-2/lib/arm64/libtangram.so (_ZN7Tangram3Map6renderEv+896)
#03 pc 0000000000a19d14  /data/app/com.test.map-2/oat/arm64/base.odex (offset 0x960000)

I execute method that

1) remove marker and draw marker
2) using mapController, set position, zoom, tilt, rotation
3) marker.setVisible(bool)

one method contains 1,2,3 step all.

And error message above is sended.

My Tangram version is 0.7.1 and device OS version is 7.1.1, device: Nexus 5X and other device such as SAMSUNG edge 6s, Nexus 6P and other device have same problem.

This problem does not happen every time, but often come up.

hjanetzek:
Hi @JunsuLime thanks for the report!  Can you share the snippet for the function that triggers the crash? What do you mean by 'remove and draw marker' exactly?

JunsuLime:
case1)

// At this code, error happened.
```java
mapController.removeMarker(selectedMarker)

for (Marker marker : myPinMarkers) {
  if (marker.isVisible()) marker.setVisible(false);
}

for (Marker marker : otherPinMarkers) {
  mapController.removeMarker(marker);
} 

for (Pin pin : trip.getPins()) {
  Marker marker = getMarker(U.SELECT_STATUS_OTHER_PIN);
  marker.setPoint(new LngLat(pin.getLng(), pin.getLat()));
  otherPinMarkers.add(marker);
}
```

```java
public Marker getMarker(int status) {
   marker = mapController.addMarker();
   marker.setStylingFromString(U.clickEffectStyle);
   marker.setDrawable(U.selectedCircleDrawable);
   marker.setDrawOrder(0);
   return marker;
}
```


case2)

// At this code, error happened.
```java
// this code's input lngLat is come from gps location
// and I converted location -> LngLat
[...]

PR #1546, opened 2017-06-21
Fix marker bugs

hjanetzek:
Fix bug https://github.com/tangrams/tangram-es/issues/1540


core/src/marker/markerManager.cpp
```diff
@@ -60,32 +64,19 @@ bool MarkerManager::remove(MarkerID markerID) {
...
-    // Build the feature mesh for the marker's current geometry.
-    buildGeometry(*marker, m_zoom);
-
-    return true;
-}
-
-bool MarkerManager::setStylingFromPath(MarkerID markerID, const char* path) {
-    Marker* marker = getMarkerOrNull(markerID);
-    if (!marker) { return false; }
+    m_dirty = true;
```
tallytalwar: 👍

Issue #1511, opened 2017-05-18
Convert Objective-C CGPoint struct parameters to pointer refs in shouldRecognize delegate callbacks


msmollin:
This allows the touch point location to be modified before processing, enabling scenarios like changing the location of the pinch gesture while location tracking is enabled.

https://github.com/tangrams/tangram-es/blob/master/platforms/ios/src/TangramMap/TGMapViewController.h#L60

karimnaaji:
Resolved via #1548

PR #1548, opened 2017-06-21
Add rotation and pinch focus delegates

karimnaaji:
Alternative for https://github.com/tangrams/tangram-es/issues/1511. Instead of giving a pointer to the location struct for all gesture delegates to allow for modification in the gesture pipeline, explicitly add two delegates `pinchFocus` and `rotationFocus`.

Usage example:
```objective-c
- (CGPoint)pinchFocus:(TGMapViewController *)view recognizer:(UIGestureRecognizer *)recognizer {
    TGGeoPoint currentLocation;
    CGPoint currentLocationOnScreen = [view lngLatToScreenPosition:currentLocationLngLat];
    // the position is within viewport
    if (!isnan(currentLocationOnScreen.x) && !isnan(currentLocationOnScreen.y)) {
        return currentLocationOnScreen;
    }
    // default behavior
    return [recognizer locationInView:view.view];
}
```

### DIFF
--- a/platforms/ios/src/TangramMap/TGMapViewController.h
+++ b/platforms/ios/src/TangramMap/TGMapViewController.h
@@ -53,6 +53,7 @@ NS_ASSUME_NONNULL_BEGIN
 /**
  Whether the map view should handle a single tap gesture.
 
+ @param view the map view attached to the recognizer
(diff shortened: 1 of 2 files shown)


msmollin:
That's a much better approach I think.

Issue #1512, opened 2017-05-18
Expose iOS gesture recognizers publicly to allow for modification


msmollin:
Currently the gesture recognizer chain is internal and no reference to them is maintained, which makes it difficult to either disable or override certain behaviors. Implementing developers may for example want to disable simultaneous gesture recognition or setup dependency chains so if a pinch gesture is recognizing we shouldn't go into a shove or a pan. We should maintain publicly accessible references to the gesture recognizers to allow for this.
https://github.com/tangrams/tangram-es/blob/master/platforms/ios/src/TangramMap/TGMapViewController.mm#L531

There are probably some sensible defaults to setup for this as well. Open to thoughts on that.

karimnaaji:
Resolved via #1550.

PR #1550, opened 2017-06-21
iOS gesture recognizers

karimnaaji:
Publicly expose iOS gesture recognizers to enable user modifications.

Fixes https://github.com/tangrams/tangram-es/issues/1512

### DIFF
--- a/platforms/ios/src/TangramMap/TGMapViewController.h
+++ b/platforms/ios/src/TangramMap/TGMapViewController.h
@@ -375,21 +375,44 @@ NS_ASSUME_NONNULL_BEGIN
  */
 @property (strong, nonatomic) NSURL* resourceRoot;
 
+/// The selector that must be used as an action on the tap gesture recognizer
 @property (readonly, nonatomic) SEL respondToTapGestureAction;
+/// The selector that must be used as an action on the double tap gesture recognizer
 @property (readonly, nonatomic) SEL respondToDoubleTapGestureAction;
+/// The selector that must be used as an action on the pan gesture recognizer
 @property (readonly, nonatomic) SEL respondToPanGestureAction;
+/// The selector that must be used as an action on the pinch gesture recognizer
 @property (readonly, nonatomic) SEL respondToPinchGestureAction;
+/// The selector that must be used as an action on the rotation gesture recognizer
 @property (readonly, nonatomic) SEL respondToRotationGestureAction;
+/// The selector that must be used as an action on the shove gesture recognizer
 @property (readonly, nonatomic) SEL respondToShoveGestureAction;
+/// The selector that must be used as an action on the long press gesture recognizer
 @property (readonly, nonatomic) SEL respondToLongPressGestureAction;
 
-/// Must be set after viewDidLoad in the view lifecycle to override default
+/// The following recognizers must be set after viewDidLoad in the view lifecycle in order to override default.
+/// Only one instance of the recognizer type will be used on the view at once.
+/// Adding any recognizer removes the default or previoulsy added recognizer on the map view.
+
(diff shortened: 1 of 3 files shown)


platforms/ios/src/TangramMap/TGMapViewController.h
```diff
@@ -343,6 +343,46 @@ NS_ASSUME_NONNULL_BEGIN
  */
 @property (strong, nonatomic) NSURL* resourceRoot;
 
+/// The selector that must be used as an action on the tap gesture recognizer
+@property (readonly, nonatomic) SEL respondToTapGestureAction;
```
msmollin: Vera nice.

platforms/ios/src/TangramMap/TGMapViewController.mm
```diff
@@ -535,56 +535,135 @@ - (void)setCameraType:(TGCameraType)cameraType
...
+    [self.view addGestureRecognizer:_pinchGestureRecognizer];
+    [self.view addGestureRecognizer:_rotationGestureRecognizer];
+    [self.view addGestureRecognizer:_shoveGestureRecognizer];
+    [self.view addGestureRecognizer:_longPressGestureRecognizer];
+}
 
-    //7. Long press
-    UILongPressGestureRecognizer* longPressRecognizer = [[UILongPressGestureRecognizer alloc]
-                                                         initWithTarget:self action:@selector(respondToLongPressGesture:)];
+- (void) setTapGestureRecognizer:(UITapGestureRecognizer *)recognizer
```
msmollin: Minor nitpick: I think the rest of the file is styled like
`- (void)setTapGestureRecognizer`
whereas you have a space after the `)` .

platforms/ios/src/TangramMap/TGMapViewController.mm
```diff
@@ -535,56 +535,135 @@ - (void)setCameraType:(TGCameraType)cameraType
...
-    [self.view addGestureRecognizer:doubleTapRecognizer];
-    [self.view addGestureRecognizer:panRecognizer];
-    [self.view addGestureRecognizer:pinchRecognizer];
-    [self.view addGestureRecognizer:rotationRecognizer];
-    [self.view addGestureRecognizer:shoveRecognizer];
-    [self.view addGestureRecognizer:longPressRecognizer];
+- (void) setDoubleTapGestureRecognizer:(UITapGestureRecognizer *)recognizer
+{
+    if (_doubleTapGestureRecognizer) {
+        [self.view removeGestureRecognizer:_doubleTapGestureRecognizer];
```
msmollin: A thought: this code conceivably allows people to nil out the recognizers entirely. Is that something we want to allow?

karimnaaji: It is within the block `NS_ASSUME_NONNULL_BEGIN`, would this still be allowed? I thought this would give a compile or warning error. 
Otherwise I think we shouldn't allow for it to be `nil`; the correct way to cancel a gesture recognition is to use the delegate and `shouldRecognize*` return whether or not the gesture would be allowed.

msmollin: That's *just* for swift to understand what you're trying to do. So yes, from Swift, because of those annotations we wouldn't be able to nil out the recognizers. However, you could still do it from Objective-C haha :) 

https://developer.apple.com/swift/blog/?id=25

msmollin: I think it would be fine to document it as such, but if you want you can enforce it at code level on the recognizer setters.

msmollin:
👍  LGTM. Not letting me approve the review for whatever reason....

PR #1558, opened 2017-07-03
Update harfbuzz

hjanetzek:


matteblair:
@hjanetzek Notice any changed behavior from this?

hjanetzek:
@matt it looks all correct with this Burmese font http://www.rfa.org/burmese/help/font_download_english.html

tallytalwar:
Ok I am not seeing the Burmese issue on android now on master. But both `master` and this branch shows issue on ios. Looks like @hjanetzek initial guess of wrong font being picked on ios could be the issue here.

Irrespective its good to update harfbuzz.

hjanetzek:
@tallytalwar @matteblair have you tried the ios font for Burmese on osx? I guess the same font used on ios is available there. If it works the problem could be with extracting the font header/tables in TGFontConverter.mm

tallytalwar:
@hjanetzek I can check that by explicitly bundling this font in the ios demo app. Stay tuned.

tallytalwar:
Merging this in, but investigation for burmese on ios still continues.

tallytalwar:
@hjanetzek with respect to your suggestion of trying out `Myanmar Sangam MN` font for osx demo.

I found the font file (`ttc`) sitting here: `/Library/Fonts/Myanmar Sangam MN.ttc`. Used https://transfonter.org/ttc-unpack to convert this `ttc` to `ttf`. And following is the result on osx demo:

screen shot 2017-07-07 at 3 33 51 pm

So possible its the font file on ios/osx!

PR #1560, opened 2017-07-04
Markers issue #1540 

hjanetzek:


core/src/map.cpp
```diff
@@ -379,17 +379,23 @@ bool Map::update(float _dt) {
...
         }
 
+        // FIXME: Set of labels needs to be updated when markers were added or removed.
+        // Otherwise Labels::m_labels may point to deleted Label instances.
+#if 0
         if (impl->view.changedOnLastUpdate() ||
-            impl->tileManager.hasTileSetChanged()) {
-
+            impl->tileManager.hasTileSetChanged() ||
+            markersNeedUpdate) {
```
tallytalwar: Maybe make sense to add a `dirty` state to `MarkerManager` and check that instead of running these updates every Tangram::update!

tallytalwar: Something on the lines of: 36a1372e70e9d4df3e45eefa30453cd708581273

hjanetzek:
Closing in favor of #1546

PR #1575, opened 2017-07-18
Get property value as a string instead of just the string properties

### LINKED ISSUE #1574: Allow non-string feature properties to be used for `text_source` parameter
Currently non-string feature properties are not allowed to be used in `text_source` parameter which is a divergent behavior from JS tangram engine.

Refer: https://github.com/tangrams/tangram-es/blob/master/core/src/style/textStyleBuilder.cpp#L174

tallytalwar:
fixes #1574

### DIFF
--- a/core/include/tangram/data/properties.h
+++ b/core/include/tangram/data/properties.h
@@ -8,6 +8,9 @@ namespace Tangram {
 class Value;
 struct PropertyItem;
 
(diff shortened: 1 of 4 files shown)


tallytalwar:
@hjanetzek added another commit to take care of JS function returning a numeric feature property. Can you review and merge?

PR #1576, opened 2017-07-18
Fix API key injection in desktop and RPi demos

matteblair:
Desktop demo apps now check for a MAPZEN_API_KEY environment variable at run-time and apply it to the scene file. If this variable isn't found, the app continues but prints a warning with instructions to get an API key and use it with the app.

### DIFF
--- a/platforms/common/glfwApp.cpp
+++ b/platforms/common/glfwApp.cpp
@@ -29,6 +29,8 @@ std::string sceneFile;
 std::string markerStylingPath = "layers.touch.point.draw.icons";
 std::string polylineStyle = "{ style: lines, interactive: true, color: red, width: 20px, order: 5000 }";
 
+std::string mapzenApiKey;
+
 GLFWwindow* main_window = nullptr;
 Tangram::Map* map = nullptr;
 int width = 800;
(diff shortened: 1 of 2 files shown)


platforms/common/glfwApp.cpp
```diff
@@ -64,16 +66,20 @@ void create(std::shared_ptr<Platform> p, std::string f, int w, int h) {
...
-    LOG("Environment variable MAPZEN_API_KEY not set. Kindly set this environment variable and relaunch.");
-    exit(1);
-#endif
+    char* mapzenApiKeyEnvVar = getenv("MAPZEN_API_KEY");
+    if (mapzenApiKeyEnvVar && strlen(mapzenApiKeyEnvVar) > 0) {
+        mapzenApiKey = mapzenApiKeyEnvVar;
+    } else {
+        LOGW("No API key found!\n\nMapzen data sources require an API key. "
+             "Sign up for a free key at http://mapzen.com/developers and then set it from the command line with: "
+             "\n\n\texport MAPZEN_API_KEY=YOUR_KEY_HERE\n");
```
tallytalwar: Should we exit the demo app here?

I guess the decision on how, user sees nothing on map vs app just exiting on the user. I might lean towards exiting!

tallytalwar: Also if a user is using xcode/clion to launch the app from osx, they need to set the environment variable through `launchctl`. So maybe just mention to set the environment variable and refer the readme link here, which has this described.

matteblair: The app might be loading a scene that doesn't use Mapzen data sources at all, in which case it's totally fine that there's no key.

matteblair: Good point, that could be confusing. Maybe it's easier to just link to https://github.com/tangrams/tangram-es/blob/master/platforms/osx/README.md

tallytalwar: Yeah thats a good point.

tallytalwar: Yup thats what I would expect. Once this gets updated will merge this PR.

PR #1577, opened 2017-07-19
Load scene from string

hjanetzek:
Scene strings can now be tested like this:
```
 ./tangram -s "{ import: [ refill-style.yaml, themes/color-blue.yaml] }" /home/jeff/work/tangram-es/styles/refill/
```

Added ios interfaces.

### DIFF
--- a/platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
+++ b/platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
@@ -331,7 +331,7 @@ public class MapController implements Renderer {
     }
 
     /**
-     * Load a new scene file asynchronously.
+     * Load a new scene synchronously, provided an explicit yaml scene string to load
      * If scene updates triggers an error, they won't be applied.
      * Use {@link #setSceneLoadListener(SceneLoadListener)} for notification when the new scene is
      * ready.
@@ -344,6 +344,25 @@ public class MapController implements Renderer {
         String[] updateStrings = bundleSceneUpdates(sceneUpdates);
         checkPointer(mapPointer);
         int sceneId = nativeLoadSceneYaml(mapPointer, yaml, resourceRoot, updateStrings);
+        removeAllMarkers();
+        requestRender();
+        return sceneId;
+    }
+
+    /**
+     * Load a new scene asynchronously, provided an explicit yaml scene string to load
+     * If scene updates triggers an error, they won't be applied.
+     * Use {@link #setSceneLoadListener(SceneLoadListener)} for notification when the new scene is
+     * ready.
+     * @param yaml YAML scene String
+     * @param resourceRoot base path to resolve relative URLs
+     * @param sceneUpdates List of {@code SceneUpdate}
+     * @return Scene ID An identifier for the scene being loaded, the same value will be passed to
+     */
+    public int loadSceneYamlAsync(String yaml, String resourceRoot, List<SceneUpdate> sceneUpdates) {
+        String[] updateStrings = bundleSceneUpdates(sceneUpdates);
+        checkPointer(mapPointer);
+        int sceneId = nativeLoadSceneYamlAsync(mapPointer, yaml, resourceRoot, updateStrings);
         removeAllMarkers();
         requestRender();
         return sceneId;
@@ -1141,6 +1160,7 @@ public class MapController implements Renderer {
     private synchronized native int nativeLoadScene(long mapPtr, String path, String[] updateStrings);
     private synchronized native int nativeLoadSceneAsync(long mapPtr, String path, String[] updateStrings);
     private synchronized native int nativeLoadSceneYaml(long mapPtr, String yaml, String resourceRoot, String[] updateStrings);
+    private synchronized native int nativeLoadSceneYamlAsync(long mapPtr, String yaml, String resourceRoot, String[] updateStrings);
     private synchronized native void nativeSetupGL(long mapPtr);
     private synchronized native void nativeResize(long mapPtr, int width, int height);
     private synchronized native boolean nativeUpdate(long mapPtr, float dt);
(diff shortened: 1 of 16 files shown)


platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
```diff
@@ -298,6 +297,26 @@ public int loadSceneFile(String path, List<SceneUpdate> sceneUpdates) {
...
+     * Load a new scene file asynchronously.
+     * If scene updates triggers an error, they won't be applied.
+     * Use {@link #setSceneLoadListener(SceneLoadListener)} for notification when the new scene is
+     * ready.
+     * @param yaml YAML scene String
+     * @param resourceRoot base path to resolve relative URLs
+     * @param sceneUpdates List of {@code SceneUpdate}
+     * @return Scene ID
+     */
+    public int loadSceneYaml(String yaml, String resourceRoot, List<SceneUpdate> sceneUpdates) {
```
tallytalwar: With my ios updates will also add a synchronous API call for this.

tallytalwar:
Ok this is good from ios side.

Added the necessary interface for `loadSceneYaml` and `loadSceneYamlAsync`. Also did some cleanup.

hjanetzek:
@tallytalwar looks good - thanks!

platforms/ios/src/TangramMap/TGMapViewController.h
```diff
@@ -530,6 +532,32 @@ NS_ASSUME_NONNULL_BEGIN
...
+ to be applied to be applied to the scene.
+ If a scene update error happens, scene updates won't be applied.
+
+ @param yaml Yaml scene string
+ @param resourceRoot Base path to resolve relative urls within the yaml scene string
+ @param sceneUpdates a list of `TGSceneUpdate` to apply to the scene
+
+ @return the integer (SceneID) associated with this scene load or -1 if scene can not be loaded.
+*/
+- (int)loadSceneYaml:(NSString *)yaml resourceRoot:(NSString *)resourceRoot sceneUpdates:(NSArray<TGSceneUpdate *> *)sceneUpdates;
```
karimnaaji: Both methods can be renamed:
```
- (int)loadSceneFromYaml:(NSString *)yaml resourceRoot:(NSString *)resourceRoot updates:(NSArray<TGSceneUpdate *> *)updates
```
 to be compliant with objective-c guidelines.

platforms/ios/src/TangramMap/TGMapViewController.h
```diff
@@ -530,6 +532,32 @@ NS_ASSUME_NONNULL_BEGIN
 - (int)loadSceneFileAsync:(NSString *)path sceneUpdates:(NSArray<TGSceneUpdate *> *)sceneUpdates;
 
 /**
+ Loads a scene synchronously using an explicitly specified yaml string, with a list of updates
+ to be applied to be applied to the scene.
```
karimnaaji: `to be applied` twice.

platforms/ios/src/TangramMap/TGMapViewController.h
```diff
@@ -530,6 +532,32 @@ NS_ASSUME_NONNULL_BEGIN
...
+ @param resourceRoot Base path to resolve relative urls within the yaml scene string
+ @param sceneUpdates a list of `TGSceneUpdate` to apply to the scene
+
+ @return the integer (SceneID) associated with this scene load or -1 if scene can not be loaded.
+*/
+- (int)loadSceneYaml:(NSString *)yaml resourceRoot:(NSString *)resourceRoot sceneUpdates:(NSArray<TGSceneUpdate *> *)sceneUpdates;
+
+/**
+ Loads a scene asynchronously using an explicitly specified yaml string, with a list of updates
+ to be applied to be applied to the scene.
```
karimnaaji: `to be applied` twice.

platforms/ios/src/TangramMap/TGMapViewController.h
karimnaaji: We could have a convenience method defined as such:
```
- (int)loadSceneFromYaml:(NSString *)yaml updates:(NSArray<TGSceneUpdate *> *)updates;
```
where `resourceRoot` defaults to the `Resource` application folder (`[[NSBundle mainBundle] resourceURL]`).

platforms/ios/src/TangramMap/TGMapViewController.h
```diff
@@ -264,6 +264,8 @@ NS_ASSUME_NONNULL_END
  `-[TGMapViewController loadSceneFile]` or
  `-[TGMapViewController loadSceneFileAsync:]` or
  `-[TGMapViewController loadSceneFileAsync:sceneUpdates:]` or
+ `-[TGMapViewController loadSceneFileYaml:sceneUpdates:]` or
```
karimnaaji: Missing `resourceRoot` from the description.

platforms/ios/src/TangramMap/TGMapViewController+Internal.h
```diff
@@ -20,7 +20,9 @@ NS_ASSUME_NONNULL_BEGIN
 
 - (void)clearDataSource:(std::shared_ptr<Tangram::TileSource>)tileSource;
 
-- (Tangram::SceneReadyCallback)setSceneReadyListener;
+- (Tangram::SceneReadyCallback)sceneReadyListener;
+
+- (std::vector<Tangram::SceneUpdate>)unpackSceneUpdates:(NSArray<TGSceneUpdate *> *)sceneUpdates;
```
karimnaaji: No need to add those to the internal header if they are not needed anywhere else than in the implementation file. They should be defined [here](https://github.com/tangrams/tangram-es/blob/89988a8241870a050bd4f8594b927b507ffc75aa/platforms/ios/src/TangramMap/TGMapViewController.mm#L26) instead.

platforms/ios/src/TangramMap/TGMapViewController.h
matteblair: I'll do that in https://github.com/tangrams/tangram-es/pull/1599 after merging this

matteblair: Will do this also in #1599

matteblair: Actually I won't be adding this - for consistency we'd need both sync and async versions and that's getting the API a little pudgy in my opinion.

nvkelso:
🎉

PR #1582, opened 2017-07-21
Fix render request on Linux and RPI

hjanetzek:
Credits to emacs+rtags highlighting of active defines!

### DIFF
--- a/platforms/common/glfwApp.cpp
+++ b/platforms/common/glfwApp.cpp
@@ -54,6 +54,15 @@ Tangram::MarkerID polyline = 0;
 
 bool keepRunning = true;
 
+void loadSceneFile() {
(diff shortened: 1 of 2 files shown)


tallytalwar:
I think @matteblair found same thing yesterday in https://github.com/tangrams/tangram-es/compare/fix-rpi.

Maybe you can cherry-pick the scene file loading refactoring to his branch.

Otherwise this is all good. 👍

matteblair:
Yep, that was the primary bug in the RPi demo actually. I resolved it by creating a `rpiPlatform.cpp` separate from `linuxPlatform.cpp`, obviating the need for that `#define`.

hjanetzek:
ok

Issue #1569, opened 2017-07-15
Windows & wxWidgets platform


tehKaiN:
Hi there,

I've ported tangram-es to Windows during this weekend and I'd like to contribute changes, but I'm not quite certain how I should do this.

Although I've tried to keep it clean as possible, some changes may be breaking for other platforms. Unfortunately I don't have environment to test other platforms, so I guess direct merge to master would be inappropriate 'cuz I won't be able to fix everything by myself.

My windows branch is here: https://github.com/tehKaiN/tangram-es/tree/platform-windows

Also, I'd like to create tangram-es wxWidgets widget. After it's done, should I contribute it later to this repo as additional platform or keep it on separate repo?

matteblair:
Hey this looks great! 👏 

Let's start with the Windows platform, I'll have more questions about wxWidgets I think. 

I just created a branch in this repo called `windows-platform`: https://github.com/tangrams/tangram-es/tree/windows-platform. You should be able to make a pull request from the branch in your fork onto our branch. Then we can test it out and quickly adjust anything that might break other platforms :)

Question about your build configuration: It looks like you have CMake searching for libCurl using `findpackage`. As far as I know Curl is not typically installed on Windows, did you have to install Curl in a particular way for CMake to be able to find it?

tehKaiN:
Ok, here it is. As of curl, I've added it's path to CMAKE_PREFIX_PATH as findpackage uses it to find packages. :)

tehKaiN:
So I did some work on wxWidgets support. https://github.com/tehKaiN/tangram-es/tree/platform-wxwidgets

Perhaps it would be wise to just have DLL platform for use with any other framework/lib/whatever, but still linux/rpi platforms are more like app stubs to begin development with.

wxtangram.cpp/.h implements wxTangram widget which inherits from wxGLCanvas widget. Main file containing app startup and widget usage is yet to be commited. This is still far from complete 'cuz many basic event handlers analogous to glfwApp are missing. Honestly, IDK if they should be included or externally implemented by devs, but I guess it's not really gonna be immutable and read-only map control implementation, but just app stub.

tehKaiN:
So right now I'm trying to compile it using Windows ABI since I have to link my project with msvc-compiled DLLs. MSVC is too poor for compiling tangram-es, since it doesn't support many cpp features. Tested under build tools 2015 and 2017 (aka 14.0 and 14.1).

[...]

PR #1585, opened 2017-07-25
Platform windows

tehKaiN:
So, here's a pull request from my code mentioned here: https://github.com/tangrams/tangram-es/issues/1569

Not sure if it won't break other platforms though.


matteblair:
Woo! Thanks, I'll go ahead an merge this to our branch so we can tweak it for other platforms and set up CI.

PR #1589, opened 2017-07-28
Fix ios demo with cmake 3.9.0

tallytalwar:
Refer the following for details on the issue with prior versions of cmake
https://gitlab.kitware.com/cmake/cmake/issues/16680
https://cmake.org/pipermail/cmake/2016-August/064125.html
https://cmake.org/cmake/help/v3.9/prop_sf/MACOSX_PACKAGE_LOCATION.html

### DIFF
--- a/toolchains/iOS.cmake
+++ b/toolchains/iOS.cmake
@@ -7,6 +7,15 @@ set(TANGRAM_FRAMEWORK ${PROJECT_SOURCE_DIR}/${TANGRAM_FRAMEWORK})
 set(EXECUTABLE_NAME "tangram")
 set(FRAMEWORKS CoreGraphics CoreFoundation QuartzCore UIKit OpenGLES Security CFNetwork GLKit)
 
+# NB:cmake versions before 3.9.0 had an issue where the path specified for MACOSX_FRAMEWORK_LOCATION
+# was prepended by "../" when set to something other than "Resources" which is what we require here.
+# Refer: https://gitlab.kitware.com/cmake/cmake/issues/16680 for the actual cmake issue
+if(${CMAKE_VERSION} VERSION_LESS "3.9.0")
+    set(MACOSX_FRAMEWORK_LOCATION ${EXECUTABLE_NAME}.app/Frameworks)
+else()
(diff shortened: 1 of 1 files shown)


matteblair:
The documents you've linked to all discuss `MACOSX_PACKAGE_LOCATION` but the variable that's changed in the diff is `MACOSX_FRAMEWORK_LOCATION`. How did you determine that these are related?

tallytalwar:
`MACOSX_FRAMEWORK_LOCATION` is a variable I am introducing which is then used to set `MACOSX_PACKAGE_LOCATION`. Refer: https://github.com/tangrams/tangram-es/pull/1589/files#diff-d7b4e80a2e5d82144c05afabb3e015b6L40

matteblair:
Do you know which version of CMake is used on CircleCI? It would be nice to be able to test this automatically going forward.

tallytalwar:
I am not sure, the docs mention `2.8` which is the default on a trusty image! But it definitely has to be at least `3.0` since thats our minimum requirement and our builds do not complain.

matteblair:
From the first link on Gitlab, it sounds like
```
set(MACOSX_FRAMEWORK_LOCATION Resources/Frameworks)
```
would have the desired result for CMake versions before and after 3.9, unless I misunderstand the comments there.

matteblair:
If a single value works then we should use that instead of branching by version. Either way it would be good to have a comment here documenting that there is some care required with this property, and maybe a link to that gitlab issue.

tallytalwar:
I am not sure `MACOSX_PACKAGE_LOCATION Resources/Frameworks` will work for both 3.8.x and 3.9.0. From the same gitlab link:

>The 'Resources' value seems to have a special meaning on iOS and the files are actually placed in the bundle root. But if use 'Resources/AnotherFolder', it gets prepended with '../' as well.

tallytalwar:
Also gave it a try and 3.8.x errors with missing framework:

3.8.x:
screen shot 2017-07-28 at 5 43 46 pm

tallytalwar:
Yeah I updated to include a comment (which I had previously mentioned in the commit comments and in the PR). Anyways I do not think there is any other way around this, so we have to include this version check! :|

PR #1594, opened 2017-08-01
Naming convention for debug version of iOS snapshots

karimnaaji:
Make sure to reflect the build type of available s3 frameworks in the naming of available frameworks on ios.mapzen.com.


msmollin:
LGTM

PR #1595, opened 2017-08-01
Enable 'tms' flag for data sources

### LINKED ISSUE #971: Support TMS-style tiled data sources
JS Work: https://github.com/tangrams/tangram/pull/401
More information: https://alastaira.wordpress.com/2011/07/06/converting-tms-tile-coordinates-to-googlebingosm-tile-coordinates/

matteblair:
Follows the behavior described in https://github.com/tangrams/tangram/issues/359

A simple change, but I haven't yet found a data source to test it with (the one used in the issue above appears dead now).

Resolves #971

### DIFF
--- a/core/src/scene/sceneLoader.cpp
+++ b/core/src/scene/sceneLoader.cpp
@@ -1066,6 +1066,11 @@ void SceneLoader::loadSource(const std::shared_ptr<Platform>& platform, const st
         isMBTilesFile = urlLength > extLength && (url.compare(urlLength - extLength, extLength, extStr) == 0);
     }
 
+    bool isTms = false;
+    if (auto tmsNode = source["tms"]) {
+        getBool(tmsNode, isTms);
+    }
+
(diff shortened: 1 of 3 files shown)


core/src/scene/sceneLoader.cpp
```diff
@@ -1066,6 +1066,11 @@ void SceneLoader::loadSource(const std::shared_ptr<Platform>& platform, const st
         isMBTilesFile = urlLength > extLength && (url.compare(urlLength - extLength, extLength, extStr) == 0);
     }
 
+    bool isTms = false;
+    if (auto tmsNode = source["tms"]) {
```
tallytalwar: Probably `const auto&` will be better here. Adding!

matteblair: Mmmm we've been back and forth on this with YAML Nodes; I think the conclusion we reached is that Nodes should be passed by value, since they themselves are just a reference to underlying data.

PR #1596, opened 2017-08-02
Fix: Error on when path for scene update is malformed or incorrect

tallytalwar:
A node path in a yaml tree is "invalid" when it is either malformed (missing apt delimiter) or has missing missing nodes in between a path.

A path is however still "valid" if the last node in the path is missing, at which point during a scene update will get added to the scene config.

### DIFF
--- a/core/src/scene/sceneLoader.cpp
+++ b/core/src/scene/sceneLoader.cpp
@@ -97,9 +97,10 @@ bool SceneLoader::applyUpdates(const std::shared_ptr<Platform>& platform, Scene&
             try {
                 // Dummy node to trigger YAML exception on YAML syntax errors
                 auto parse = YAML::Load(update.path);
-                Node node = YamlPath(update.path).get(root);
+                Node node = Node();
+                auto validPath = YamlPath(update.path).get(root, node);
 
-                if (node && node.Scalar().empty() && node != root) {
+                if (!validPath) {
                     scene.errors.push_back({update, Error::scene_update_path_not_found});
                     return false;
                 } else {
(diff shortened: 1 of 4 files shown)


core/src/scene/sceneLoader.cpp
```diff
@@ -98,9 +98,10 @@ bool SceneLoader::applyUpdates(const std::shared_ptr<Platform>& platform, Scene&
             try {
                 // Dummy node to trigger YAML exception on YAML syntax errors
                 auto parse = YAML::Load(update.path);
-                Node node = YamlPath(update.path).get(root);
+                bool validPath = false;
+                auto node = YamlPath(update.path).get(root, validPath);
 
-                if (node && node.Scalar().empty() && node != root) {
+                if (!validPath || (node && node.Scalar().empty() && node != root)) {
```
tallytalwar: @karimnaaji can you elaborate on `node && node.Scalar().empty() && node != root`

tallytalwar:
Hold on this, doing some refactoring and adding couple of unit tests.

core/src/scene/sceneLoader.cpp
karimnaaji: Discussed on slack, this test is similar to `validPath` and can be removed. For convenience, the interface to test `YamlPath` validity could be changed for that:

```c++
bool YamlPath::get(YAML::Node node, Node* node);
```
so it's used as such:
```c++
Node n;
if (YamlPath(update.path, &n).get()) {
   // valid
} else {
  // invalid
}```

**Comment by tallytalwar** [comment]
Ok this is much better now. Will update with some unit test tomorrow am.

**Comment by matteblair** [comment]
Going to add some tests for this and then merge.

File: core/src/scene/sceneLoader.cpp [hunk 2]
```diff
@@ -94,21 +94,12 @@ bool SceneLoader::applyUpdates(const std::shared_ptr<Platform>& platform, Scene&
         }
 
         if (value) {
-            try {
-                // Dummy node to trigger YAML exception on YAML syntax errors
-                auto parse = YAML::Load(update.path);
```
tallytalwar: I think we will not hit a yaml syntax error without this case. Its good to inform the user of this error.

matteblair: This check isn't useful - the coded path for a yaml node need not be valid yaml. It might be useful to check whether the individual tokens in the path are valid yaml, but this doesn't check that.

PR #1597, opened 2017-08-02
Add Proguard configuration

### LINKED ISSUE #1537: Proguard for Android
It would be nice to have a maintained proguard file for release builds on android that is updated when https://github.com/tangrams/tangram-es/blob/master/platforms/android/tangram/src/main/cpp/platform_android.cpp changes.

Here is the configuration I've been using: [proguard-tangram.txt](https://github.com/tangrams/tangram-es/files/1078429/proguard-tangram.txt)

matteblair:
Resolves #1537 

We now need to be diligent about keeping `tangram-proguard-rules.txt` up to date if anything is renamed in Java or accessed by name in C++.


matteblair:
Update: Rather than listing all of the code members to keep in the minify process, we can use the [Keep annotation](https://developer.android.com/reference/android/support/annotation/Keep.html) to denote this information directly on the declarations of the members.

This seems like a better solution to me, because it doesn't add another place that we need to update when refactoring.

(The native methods of `MapController` had `-keep` rules in the proguard file, but no `@Keep` annotations here. The proguard tool seems to be smart enough to not rename `native` methods, you can verify in the proguard mapping output that the identifiers are unchanged.)

matteblair:
Also: the easiest way to test proguard behavior is to add `minifyEnabled true` in `android.buildTypes.debug` in the demo app `build.gradle`.

hjanetzek:
Yes, lets merge this after scene-complete and add the missing annotations then

PR #1598, opened 2017-08-02
Android tools update

matteblair:
**Updates the Gradle wrapper version from 2.14.1 to 3.3 and the Gradle Android build plugin version from 2.2.3 to 2.3.3**

This enables "instant run", which re-deploys apps super fast after small changes, and makes C++ build output visible in the Android Studio console. It also prevents the native debugger from stopping at breakpoints on some machines (read: Varun's), while still working correctly on other machines (read: mine). More data points here will be useful.

**Updates the target Android SDK to 26**

O! O! Android O!

**Update support-annotations dependency version from 25.0.0 to 25.3.1**

Seems like good hygiene.

### DIFF
--- a/platforms/android/build.gradle
+++ b/platforms/android/build.gradle
@@ -5,7 +5,7 @@ allprojects {
       jcenter()
     }
     dependencies {
-      classpath 'com.android.tools.build:gradle:2.2.3'
+      classpath 'com.android.tools.build:gradle:2.3.3'
       classpath 'com.github.dcendents:android-maven-gradle-plugin:1.5'
     }
   }
--- a/platforms/android/gradle/wrapper/gradle-wrapper.properties
+++ b/platforms/android/gradle/wrapper/gradle-wrapper.properties
@@ -1,6 +1,6 @@
-#Wed Sep 14 20:10:31 EDT 2016
+#Wed Aug 02 15:25:27 PDT 2017
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-2.14.1-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-3.3-all.zip
--- a/platforms/android/tangram/build.gradle
+++ b/platforms/android/tangram/build.gradle
@@ -7,12 +7,12 @@ version = VERSION_NAME
 apply from: 'versioning.gradle'
 
 android {
-  compileSdkVersion 25
-  buildToolsVersion '25.0.2'
+  compileSdkVersion 26
+  buildToolsVersion '25.0.3'
 
   defaultConfig {
     minSdkVersion 15
-    targetSdkVersion 25
+    targetSdkVersion 26
     versionCode buildVersionCode()
     versionName VERSION_NAME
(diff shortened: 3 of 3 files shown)


tallytalwar:
Its correct that native debugging stops working for for me when updating to the latest gradle plugin for android studio.

A note for gathering more data points: make sure to update `ndk`, `cmake` and other sdk tools along with making sure your android studio is updated to latest `2.3.3`, before trying this change.

tallytalwar:
Ok as per https://developer.oculus.com/documentation/mobilesdk/latest/concepts/mobile-studio-debug/, we need to specify a symbols directory path (**absolute path**) to the architecture `so` files generated from the native build.

Once I added this path, native debugging started working for me.
screen shot 2017-08-02 at 5 50 07 pm

As a note in the previous version I (along with other members of the team) had to include a path which includes the our native symbols, but it used to work recursively so something like `/Users/varuntalwar/Development/tangram-es/platforms/android/` used to work.

matteblair:
One unintended side effect of updating the Gradle build plugin is that our Android build logs have become super long. The CMake build output is printed to the Gradle log now and some of our dependencies are generating _excessive_ amounts of warnings (I'm looking at you hb-icu-freetype). We should silence some of the warnings on 3rd party libraries - otherwise the build log is far too noisy to be useful.

karimnaaji:
Worked all good for me.

matteblair:
Hmm so there's definitely a build issue on TravisCI when we use NDK version 15. From some reading I strongly suspect that it has to do with [Unified Headers](https://android.googlesource.com/platform/ndk/+/ndk-r15-release/docs/UnifiedHeaders.md), which were enabled by default in NDK 15. I'm not yet sure exactly what condition causes the error and why it only happens on Travis (locally I can build with NDK 15 successfully).

matteblair:
@tallytalwar I'm ready to merge this now. Did you find out anything about the NDK r15 build issues with SQLite? If there's a solution that allows us to build with the newer NDK then I'd rather do that.

tallytalwar:
@matteblair Nopes updating to newer sqlitecpp did not resolve the issue with 15c ndk version. The only way I could get android builds to run for the time being was to set the deprecated headers cmake option `-DANDROID_DEPRECATED_HEADERS=ON` in the gradle config:

```
externalNativeBuild {
      cmake {
        targets 'tangram'
        arguments '-DPLATFORM_TARGET=android',
                  '-DANDROID_TOOLCHAIN=clang',
                  '-DANDROID_STL=c++_shared',
                  '-DANDROID_DEPRECATED_HEADERS=ON'
        cppFlags '-std=c++14',
                 '-pedantic',
                 '-fPIC',
                 '-fexceptions',
                 '-frtti',
                 //warnings
                 '-Wall',
                 '-Wignored-qualifiers',
                 '-Wtype-limits',
                 '-Wmissing-field-initializers'
```

Reference: https://android.googlesource.com/platform/ndk/+/ndk-r15-release/docs/UnifiedHeaders.md
Sqlite3 issue on android ndk project: https://github.com/android-ndk/ndk/issues/477

tallytalwar:
@sarahlensing it will be good for you also to give this a shot. Thanks.

tallytalwar:
Included a commit, to include deprecated headers for fix sqlite build errors with ndk 15.2 or higher. As a note, this is a temporary solution and we will have to address the main issue at a later date. Will create a followup issue for this.

PR #1599, opened 2017-08-04
iOS tweaks

matteblair:
- Cleaning up naming and documentation to be more Cocoa-ey
- Fixes a potential bug where TGMapViewController would fail to remove markers after a scene load if mapViewDelegate is nil

### DIFF
--- a/platforms/ios/src/TangramMap/TGMapViewController.h
+++ b/platforms/ios/src/TangramMap/TGMapViewController.h
@@ -260,17 +260,25 @@ NS_ASSUME_NONNULL_END
 - (void)mapView:(nonnull TGMapViewController *)view didCaptureScreenshot:(nonnull UIImage *)screenshot;
 
 /**
- If set by the user, this is called after a
- `-[TGMapViewController loadSceneFile]` or
- `-[TGMapViewController loadSceneFileAsync:]` or
- `-[TGMapViewController loadSceneFileAsync:sceneUpdates:]` or
- `-[TGMapViewController loadSceneFileYaml:resourceRoot:sceneUpdates:]` or
- `-[TGMapViewController loadSceneFileYamlAsync:resourceRoot:sceneUpdates:]` or
- `-[TGMapViewController applySceneUpdate]` is completed.
+ Called after a scene has been loaded or updated.
 
- @param mapView a pointer to the map view
- @param sceneID sceneID corresponding to which this callback will be called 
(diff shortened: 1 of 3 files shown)


platforms/ios/src/TangramMap/TGMapViewController.h
```diff
@@ -296,8 +302,9 @@ NS_ASSUME_NONNULL_BEGIN
  an API key</a> and load it through your application:
 
  ```swift
- let sceneURL = "https://mapzen.com/carto/walkabout-style-more-labels/walkabout-style-more-labels.yaml";
- view.loadSceneFile(sceneURL, sceneUpdates: [ TGSceneUpdate(path: "sources.mapzen.url_params", value: "{ api_key: \(YOUR_API_KEY) }") ]);
+ let sceneURL = URL("https://mapzen.com/carto/walkabout-style-more-labels/walkabout-style-more-labels.yaml");
+ let sceneUpdates = [ TGSceneUpdate(path: "sources.mapzen.url_params", value: "{ api_key: \(YOUR_API_KEY) }") ];
+ view.loadScene(from: sceneURL, with: sceneUpdates);
```
tallytalwar: Ok I do not know this at all. But a reference for how objective-c to swift naming transitions work will be good here.

platforms/ios/src/TangramMap/TGMapViewController.mm
```diff
@@ -131,10 +131,10 @@ - (void)clearDataSource:(std::shared_ptr<Tangram::TileSource>)tileSource
 
 - (Tangram::SceneReadyCallback) sceneReadyListener {
     return [=](int sceneID, auto sceneError) {
-        if (!self.mapViewDelegate || ![self.mapViewDelegate respondsToSelector:@selector(mapView:didLoadScene:withError:)]) { return; }
-
         [self.markersById removeAllObjects];
```
karimnaaji: Good catch, we would benefit from leaving`self.renderRequested = YES;` before checking whether a `mapViewDelegate` is set.

platforms/ios/src/TangramMap/TGMapViewController.h
matteblair: https://developer.apple.com/library/content/documentation/Swift/Conceptual/BuildingCocoaApps/InteractingWithObjective-CAPIs.html#//apple_ref/doc/uid/TP40014216-CH4-ID35

PR #1600, opened 2017-08-04
fix: use GetStringUTFLength when converting Java UTF16 to UTF8 C-strings

cleeus:
The Android JNI bindings query jniEnv->GetStringLength to determine the size of the allocated std::string but then copy UTF8 bytes which could be longer should non-ASCII input occur. jniEnv->GetStringUTFLength should be used instead.

### DIFF
--- a/platforms/android/tangram/src/main/cpp/androidPlatform.cpp
+++ b/platforms/android/tangram/src/main/cpp/androidPlatform.cpp
@@ -141,7 +141,7 @@ void onUrlFailure(JNIEnv* _jniEnv, jlong _jCallbackPtr) {
 }
 
 std::string stringFromJString(JNIEnv* jniEnv, jstring string) {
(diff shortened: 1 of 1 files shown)


cleeus:
Hi all,

the travis-ci build fails with:

> To run the demo, you must provide a Mapzen API key as an environment    variable named 'MAPZEN_API_KEY' or a Gradle property named 'mapzenApiKey'. Expression: (apiKey != null). Values: apiKey = null

so nothing related to this patch

tallytalwar:
@cleeus  

Thanks for the PR (number 1600 :D). This is good.

Yeah CI builds fail for any external contribution precisely because of the reason you mentioned. I created #1602 with your commit which takes care of the CI. Again thanks for catching this issue.

tallytalwar:
Closing this in favor of #1602

PR #1602, opened 2017-08-07
fix: use GetStringUTFLength when converting Java UTF16 to UTF8 C-strings

tallytalwar:
Internal PR to resolve CI for #1600 

Thanks @cleeus for catching this issue.


matteblair:
This is certainly _more_ correct than before, but it makes me realize that there's more we need to do to ensure correct string conversion from Java to C++.

Java strings are normally stored as UTF16 and `GetStringUTFChars` actually returns a string in _modified UTF8_: https://developer.android.com/training/articles/perf-jni.html#UTF_8_and_UTF_16_strings

To correctly support the full Unicode range, we should be converting strings to UTF8 when going from Java to C++ and converting to UTF16 when going from C++ to Java (example: https://github.com/mapbox/jni.hpp/blob/master/include/jni/string.hpp#L28-L32).

matteblair:
Superseded by #1628

PR #1606, opened 2017-08-09
OSX Demo Revamp

### LINKED ISSUE #1571: Produce MacOS desktop app bundle on CI
To easily test any version of the Tangram ES core library, it will be helpful for our CI builds to include a MacOS desktop target. We can store the output on an S3 bucket so that the build artifacts are easily cataloged and publicly accessible.

matteblair:
**Display Tangram ES version in GLFW demo title bar**

screen shot 2017-08-09 at 11 59 58 am

 **Add menu shortcuts for managing scene file and API key**

screen shot 2017-08-09 at 11 58 28 am

 - **API Key...** opens a dialog box that allows you to edit and save a Mapzen API key. This key is saved in user preferences and used for every version of the Tangram ES app.
 - **Edit Scene** opens the current scene file/URL using the system default application for the file type.
 - **Open Scene...** opens a file-open panel where you can pick a scene file to load.

**Add MacOS demo builds to CircleCI config**

 - Hosted on ios.mapzen.com

Resolves #1571

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -32,7 +32,7 @@ deployment:
       - aws s3 cp ${CIRCLE_ARTIFACTS}/pod.zip s3://ios.mapzen.com/tangram-snapshots/tangram-snapshot-debug-${CIRCLE_BUILD_NUM}.zip
       - aws s3 cp ${CIRCLE_ARTIFACTS}/pod.zip s3://ios.mapzen.com/tangram-latest-debug.zip
       # Build the MacOS demo app and package it into demo.zip
-      - make -j osx
+      - make osx -j MACOSX_DEPLOYMENT_TARGET=10.10.0
       - cd build/osx/bin && zip -r ${CIRCLE_ARTIFACTS}/demo.zip tangram.app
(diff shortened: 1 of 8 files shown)


bcamper:
woo!

matteblair:
Maybe we should host all of our Tangram builds on `tangram.mapzen.com`?

bcamper:
That could be nice. Not a blocker though.

tallytalwar:
One thing good to have here will be include "build number" along with the version number. Since we will be deploying this for every master build I suppose.

karimnaaji:
@matteblair we need to set the flag `-mmacosx-version-min` in the build toolchain. I'm getting this error while running otherwise

screen shot 2017-08-10 at 10 32 49

matteblair:
@karimnaaji Clearly you need to update your OS ;)

I rebuilt the demo targeting 10.10, see if it works now - the linker spat out some warnings.

Is macOS 10.10 a reasonable target? Is anyone using an older version?

PR #1614, opened 2017-08-15
More OSX demo tweaks

matteblair:
A few more conveniences for the OSX demo app
 - If the current scene file is from an HTTP URL, the **Edit Scene** menu item will open it in a browser.
 - The menu now has a **Reload Scene** item (this already worked via GLFW keyboard input, but having it in the menu is nicer UX).

TODO:
 - ~Relative resource URLs in scene files don't seem to work with the Open panel.~ _fixed_
 - ~View stdout (maybe log it to a file, then it's easy to send with a bug report)~ _won't fix here_
 - ~Open HTTP URLs (maybe support pasting URLs)~ _won't fix_
 - ~Scenes without `global.sdk_mapzen_api_key` fail to load because the scene update fails.~ _won't fix here_
 - ~Double-click should zoom the map instead of centering.~ _fixed_
 - ~Should be able to paste an API key into the alert box.~ _fixed_

### DIFF
--- a/platforms/common/glfwApp.cpp
+++ b/platforms/common/glfwApp.cpp
@@ -251,10 +251,15 @@ void mouseButtonCallback(GLFWwindow* window, int button, int action, int mods) {
 
     if ((time - last_time_released) < double_tap_time) {
         // Double tap recognized
-        Tangram::LngLat p;
-        map->screenPositionToLngLat(x, y, &p.longitude, &p.latitude);
-        map->setPositionEased(p.longitude, p.latitude, 1.f);
-
+        const float duration = 0.5f;
+        Tangram::LngLat tapped, current;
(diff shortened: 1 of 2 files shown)


matteblair:
Resolves #1648

matteblair:
Resolves https://github.com/tangrams/tangram-es/issues/1649

Supporting keyboard shortcuts for copy and paste was shockingly complex. In the future I'd prefer to implement the API Key interface with a cross-platform UI library like Dear Imgui, which would handle this kind of trickiness for us.

tallytalwar:
Yeah it does make sense to have a single platform agnostic desktop app. However as discussed with Matt IRL, some of the integration like file dialog, etc will be more user-facing if done using the platform specific libraries. But more later.

PR #1615, opened 2017-08-15
blend: translucent option

hjanetzek:


bcamper:
I was imagining this would be `blend: translucent` rather than a new `blend_translucent: true` keyword -- does that not fit the pipeline here?

hjanetzek:
Ah I thought we want to allow combination with other blending functions for add and multiply. I'll change it to `blend: translucent` option. Then we can add `translucent-multiply` as needed.

bcamper:
Well let's wait for discussion here :)

I can see the usefulness of translucent labels, for example (I am less concerned about supporting combination of translucent/multiply/add etc.).

nvkelso:
Is this more a 2nd pass compositing operation for an entire group of features that share the same style? So all the features using that style are rendered, and then that resulting image is then composited with the other things behind it? If a building occludes another building with the same style, only the non-occluded portions of the buildings will be rendered in the 1st pass, and the 2nd pass makes them transparent over other things like roads? 

This seems similar to Photoshops ability to set a blend mode on a parent "layer" in addition to per feature "draw style" / "child" layer. 

**Photoshop blend multiply on layer:**

screen shot 2017-08-16 at 08 58 51

**Photoshop blend normal on layer:**

screen shot 2017-08-16 at 08 59 09

bcamper:
Thinking about this more, it seems like it does make sense for translucent to just be a new blend mode, and it looks like you've updated it here to work this way. So looks good to me! Looking forward to adding to JS too :)

hjanetzek:
Yes, we discussed it only last \#pragma meeting so far. Since some blending modes conflict with the depth mask options required for 'translucent' drawing, we thought it's best to make it an additional blend mode.

bcamper:
In JS, I put `translucent` after `opaque` but before `add`/`multiply`. I'm unsure how much if any impact this has vs. it coming after (always have a hard time thinking through blend order) -- either works for me but we should be consistent.

https://github.com/tangrams/tangram/pull/587/files#diff-e90eeb4f0c5a50cd03b669245e7e0724R679

matteblair:
Yeah it feels like a toss-up whether `translucent` should go before or after `add`/`multiply`, but it should probably go __after__ `inlay`, otherwise `inlay` geometry won't be visible through translucent buildings.

bcamper:
Ah, yeah. The order you have makes sense, I'll go with that.

matteblair:
Some notes about using multiple translucent styles in the same area:

The order in which translucent styles are rendered impacts the output. I spent a lot of time trying to figure out whether this was a bug, but I think this is the most "natural" behavior and could actually be useful in some situations.

Here are images from a scene with buildings rendered in two identical translucent styles. With the red style rendered first, you get:
screen shot 2017-08-30 at 7 45 32 pm

And with the blue style rendered first:
screen shot 2017-08-30 at 7 45 20 pm

This is made possible by clearing the stencil buffer before each translucent style renders. The code for that struck me as odd, so I tried clearing the stencil buffer just once per frame:
screen shot 2017-08-30 at 8 01 39 pm

Definitely less useful :) So for future reference, this is why we clear the stencil buffer per-style.

bcamper:
Note: updated JS blend order to match in https://github.com/tangrams/tangram/pull/587/commits/99bad9ece889d5668a0b4f470c8f5a818e641572

PR #1621, opened 2017-08-16
Cancelled URL request errors are correctly silenced on iOS and OSX

matteblair:
This fixes the implementation of #1611 and adds the same change for OSX.

NSError 'domain' is a NSString value that can't be correctly compared using ==.


msmollin:
+1

PR #1625, opened 2017-08-18
OSX system font

tallytalwar:
First step to debug and fix AAT system fonts with harfbuzz. This work coming in another PR.

OSX platform does not use bundled fallback fonts but relies on the system fonts for fallback.


platforms/common/TGFontConverter.h
```diff
@@ -0,0 +1,24 @@
+//
```
karimnaaji: We can remove these files from `platforms/ios/src/TangramMap/` now.

tallytalwar: Yup will do. And yes it has no dependency on harfbuzz updates. I had just thought, that PR will go in first :D.

Also we should wait on `0.8.0` to be released before we merge this or any other PR in.

platforms/osx/src/osxPlatform.mm
```diff
@@ -93,18 +96,125 @@ void initGLExtensions() {
...
+    CGFontRelease(fontRef);
+
+    if (data.empty()) {
+        LOG("CoreGraphics font failed to decode");
+    }
+
+    return data;
+}
+
+bool allowedFamily(NSString* familyName) {
```
karimnaaji: We could share the list of allowed families between iOS and OSX if those are the same.

tallytalwar: Yeah I can move this to a common file and share the code.

tallytalwar: I will take care of this in the coretext usage work I am doing.

Issue #1553, opened 2017-06-24
Fallback to lower zoom levels


rwrx:
I was playing a little bit with mbtiles and one improved came into my mind. One of my testing scenarios was that I had mbtiles file with 0 - 6 zooms for whole world. When my app started zoomed to let's say 10 zoom level it won't show any tile until I zoom out to 6 zoom level. Then it is able to fetch 6 zoom level and when I zoom again to 10 zoom level it will still show that tile from 6 zoom level. 

As an user this seems to me as strange behavior. As developer I understand why is this happening, but for user this is strange I think. 

Is it possible to add fallback to fetch tiles from lower zoom levels when map is zoomed that much that mbtiles doesn't contain appropriate tiles? So it will serve tile from lower zoom level. From using of Google Maps it seems to me that they are doing fallback to tiles from lower zoom levels when tiles from higher zoom levels are not available.

hjanetzek:
Hi @rwrx, you can set the maximal zoom that should be loaded by a tile source with `max_zoom` parameter. 

It's a good idea to try lower zoom-level tiles when a tile fails to load. This could be useful when mbtiles contains the whole world at low zoom and some small area at high zoom-level

rwrx:
Thank you. Yes, I am aware of max_zoom parameter but it does not work in this scenario because I need to have max_zoom set at least to 15 but mbtiles file may contain only tiles for zoom 6 for example.

I can add code for trying lower zoom-level tiles and maybe create pull request. Or would rather somebody from Tangram ES team do this?

rwrx:
@hjanetzek Firstly I have tried to add trying lower zoom-levels into `MBTilesDataSource::loadTileData()` method when `MBTilesDataSource::getTileData()` fails to load. But it is somehow needed to properly cut tile from lower zoom-level to fit this tile correctly. Is this possible? Also then I tried to add this fallback into `TileManager` class, but I am not sure how to do this correctly without much changing code. It seems to me that this couldn't be simply added due to code architecture. How this could be done? Thank you!

rwrx:
I haven't tried to add fallback already, but now I have some time to do it. Is it possible to give me some advice how this could be done effectively? As I didn't digged deeply into tile rendering architecture in Tangram-ES but as I have looked into it previously it seemed to me that currently from architectural point of view this cannot be done without changing architecture a bit. Am I right?

rwrx:
Will move discussion here: https://github.com/tangrams/tangram-es/pull/1626

PR #1626, opened 2017-08-20
Using fallback tile when tile is not available in DataSource.

rwrx:
I have added fallback to tiles from lower zoom levels when tiles for higher zoom levels are not available. I have tested this on different repo on Android device with mbtiles Data Source. Sometimes it is not able to get tiles when using `MemoryCacheDataSource` before `MBTilesDataSource`, however I haven't found why. Maybe you will see it immediately why in some cases it does not work as I was also unable to attach native debugger to Android device.

This also closes https://github.com/tangrams/tangram-es/issues/1553


matteblair:
@hjanetzek has looked into this feature and it seems possible, but would require a lot of refactoring of "proxy tile" management to implement in a reasonable way. We can keep this PR open but it's going to be a while before we'll have a chance to put the necessary time into this.

rwrx:
Thank you a lot that you have looked into this. So as I understand when it needs refactoring of proxy tile this is why it didn't worked cirrectly in some cases? I wasn't able to track down why it doesn't work all time. So will you include this in 0.8.1?

rwrx:
Hi @matteblair do you have any updates regarding this pull request? Thank you.

PR #1633, opened 2017-08-29
Fix JobQueue thread safety

cleeus:
The implementation of JobQueue does not hold up to the thread safety promises in the header comments. I wrote a simple test case that stresses JobQueue a bit and easily crashes. The fixed JobQueue swaps out the m_jobs vector inside a lock and then processes the jobs.

### DIFF
--- a/core/src/util/jobQueue.cpp
+++ b/core/src/util/jobQueue.cpp
@@ -13,24 +13,39 @@ void JobQueue::add(Job job) {
 
     if (!m_stopped) {
         std::lock_guard<std::mutex> lock(m_mutex);
-        m_jobs.push_back(job);
+        m_jobs.push_back(std::move(job));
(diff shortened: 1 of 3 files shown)


cleeus:
I just added some cosmetic fixes - this should be done now.

Btw, I think this is an actual issue. If I understand it correctly, in Map::loadSceneAsync and Map::updateSceneAsync a call to JobQueue::add(...) is scheduled to Map::Impl::asyncWorker via Map::runAsyncTask. When the asyncWorker executes JobQueue::add, that should be a racy with JobQueue::runJobs in Map::update.

tallytalwar:
Thanks @cleeus for looking at this issue and submitting a complete PR with a test case to catch the issue. Merged in master now, will be available in the next release.

Issue #1513, opened 2017-05-18
Myanmar/Burmese script rendering incorrectly


bcamper:
Based on on-going work on [Tangram JS](https://github.com/tangrams/tangram/pull/555), I was checking Tangram ES text rendering and noticed that Myanmar/Burmese script is currently not rendering properly. In the screens below, the characters with "dotted lines" should not be shown -- they should be combined via text shaping with other characters. Perhaps there's a text shaping option we don't have enabled?

You can reproduce this by going to Yangon, the capital of Myanmar.

Note: I have tested this on **iOS**, but have not yet verified on other platforms. 

![img_5471](https://cloud.githubusercontent.com/assets/16733/26209028/a85121ca-3bb9-11e7-9e92-f16311d491be.PNG)

![img_5468](https://cloud.githubusercontent.com/assets/16733/26209063/c69546ca-3bb9-11e7-9857-d4d0b3f5d9e3.PNG)

tallytalwar:
Some results on testing on this here: https://github.com/tangrams/tangram-es/pull/1558#issuecomment-313808942

matteblair:
This issue doesn't seem to occur on Android, see the attached screen capture.

I suspected this might be a result of the font family pruning we added on iOS recently (https://github.com/tangrams/tangram-es/blob/master/platforms/ios/src/TangramMap/iosPlatform.mm#L116-L125) but even with all font files allowed we still get missing code points in Myanmar on iOS.

![screenshot_20170718-173413](https://user-images.githubusercontent.com/3371850/28345458-76bde17e-6bdf-11e7-8d81-9a846c84dbaf.png)

tallytalwar:
Couple of screen shots showing issues with "Tibetan" and "Thai" scripts. In these the `dotted circle` appears behind a glyph.

Tibetan Script (dzongkha):
![image uploaded from ios](https://user-images.githubusercontent.com/360641/29422397-ef936952-832c-11e7-9633-3e372f79759e.jpg)

Thai:
![image uploaded from ios 1](https://user-images.githubusercontent.com/360641/29422427-01974b00-832d-11e7-8f47-61cf25c3bbc4.jpg)

matteblair:
Apple Maps on macOS has the same issues with Burmese, which strongly suggests to me that this isn't an artifact of our handling of text shaping or font loading, but rather something incorrect in the font itself.

[...]

PR #1634, opened 2017-08-31
Harfbuzz freetype coretext

tallytalwar:
The work here makes alfons use coretext for harfbuzz font creation to get shaped glyph from coretext instead of freetype. We still use freetype face for rendering and font atlas texture lookup purposes.

All font loading happens through the `FontSourceHandle` interface. This updates the previous implementation of `systemFont` which was passing around raw font data. `FontSourceHandle` also maintains if a font is a native apple font, to pass that information to `alfons::InputSource` aptly.

Addresses #1513


tallytalwar:
This is ready for review now.

core/include/tangram/platform.h
```diff
@@ -14,10 +14,14 @@ using FontSourceLoader = std::function<std::vector<char>()>;
...
 struct FontSourceHandle {
     FontSourceHandle(std::string _path) : path(_path) {}
+    FontSourceHandle(std::string _path, bool _appleFont) : path(_path), appleFont(_appleFont) {}
     FontSourceHandle(FontSourceLoader _loader) : load(_loader) {}
+    FontSourceHandle() {}
 
     std::string path;
-    FontSourceLoader load;
+    FontSourceLoader load = nullptr;
+    bool appleFont = false;
```
hjanetzek: Can we rename this flag to `systemFontName` or `fontName` ?

tallytalwar: updated aptly to reflect the usage.

tallytalwar:
Updated submodules to master commits for alfons and harfbuzz-icu-freetype and addressed the comment @hjanetzek mentioned.
Will merge once approved!

tallytalwar:
Rebased with master, merging when CI goes green.

matteblair:
@hjanetzek You reviewed this, right? Can you give this a ✅ if it looks good?

core/include/tangram/platform.h
```diff
@@ -13,11 +13,15 @@ using UrlCallback = std::function<void(std::vector<char>&&)>;
 using FontSourceLoader = std::function<std::vector<char>()>;
 
 struct FontSourceHandle {
-    FontSourceHandle(std::string _path) : path(_path) {}
+    FontSourceHandle(std::string _path) : pathOrFontName(_path) {}
+    FontSourceHandle(std::string _pathOrFontName, bool _isFontName) : pathOrFontName(_pathOrFontName), isFontName(_isFontName) {}
```
karimnaaji: Leaving this as a note for future cleanup on this: Using `Url` class here would be cleaner overhaul, only adding a dependency for this class from platform code.

tallytalwar: As discussed in slack, makes sense to do this cleanup with `fontConfig` work for linux platforms.

PR #1294, opened 2017-02-13
fix: alfons cluster shaping issue

hjanetzek:

Issue #1293, opened 2017-02-13
Incorrect shaped text


tallytalwar:
While looking at curved text PR, I noticed some of hindi text are incorrectly shaped and specifically missing last glyph (shaped).

Example: 
- Delhi (दिल्ली) is rendered as: दिल्ल
- Kolkata (कोलकाता) is rendered as: कोलकात
- Noida (नोएडा) is rendered as: नोएड
- Mussoorie (मसूरी) is rendered as: मसूर
- Ambala (अम्बाला) is rendered as: अम्बाल

The common error in the above example is that these are missing last (shaped) glyph. I have not found this last shaped glyph missing in other labels, example in the case of river Yamuna, it correctly renders it as यमुना. But thats the only correct one I am seeing so far.

Investigating this a bit more to see where in the code it messes up rendering the last shaped glyph. Will update more here.

cc. @hjanetzek

hjanetzek:
Looking into it - seems to be the TextBatch draw logic for glyph clusters. Fortunately not TextShaper :)

tallytalwar:
Ahh I am looking at the shaper code,  which seems to be doing the correct
thing with delhi example. But I got familiar with this code a bit :).

tallytalwar:
No doubt I am seeing it only for the place labels and not for anything else like all the river labels are good.

tallytalwar:
Looks like this piece of code: https://github.com/tangrams/tangram-es/blob/master/core/src/text/textUtil.cpp#L37-L55 is incorrectly breaking on `ल` for `दिल्ली`.

tallytalwar:
Resolved by #1294

PR #1637, opened 2017-09-05
Include all shaped glyph from the last cluster

tallytalwar:
previously the `lastShape` was not updated to include the shape of the last cluster (which is set to
must break in alfons) and hence last shape of the last cluster were not drawn.

Old behavior (Refer to continents names ending in vowels for example):
screen shot 2017-09-05 at 4 06 49 pm

Fixed behavior:
screen shot 2017-09-05 at 4 06 04 pm

This fixes a regression, which was previously reported and fixed (partially) in #1293 and #1294 respectively.


karimnaaji:
Needs rebase and good to go.

PR #1641, opened 2017-09-07
Misc font handling fixes

tallytalwar:
- match alfons default style name in ES (`normal`->`regular`)
- get font with apt font size.
We load 3 sizes for fonts, `16`,`28` and `40`. Make sure to get the font in this range.
- respect `normal` as default weight property, should map to numeric `400`. and `bold` should map to numeric `700`.

### DIFF
--- a/core/src/text/fontContext.cpp
+++ b/core/src/text/fontContext.cpp
@@ -316,7 +316,7 @@ std::shared_ptr<alfons::Font> FontContext::getFont(const std::string& _family, c
 
     // Pick the smallest font that does not scale down too much
     float fontSize = BASE_SIZE;
-    for (int i = 0; i < MAX_STEPS; i++) {
+    for (int i = 0; i < MAX_STEPS-1; i++) {
         sizeIndex = i;
 
         if (_size <= fontSize) { break; }
--- a/core/src/text/fontContext.cpp
+++ b/core/src/text/fontContext.cpp
@@ -9,14 +9,13 @@
 #include <memory>
 #include <regex>
 
-#define BASE_SIZE 16
-#define STEP_SIZE 12
-#define MAX_STEPS 3
 #define SDF_WIDTH 6
 
 #define MIN_LINE_WIDTH 4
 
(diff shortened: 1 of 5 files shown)


tallytalwar:
This is good to be reviewed, @karimnaaji can you have a look.

core/src/text/fontContext.cpp
```diff
@@ -316,7 +316,7 @@ std::shared_ptr<alfons::Font> FontContext::getFont(const std::string& _family, c
 
     // Pick the smallest font that does not scale down too much
     float fontSize = BASE_SIZE;
-    for (int i = 0; i < MAX_STEPS; i++) {
+    for (int i = 0; i < MAX_STEPS-1; i++) {
```
karimnaaji: There should be a face for range `0..MAX_STEPS-1`, why change this to go from `0..MAX_STEPS-2`? 
(Those faces are added from line https://github.com/tangrams/tangram-es/pull/1641/files#diff-4cb2204fedb2ba4ea43314c75ab033f0R47).

tallytalwar: Yeah this was interesting, in this loop we are trying to figure out what should be `fontSize` for the face we need to fetch.

By default we load faces for `16`, `28` and `40` sizes. However previously when the loop was from `0 .. MAX_STEP` the `fontSize` is incremented to `52` for any `_size` which is greater than `40`.

karimnaaji: Ah nice catch, I see now. The final version of this code is rather confusing now though, we should refactor that to make it more clear. We could remove `BASE_SIZE`, `MAX_STEPS`, `STEP_SIZE` and simply leave that in an array.

Maybe try something similar to this:
```c++
std::vector<int> fontRasterSizes = {16, 28, 40};
auto it = std::lower_bound(fontRasterSizes.begin(), fontRasterSizes.end(), _size);
```

tallytalwar: Yup should refactor the usage here. Addressed below.

PR #1652, opened 2017-09-16
Fix: text style should only draw when a `font` draw rule is found

tallytalwar:
Documentation: https://github.com/tangrams/tangram-docs/blob/gh-pages/pages/Styles-Overview.md#text-draw-group-requirements

Fixes issue with rogue black labels in tron. (cc @nvkelso)

### DIFF
--- a/core/src/style/textStyleBuilder.h
+++ b/core/src/style/textStyleBuilder.h
@@ -57,20 +57,7 @@ public:
     bool handleBoundaryLabel(const Feature& _feat, const DrawRule& _rule,
                              const TextStyle::Parameters& _params);
 
-    bool checkRule(const DrawRule& _rule) const override {
-        if (_rule.hasParameterSet(StyleParamKey::text_font_family) ||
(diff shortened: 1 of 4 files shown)


core/src/style/textStyleBuilder.h
```diff
@@ -57,7 +57,19 @@ class TextStyleBuilder : public StyleBuilder {
     bool handleBoundaryLabel(const Feature& _feat, const DrawRule& _rule,
                              const TextStyle::Parameters& _params);
 
-    bool checkRule(const DrawRule& _rule) const override { return true; }
+    bool checkRule(const DrawRule& _rule) const override {
```
karimnaaji: We should move that to implementation file.

tallytalwar: 👍

tallytalwar: Updated, merging when CI goes green.

PR #1657, opened 2017-09-18
fixes an issue with zipped import asset with parent yaml

tallytalwar:
It looks like there was an incorrect assumption that an imported "relative path" of a zipped bundled must have a parent which itself is a zipped bundle, so incorrect code path was being taken.

Example scene file which will segfault without this fix:

```
import:
   - refill-style.zip
```


tallytalwar:
@matteblair its a small fix, which might affect the all things url work you are doing. But should definitely not be a big rebase, just 1 line change.

PR #1658, opened 2017-09-20
path resolution for paths without file scheme

tallytalwar:
- Fixes an issue for path resolution when a file resource has a url defined with no scheme.
Noticable when a remote scene import happens after a local scene import, resulting in incorrect file
resource path resolution.

Note: A complete overhaul of the asset loading functionality is happening for 0.9.0, but we need to address the bug for the next release.

### DIFF
--- a/core/include/tangram/platform.h
+++ b/core/include/tangram/platform.h
@@ -76,6 +76,8 @@ public:
 
 protected:
 
+    // A stop gap solution to load assets having `file` scheme in the path
+    std::string removeFileScheme(const char* _path) const;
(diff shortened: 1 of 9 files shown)


matteblair:
As far as I can tell, the current behavior won't create problems for either the iOS or Android SDKs because neither SDK loads scenes from relative paths (on Android, any relative paths get interpreted as absolute `asset://` paths and on iOS we require a `NSURL` and take it's `absoluteString` property). I can see how this can be problematic for some situations in the desktop app, but considering the rather small scope of potential problems I'd prefer to wait for the proper fix rather than applying this temporary work-around. Does that make sense?

tallytalwar:
Closing in favor of #1668

PR #1663, opened 2017-09-25
Wrap output longitudes to canonical range

### LINKED ISSUE #1097: MapController::getPosition return LngLat with negative Longitude
Tangram-ES Version 0.4.5

MapController::getPosition and MapController::screenPositionToLngLat may return a LngLat in which (at least) the longitude is negative.

I got this above Hamburg, when having the map centered at lat=53,58220643220753, lon=9,939129168887803. Lon was -350.0608708311122 instead.

I know this is still mathematically correct, but having a coordinate whose longitude is not within the -180°..+180° bounds is certainly unexpected. If you intend to not fix this by normalizing the angle, it would be good if this were documented.

matteblair:
Resolves #1097

### DIFF
--- a/core/include/tangram/util/types.h
+++ b/core/include/tangram/util/types.h
@@ -29,6 +29,21 @@ struct LngLat {
                latitude == _other.latitude;
     }
 
+    // Get a LngLat with an equivalent longitude within the range (-180, 180].
+    LngLat wrapped() const {
+        return LngLat(wrapLongitude(longitude), latitude);
+    }
+
+    // Get an equivalent longitude within the range (-180, 180].
+    static double wrapLongitude(double longitude) {
+        if (longitude > 180.0) {
+            return longitude - (int)((longitude + 180.0) / 360.0) * 360.0;
+        } else if (longitude <= -180.0) {
+            return longitude - (int)((longitude - 180.0) / 360.0) * 360.0;
+        }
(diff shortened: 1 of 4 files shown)


matteblair:
I haven't changed the default behavior of `Tangram::LngLat` for now, but in the future it might be easier to enforce checks like this if `LngLat` constructors applied this wrapping automatically.

tallytalwar:
Hmm, maybe it does make sense to do wrap the lat lon in constructor itself to start with instead of doing it at a later stage. Do you have any thoughts on how it will not be a good idea to do it now? I do not see it being a larger code change.

matteblair:
Yes we could do this in the `LngLat` constructor now and it probably wouldn't break anything, but for that to really be useful we'd also have to make `LngLat` immutable (i.e. members are `const`). Unless we do that we won't really know for sure that `LngLat` members are validated (they could have been set to an invalid value after construction). 

That seemed like a bigger change that I didn't want to bother with right now.

tallytalwar:
Ahh.. its a struct with public access to `longitude` and `latitude`. I am fine with this for the time being in that case.

tallytalwar:
Merging this with a note to refactor this code to make `LngLat` members immutable and provide apt accessors along with `LngLat` constructor taking care of wrapping the `longitude` values aptly.

PR #1664, opened 2017-09-26
Remove 'armeabi' and 'arm64-v8a' architectures from Android builds

matteblair:
Significantly reduces Android library build time and output size at a negligible functional cost. Full release AAR size is reduced from 19.9MB to 5.7MB, a __size reduction of 71%__.

Dropped architectures:
 - __armeabi__, the ABI for ARMv6 CPUs, is no longer supported by Android as of 4.4 (according to the compatibility requirements, section 3.3.1: https://source.android.com/compatibility/4.4/android-4.4-cdd.pdf). Our minimum supported Android version is 4.0.3, so it is theoretically possible that an ARMv6 device could install our AAR and fail to load a compatible native library at runtime, but I expect this case to be vanishingly rare. We could consider increasing our minimum Android version to 4.4, but that would exclude ~7% of currently active devices (according to the Android Dashboards: https://developer.android.com/about/dashboards/index.html).
 - __arm64-v8a__, the ABI for 64-bit ARM CPUs, has backward compatibility with binaries using the 32-bit `armeabi-v7a` ABI (according to: https://en.wikipedia.org/wiki/ARM_architecture#ARMv8-A). There are possible performance benefits to running ARM64 binaries on a 64-bit ARM device and I'd like to do some benchmarks to make sure we're not making a dramatic sacrifice on 64-bit devices.

Kept architectures:
 - __armeabi-v7a__, the ABI for ARMv7 CPUs, is the target for the vast majority of active Android devices.
 - __x86__, the ABI for 32-bit x86 CPUs, is used rarely for mobile devices (mostly niche products like the ZenFone) but is immensely helpful for running in the Android emulator. The x86 emulator is at least 10x faster than the ARM emulator and is strongly recommended by Android Studio.

### DIFF
--- a/platforms/android/tangram/build.gradle
+++ b/platforms/android/tangram/build.gradle
@@ -62,7 +62,7 @@ android {
       // Default configuration
     }
     full {
-      externalNativeBuild.cmake.abiFilters 'armeabi', 'arm64-v8a', 'x86'
+      externalNativeBuild.cmake.abiFilters 'x86'
     }
   }
 


westnordost:
👍

westnordost:
Please do not increase the min version to 4.4 though, there are still too many people using these versions.

The device catalogue in the Google Play Console shows me that 10699 different devices are currently compatible with my app. Without armeabi, it's only two devices less (and no device of the currently installed 5k).
That is less than **0.02%** of devices tangram-es becomes incompatible with and thus dimensions away from the ~7% that would be lost by a hard increase of minSdkVersion to 19.

westnordost:
Also, FYI, about roughly 55% the devices seem to be currently based on arm64-v8a and about 1% on x86.

tallytalwar:
Thanks @westnordost. Its good to know the stats from streetcomplete. As you mentioned there are fair bit of devices running `arm64-v8a`, so it might be beneficial for us to keep that architecture.

However as @matteblair mentioned it does make sense to do a performance benchmark to gauge how much of performance improvement we get with 64bit arch. If its significant enough it does make sense to only get rid of `armeabi` and stick with `arm64-v8a`.

Will read more on this and update.

tallytalwar:
Confirmed from multiple documentations (android, arm and intel), that devices having 64bit archs are backward compatible with 32bit ABIs. So with respect to compatibility and device support we should be good here.

With respect to performance and power utilization, documented evidence points to about 5 to 10% performance and power improvement. I will run some tests with tangram on a device running arm64-v8a to get some concrete evidence with `arm64-v8a` and `armeabi-v7a` explicitly.

Reference: https://androidbycode.wordpress.com/tag/armeabi-v7a/, https://software.intel.com/en-us/android/articles/how-to-develop-and-evaluate-64-bit-android-apps-on-intel-x86-platforms

ac87:
@tallytalwar As I have a dependency on SQLCipher which does not have 64 bit support. All the work I've done with Tangram-es has been excluding the arm64-v8a lib and I have not seen any problems with this across a variety of newer devices.

Be very interested in seeing the results of your arm64-v8a vs armeabi-v7a benchmark.

tallytalwar:
Ok did some basic benchmarking with just scene loading with different house scene files and compared the approx numbers for armeabi-v7a and arm64-v8a. 


Scene | CPU (max) % usage |   | Memory (peak) - Mb |   | GPU (max) - ms |  
-- | -- | -- | -- | -- | -- | --
  | armeabi-v7a | arm64-v8a |   | armeabi-v7a | arm64-v8a |   | armeabi-v7a | arm64-v8a |  
default | 50 | 44 | 12 | 10 | 8.5 | 15 | 67 | 58 | 13.43283582
bubble | 50 | 40 | 20 | 10 | 10 | 0 | 67 | 67 | 0
refill | 45 | 42 | 6.666666667 | 11 | 9.5 | 13.63636364 | 67 | 60 | 10.44776119
walkabout | 40 | 38 | 5 | 10.2 | 9 | 11.76470588 | 67 | 67 | 0
tron | 55 | 50 | 9.090909091 | 11.2 | 10 | 10.71428571 | 67 | 67 | 0
cinnabar | 40 | 38 | 5 | 11.2 | 11 | 1.785714286 | 67 | 67 | 0
zinc | 40 | 38 | 5 | 11.2 | 10.5 | 6.25 | 67 | 67 | 0
  |   |   | 8.965367965 |   |   | 8.450152788 |   |   | 3.411513859



To Summarize on an average wrt just scene loading for tangram using the demo app:
CPU utilization is optimized by about 8%
Memory utilization is optimized by about 8%
GPU render times are optimized negligibly by a factor of 3%

Will do some other benchmarking with scene updates, etc. But this gives a ballpark of optimization we get with using arm64-v8a vs using armeabi-v7a.

matteblair:
Thanks for checking out some benchmarks @tallytalwar! Can you explain more about how you got these numbers? I'm curious because the CPU and memory usage overall look lower than I would expect, and memory use should almost certainly be higher for the 64-bit architecture (due to pointers and struct alignment being larger). Maybe I'm overlooking something?

westnordost:
I wonder, do you consider a speed improvement of ~8% as significant?

My gut opinion would be (still) "yes" as I usually tend to value lower CPU usage always over lower disk drive / memory usage. The latter is just usually available in plenty, the former a very limited resource.
Plus, in any case, if you'd revisit this topic in a few years, the 55% of 64bit users will have grown into a vast majority.

matteblair:
I tested this change on a Galaxy S7, in Release configuration, using the following diff to print a running average of theoretical FPS:
```
--- a/platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
+++ b/platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
@@ -1254,6 +1254,9 @@ public class MapController implements Renderer {
     private Map<Long, Marker> markers = new HashMap<>();
     private Handler uiThreadHandler;
 
+    private long frameCounter = 0;
+    private long frameTimeSum = 0;
+
     // GLSurfaceView.Renderer methods
     // ==============================
 
@@ -1269,11 +1272,20 @@ public class MapController implements Renderer {
             return;
         }
 
+        long frameStartTime = System.nanoTime();
         boolean viewComplete;
         synchronized(this) {
             viewComplete = nativeUpdate(mapPointer, delta);
             nativeRender(mapPointer);
         }
+        frameTimeSum += System.nanoTime() - frameStartTime;
+
+        frameCounter++;
+        long frameAverageWindow = 60;
+        if (frameCounter % frameAverageWindow == 0) {
+            long fps = frameCounter * 1000000000 / frameTimeSum;
+            android.util.Log.d("Tangram", String.valueOf(fps) + " FPS");
+        }
 
         if (viewComplete) {
```

Using a library compiled for `armeabi-v7a` I recorded averages of roughly 75-85 FPS.

Using a library compiled for `arm64-v8a` I recorded averages of roughly 75-85 FPS.

Any difference between the two was not significant compared to normal variation in application performance. 

Based on these results, I'm in favor of moving forward with this PR as-is.

tallytalwar:
Sure in my results also apart from cpu and memory improvement the fps rate was pretty consistent. 

I also did some more reading on this and probably make sense to re-evaluate this when the state of the industry and adoption of arm64-v8a increases. 
http://www.informit.com/articles/article.aspx?p=1843892

Issue #1645, opened 2017-09-11
Dynamic point textures


nvkelso:
Tangram ES support for JS issue tracked in https://github.com/tangrams/tangram/pull/588.

Relaxes the requirement that a `points`-based style be bound to only one (or no) texture.

With this change, a scene author can freely re-assign (or un-assign) the texture that the style is bound to, within a `draw` group, using the `texture` parameter.

_Note: one motivation for this feature from the Mapzen cartography perspective is easier support for additional textures in the future, such as expanded road shields for US states and countries, where it is impractical to fit all these shields into a single texture (and/or some textures may be loaded lazily at run-time, only when those regions are visited)._

**For example:**

```
layers:
  data: ...
   
  landuse:
    filter: ...
    draw:
      points:
        texture: landuse.png # render landuse icons with one texture
        ...

  pois:
    filter: ...
    draw:
      points:
        texture: pois.png # render POI icons with a different texture
        ...

    zoos:
      filter: ...
      draw:
        points:
          texture: zoos.png # override in sublayer to render zoos with a third texture

```

This change is fully backwards compatible with existing `styles`: if a style has set a `texture`, this simply becomes the default value, but that value can then be overridden if desired within `layers`.

If no `texture` value is set for the final merged `draw` group (either because it was not set by the `style` or by a parent layer's `draw` group, or it was explicitly *un-set* in a child `draw` group with `texture: null`), then the feature will be drawn with the built-in points style (e.g. as regular `points` would be drawn, with no texture bound).

Extending the example above:

```
  buildings:
    filter: ...
    draw:
      points:
        color: red # no texture is set by parent style or layer, so buildings will draw as simple points
        ...

  pois:
    filter: ...
    draw:
      points:
        texture: pois.png # default POI texture

    parks:
      filter: ...
      draw:
        points:
          texture: null # texture is explicitly un-set, so parks will render as simple points
          color: green
          ...
```

Internally, this feature uses the recently added "mesh variant" concept, where the same style can create sub-meshes that each have their own uniform state or vertex layouts. All vertex layouts must be compatible with the same underlying shader program, but a variant can set a vertex attribute to a static value to lower memory consumption, if that attribute has a suitable default value. For this PR, the variants will vary the uniform texture sampler that is bound for sprites, and the vertex layout can set static values of zero for point outlines where they are not needed.

tallytalwar:
Addressed by #1665

PR #1665, opened 2017-09-26
Dynamic PointStyle textures

hjanetzek:
https://github.com/tangrams/tangram-es/issues/1645

### DIFF
--- a/core/src/scene/sceneLoader.cpp
+++ b/core/src/scene/sceneLoader.cpp
@@ -1555,8 +1555,8 @@ void SceneLoader::parseStyleParams(Node params, const std::shared_ptr<Scene>& sc
             if (value.IsScalar()) {
                 auto strVal = value.Scalar();
                 out.push_back(StyleParam{ StyleParamKey::texture, strVal });
-            } else {
-                ;// TODO load texture
+            } else if (value.IsNull()){
+                out.push_back(StyleParam{ StyleParamKey::texture, "" });
             }
             continue;
         }
--- a/core/src/scene/sceneLoader.cpp
+++ b/core/src/scene/sceneLoader.cpp
@@ -308,7 +308,8 @@ bool SceneLoader::applyConfig(const std::shared_ptr<Platform>& _platform, const 
     return true;
 }
 
-void SceneLoader::loadShaderConfig(const std::shared_ptr<Platform>& platform, Node shaders, Style& style, const std::shared_ptr<Scene>& scene) {
+void SceneLoader::loadShaderConfig(const std::shared_ptr<Platform>& platform, Node shaders, Style& style,
+                                   const std::shared_ptr<Scene>& scene) {
 
     if (!shaders) { return; }
 
@@ -416,7 +417,9 @@ glm::vec4 parseMaterialVec(const Node& prop) {
     return glm::vec4(0.0);
 }
 
-void SceneLoader::loadMaterial(const std::shared_ptr<Platform>& platform, Node matNode, Material& material, const std::shared_ptr<Scene>& scene, Style& style) {
+void SceneLoader::loadMaterial(const std::shared_ptr<Platform>& platform, Node matNode, Material& material,
+                               const std::shared_ptr<Scene>& scene, Style& style) {
+
     if (!matNode.IsMap()) { return; }
 
     if (Node n = matNode["emission"]) {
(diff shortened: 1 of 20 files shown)


core/src/scene/sceneLoader.cpp
```diff
@@ -1545,18 +1544,28 @@ void SceneLoader::parseStyleParams(Node params, const std::shared_ptr<Scene>& sc
...
             parseTransition(prop.second, scene, key, out);
             continue;
         }
 
+        if (key == "texture") {
+            if (value.IsScalar()) {
+                auto strVal = value.Scalar();
+                out.push_back(StyleParam{ StyleParamKey::texture, strVal });
+            } else {
+                ;// TODO load texture
```
tallytalwar: Is this a documented use case, where a texture could be defined directly as a style parameter instead of in the scene and referred as a style parameter?

core/src/style/pointStyleBuilder.cpp
```diff
@@ -105,13 +105,6 @@ void PointStyleBuilder::setup(const Marker& _marker, int zoom) {
 }
 
 bool PointStyleBuilder::checkRule(const DrawRule& _rule) const {
```
tallytalwar: What is the reason of always making point style pass irrespective of any color or texture parameter set?

core/src/style/pointStyleBuilder.cpp
```diff
@@ -275,38 +270,62 @@ void PointStyleBuilder::addLabel(const Point& _point, const glm::vec4& _quad,
...
+    if (!_params.texture.empty()) {
 
-        if (!m_style.spriteAtlas()->getSpriteNode(_params.sprite, spriteNode) &&
-            !m_style.spriteAtlas()->getSpriteNode(_params.spriteDefault, spriteNode)) {
-            return false;
-        }
+        if (_params.texture == "null") {
+            texture = nullptr;
+        } else {
+            if (!m_style.textures()) {
```
tallytalwar: I am not sure this is required, this check is mandating that the style must define a texture for it to be overriden in the draw group. Based on the description of the problem in #1645 that does not seem to be case. Can you clarify?

matteblair:
I've been testing this a bit and have two questions:

1. Setting new textures in `draw` seems to work, but I don't see the expected behavior when I *unset* a texture at the `draw` level. Setting a null or empty value for the texture parameter should cause the point to draw with the default point texture and that doesn't seem to happen.

2. This can be very, very inefficient. By setting  a custom texture on one of the POI layers in the default scene, I produced over 10 times as many draw calls to render the points style. These are expensive draw calls too because they rebind textures and they don't use VAOs (also: why don't we use VAOs for point styles??). Is there a way to sort our points by texture before we submit them for drawing? We *could* merge this without improving the batching, but it would be better to address this performance bottleneck before users hit it.

tallytalwar:
> Setting new textures in draw seems to work, but I don't see the expected behavior when I unset a texture at the draw level. Setting a null or empty value for the texture parameter should cause the point to draw with the default point texture and that doesn't seem to happen.

This does seem to work for me and as described in the issue:
>If no texture value is set for the final merged draw group (either because it was not set by the style or by a parent layer's draw group, or it was explicitly un-set in a child draw group with texture: null), then the feature will be drawn with the built-in points style (e.g. as regular points would be drawn, with no texture bound).

tallytalwar:
Ok I see, I can repro what @matteblair mentioned above in `1`.

Refer the following scene file diff:

```
diff --git a/scenes/scene.yaml b/scenes/scene.yaml
index 627d2e0b..6bbf74af 100644
--- a/scenes/scene.yaml
+++ b/scenes/scene.yaml
@@ -60,6 +60,9 @@ textures:
             arrow: [1, 1175, 22, 22]

 styles:
+    icons:
+        base: points
+        texture: pois
     heightglow:
         base: polygons
         blend: translucent
@@ -307,8 +310,8 @@ layers:
             # oneway arrows and shields are distinct groups!
             filter: { oneway: yes, shield_text: false, $zoom: { min: 17 } }
             draw:
-                points:
-                    texture: pois
+                icons:
+                    texture: null
                     flat: true
                     sprite: arrow
                     size: [[17, 18px], [18, 20px], [20, 32px]]
```

One would expect no arrows will be shown as the texture is explicitly set to `null` in the draw rules, but it does render the arrows.

core/src/scene/sceneLoader.cpp
hjanetzek: Yes, I was not sure about this. Though regarding consistency we have 'Load inline material  textures' in SceneLoader.

core/src/style/pointStyleBuilder.cpp
hjanetzek: Updated checkRule to also check dynamic textures. Thanks!

core/src/style/pointStyleBuilder.cpp
hjanetzek: This is just a nullptr check before accessing the global texture map :)

matteblair:
The CI build errors should fix themselves after rebasing: https://circleci.com/gh/tangrams/tangram-es/2297

There must be some issue with the Xcode 8.1 image we previously used on CircleCI.

matteblair:
Would it be simpler to use the "style-default draw rule" mechanism to handle the dynamic texture behavior? That is, we may not need a new "default texture" field in PointStyle, we can store that information in the "default draw rule" for that style and let the merging take care of the per-draw-rule behavior.

hjanetzek:
@matteblair good point! Though looking into it this seems to be slightly less efficient and it would not simplify the implementation.

PR #1666, opened 2017-09-27
Fix iOS Headers and add custom URL Configuration initializer

msmollin:
The iOS Header implementation is currently non-functional as NSURLSession
makes a copy of the configuration (as documented). Fix for this is to
reset the URL session when new headers are installed. An alternative
instead of setting headers is to pass in a configuration at creation time.

### DIFF
--- a/platforms/ios/src/TangramMap/TGHttpHandler.mm
+++ b/platforms/ios/src/TangramMap/TGHttpHandler.mm
@@ -34,6 +34,18 @@
     return self;
 }
 
+
+- (instancetype)initWithSessionConfiguration:(NSURLSessionConfiguration *)configuration
+{
+    self = [super init];
+
+    if (self) {
+        [self initialSetupWithConfiguration:configuration];
+    }
+
+    return self;
+}
+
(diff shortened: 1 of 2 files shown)


platforms/ios/src/TangramMap/TGHttpHandler.mm
```diff
@@ -74,8 +86,7 @@ - (NSMutableDictionary *)HTTPAdditionalHeaders {
 - (void)setAdditionalHTTPHeaders:(NSMutableDictionary *)HTTPAdditionalHeaders {
     _HTTPAdditionalHeaders = HTTPAdditionalHeaders;
     self.configuration.HTTPAdditionalHeaders = HTTPAdditionalHeaders;
-    //Probably unnecessary, but for safety sake
-    self.session.configuration.HTTPAdditionalHeaders = HTTPAdditionalHeaders;
+    self.session = [NSURLSession sessionWithConfiguration:self.configuration];
```
matteblair: To be consistent, we would want to do this in `setCachePath` as well, right?

msmollin: Agreed. Fixed.

tallytalwar:
Just reading more on it, will rebase and merge when I am done.

Issue #1660, opened 2017-09-21
Address build issues with sqlite3 for ndk 15+


tallytalwar:
Currently we are including deprecated headers to avoid running into sqlite build issues with ndk 15.2+. But we will have to address this issue sooner than later as these headers will be excluded in the future ndk versions.

Reference: https://android.googlesource.com/platform/ndk/+/ndk-r15-release/docs/UnifiedHeaders.md
Sqlite3 issue on android ndk project: android-ndk/ndk#477

cleeus:
Right on time. I just got bitten by this.
I tried to build on iOS11. On iOS the OS-version of sqlite is used, which comes in a newer version which requires a newer SQLiteCpp because of slight header incompatibilities. So I tried to update SQLiteCpp. SQLiteCpp 2.2.0 comes with sqlite-3.20.1 bundled. When building on Android this fails because the new sqlite uses mmap.

I was able to fix/work around this by adding
target_compile_definitions(sqlite3 PRIVATE -DSQLITE_MAX_MMAP_SIZE=0 -DSQLITE_OMIT_WAL)
to the core/deps/SQLiteCpp/sqlite3/CMakeLists.txt. This disables the use of mmap in sqlite3. I'm not sure of this is the right fix.

tallytalwar:
Thanks @cleeus. This could be good, will try it out. Thanks

4brunu:
I was having the same issue, and this is what fixed for me.
https://github.com/couchbaselabs/couchbase-lite-libsqlcipher/pull/23

tallytalwar:
Resolved by #1667

PR #1667, opened 2017-09-28
Fix android with ndk15c without deprecated flags

tallytalwar:
- With ndk15c _FILE_OFFSET_BITS=64 took affect and required the availability of certain 64 bit
functions, which are not available in api levels below 21. (We require our min api level to be 15)
- This _FILE_OFFSET_BITS was explicitly set in sqlitecpp to 64 and hence caused build failures
because of missing symbols for the 64 bit equivalent methods for `lseek`, etc.

- For our android builds to work for api level 15+ (below 21) we are required to specify
_FILE_OFFSET_BITS=32 for sqlitecpp

A detailed explaination for this issue is available here:
https://android.googlesource.com/platform/bionic/+/master/docs/32-bit-abi.md

And I quote the main description from above:

> Android support for _FILE_OFFSET_BITS=64 (which turns off_t into off64_t and replaces each off_t function with its off64_t counterpart, such as lseek in the source becoming lseek64 at runtime) was added late. Even when it became available for the platform, it wasn‘t available from the NDK until r15. Before NDK r15, _FILE_OFFSET_BITS=64 silently did nothing: all code compiled with that was actually using a 32-bit off_t. With a new enough NDK, the situation becomes complicated. If you’re targeting an API before 21, almost all functions that take an off_t become unavailable. You‘ve asked for their 64-bit equivalents, and none of them (except lseek/lseek64) exist. As you increase your target API level, you’ll have more and more of the functions available. API 12 adds some of the <unistd.h> functions, API 21 adds mmap, and by API 24 you have everything including <stdio.h>. See the linker map for full details.

Addressed #1660 

Note: Should go in after `0.8.1` release.

### DIFF
--- a/core/deps/CMakeLists.txt
+++ b/core/deps/CMakeLists.txt
@@ -78,6 +78,14 @@ add_subdirectory(SQLiteCpp)
 # Extensions aren't needed for MBTiles and aren't available in older versions of sqlite3.
 target_compile_definitions(SQLiteCpp PRIVATE SQLITE_OMIT_LOAD_EXTENSION)
 
+# needed for sqlite3 to work for ndk15c+ and android api level < 21
+# refer:
+# https://github.com/android-ndk/ndk/issues/477 and
+# https://android.googlesource.com/platform/bionic/+/master/docs/32-bit-abi.md
+if (ANDROID)
+    target_compile_definitions(sqlite3 PRIVATE _FILE_OFFSET_BITS=32)
+endif()
+
 ## double-conversion ##
 #######################
 add_subdirectory(double-conversion)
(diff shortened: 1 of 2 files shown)


core/CMakeLists.txt
```diff
@@ -74,6 +74,14 @@ else()
   target_link_libraries(${LINK_LIBRARIES})
 endif ()
 
+# needed to SQLITECPP to work for ndk15c+ and android api level < 21
+# refer:
+# https://github.com/android-ndk/ndk/issues/477 and
+# https://android.googlesource.com/platform/bionic/+/master/docs/32-bit-abi.md
+if (ANDROID)
+    add_definitions(-D_FILE_OFFSET_BITS=32)
```
matteblair: Can we add this flag to just the SQLiteCpp project and not the whole core library? We should be able to add it here, just like this other definition: https://github.com/tangrams/tangram-es/blob/master/core/deps/CMakeLists.txt#L79

tallytalwar: Sure, will update

tallytalwar:
Hmm.. interestingly this solution stopped working with updates to SQLiteCpp. There could be more here, but so far nothing successful with updated SQLiteCpp

PR #1668, opened 2017-09-28
Fix path resolution for local paths mixed with absolute URLs

matteblair:
I think this solves the same issue as https://github.com/tangrams/tangram-es/pull/1658 but with a more limited code change.


tallytalwar:
Yeah considering this only (in practice at least) affects desktop versions, I think this should suffice as a stop-gap solution.

Thanks

matteblair:
Honestly I mostly did this to just make rebasing my other branch easier - so thanks :)

PR #1676, opened 2017-10-04
Support iOS builds with Xcode 9 and iOS 11 SDK

matteblair:
Currently Tangram ES cannot make iOS builds using Xcode 9 for two reasons:
 - iOS 11 doesn't support 32-bit architectures and Xcode 9 compiles with `-miphoneos-version-min=11.0` by default, so linking fails for 32-bit builds.
 - The version of SQLite3 shipped with iOS 11 creates a build error for the version of SQLiteCpp currently used in Tangram ES (more info: https://github.com/SRombauts/SQLiteCpp/commit/1a2c7cbba707ea31affa056925a013b1041738cb).

This PR addresses both of those issues and updates our CircleCI configuration to use Xcode9 for iOS builds.

I've set the target iOS version to 10.0 for now, this is easily adjustable if we find a reason to change it.

I believe that fixing the issue with SQLiteCpp means that it is no longer compatible with versions of SQlite3 below 3.19 - which implies that the iOS 11 SDK will required to build Tangram ES.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,6 +1,6 @@
 machine:
   xcode:
-    version: 8.1
+    version: 9.0
   environment:
     # Dummy values, Circle won't run without a project and scheme.
     XCODE_PROJECT: build/tangram.xcodeproj
--- a/circle.yml
+++ b/circle.yml
@@ -7,7 +7,7 @@ machine:
     XCODE_SCHEME: phony
 dependencies:
   pre:
-    - brew install awscli
+    - brew install awscli cmake
     - gem install jazzy
 checkout:
   post:
--- a/toolchains/iOS.framework.cmake
+++ b/toolchains/iOS.framework.cmake
@@ -2,6 +2,8 @@ include(${CMAKE_SOURCE_DIR}/toolchains/iOS.toolchain.cmake)
 
 set(FRAMEWORK_VERSION "0.8.2-dev")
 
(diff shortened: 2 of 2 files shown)


matteblair:
@msmollin Thanks! Instead of increasing the deployment target in the podspec, I changed the deployment target in CMake to match it. This puts it at the same deployment target version as the Mapzen iOS SDK (9.3). There's really no reason for it to be higher!

tallytalwar:
We wanted to give this a try with xcode versions before 9 in line of the statement above from the description:

```
I believe that fixing the issue with SQLiteCpp means that it is no longer compatible with versions of SQlite3 below 3.19 - which implies that the iOS 11 SDK will required to build Tangram ES.
```

msmollin:
Is that what that line meant? I just figured that was an "FYI - we need xcode 9"

matteblair:
Tested on CircleCI, confirmed that builds will break with earlier SDKs now: https://circleci.com/gh/tangrams/tangram-es/2280

So yes, this effectively means we require Xcode 9 for iOS now. If it becomes a problem we can figure out a solution to enable older SDKs at that point.

msmollin:
I don't see that as a big deal. We can still support 9.3 as a deployment target, so we're not orphaning anyone.

tallytalwar:
We should update the readme, to reflect this in that case.

PR #1683, opened 2017-10-10
Change color model ARGB to RGBA

AndreyArapov:


matteblair:
Thanks for the submission @AndreyArapov! We do indeed have a bug here. I'm going to submit a slightly different change that should make this code more maintainable in the future.

matteblair:
Resolved by https://github.com/tangrams/tangram-es/pull/1689

PR #1687, opened 2017-10-11
Remove assertion in Android demo build config requiring API key

matteblair:
This currently causes all CI builds from external contributors to fail.


matteblair:
@tallytalwar That's reasonable. I just pushed a commit to log a warning when no API key is found in the demo.

PR #1689, opened 2017-10-11
Fix ordering of color components returned from JS function as array

matteblair:
The current code puts color components into the wrong parts of the resulting integer - making the final colors wrong. For example, if you set the `water` layer color to `function() { return [ 0.0, 0.25, 0.5, 1.0 ]; }` in the demo scene, you currently get:

screen shot 2017-10-11 at 4 11 39 pm

The correct color, using the new code, is:

screen shot 2017-10-11 at 4 10 01 pm

The specific way that color components are converted to integers appears to match the behavior of Tangram JS by interpreting the doubles modulo 1.0, i.e. `[0., 0., 1.5, 1.]` produces the same result as `[0., 0., 0.5, 1.]`. 

Thanks to @AndreyArapov for bringing this to our attention :)

### DIFF
--- a/core/src/scene/styleContext.cpp
+++ b/core/src/scene/styleContext.cpp
@@ -432,10 +432,7 @@ void StyleContext::parseStyleResult(StyleParamKey _key, StyleParam::Value& _val)
                     duk_pop(m_ctx);
                 }
 
-                _val = (((uint32_t)(255.0 * a) & 0xff) << 24) |
-                       (((uint32_t)(255.0 * r) & 0xff)<< 16) |
-                       (((uint32_t)(255.0 * g) & 0xff)<< 8) |
-                       (((uint32_t)(255.0 * b) & 0xff));
+                _val = Color::fromNormalized(r, g, b, a).abgr;
                 break;
             }
             default:
--- a/core/src/scene/styleContext.cpp
+++ b/core/src/scene/styleContext.cpp
@@ -432,7 +432,7 @@ void StyleContext::parseStyleResult(StyleParamKey _key, StyleParam::Value& _val)
                     duk_pop(m_ctx);
                 }
 
-                _val = Color::fromNormalized(r, g, b, a).abgr;
+                _val = ColorF(r, g, b, a).toColor().abgr;
                 break;
             }
             default:
--- a/core/src/gl/framebuffer.cpp
+++ b/core/src/gl/framebuffer.cpp
@@ -20,7 +20,7 @@ FrameBuffer::FrameBuffer(int _width, int _height, bool _colorRenderBuffer) :
 
 }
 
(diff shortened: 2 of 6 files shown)


core/src/scene/styleContext.cpp
```diff
@@ -432,10 +432,7 @@ void StyleContext::parseStyleResult(StyleParamKey _key, StyleParam::Value& _val)
                     duk_pop(m_ctx);
                 }
 
-                _val = (((uint32_t)(255.0 * a) & 0xff) << 24) |
-                       (((uint32_t)(255.0 * r) & 0xff)<< 16) |
-                       (((uint32_t)(255.0 * g) & 0xff)<< 8) |
-                       (((uint32_t)(255.0 * b) & 0xff));
+                _val = Color::fromNormalized(r, g, b, a).abgr;
```
tallytalwar: I could be completely missing something here!

You are initializing the `Color`'s anonymous union with `r`, `g`, `b`, `a` constructor then I am not sure how you can access `abgr` from the anonymous union?

matteblair: Setting the `uint8_t` members of the union to the component values initializes the memory occupied by the `uint32_t` member in the same way as packing them with bit shifting, like this code was trying to do. Strictly speaking, this only works on little-endian systems, but we don't currently target (or project to target) big-endian systems.

tallytalwar: Ahh make sense, I would however refrain more such memory trickery and provide a method to do the apt conversion.

To quote the documentation "The details of that allocation are implementation-defined, and it's undefined behavior to read from the member of the union that wasn't most recently written". http://en.cppreference.com/w/cpp/language/union

tallytalwar: It actually gets interesting, https://stackoverflow.com/questions/11373203/accessing-inactive-union-member-and-undefined-behavior

matteblair: I got this trick from code for a game and it's always seemed to work, but you may be right! It's not totally obvious whether this is UB: https://stackoverflow.com/questions/11373203/accessing-inactive-union-member-and-undefined-behavior
In short: this trick is allowed in C, and the C++ spec doesn't outright forbid it but there are only a few cases where it isn't UB.

matteblair:
This approach is working well, it produces the expected assembly on every compiler I've checked. I also took the opportunity to refactor `Color` a bit. Ideally we'd call `Color` something more specific like `Color32` but I didn't want to change all the occurrences of its name.

tallytalwar:
I would have gone with the documentation, but if you are saying that assembly generated on every compiler you tried for this approach is same then I am fine going forward with this. Maybe add a comment to mention that this will break for big-endian targets.

matteblair:
Are there still issues to resolve with this? If not can someone approve the PR?

PR #1697, opened 2017-10-18
fix conversion of empty strings in android jni wrapper

cleeus:
Hi all,

the new Android JNI stringFromJstring and jstringFromString helper functions dereference the 0th element of C++ string objects even when the strings are empty. While this most of the times does not fail in practice, it's still an out of bounds access and must be checked.


matteblair:
I don't think this is actually a bug? Dereferencing a `std::string` at `size` returns a reference to the null character (http://en.cppreference.com/w/cpp/string/basic_string/operator_at)

tallytalwar:
Yeah thats a good point. I forgot about that.

tallytalwar:
Hmm, I am unsure of the `GetStringRegion` behavior though. 

Nothing is documented here,
http://docs.oracle.com/javase/7/docs/technotes/guides/jni/spec/functions.html. But possible it misbehaves or throws an index out of bound exception on specifying a pointer to the null character reference (should be fine though). @cleeus did you run into this exception?

cleeus:
Oh, they changed that in C++11 ... didn't check/know that ... sorry for the noise.
This was just code review - no actual bugs observed.

PR #1698, opened 2017-10-18
Allow curl to follow a fixed number of redirects

cleeus:
Hi all,

I noticed that the UrlClient/curl wrapper does not follow redirects which might be required by some map services. This enables curls automatic redirect handling, see https://curl.haxx.se/libcurl/c/CURLOPT_FOLLOWLOCATION.html for reference.


tallytalwar:
This does make sense, probably @matteblair can cherry-pick this commit in #1659. Else we can directly merge this. Will leave this for him to decide.

matteblair:
Sounds like an improvement! This change shouldn't conflict with the other PR. This should be good to merge after rebasing. I'll also suggest a redirect limit of 20, as that seems to be the limit used by OkHTTP and many browsers (https://github.com/square/okhttp/blob/54b583ed57ef1be8a56c346f3c466f3d00b37e4c/okhttp/src/main/java/okhttp3/internal/http/RetryAndFollowUpInterceptor.java#L62).

cleeus:
changed to 20 redirects

matteblair:
Merged into master with 36e76339d0c2f90c611c1e7db533f16b9c805539 and 7865a3a0b508325e8450e9242172c796146e7e31

Thanks @cleeus !

PR #1701, opened 2017-10-18
Fix iOS memory leak

### LINKED ISSUE #1690: TGMapViewController not deallocated on iOS
Hello,

I have a memory problem with TGMapViewController.
If I put a TGMapViewController in the stack of navigation controller and after I pop it, the TGMapViewController is not deallocated.

TO REPRODUCE THE ISSUE, FOLLOW THESE STEPS:
I took the example project and I added a navigation controller. When i pop the dealloc method is not called and the memory remains high.

Does anyone have a solution for this?

Thank you !

karimnaaji:
- Add content of incorrectly removed `tearDownGL` method call
- Use weak references for `TGMapViewController`

Fixes #1690

### DIFF
--- a/platforms/ios/src/TangramMap/TGMapViewController.mm
+++ b/platforms/ios/src/TangramMap/TGMapViewController.mm
@@ -105,7 +105,8 @@ __CG_STATIC_ASSERT(sizeof(TGGeoPoint) == sizeof(Tangram::LngLat));
                     dataLayerName, "", generateCentroid);
     self.map->addTileSource(source);
 
-    TGMapData* clientData = [[TGMapData alloc] initWithMapView:self name:name source:source];
+    __weak TGMapViewController* weakSelf = self;
+    TGMapData* clientData = [[TGMapData alloc] initWithMapView:weakSelf name:name source:source];
     self.dataLayersByName[name] = clientData;
 
     return clientData;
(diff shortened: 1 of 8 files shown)


platforms/ios/src/TangramMap/TGMapViewController.mm
```diff
@@ -140,17 +141,18 @@ - (void)clearDataSource:(std::shared_ptr<Tangram::TileSource>)tileSource
 }
 
 - (Tangram::SceneReadyCallback)sceneReadyListener {
-    return [=](int sceneID, auto sceneError) {
-        [self.markersById removeAllObjects];
-        [self renderOnce];
+    __weak TGMapViewController* weakSelf = self;
+    return [weakSelf](int sceneID, auto sceneError) {
+        [weakSelf.markersById removeAllObjects];
```
msmollin: Might want to check here to make sure `weakSelf` is not nil.

karimnaaji: Right, but thinking more about this I think that to be truly safe using `weakSelf` we would need to retain the pointer from the lambda function block by creating a temporary strong reference on it. Since it is likely that the lambda is called from a different thread, a statement testing whether `weakSelf` is valid doesn't mean that the next statement will still have a valid `weakSelf` pointer.

karimnaaji: Updated in https://github.com/tangrams/tangram-es/pull/1701/commits/f6649efd1e7b853c9a4a66093efbf2849de691e4.

msmollin:
👍

Issue #1636, opened 2017-09-01
Enable point outlines


matteblair:
Draw groups using the `points` style with no texture attached should be able to specify an outline, as described in the documentation: https://mapzen.com/documentation/tangram/draw/#outline

This is implemented in Tangram JS here: https://github.com/tangrams/tangram/pull/508, https://github.com/tangrams/tangram/pull/509

matteblair:
Resolved by #1702

PR #1702, opened 2017-10-18
Point outlines

hjanetzek:
Still needs some tweaking when outline is larger than the point radius.

Todo:
- handle `outline_visible: false` 

https://github.com/tangrams/tangram-es/issues/1636


matteblair:
Good stuff! This seems to be working correctly for the most part and some of the new code actually looks simpler. 

A case that we'll need to look into more closely is outlines with partial opacity. Here's a test scene file that illustrates the issue:
```yaml
global: {}
scene: { background: { color: grey } }
sources:
    mapzen:
        type: TopoJSON
        url: https://tile.mapzen.com/mapzen/vector/v1/all/{z}/{x}/{y}.topojson
        url_params: { api_key: global.sdk_mapzen_api_key }
layers:
    pois:
        data: { source: mapzen }
        draw:
            points:
                texure: null
                color: [1, 0, 0, 1]
                size: 30px
                outline:
                    width: 10px
                    color: [0, 0, 1, 0.5]
```
In Tangram Play, the outline is blended over the fill:

screenshot-2017-11-3 tangram play

In Tangram ES, the outline replaces the fill:

screen shot 2017-11-03 at 2 10 48 pm

core/shaders/point.vs
```diff
@@ -33,6 +35,9 @@ attribute LOWP vec4 a_stroke;
 attribute float a_scale;
 #else
 attribute vec3 a_position;
+attribute vec4 a_outline_color;
+attribute float a_outline_edge;
+attribute float a_aa_factor;
```
matteblair: If I understand this parameter, it seems like `a_aa_factor` could be a uniform instead of an attribute.

matteblair:
There are also some differences here regarding the dimensions of outlined points. Using the same example from before to illustrate:
screenshot-2017-11-3 tangram play

screen shot 2017-11-03 at 2 10 48 pm

I think the way point outlines work in Tangram JS is essentially imitating an SVG stroke on a circle path. The SVG behavior is documented here with some good examples to play with: https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/stroke

hjanetzek:
@matteblair Thanks I'll go over it again.

It seems there is currently no way to define a half transparent fill and outline where the outline does not blend with fill. E.g. like
```
<svg height="100" width="100">
    <circle opacity="0.5" fill="blue" cx="50" cy="50" r="40" stroke="red" stroke-opacity="1" stroke-width="10" fill="black" />
</svg>
```

matteblair:
Hmm you're right, this does seem to be a limitation of our current blending implementation. 

Varying the overall opacity is probably more frequently useful than blending the stroke into the fill, so I'm inclined to find a way to make that work. However, I don't think we can cover all of the opacity options of SVG fill and stroke with just the two color values that we currently have - we would need another value for the overall opacity of both fill and stroke.

@bcamper Any thoughts on this? Does the problem here make sense?

matteblair:
Looks like this now matches the Tangram JS behavior for point outlines - thanks Hannes!

I've pushed one change to `point.vs` and `point.fs` to fix a GLSL linking error on iOS.

But I still see some unexpected results on my two Android devices here:
screenshot_20171106-172126

There are no GLSL errors reported, but it looks like maybe `u_sprite_mode` has the correct value in the fragment shader but not the vertex shader?

hjanetzek:
Ok that looks interesting.. Did you try what happens when there is only the `u_sprite_mode == 0` branch part in point.vs?

hjanetzek:
@matteblair could you test again on Android?

matteblair:
Yeah this is very strange, when I comment out the branch in `point.vs` such that only the `u_sprite_mode == 0` path is present I still get this quarter-circle output. 

Are you able to reproduce this on an Android device too @hjanetzek?

matteblair:
Oh I just saw your change, let me try again.

matteblair:
Yeah I'm seeing the same results with the `highp` change applied.

matteblair:
Aha! That change from `u16vec2` to `i16vec2` fixes it, good catch.

This should be good to go after reverting the POI changes in the default scene file.

matteblair:
And we can probably revert the change to `highp float` in the point vertex shader - better not to introduce more changes than needed, I think.

PR #1705, opened 2017-10-19
Deep copy YAML nodes when copying scene configuration

### LINKED ISSUE #1677: Multiple scene updates followed by a scene load causes a crash

- Make sample iOS project using Tangram-es 0.8.1 release
- Load a scene file from disk (did not test from web)
- Apply 3+ scene updates and an async scene load in that order

**RESULT:**

Crash!

**EXPECTED RESULT:**

Not to crash!

**ENVIRONMENT:**

-iOS 10.x both simulator and device.

**OTHER:**

I tried to reproduce this with the debug builds available on ios.mapzen.com but to no avail. It seems the debug builds either take a slightly different code path, OR are just so much slower that tangram can cope with all the scene updates. I've attached the view controller code i used to do it to speed up the process. 
[ViewController.swift.zip](https://github.com/tangrams/tangram-es/files/1354501/ViewController.swift.zip)

tallytalwar:
- this could cause the underlying scene config data could be modified by multiple threads, which
actually happens in the case of `loadSceneAsync` and `updateScene`!

Fixes #1677

### DIFF
--- a/core/src/scene/scene.cpp
+++ b/core/src/scene/scene.cpp
@@ -49,7 +49,7 @@ void Scene::copyConfig(const Scene& _other) {
 
     m_featureSelection.reset(new FeatureSelection());
 
-    m_config = _other.m_config;
+    m_config = YAML::Clone(_other.m_config);
     m_fontContext = _other.m_fontContext;
 
     m_url = _other.m_url;


tallytalwar:
@msmollin can you also give this a try on ios sdk. This should resolve the crash you were seeing. I do not see any crashes with this fix on ios demo or on osx.

msmollin:
👍 recreated the environment that was crashing consistently in the Mapzen iOS SDK, and cannot cause the crash now.

matteblair:
Can we run some timers on scene loading before and after this change? Recursively cloning the entire configuration for a scene could be pretty slow.

tallytalwar:
@matteblair will update with these numbers next.

tallytalwar:
Ok 2 things here:

1. Since having a `updateSceneAsync` call just after a `loadSceneAsync` call crashes on master the repro scenario was not the "exact" same.
2. Since scene config copy only happens on update, (1) can be avoided by explicitly doing a `updateSceneAsync` by triggering a key event.

And as obvious it is, since in current master code, yaml scene config is copied as a reference during update it takes `0.2 milliseconds` to do the `Scene::copyConfig` call, whereas doing a deep copy of the yaml config (this PR) takes on an average of `20 milliseconds` (about 100 times slower) to complete `Scene::copyConfig` call.

Even though its 100 times slower, overall the time updating is reasonable and its also running on an async thread. The test scene I was using is walkabout which is also fairly evolved.

PR #1710, opened 2017-10-27
Fix missing initialization of Map::Impl::sceneLoadTasks

cleeus:
Hi all,

sometimes the Map::update method never returns true. I debugged it and found that the sceneLoadTasks counter is not properly initialized. Simple fix.

regards,
Kai


hjanetzek:
@varun the benchmark script requires MAPZEN_API_KEY to be set now. I've pushed a fix that should make CI work without key.

hjanetzek:
Thanks @cleeus!

Issue #1706, opened 2017-10-20
Point features are sometimes built with incorrect UVs before textures load.


matteblair:
**TO REPRODUCE THE ISSUE, FOLLOW THESE STEPS:**

 - In the default `scene.yaml` test scene, change the `url` of the `pois` texture from `img/poi_icons_32.png` to `https://github.com/tangrams/tangram-es/raw/master/scenes/img/poi_icons_32.png` (the same texture, but remote).
 - Build and run the desktop demo application, loading `scene.yaml`.

**RESULT:**

In some fraction of runs (more than half in my case, but probably variable depending on your network connection) the point features using the `pois` texture will not be visible in part or all of the initial view. However, features built shortly after the initial scene load _are_ visible as expected.

![screen shot 2017-10-20 at 10 49 46 am](https://user-images.githubusercontent.com/3371850/31834933-92977aba-b584-11e7-872d-716de32848ee.png)

**EXPECTED RESULT:**

Features using the `pois` texture should be visible regardless of when they are built.

![screen shot 2017-10-20 at 10 50 45 am](https://user-images.githubusercontent.com/3371850/31834976-b6ef2700-b584-11e7-9592-5860df76460e.png)

**ENVIRONMENT:**

 - macOS 10.12.6
 - Current `master` branch

**OTHER:**

I believe what's happening is:

 - Scene begins loading and encounters a texture to load.
 - An empty Texture and SpriteAtlas are created as placeholders.
 - The async request for the texture data starts.
 - Scene loading finishes and the Map starts loading and building tiles.
 - PointStyleBuilder encounters a feature using the texture that is still loading and gets the UV parameters from the texture's SpriteAtlas, which are still the default 0 values.
 - The async texture request completes and updates the SpriteAtlas with the correct UVs.

The most straightforward solution seems to be delaying tile building until the scene's textures are all finished loading.

matteblair:
@hjanetzek This seems to have been introduced by the changes in #1665, though I'm still trying to figure out exactly how.

matteblair:
Resolved by #1721

PR #1721, opened 2017-11-09
Wait for pending textures before creating tiles

hjanetzek:
Should fix use of incomplete texture atlas and concurrent read/writes of texture data.

https://github.com/tangrams/tangram-es/issues/1706

### DIFF
--- a/core/src/map.cpp
+++ b/core/src/map.cpp
@@ -391,8 +391,9 @@ void Map::resize(int _newWidth, int _newHeight) {
 
 bool Map::update(float _dt) {
 
-    // Wait until font resources are fully loaded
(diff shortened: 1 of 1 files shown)


matteblair:
I really expected this to fix the missing sprite UV issue, but it still seems to occur! I might not fully understand the problem.

matteblair:
The latest change here seems to resolve the issue.

The reason the pending resource check wasn't working before is that we were checking the wrong scene. In an async scene load, the new scene isn't set on the `impl` until the `JobQueue` is run, so while the new scene is being loaded, the pending resource check is actually checking the _previous_ scene object, not the new one that we care about.

PR #1730, opened 2017-11-17
Store shader uniform textures in scene config

matteblair:
Restores correct behavior for textures specified in a shader uniform. This regression was introduced in https://github.com/tangrams/tangram-es/commit/4b4a8dbabe54c5fdcf413251d1c3bfcfba8d2738. The regression causes styles using texture uniforms to render incorrectly (notably, this impacts several styles in walkabout).


hjanetzek:
Oh good catch! - I would add it to scene textures in `loadTexture` to be consistent with the other overload

matteblair:
Superseded by #1731

PR #1733, opened 2017-11-21
Update android builds

### LINKED ISSUE #1732: Update android build tools and gradle plugin
With android studio 3.0 we get new set of build tools and newer gradle plugin.

Follow the following for gradle updates: https://developer.android.com/studio/build/gradle-plugin-3-0-0-migration.html

tallytalwar:
- Updates gradle configs to use new dependency configuration
 - Uses gradle 3.0.1
 - needs android studio 3.0 (Updated readme)
- Updates build tools to 26.0.2

Followed migration instructions from: https://developer.android.com/studio/build/gradle-plugin-3-0-0-migration.html

Closes #1732

### DIFF
--- a/platforms/android/build.gradle
+++ b/platforms/android/build.gradle
@@ -3,9 +3,10 @@ allprojects {
     repositories {
       mavenCentral()
       jcenter()
+      google()
     }
     dependencies {
-      classpath 'com.android.tools.build:gradle:2.3.3'
+      classpath 'com.android.tools.build:gradle:3.0.1'
       classpath 'com.github.dcendents:android-maven-gradle-plugin:1.5'
     }
   }
@@ -14,6 +15,7 @@ allprojects {
     repositories {
       mavenCentral()
       jcenter()
+      google()
     }
   }
 }
--- a/platforms/android/demo/build.gradle
+++ b/platforms/android/demo/build.gradle
@@ -2,7 +2,7 @@ apply plugin: 'com.android.application'
 
 android {
   compileSdkVersion 26
-  buildToolsVersion '25.0.3'
+  buildToolsVersion '26.0.2'
 
   def apiKey = project.hasProperty('mapzenApiKey') ? mapzenApiKey : System.getenv('MAPZEN_API_KEY')
 
--- a/platforms/android/demo/build.gradle
+++ b/platforms/android/demo/build.gradle
@@ -10,6 +10,8 @@ android {
     minSdkVersion 15
     targetSdkVersion 26
     buildConfigField 'String', 'MAPZEN_API_KEY', '"' + apiKey + '"'
+    // provide dimension strategy for tangram library's dim
+    missingDimensionStrategy 'tangramAbi', 'slim'
   }
 
   sourceSets.main {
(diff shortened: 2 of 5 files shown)


tallytalwar:
@matteblair I think we can bring this in 0.9.0.

platforms/android/demo/build.gradle
```diff
@@ -2,14 +2,16 @@ apply plugin: 'com.android.application'
 
 android {
   compileSdkVersion 26
-  buildToolsVersion '25.0.3'
+  buildToolsVersion '26.0.2'
```
karimnaaji: Android build tools 27 are available, should we switch to that?

matteblair: From the Gradle plugin [release notes](https://developer.android.com/studio/releases/gradle-plugin.html#3-0-0):

> With this update, you no longer need to specify a version for the build tools—the plugin uses the minimum required version by default. So, you can now remove the android.buildToolsVersion property.

I suggest we just remove it.

platforms/android/tangram/build.gradle
```diff
@@ -45,6 +45,8 @@ android {
     }
   }
 
+  flavorDimensions "tangramAbi"
```
matteblair: With the "variant-aware dependency management" in the new plugin, I'm no longer sure that product flavors are the right way for us to manage ABI variants. 

From the perspective of an application that depends on `tangram`, all variants of the library should provide all available ABIs. This is because ABI variants in applications are typically not managed with product flavors, but with [separate APKs](https://developer.android.com/studio/build/configure-apk-splits.html). Using flavors for this wasn't previously a problem because flavors weren't exposed when the library was used as a dependency.

Moving forward, I think we should put all of our supported ABIs in the `defaultConfig` and not separate them by flavor. There's clearly a benefit to building only a single ABI during a development for a fast build/run cycle - and luckily for us a feature has snuck into Android Studio that does just that. When a run configuration is started from Studio, it will build only the ABIs needed for the deployment device (see [this thread](https://code.google.com/p/android/issues/detail?id=221098#c20) for discussion).

platforms/android/demo/build.gradle
```diff
@@ -2,14 +2,16 @@ apply plugin: 'com.android.application'
...
 
   def apiKey = project.hasProperty('mapzenApiKey') ? mapzenApiKey : System.getenv('MAPZEN_API_KEY')
 
   defaultConfig {
     minSdkVersion 15
     targetSdkVersion 26
     buildConfigField 'String', 'MAPZEN_API_KEY', '"' + apiKey + '"'
+    // provide dimension strategy for tangram library's dim
+    missingDimensionStrategy 'tangramAbi', 'slim'
```
matteblair: I'm not totally clear on how flavors are published from a library like ours. Will all projects using `tangram` need to add this `missingDimensionStrategy`? Or just projects using the 3.0 plugin? Will this make any difference for projects using older plugin versions?

platforms/android/build.gradle
```diff
@@ -3,9 +3,10 @@ allprojects {
     repositories {
       mavenCentral()
       jcenter()
+      google()
```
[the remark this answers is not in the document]
tallytalwar: @matteblair was there a reason you left `jcenter` in here, in your revisions to this PR?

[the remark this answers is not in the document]
matteblair: I vaguely remember a CI build failing to find a dependency when I removed it, but I'm not totally certain.

PR #1734, opened 2017-11-21
fix: preserve relative paths during Url dot-removal

cleeus:
Hi all,

I hit an issue when working with scene imports from local paths. Given a filesystem of:
* scene.yaml with import: foo/bar.yaml
* foo/bar.yaml with import ../bar/baz.yaml
* bar/baz.yaml

The relative url in "import: ../bar/baz.yaml" will be transformed to "/bar/baz.yaml" during dot removal. This of course is an absolute path that probably doesn't exist.

### DIFF
--- a/core/src/util/url.cpp
+++ b/core/src/util/url.cpp
@@ -453,6 +453,8 @@ size_t Url::removeDotSegmentsFromRange(std::string& str, size_t start, size_t co
     size_t pos = start; // 'input' position.
     size_t out = pos; // 'output' position.
 
+    const bool had_leading_slash = count != 0 ? str[start] == '/' : true;
+
     while (pos < end) {
         if (pos + 2 < end &&
                    str[pos] == '.' &&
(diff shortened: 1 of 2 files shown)


cleeus:
hmmm ... this might be obsolete ... I have to read through your new URL handling code ...

matteblair:
Hi, thanks for the proposal. I can see how the examples you've given seem to have issues, but I think the `Url` class is working as intended here. `Url::removeDotSegmentsFromString` follows the procedure described in [IETF RFC 3986](https://tools.ietf.org/html/rfc3986#section-5.2.4). Using the steps described there on one of your examples:
```
STEP    OUTPUT      INPUT
1                   a/../b
2E      a           /../b
2C                  /b
2E      /b          
```
We see that `Url` is currently producing the correct result, so I don't think changes are needed in this particular function. 

However it's still possible that there are oversights in our use of URLs elsewhere in Tangram ES, so let us know if you find that to be the case.

cleeus:
Thanks for the thorough review. So the issue with my scenario is that the root scene.yaml is a relative path and not a file:// url with an absolute path. Since relative filesystem paths seem to be replaced with absolute file:// urls after your big url rewrite, my problem is now by definition elsewhere. I'll look for another solution.

PR #1735, opened 2017-11-22
Fix wait for scene textures

hjanetzek:
- render default scene directly after load on android
  (without moving manually the map, etc.)
- cleaups


tallytalwar:
I think I also noticed this on osx on master yesterday. Will test this PR.

Issue #1709, opened 2017-10-25
Add texture 'density' parameter


matteblair:
As discussed here: https://github.com/tangrams/tangram/pull/234 texture defninitions will get a new `density` parameter that re-interprets the coordinates in sprite atlases so that multiple resolutions of texture can be provided without tedious duplication of sprites.

matteblair:
Resolved by #1742

PR #1742, opened 2017-12-01
Add density option for scaling of sprite UVs

hjanetzek:
https://github.com/tangrams/tangram-es/issues/1709

Includes small change: using 'const Node&' to not create new map entries


tallytalwar:
I do not think its working for icons examples from https://github.com/tangrams/icons/tree/master/spritesheet. More updates on this tomorrow :D

tallytalwar:
Wait, this seems to work, the atlas definitions in the above icons repo should be corrected and include the density parameter.

Issue #1679, opened 2017-10-05
Polygon MapData Flickers


sarahlensing:
**TO REPRODUCE THE ISSUE, FOLLOW THESE STEPS:**

```java
mapView.getMap(
        new MapController.SceneLoadListener() {
          @Override public void onSceneReady(int sceneId, SceneError sceneError) {
            
            // 1. Add MapData layer
            MapData mapData = mapController.addDataLayer("mz_default_polygon");

            // 2. Create coordinates
            List<LngLat> ring = new ArrayList();
            ring.add(new LngLat(-73.9903, 40.74433));
            ring.add(new LngLat(-73.984770, 40.734807));
            ring.add(new LngLat(-73.998674, 40.732172));
            ring.add(new LngLat(-73.996142, 40.741050));
            ring.add(new LngLat(-73.9903, 40.74433));
            List coordinates = new ArrayList();
            coordinates.add(ring);

            // 3. Add a polygon & request render
            mapData.addPolygon(coordinates, null);
            mapController.requestRender();

            // 4. Configure map zoom and position
            mapController.setZoom(15f);
            mapController.setPosition(new LngLat(-73.9918, 40.73633));
          }
        });
```

**RESULT:**

The map & polygon flicker and the polygon is temporarily clipped

**EXPECTED RESULT:**

No map & polygon flickering or clipping of polygon

**ENVIRONMENT:**

 - Android 8.0
 - Tangram 0.8.1

**OTHER:**

This gif shows the polygon clipping well but not the map flicker:

![polygon_flicker](https://user-images.githubusercontent.com/494323/31253116-cfac0236-a9e0-11e7-961f-4d172d6cc9c9.gif)

PR #1743, opened 2017-12-04
Fix client source flickering

hjanetzek:
https://github.com/tangrams/tangram-es/issues/1679


hjanetzek:
Oh hold on - I think we removed also all tiles from clientDataSources since they could reference raster sources... Looking for another solution.

PR #1744, opened 2017-12-04
Fix apple default system font

tallytalwar:
- Some cleanups in font usages.
- Update Alfons submodule:
    - Fixes AppleSystemFontUI issue
    - Uses newer harfbuzz 1.7.2 coretext font creation API
- Harfbuzz-icu-freetype updates
    - Updates to use harfbuzz 1.7.2, which fixes an issue with default osx system font which uses kern tables

Depends on:
- https://github.com/tangrams/harfbuzz-icu-freetype/pull/8
- https://github.com/hjanetzek/alfons/pull/17


tallytalwar:
@karimnaaji can you test this out on osx please?

One way to force the osx to use default system font (San Francisco) is, to comment out all fallback font in `osxPlatform.mm` and modify the following to not have a scene specified fontfamily:

https://github.com/tangrams/tangram-es/blob/e7529f6cf1dd778939a04deacd2bb83a6d298948/core/src/style/textStyleBuilder.cpp#L597

PR #1747, opened 2017-12-08
This will allow null values for style params

tallytalwar:
- style mixing will automatically override the none_type here, from a
sublayer mixing, unlike previously when the style param was removed from
the draw rule set for the layer

Refer the discussion here: https://github.com/tangrams/refill-style/pull/82#discussion_r155866056

### DIFF
--- a/core/src/scene/sceneLoader.cpp
+++ b/core/src/scene/sceneLoader.cpp
@@ -1566,6 +1566,7 @@ void SceneLoader::parseStyleParams(Node params, const std::shared_ptr<Scene>& sc
         }
         default:
             LOGW("Style parameter %s must be a scalar, sequence, or map.", key.c_str());
+            out.emplace_back(StyleParam::getKey(key));
         }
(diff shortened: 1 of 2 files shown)


core/src/scene/sceneLoader.cpp
```diff
@@ -1566,6 +1566,7 @@ void SceneLoader::parseStyleParams(Node params, const std::shared_ptr<Scene>& sc
         }
         default:
             LOGW("Style parameter %s must be a scalar, sequence, or map.", key.c_str());
+            out.emplace_back(StyleParam::getKey(key));
```
matteblair: This should probably go into a `case NodeType::Null`, otherwise `Undefined` nodes would get the same treatment. Also we won't need to print that warning for `Null` nodes now.

tallytalwar: I forgot about `Undefined` nodes. 👍

PR #1748, opened 2017-12-08
Fix calculation of View::pixelsPerMeter()

matteblair:
This is probably an *extremely* old bug that impacts shaders using the 'u_pixels_per_meter' uniform, like the "waves" shaders in refill.

Before there was noticeable crawling of the pattern:
waves-before

Now it's steady while panning:
waves-after


tallytalwar:
Though this looks correct, and surprising that we have not caught this before, but I am still not seeing the difference on my machine when comparing master and this branch.

Issue #1759, opened 2018-01-08
max_tilt ignored for all camera types except 'perspective'


m-sasha:
The docs say:

> For cameras with type: **isometric** the tilt is **further** constrained at high zooms to prevent the viewing plane from intersecting the ground.

But the code (view.cpp:235) says:

```   
    // Ensure valid pitch angle.
    {
        float maxPitchRadians = glm::radians(getMaxPitch());
        if (m_type != CameraType::perspective) {
            // Prevent projection plane from intersecting ground plane.
            maxPitchRadians = atan2(m_pos.z, m_height * .5f);
        }
        m_pitch = glm::clamp(m_pitch, 0.f, maxPitchRadians);
    }
```

Which is wrong on two counts: it restricts the tilt for all camera types **other than isometric**, and it doesn't further restrict it (via `maxPitchRadians = min(maxPitchRadians, atan2(...))`) but simply sets it.

m-sasha:
Also, there appears to be confusion between degrees and radians. The code above calls `radians()` on `View.m_maxPitch`, which is defined with:

```
float m_maxPitch = 0.5 * PI;
```

PR #1760, opened 2018-01-09
Apply 'max_tilt' to isometric and flat cameras

matteblair:
Resolves https://github.com/tangrams/tangram-es/issues/1759

 - As noted in the issue, `m_maxPitch` was being ignored for isometric and flat cameras. Now it is used in conjunction with the ground-plane-intersection pitch.
 - The units of `m_maxPitch` were not clear and the initial value suggested that it was radians. It is in fact degrees, and the code makes that clear now. (I think that initial value is overridden in most cases.)

### DIFF
--- a/core/src/view/view.cpp
+++ b/core/src/view/view.cpp
@@ -237,7 +237,8 @@ void View::update(bool _constrainToWorldBounds) {
         float maxPitchRadians = glm::radians(getMaxPitch());
         if (m_type != CameraType::perspective) {
             // Prevent projection plane from intersecting ground plane.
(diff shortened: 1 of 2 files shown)


m-sasha:
The suggested patch adjusts the max pitch for all cameras *except* perspective. Is that the intent?

matteblair:
The docs should be amended in this case to say that this constraint is applied to `flat` cameras as well. Both `isometric` and `flat` have the same problem that the view plane can intersect the ground plane at high tilts, clipping through geometry in an undesirable way.

m-sasha:
I see. Thanks!

PR #1768, opened 2018-02-02
Fix: flat sprite texture

hjanetzek:
- keep w-ordinate for correct texture perspective


matteblair:
Nice fix, this is smart :)

PR #1771, opened 2018-02-06
Migrate iOS and macOS builds to CircleCI 2.0

matteblair:
WIP


tallytalwar:
Sorry for the delay here, will get to this today.

We might have to also look at github circle linkage here:
screen shot 2018-02-14 at 3 42 58 pm

matteblair:
@tallytalwar I've just updated the Github branch settings to use the new status check from CircleCI. We can now require a status check from any of the "jobs" in our CI configuration.

.circleci/config.yml
```diff
@@ -1,28 +1,39 @@
...
+      XCODE_SCHEME: phony
+    steps:
+      # Check out repository with submodules.
+      - checkout
+      - run: git submodule update --init
+      # Install dependencies.
+      - run: brew install awscli cmake
+      - run: gem install jazzy
       # Build the framework in debug mode and package it into pod.zip
-      - make clean-ios
```
tallytalwar: Probably make sense to leave `clean-ios` in there, just in case a previous build is cached.

.circleci/config.yml
```diff
@@ -31,24 +42,68 @@ deployment:
...
+    jobs:
+      - build-ios:
+          filters:
+            branches:
+              ignore: master
+      - build-deploy-macos-snapshot:
+          filters:
+            branches:
+              only: master
+      - build-deploy-ios-snapshot:
```
tallytalwar: `workflow` feature in the newer circle version is nice. I think based on the recommendation in https://circleci.com/docs/2.0/migrating-from-1-2/, we should split `build-universal-debug` and `deploy-snapshot`. As mentioned in the docs, this will prevent us to not run the entire build when just the deployment failed.

>`It might make sense to start by just splitting out a deploy job to prevent you from having to re-run the entire build when only the deployment fails.`

tallytalwar: Similar split the `build-deploy-ios-release` workflow to `build-universal-release` and `deploy-release`.

tallytalwar: For the above to work, we will need to use the `requires` key on the workflow jobs' `build` and `deploy` to run sequentially.

PR #1774, opened 2018-02-13
[Android] Efficiency + Help developpers

Dean79000:
-  Help developpers : adding NonNull and Nullable annonations : Show errors before runtime
- Improve efficiency : final keyword, and small code changes
- Update OkHttp from 3.5 to 3.9.1
- Update Android Support Annotation lib from version 25 to 26

### DIFF
--- a/platforms/android/tangram/build.gradle
+++ b/platforms/android/tangram/build.gradle
@@ -61,7 +61,7 @@ android {
 }
 
 dependencies {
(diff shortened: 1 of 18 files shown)


matteblair:
Hi @Dean79000, thanks for the PR! The general idea here makes sense to me, compile-time errors are always better than run-time errors :) I'll take a look this week.

PR #1781, opened 2018-03-22
Update URLs and references from Mapzen to Nextzen

matteblair:
Pretty straight-forward replacement.


tallytalwar:
Testing on android last and merging.

matteblair:
@tallytalwar On my end the build script seems to be finding the `nextzenApiKey` property from my `~/.gradle/gradle.properties` as expected. I believe a `nextzenApiKey` project property would be first priority, then the global `nextzenApiKey`, then the environment variable `NEXTZEN_API_KEY`. Is this not what you're observing?

PR #1784, opened 2018-03-30
TGMarker manually converts image data to un-premultiplied RGBA

### LINKED ISSUE #1782: translucent png images for markers has a dark tint

 I am trying to add a png image that has an opacity of %20 percent. 

**RESULT:**

The png is coming out dark. It seems anytime you add an image to a marker with an alpha other than 1, it darkens the image. 

img_7708

**EXPECTED RESULT:**

I expect the image to have a fully transparent background and should not tint my png. It should look like this: 

screen shot 2018-03-26 at 10 45 43 am

**ENVIRONMENT:**

 - Xcode 9.3 swift

matteblair:
Resolves #1782 

The process of converting a `UIImage` to raw pixel data has become pretty laborious, but I couldn't find a more efficient way of extracting texture data. This is something to keep an eye on to make sure it doesn't become a performance bottleneck.

### DIFF
--- a/platforms/ios/src/TangramMap/TGMarker.mm
+++ b/platforms/ios/src/TangramMap/TGMarker.mm
@@ -154,6 +154,7 @@
     bitmap.resize(w * h);
 
     CGColorSpaceRef colorSpace = CGImageGetColorSpace(cgImage);
+    // iOS only supports contexts with pre-multiplied alpha, so we transform it below.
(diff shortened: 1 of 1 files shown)


matteblair:
@tallytalwar can you take a look at this on Monday? The current Bintray slowness is causing Android builds to fail, hopefully resolved soon.

tallytalwar:
This is fine for the time being. But it does make sense to handle the logic by switching the rendering mode, what you mentioned in the comments. Will make sense to create an issue for the same so that its tracked.

PR #1786, opened 2018-04-05
Use sprite texture density aptly

tallytalwar:
Fixes the usage of the texture density property.

Based on the description https://github.com/tangrams/tangram/pull/234, and discussion with the JS and cartography crew, density defines divisor which is used to get the size of the sprites in css pixel units. The below implementation reflects this change and matches the behavior with the JS implementation.

### DIFF
--- a/core/src/gl/texture.cpp
+++ b/core/src/gl/texture.cpp
@@ -21,17 +21,19 @@
 
 namespace Tangram {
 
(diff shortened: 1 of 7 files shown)


tallytalwar:
Wait on this, I released, `density` can be applied and used for any texture and not just the use case of a spriteAtlas. Updating.

hjanetzek:
Yes - I also think density should be a Texture member if possible.

PR #1788, opened 2018-04-09
Implement % and auto sizing for point style

tallytalwar:
Refer the [docs](https://github.com/tangrams/tangram-docs/blob/2de1ea6941d492a6008f52ebbfda7635ecac8379/pages/draw.md#size-) for details.

- Introduced `SizeValue` as a `StyleParam::Value`
- Explicit handling of size stop evaluation
- Fix to take care of mixed dimension size stops

Explanation:
Since we no longer have a single "Unit" (pixels) for size style parameter, we can not simply encode size parameter values from style sheet in a `vec2`. We will need to have the unit specified for the size in the stylesheet. 
We will also require `cssSize` and `aspect ratio` of the texture/sprite being used for the feature drawn using a point style. This means we can not pre-calculate stops as `vec2` values before build the quad for the point being drawn, as `cssSize` and `aspect ratio` of the texture/sprites can only be determined by the style builders.

TODO:
- [x] A bit more code cleanup
- [x] Fix/add unit tests 
- [x] Remove JS function string value support for `StyleParam::size` as its not supported in TangramJS

### DIFF
--- a/core/src/scene/stops.cpp
+++ b/core/src/scene/stops.cpp
@@ -456,16 +456,18 @@ auto Stops::nearestHigherFrame(float _key) const -> std::vector<Frame>::const_it
 }
 
 void Stops::eval(const Stops& _stops, StyleParamKey _key, float _zoom, StyleParam::Value& _result) {
+
+    /* StyleParam::size stops can not have a generic evaluation, and
+     * requires more context and is handled in the pointStyleBuilder
+     */
+    if (StyleParam::isSize(_key)) { return; }
+
     if (StyleParam::isColor(_key)) {
         _result = _stops.evalColor(_zoom);
     } else if (StyleParam::isWidth(_key)) {
         _result = _stops.evalExpFloat(_zoom);
     } else if (StyleParam::isOffsets(_key)) {
         _result = _stops.evalVec2(_zoom);
-    } else if (StyleParam::isSize(_key)) {
-        // TODO: What should be done here!
-        // Size stop evaluation happens during pointStyleBuilder rule evaluation
-      //  _result = _stops.evalSize(_zoom);
     } else {
         _result = _stops.evalFloat(_zoom);
     }
--- a/core/src/scene/stops.cpp
+++ b/core/src/scene/stops.cpp
@@ -76,7 +76,7 @@ auto Stops::FontSize(const YAML::Node& _node) -> Stops {
     return stops;
 }
 
-auto Stops::Sizes(const YAML::Node& _node, const std::vector<Unit>& _units) -> Stops {
+auto Stops::Sizes(const YAML::Node& _node, uint8_t _units) -> Stops {
     Stops stops;
 
     // mixed dim stops not allowed for sizes (except when 1D stop uses %)
@@ -91,8 +91,8 @@ auto Stops::Sizes(const YAML::Node& _node, const std::vector<Unit>& _units) -> S
 
     auto constructFrame = [&](const auto& _frameNode, size_t offset, StyleParam::ValueUnitPair& _result) -> bool {
         if (StyleParam::parseSizeUnitPair(_frameNode.Scalar(), offset, _result)) {
-            if (std::find(_units.begin(), _units.end(), _result.unit) == _units.end()) {
-                LOGW("Size StyleParam can only take in pixel, % or auto values in: %s", Dump(_node).c_str());
+            if ( !(_units & _result.unit) ) {
+                LOGW("Size StyleParam can only take in pixel, %% or auto values in: %s", Dump(_node).c_str());
                 return false;
             }
         } else {
@@ -164,7 +164,7 @@ auto Stops::Sizes(const YAML::Node& _node, const std::vector<Unit>& _units) -> S
     return stops;
 }
 
-auto Stops::Offsets(const YAML::Node& _node, const std::vector<Unit>& _units) -> Stops {
+auto Stops::Offsets(const YAML::Node& _node, uint8_t _units) -> Stops {
     Stops stops;
     if (!_node.IsSequence()) {
         return stops;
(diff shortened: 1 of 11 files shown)


core/src/scene/styleContext.cpp
```diff
@@ -361,6 +361,14 @@ void StyleContext::parseStyleResult(StyleParamKey _key, StyleParam::Value& _val)
             case StyleParamKey::text_source_right:
                 _val = value;
                 break;
+            case StyleParamKey::size: {
```
matteblair: Tangram JS doesn't seem to support returning size strings from parameter functions, so we should keep that consistent.

tallytalwar: This is to handle `auto` and `%`. I need to check if this use case is supported on JS via functions.

tallytalwar: If not just simplifies things for us.

matteblair: Looks like it doesn't, based on http://tangram.city/play

tallytalwar: thanks will take it out.

tallytalwar: addressed. Thanks

tallytalwar:
Up for review now. I feel the StyleParam interface is being abused for a while now and has outgrown its initial design. Will start thinking on it for separate work, but not touching any of that for the purposes of sticking to the task of being parity with JS features.

core/src/style/pointStyleBuilder.cpp
```diff
@@ -329,39 +299,106 @@ bool PointStyleBuilder::getUVQuad(Parameters& _params, glm::vec4& _quad, Texture
...
+        const auto &atlas = _texture->spriteAtlas();
+        if (atlas) {
             if (!atlas->getSpriteNode(_params.sprite, spriteNode) &&
                 !atlas->getSpriteNode(_params.spriteDefault, spriteNode)) {
                 return false;
             }
+            cssSize = spriteNode.m_size * _texture->invDensity();
+            aspectRatio = spriteNode.m_size.x / spriteNode.m_size.y;
+        } else {
+            // missing sprite atlas for texture but sprite specified in draw rule
```
matteblair: I'm not clear how this determines that a sprite was specified in the draw rule.

core/src/style/pointStyleBuilder.cpp
```diff
@@ -329,39 +299,106 @@ bool PointStyleBuilder::getUVQuad(Parameters& _params, glm::vec4& _quad, Texture
...
+            if (std::isnan(size[0].value) && std::isnan(size[1].value)) {
+                _params.size = cssSize;
+            } else if (size[0].value == 0.f || std::isnan(size[1].value)) {
+                _params.size = glm::vec2(size[0].value);
+            } else {
+                _params.size = glm::vec2(size[0].value, size[1].value);
             }
+        } else {
+            // _texture/spriteNode must have been set!
+            assert(useTextureInfo);
```
matteblair: Do we really want to assert here? This seems like a scene authoring error and that should not stop the program. Maybe I misunderstand this logic?

matteblair:
Testing this feature has illuminated a more general problem with the way we're evaluating style parameters from JS functions in `StyleContext`: https://github.com/tangrams/tangram-es/blob/master/core/src/scene/styleContext.cpp#L355-L367

The logic here checks the type of the JS return value first, then based on that type, tries to assign it with the given style key. The problem here is in the `duk_is_string` block, where the `default` case calls the generic `StyleParam::parseString` function. Since any parameter can be parsed from a suitable string representation, this means that a JS function can return strings for any parameter and produce a successful output. For example, a point size can currently be specified as `function() { return '32px'; }` and you'll get a 32px point. It even works with the way that we stringify arrays: `function() { return '32, 16'; }` produces a 32px by 16px point. Clearly this isn't how the docs describe this parameter, and it's not how we meant it to work!

The obvious solution to try is to just remove the `parseString` from the default case. Unfortunately I am not sure whether any other parameters depend on this behavior, so we'll want to carefully check for any changes in styling when we remove it.

core/src/scene/stops.cpp
```diff
@@ -94,40 +113,53 @@ auto Stops::Sizes(const YAML::Node& _node, const std::vector<Unit>& _units) -> S
...
-                        LOGW("Size StyleParam can only take in pixel values in: %s", Dump(_node).c_str());
-                    }
-                }
 
-                stops.frames.emplace_back(key, sizeValue.value);
-            } else {
-                LOGW("could not parse node %s\n", Dump(frameNode[1]).c_str());
+            StyleParam::SizeValue sizeValue;
+            // default size values
+            sizeValue.fill({NAN, Unit::pixel});
```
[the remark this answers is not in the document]
tallytalwar: Will see what can be done here.

core/src/style/pointStyleBuilder.cpp
tallytalwar: Make sense, seems to be a job of a linter, but will change this to be a conditional check.

core/src/style/pointStyleBuilder.cpp
tallytalwar: Yup this is incorrect, will update.

tallytalwar:
Addressed your comments/suggestions @matteblair.

core/src/style/pointStyleBuilder.cpp
tallytalwar: Yeah I missed to update this earlier. Fixed now.

matteblair:
Reviewing again now. @tallytalwar did you decide against giving `SizeValue` a default constructor?

matteblair:
@tallytalwar I took a while to understand the changes here and I think I've been able to simplify it a little. Take a look at the changes I just pushed and let me know if we should edit or revert.

core/src/scene/stops.cpp
```diff
@@ -113,53 +113,29 @@ auto Stops::Sizes(const YAML::Node& _node, uint8_t _units) -> Stops {
...
-            if (!validStop || !constructFrame(sequenceNode[1], 0, sizeValue[1])) {
+            if (!constructFrame(sequenceNode[0], sizeValue.x) ||
+                !constructFrame(sequenceNode[1], sizeValue.y)) {
                 continue;
             }
-
+            has2DSize = true;
             stops.frames.emplace_back(key, sizeValue);
         }
+        if (has1DSize && has2DSize) {
```
tallytalwar: The difference here is the early exit. If one of the frames is "bad", we can save on some evaluations.

tallytalwar:
Ok All good. If you can approve it then I will merge it.

matteblair:
Probably makes sense to squash some of these commits in here, I'll let you merge @tallytalwar

PR #1792, opened 2018-04-16
Reduced build times on Travis CI

matteblair:
This is part refactoring, part optimization.
 - Reformats `.travis.yml` to be more clean and consistent
 - Edits Makefile to use generic `cmake --build` command so that other generators can be used
 - Uses `ninja` to execute builds for Linux, benchmark, and test targets on Travis (this doesn't actually save as much time as I'd hoped - `ninja` seems to mostly benefit incremental rebuilds)
 - Directs Travis to only build the armv7 ABI for Android PR builds (this uses an undocumented Gradle property that Android Studio injects when running builds for a device target - it may not be guaranteed to work forever but it cuts build times almost in half)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: android
-
-sudo: enabled
+os: linux
+dist: trusty
+sudo: false
 
 branches:
     only:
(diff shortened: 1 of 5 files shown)


matteblair:
Superseded by #1793

PR #1793, opened 2018-04-16
Run all builds on CircleCI

matteblair:
To improve CI build speed and convenience, this PR moves all of our automated builds, tests, and deployments to CircleCI.

 - Android build times for PRs are under 4 minutes
 - Linux builds with tests and benchmarks ~2 minutes faster than on Travis, can likely be much faster if we use a Docker container
 - CircleCI interface is much easier to for navigating and interpreting logs
 - CircleCI structure fits better with our needs (heterogeneous tasks with multiple languages)
 - All of our builds in one place

TODO:
- [x] Condense git history
- [x] Create a Docker image for linux builds?
- [x] Update Github settings for PR status checks
- [x] Update build status badges


tallytalwar:
We can do some cleanup here remove the following commits from this PR:

7f4b712570edd20032fa7d539dcbac33b76b4acb
4b461e7462063d92f302bf6ae884955d3631c742
b69e8d675b8b877c2187d6c7a65323b983fb7a79

tallytalwar:
Also we can remove `.travis.yml` all together.

.circleci/config.yml
```diff
@@ -1,5 +1,47 @@
...
+            cp bench/test_tile_10_301_384.mvt build/linux/bin/tile.mvt
+            cd ./build/linux/bin
+            for file in bench/*.out
+              do
+                echo "Running ${file}"
+                $file
+              done
+  build-android:
+    docker:
+      - image: lakoo/android-ndk:25-26.0.1-r15c
```
tallytalwar: As a note, we should replace this with circle supported android image (sdk + ndk) when this gets available from circle (soonish it seems). Refer: https://github.com/circleci/circleci-images/pull/177

.circleci/config.yml
```diff
@@ -1,5 +1,47 @@
...
+  build-deploy-android-snapshot:
+    docker:
+      - image: lakoo/android-ndk:25-26.0.1-r15c
+    steps:
+      - checkout
+      - run: git submodule update --init
+      - run: cd platforms/android && ./gradlew uploadArchives
   build-ios:
     macos:
       xcode: "9.0"
```
tallytalwar: https://circleci.com/docs/2.0/testing-ios/#supported-xcode-versions

.circleci/config.yml
```diff
@@ -91,6 +145,7 @@ workflows:
           filters:
             branches:
               only: master
+      # Run on tag pushes
```
tallytalwar: This makes me think we should automate android release deployment to bintray as well. Since we can do that now unlike earlier with SonaType!

Makefile
```diff
@@ -316,43 +297,39 @@ endif
...
+	cmake --build ${LINUX_BUILD_DIR}
 
 cmake-linux:
-	mkdir -p ${LINUX_BUILD_DIR}
-	cd ${LINUX_BUILD_DIR} &&\
+	@mkdir -p ${LINUX_BUILD_DIR}
+	@cd ${LINUX_BUILD_DIR} && \
 	cmake ../.. ${LINUX_CMAKE_PARAMS}
 
 tizen-arm: cmake-tizen-arm
```
tallytalwar: We should these tizen out. (I said it!)

scripts/travis/.travis.yaml
```diff
@@ -0,0 +1,59 @@
+language: android
```
tallytalwar: We can remove travis config and all the travis script files before we merge this PR.

.circleci/config.yml
matteblair: CircleCI does support newer versions, but in general upgrading tools causes sadness. Unless there's a compelling reason to upgrade I'd prefer to leave it the same.

.circleci/config.yml
matteblair: I agree! Sounds like a good idea for another PR.

Makefile
matteblair: Might make sense to remove the Tizen build, it's probably broken now anyway. Since that isn't the purpose of this PR I'll leave it as is for now.

.circleci/config.yml
tallytalwar: If its not a big change then could be done part of the same PR, but I am fine either way.

Makefile
tallytalwar: Yup, should be done sooner rather than later as its not officially supported and is probably broken!

.circleci/config.yml
tallytalwar: Hmm, I agree with sadness!
But I am of an opinion that CI builds should reflect the build system we use for our local builds. At least the release builds should be done using the latest supported versions and if not we should document the releases to be build on a specific xcode version like in this case.

matteblair: Updated to Xcode 9.2.0, which is what I've got locally. Seems fine!

PR #1794, opened 2018-04-17
[WIP] Best practices with atomics (1)

tallytalwar:
- Its recommended to use the member functions
- Specially for atomic counters one can get the max speed up with
std::memory_order_relaxed, as no operations/data writes/data reads are
dependent on the memory write/read orders.

I will see if I can benchmark before and after affects of this change. Based on my readings one should get some speed up as the default memory fence ordering when using `operator++`/`fetch_add` or `operator--`/`fetch_sub` is `std::std::memory_order_seq_cst`, which can be very strict for the purposes of a counter. Refer: http://en.cppreference.com/w/cpp/atomic/memory_order#Sequentially-consistent_ordering

### DIFF
--- a/core/src/data/tileSource.cpp
+++ b/core/src/data/tileSource.cpp
@@ -24,7 +24,7 @@ TileSource::TileSource(const std::string& _name, std::unique_ptr<DataSource> _so
 
     static std::atomic<int32_t> s_serial;
 
-    m_id = s_serial++;
+    m_id = s_serial.fetch_add(1, std::memory_order_relaxed);
 }
 
 TileSource::~TileSource() {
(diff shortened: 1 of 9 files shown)


cleeus:
Uhoh, it's already hard to reason about thread safety in Tangram - using memory_order_relaxed make this impossible for the mortals among us.
Did you watch

tallytalwar:
@cleeus This was just a test but the idea was to use the member function instead of overloaded operators for atomics, I am actually going to close this and reopen with just changes to use member functions.

PS. Yeah I read a lot of memory fences and watches a few videos too :D. These are interesting concepts I was not aware off :D

PR #1795, opened 2018-04-23
Thread sanitization

hjanetzek:
Fix issues shown by thread-sanitizer.


tallytalwar:
Still looking at this, few things so far:

1. `atomic` operations on smart pointers are not supported on gcc 4.9 and update iOS toolchain
2. Following seems to misbehave at least in the unit tests:
```
    std::shared_ptr<Tile> tile() { return std::atomic_load(&m_tile); }
```
3. Since we are refactoring this change, I would suggest to use `unique_ptr` for task's `tile`, which get moved to `tileManager` from the `tileTask` when the tile is ready to be drawn.

tallytalwar:
In line of minimum gcc version requirement, we should also update the minimum supported clang version for tangram-es, with this PR. `clang 3.7` had a bunch of compilation errors with this branch for me on osx.

toolchains/linux.cmake
```diff
@@ -5,8 +5,13 @@ set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fno-omit-frame-pointer")
 
 if (CMAKE_CXX_COMPILER_ID MATCHES "Clang")
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-gnu-zero-variadic-macro-arguments")
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -stdlib=libc++")
-  set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS}  -lc++ -lc++abi")
+
+  #set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -stdlib=libc++")
+  set(CMAKE_E
+    XE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS}  -lc++ -lc++abi")
```
matteblair: This looks like a typo?

matteblair:
Hey nice! This seems to have resolved the test failures and build errors. `unique_ptr` is definitely the right choice for this use in `TileTask`. I spotted what looks like a typo in a CMake file but overall it looks good, so I'll go ahead and approve. Merge when ready @hjanetzek !

PR #1798, opened 2018-04-24
Define Tangram::Point as vec2

matteblair:
The z component of Point in TileData was never used. This should reduce the size of allocations for `Feature`s.

### DIFF
--- a/core/src/data/tileData.h
+++ b/core/src/data/tileData.h
@@ -60,11 +60,11 @@ enum GeometryType {
     polygons
 };
 
-typedef glm::vec2 Point;
+using Point = glm::vec2;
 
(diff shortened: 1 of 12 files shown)


core/src/data/tileData.h
```diff
@@ -63,7 +60,7 @@ enum GeometryType {
     polygons
 };
 
-typedef glm::vec3 Point;
```
tallytalwar: While we are on it and touching this file, can you change these to use `using`. Its just cosmetic, but thats what the standard is going with these days!

matteblair: Makes sense - done.

PR #1801, opened 2018-05-03
Batch GL resource deletion

karimnaaji:
I have noticed a few cpu stalls on deletion of gl resources that can peak over 100ms for some frames on my machine (disclaimer: it was not on mobile, I don't know if it is worse or similar but worth testing!).

First run (master) panning around manhattan (look for `Map::render_runJobs`, corresponding to this specific call https://github.com/tangrams/tangram-es/blob/master/core/src/map.cpp#L509):

screen shot 2018-05-02 at 11 46 52 pm

With this change for a similar interaction:
screen shot 2018-05-02 at 11 56 31 pm

Attached are the traces from the screenshot, to load go in the tool `about:tracing` in a new chrome tab:
[traces.zip](https://github.com/tangrams/tangram-es/files/1969601/traces.zip)


tallytalwar:
On this now!

core/src/gl/mesh.cpp
```diff
@@ -34,25 +34,15 @@ MeshBase::MeshBase(std::shared_ptr<VertexLayout> _vertexLayout, GLenum _drawMode
...
-
-    m_disposer([=](RenderState& rs) mutable {
-        // Deleting a index/array buffer being used ends up setting up the current vertex/index buffer to 0
-        // after the driver finishes using it, force the render state to be 0 for vertex/index buffer
-        if (glVertexBuffer) {
-            rs.vertexBufferUnset(glVertexBuffer);
-            GL::deleteBuffers(1, &glVertexBuffer);
+    if (m_rs) {
+        if (m_glVertexBuffer) {
+            m_rs->queueBufferDeletion(m_glVertexBuffer);
```
karimnaaji: Just thought that could lock only once by changing the interface a bit, by queuing vao/index/vertex buffer at the same time, same for shaders.

matteblair: Good idea, I'll add that now.

matteblair: Added! I left VAOs separate from buffers because it was a less obvious change.

core/src/gl/renderState.cpp
```diff
@@ -37,13 +37,74 @@ RenderState::RenderState() {
...
+}
+
+void RenderState::queueTextureDeletion(GLuint texture) {
+    std::lock_guard<std::mutex> guard(m_deletionListMutex);
+    m_textureDeletionList.push_back(texture);
+}
+
+void RenderState::queueVAODeletion(size_t count, GLuint* vao) {
+    std::lock_guard<std::mutex> guard(m_deletionListMutex);
+    m_VAODeletionList.insert(m_VAODeletionList.end(), vao, vao + count);
```
tallytalwar: I think we might benefit by doing a `reserve` here before `insert`.

karimnaaji: I think `count > 1` is pretty rare? I believe there is several VAOs when meshes don't fit under 65k index limit, but maybe reserve an arbitrary size at initialization if you noticed reallocs here.

karimnaaji:
@tallytalwar fine with me!

core/src/gl/renderState.cpp
tallytalwar: We do see a few cases when meshes are >65k index limit and hence count > 1. But yes you are right there are not many which will cause significant perf issues here! I will leave it as it is right now.

PR #1808, opened 2018-05-17
Fly to animation

hjanetzek:


platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
```diff
@@ -593,6 +593,18 @@ public float getTilt() {
         return nativeGetTilt(mapPointer);
     }
 
+    /**
+     * Run flight animation to change postion and zoom  of the map
+     * @param tilt Tilt angle in radians; 0 corresponds to straight down
+     * @param duration Time in milliseconds to ease to the given tilt
+     * @param ease Type of easing to use
+     */
+    public void flyTo(final double lon, final double lat, final float zoom, final float duration, final float speed) {
```
matteblair: I think we'll also want overloads for this, one with only duration and one with neither duration nor speed. It would be nice to specify only speed as well, but that won't work without named arguments.

core/src/view/flyTo.cpp
```diff
@@ -0,0 +1,98 @@
...
+
+    auto fn = [=](float t) {
+
+                 if (t >= 1.0) {
+                     return end;
+
+                 } else if (move) {
+                     double s = S * t;
+                     glm::dvec2 pos = glm::mix(c0, c1, u(s) / u1);
+                     double zoom = start.z + (std::log(1.0 / w(s)) / std::log(2.0));
```
matteblair: `log(1.0 / x)` is just `-log(x)` right? And I think `log2` is part of the STL (http://en.cppreference.com/w/cpp/numeric/math/log2), it's used in `getMinimumEnclosingZoom` above.

platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
```diff
@@ -593,6 +593,20 @@ public float getTilt() {
...
 
+    /**
+     * Run flight animation to change postion and zoom  of the map
+     * @param position LngLat of the position to set
+     * @param zoom Zoom level; lower values show more area
+     * @param duration Time in milliseconds to ease to given zoom
+     * @param speed If duration is 0, speed is used as factor to change the duration that is
+     *              calculated for the duration of the flight path. (Recommended range 0.1 - 2.0)
+     */
+    public void flyTo(@NonNull final LngLat position, final float zoom, final float duration, final float speed) {
```
matteblair: If `duration` is in milliseconds then we should make it an `int` for consistency with the other easing methods.

matteblair: I still think overloads may be useful, but we can always add them later without breakage.

hjanetzek: Thanks for the suggestions! I'll finish this up on Monday

core/src/map.cpp
```diff
@@ -638,6 +647,42 @@ void Map::setZoomEased(float _z, float _duration, EaseType _e) {
...
+                               distance);
+    auto cb =
+        [=](float t) {
+            glm::dvec3 pos = fn(t);
+            impl->view.setPosition(pos.x, pos.y);
+            impl->view.setZoom(pos.z);
+            impl->platform->requestRender();
+        };
+
+    float duration = _duration > 0 ? _duration : distance * _speed;
```
tallytalwar: Oops I think you meant to have a `/` here?
Time = Distance / Speed!

```
float duration = _duration > 0 ? _duration : distance / _speed;
```

core/src/view/flyTo.cpp
```diff
@@ -0,0 +1,85 @@
...
+    glm::dvec2 aMeters = projection.LonLatToMeters(glm::dvec2(aLng, aLat));
+    glm::dvec2 bMeters = projection.LonLatToMeters(glm::dvec2(bLng, bLat));
+    double distance = glm::distance(aMeters, bMeters) * (1. + buffer);
+    double focusScale = distance / (2. * MapProjection::HALF_CIRCUMFERENCE);
+    double viewScale = view.getWidth() / projection.TileSize();
+    double zoom = -log2(focusScale / viewScale);
+    return zoom;
+}
+
+std::function<glm::dvec3(float)> getFlyToFunction(const View& view, glm::dvec3 start, glm::dvec3 end, double& _duration) {
```
tallytalwar: It will be good to add the link to the paper we are referring as a comment here.

tallytalwar: `_duration` should actually be called `_distance`. Its used correctly in the callee above in `map.cpp` but not in the function definition and declaration.

core/src/view/flyTo.h
```diff
@@ -0,0 +1,14 @@
...
+#include <functional>
+#include "glm/glm.hpp"
+
+namespace Tangram {
+
+class View;
+
+float getMinimumEnclosingZoom(double aLng, double aLat, double bLng, double bLat, const View& view, float buffer);
+
+std::function<glm::dvec3(float)> getFlyToFunction(const View& view, glm::dvec3 start, glm::dvec3 end, double& _duration);
```
tallytalwar: Can you add some documentation on the intent of this? And `_duration` should be `_distance` here.

core/src/view/flyTo.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+    const double w0 = std::max(width, height);
+    const double w1 = w0 / scale;
+
+    const glm::dvec2 c0{start.x, start.y};
+    const glm::dvec2 c1{end.x, end.y};
+
+    const double u1 = glm::distance(c0, c1);
+
+    auto b = [=](int i) {
```
tallytalwar: Will be good to document what these `b`, `r`, `u`, `w` and `S` represent here. One will save time to not refer to equation 9 on the implemented paper.

core/src/view/flyTo.cpp
```diff
@@ -0,0 +1,85 @@
+#include "view/flyTo.h"
+
+#include "util/mapProjection.h"
+#include "view/view.h"
+#include <cmath>
+#include "log.h"
+
+namespace Tangram {
+
+float getMinimumEnclosingZoom(double aLng, double aLat, double bLng, double bLat, const View& view, float buffer) {
```
tallytalwar: Is `getMinimumEnclosingZoom` being used? Is this required?

matteblair: I wrote this function as part of my first implementation of `flyTo`. I don't think it's used now, but we'll need it when we enable moving the camera to a target area.

PR #1812, opened 2018-05-22
[WIP] Windows platform support

tehKaiN:
Not ready yet!


Makefile
```diff
@@ -1,3 +1,17 @@
+# Multiplatform stuff
+ifdef ComSpec
+	# Windows shell
+	RMPATH = rmdir /S /Q
+	MKPATH = md
+	# A hack to suppress "already exists" error
+	MKPATH_QUIET = 2>nul || VER>nul
+else
+	# Civilized rest of the world
```
matteblair: 😂

matteblair:
Hey thanks for all the work on this port @tehKaiN ! Windows support has been missing for too long. I especially appreciate the documentation for different compiler toolchains.

It's been difficult to find time for code review on this because our core team only has macs during working hours, but I will commit that this weekend I'm going to spend time trying this out and hopefully we can get this merged.

tehKaiN:
How's it goin'? ;)

matteblair:
https://github.com/tangrams/tangram-es/tree/windows-v2

My goal here has been to get the sample app working with Visual Studio - since that's the only build toolchain that I really use on Windows :D It doesn't seem impossible, but there still some build issues with the Mapbox Variant submodule I need to figure out. 

One thing that simplified the build process greatly for me was using [Conan](https://conan.io/) to manage library dependencies like Zlib and libCURL. On Windows the complexity of managing binary dependencies is great enough that something like Conan actually makes sense, and it worked exactly as advertised.

matteblair:
In regards to merging these windows changes, I don't mind merging things as they are and continuing to work on the build configuration incrementally. This work is almost totally separate from the other platform targets, so there isn't much risk of breaking something else.

I think it will be important to add a CI configuration (probably [Appveyor](https://www.appveyor.com/)) to check our Windows builds regularly. Otherwise, we're just going to break Windows all the time - since none of us maintainers work on Windows. 

@tallytalwar @hjanetzek Thoughts?

tallytalwar:
FYI, I also started looking at this at my home pc. More eyes the better. Will update more this week.

tehKaiN:
I'm fine with that as long as it gets merged eventually. ;)

I think I'm done here. Is there anything else I should do for this PR?

tallytalwar:
Tiny update.

Thanks @tehKaiN for documenting this PR so well, specially all the possible windows platform toolchains. I have the basic setup done on my home machine and plan to finish reviewing this the following week.

I agree with @matteblair about having dependency management (can be added as a subsequent PR) and CI (with this PR or immediately after this, though should not be too hard). 
On the GL front, I am happy with the usage of `GLAD`. Less unsure of whats tangram's future with gl library is in lieu of iOS deprecating opengl-es, but we this is "totally" fine for our purposes right now.

More to followup next week on this.

tehKaiN:
Guys?

tehKaiN:
I see that now there are conflicts in some files. If anyone is willing to review my changes and eventually merge them I can fix them for you.

karimnaaji:
@tehKaiN @matteblair @tallytalwar I can review this branch if there is still interest in getting windows platform in main.

tallytalwar:
@karimnaaji go ahead please :D.

tehKaiN:
@karimnaaji the file structure and build system has changed a bit since beginning of this pull request. You want me to make this stuff compatible with master, or would you like to do it yourself?

tehKaiN:
@karimnaaji I've added required changes on top of recent master since I need this lib for my company's app and I like to be sync'd with latest stuff. You can look at it here: https://github.com/tehKaiN/tangram-es/tree/windows-v3

Unfortunately there are some crashes as described in #1569 which weren't present before.

I was hoping for wxWidgets platform support which I have implemented too, but I think incorporating it to your repo won't work in the long run - I'll refactor it as a lib with libtangram as dependency when this pull request is over.

karimnaaji:
@tehKaiN sounds good! So far I've rebased the branch locally, and tried a few things:
- MSVC+Conan: I really wanted to get MSVC build with Conan working but I think there would be a lot of complications from it for this repo, as most of the dependencies tangram relies on are not supporting that toolchain, this would result in being a pain to update those upstream every time. The advantage from that is that there is already the built dependencies Zlib/Curl/OpenSSL available on their server. 
- Conan+MinGW gcc: I've tried using Conan with mingw gcc and build Zlib/Curl/OpenSSL from source with a custom profile, but Conan didn't really like that either and was breaking in their generated compilation python code, which seems like something we could report on https://github.com/conan-io/conan but for now I think the solution you had was the simplest (MinGW + manually built dependencies), though not ideal as this is not the most common toolchain setup. Maybe not relying on curl at all would be the best (in a later step).

tehKaiN:
@karimnaaji perhaps it'd be simpler to make it work with bare CMake as rest of platforms are supported. You can resolve CURL dependency by building it using CMake and then using `make install` as mentioned in platform readme on my branches.

Having libs installed in predefined directory simplifies things a lot, just don't use default install path (Program Files) as it requires elevated privileges.

And yes, forget about MSVC for now as its support for latest C++ is a pos. Even Microsoft adds Clang to Visual Studio installers, and there's no coincidence in that. ;) Unfortunately Clang on Windows uses MSVC compatibility mode instead of GCC one as on rest of platforms, which I haven't figured out how to disable.

PR #1819, opened 2018-06-11
[wip] Camera API

hjanetzek:
To decide:
- Add option to make camera non-interruptible?
- Support chaining of animations? i.e. append next CameraPosition when the current ease was not manually canceled?


core/include/tangram/map.h
```diff
@@ -129,6 +139,8 @@ class Map {
     // The callback may be be called from the main or worker thread.
     void setSceneReadyListener(SceneReadyCallback _onSceneReady);
 
+    void setCameraAnimationListener(CameraAnimationCallback _cb);
```
matteblair: @hjanetzek what do you think about taking a camera callback as an argument to each animation call? Like
```
void setCameraPositionEased(const CameraPosition& _camera, float duration, EaseType _e, CameraAnimationCallback _cb);
```
The way I see callbacks being used is to trigger events in an app UI after the camera has finished moving to a specific location. This behavior is probably different for each call site, so rather than having one callback set for the map and switching behavior based on state inside the callback, we can make the state implicit by associating one callback with each animation call.

hjanetzek: Yes we could also do it that way. I made it analog to the `SceneReadyCallback`, though on the Android side the event already is dispatched to the callback passed to `setCameraPositionEased`. 
```
 public void setCameraPositionEased(@NonNull final CameraPosition camera, final int duration, @NonNull final EaseType ease, final CameraAnimationCallback cb) 
```

hjanetzek: At least for Android it would not help much I guess. Holding onto the java callback object through the native callback function seems more complicated.

rwrx: Hello, sorry for the interruption, but I needed some parity in Tangram-ES with Google Maps on Android so I have implemented cancel and finish callbacks into all easing animations - position, zoom, rotation and tilt. I have working implementation also in production app, but it is in private repository on GitLab, but I can commit these changes into my Tangram-ES forked GitHub repository and maybe try to merge it into this repo? 

I haven't done so before because I am not sure if you wanted to add this functionality and I am not sure if I have done it right way. And also this is rather big change to the interface. Merging my changes into this repository might be also for benefit for me because I will have less conflicts when pulling your changes into my repo.

matteblair: @rwrx Can you describe what animation features you specifically needed for your application? We weren't planning on having separate animation calls for each camera parameter but maybe there are use cases for these that I haven't foreseen.

rwrx: @matteblair I just needed api which can do position, zoom, rotation and tilt animations, because I am also using Google Maps API in my app and I needed also cancel and finish callbacks, mainly due to some logic inside app - for example during the animation I needed to disable some functionality and then right after the animation finished I needed to enable that functionality again. Here is this API in Google Maps: https://developers.google.com/android/reference/com/google/android/gms/maps/GoogleMap#animateCamera(com.google.android.gms.maps.CameraUpdate,%20com.google.android.gms.maps.GoogleMap.CancelableCallback).

I am not sure what you meant by "separate animation calls for each camera parameter", because Tangram-ES already have Map::setPositionEased(), Map::setZoomEased(), Map::setRotationEased and Map::setTiltEased.

tallytalwar: > At least for Android it would not help much I guess. Holding onto the java callback object through the native callback function seems more complicated.

This was also the reason for not doing it for `SceneReadyCallback`. But I think we did discuss the complexity of doing it to not be huge!

tallytalwar: @rwrx I see, the aim of this work is to be as close to what google and other mapping provider apis are with camera updates. Can you provide a link to your work (maybe gitlab link itself)?

rwrx: @tallytalwar I have tried to merge my GitLab repo into GitHub fork of Tangram ES, I am not completely sure if it was merged correctly (I haven't tried to compile it), but for example you can see how I have done this easing here: https://github.com/rwrx/tangram-es/blob/master/core/src/map.cpp#L612. Also support for these callbacks has changes in other files, but currently I can't remember exact parts which I have changed as I have done this one year ago. And also it is possible that these parts are merged incorrectly. I just wanted to give you some overview.

core/include/tangram/map.h
```diff
@@ -166,56 +178,54 @@ class Map {
...
-    // if duration (in seconds) is provided, tilt eases to the set value over the
-    // duration; calling either version of the setter overrides all previous calls
+    // Set the tilt angle of the view in radians; 0 corresponds to straight down
     void setTilt(float _radians);
-    void setTiltEased(float _radians, float _duration, EaseType _e = EaseType::quint);
 
     // Get the tilt angle of the view in radians; 0 corresponds to straight down
     float getTilt();
 
+    void getEnclosingViewPosition(LngLat a, LngLat b, float bufferMeters, LngLat& center, float& zoom);
```
matteblair: Does `bufferMeters` indicate padding around the target area measured in meters? I think it would be much more useful to specify this padding in pixels or at least as a fraction of the viewport.

matteblair: Also might be easier to use if it returned a `CameraPosition`.

matteblair: Actually the implementation of this looks like `bufferMeters` is a fractional value - is this just an unclear name?

platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
```diff
@@ -423,172 +428,243 @@ public void setHttpHandler(final HttpHandler handler) {
...
     /**
-     * Set the geographic position of the center of the map view
-     * @param position LngLat of the position to set
+     * Set the camera position of the map view
+     * @param camera CameraPosition to set
      */
-    public void setPosition(@NonNull final LngLat position) {
-        checkPointer(mapPointer);
-        nativeSetPosition(mapPointer, position.longitude, position.latitude);
+    public void setCameraPosition(@NonNull final CameraPosition camera) {
```
matteblair: Theoretically we can get rid of the methods that move the camera with a `CameraPosition` - using a `CameraUpdate` covers a superset of their functionality.

Pros:
 - Makes the interface simpler and easier to understand
 - Less code, easier to maintain

Cons:
 - Creates more new objects than strictly necessary, could worsen GC

platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
```diff
@@ -423,172 +428,243 @@ public void setHttpHandler(final HttpHandler handler) {
...
     }
 
     /**
-     * Set the rotation of the view with default easing
-     * @param rotation Counter-clockwise rotation in radians; 0 corresponds to North pointing up
-     * @param duration Time in milliseconds to ease to the given rotation
+     * Set the geographic position of the center of the map view
+     * @param position LngLat of the position to set
+     */
+     public void setPosition(@NonNull final LngLat position) {
```
matteblair: Do we need these direct setters? I'm guessing the idea here is to enable brevity for simple changes - but I suspect that changes to a single parameter like this are quite rare in actual applications!

matteblair:
As I'm reviewing this I remembered that I started on a different implementation of a similar interface on Android over two years ago: https://github.com/tangrams/tangram-es/commit/6cf24c25612a09c2a78ffd4752ff7451fdb1d307

matteblair:
I'm unsure about allowing an arbitrary configuration of `CameraUpdate`. e.g. If you set both a "zoom by" and a "zoom to" then what is the resulting zoom? I'm starting to see the appeal to a camera interface like google maps'.

hjanetzek:
It certainly needs to be documented how relative and absolute setters should interact. I suggest that one can set either one or the other - when relative values are set they are always relative to the current camera position and absolute values override anything that was set before.

matteblair:
Documenting the behavior is good, but I think it's better if the interface makes these details self-evident or makes the compiler enforce them.

I can think of quite a few ways to build contradictory camera updates.
- Set a zoom and a zoomBy (same with rotation or tilt)
- Set a LngLat and a bounds
- Set a LngLat and a CameraPosition
- Set a bounds and a CameraPosition
- Set a bounds and a zoom (or zoomBy)
- etc.

Rather than choosing rules to disambiguate all of these cases and then requiring users to be aware of them all, it seems easier to have methods for building the (somewhat small) set of plausible camera updates - something like: https://developers.google.com/android/reference/com/google/android/gms/maps/CameraUpdateFactory

matteblair:
I was trying to test out the callback behavior on Android with the following diff:
```diff
diff --git a/platforms/android/demo/src/main/java/com/mapzen/tangram/android/MainActivity.java b/platforms/android/demo/src/main/java/com/mapzen/tangram/android/MainActivity.java
index a8846fb6..943230e7 100644
--- a/platforms/android/demo/src/main/java/com/mapzen/tangram/android/MainActivity.java
+++ b/platforms/android/demo/src/main/java/com/mapzen/tangram/android/MainActivity.java
@@ -221,7 +221,17 @@ public class MainActivity extends AppCompatActivity implements MapController.Sce
         map.pickLabel(x, y);
         map.pickMarker(x, y);
 
-        map.updateCameraPosition(CameraUpdateFactory.setPosition(tappedPoint), 1000);
+        map.updateCameraPosition(CameraUpdateFactory.setPosition(tappedPoint), 1000, MapController.EaseType.CUBIC, new MapController.CameraAnimationCallback() {
+            @Override
+            public void onFinish() {
+                Log.d("Tangram","finished!");
+            }
+
+            @Override
+            public void onCancel() {
+                Log.d("Tangram","canceled!");
+            }
+        });
 
         return true;
     }
```

I expected the "finished" branch to run, since I never cancelled the animation. But instead, the "cancelled" branch is run immediately, and the animation _also_ runs to completion. Unless I'm misunderstanding the usage here, something is going wrong!

core/src/map.cpp
```diff
@@ -582,69 +586,206 @@ void Map::captureSnapshot(unsigned int* _data) {
...
+    float radiansDelta = glm::mod(_camera.rotation - radiansStart, (float)TWO_PI);
+    if (radiansDelta > PI) { radiansDelta -= TWO_PI; }
+
+    e.start.rotation = radiansStart;
+    e.end.rotation = radiansStart + radiansDelta;
+
+    e.start.tilt = getTilt();
+    e.end.tilt = _camera.tilt;
+
+    impl->eases.push_back(Ease{_duration,
```
tallytalwar: `emplace_back` is a bit pleasing to the eyes.

core/src/map.cpp
```diff
@@ -582,69 +586,206 @@ void Map::captureSnapshot(unsigned int* _data) {
...
+        camera = getCameraPosition();
+    }
+    if ((_update.set & CameraUpdate::SET_BOUNDS) != 0) {
+        camera = getEnclosingCameraPosition(_update.bounds[0], _update.bounds[1], _update.boundsPadding);
+    }
+    if ((_update.set & CameraUpdate::SET_LNGLAT) != 0) {
+        camera.longitude = _update.lngLat.longitude;
+        camera.latitude = _update.lngLat.latitude;
+    }
+    if ((_update.set & CameraUpdate::SET_LNGLAT) != 0) {
```
tallytalwar: Looks like this is a dup of the above condition.

core/src/map.cpp
```diff
@@ -681,77 +822,9 @@ void Map::flyTo(double _lon, double _lat, float _z, float _duration, float _spee
...
-    impl->setEase(EaseField::tilt, { _duration, cb });
-
-}
-
-float Map::getTilt() {
-
-    return impl->view.getPitch();
+    cancelCameraAnimation();
 
+    impl->eases.push_back(Ease{duration, cb});
```
tallytalwar: same, `emplace_back` is pleasing to eyes. (lesser `{}`)

core/src/map.cpp
```diff
@@ -582,69 +586,202 @@ void Map::captureSnapshot(unsigned int* _data) {
...
+    impl->view.setPitch(_radians);
+    impl->platform->requestRender();
+}
 
+float Map::getTilt() {
+    return impl->view.getPitch();
 }
 
-void Map::setZoomEased(float _z, float _duration, EaseType _e) {
+CameraPosition Map::getEnclosingCameraPosition(LngLat _a, LngLat _b, int _buffer) {
```
matteblair: Seems like it should be pretty simple to allow separate buffer values for each dimension - that would match the behavior of other SDKs.

platforms/android/tangram/src/main/java/com/mapzen/tangram/CameraPosition.java
```diff
@@ -0,0 +1,53 @@
...
+        if (o == null || getClass() != o.getClass()) return false;
+        CameraPosition that = (CameraPosition) o;
+        return Double.compare(that.longitude, longitude) == 0 &&
+                Double.compare(that.latitude, latitude) == 0 &&
+                Float.compare(that.zoom, zoom) == 0 &&
+                Float.compare(that.rotation, rotation) == 0 &&
+                Float.compare(that.tilt, tilt) == 0;
+    }
+
+    public void set(CameraPosition camera) {
```
matteblair: Probably want a `@NonNull` here

PR #1824, opened 2018-06-15
[Android] Drop map usage for marker, raise minSDK to 19

Dean79000:
Raise Android minSDK to 19 (Android 4.4) :  Current minSDK 15 has been released in 2011, and it not be used anymore (0.4%  -> near same usage of Android 2.3).

This allow to use "new" (5 year ago) classes, permits to reduce memory footprint.

### DIFF
--- a/platforms/android/demo/build.gradle
+++ b/platforms/android/demo/build.gradle
@@ -6,7 +6,7 @@ android {
   def apiKey = project.hasProperty('nextzenApiKey') ? nextzenApiKey : System.getenv('NEXTZEN_API_KEY')
 
   defaultConfig {
-    minSdkVersion 15
+    minSdkVersion 19
     targetSdkVersion 27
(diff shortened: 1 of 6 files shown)


matteblair:
Hey thanks for bringing `minSdkVersion` to our attention - clearly we haven't re-assessed it for a while. Google reports SDK version usage here in a convenient chart: https://developer.android.com/about/dashboards/

With SDK version 15 usage at 0.4% I am totally on-board with dropping support going forward. However the other SDK versions below 19 still have 4.3% usage, which - though small - isn't negligible. 

My preference is to raise `minSdkVersion` to 16 and use the features it provides - while keeping SDK-level checks for newer features. Although these checks can be a little ugly in the code, they ensure that only the older SDKs use the less-efficient implementations (which I think is still better than not running at all).

PR #1826, opened 2018-06-26
Angle style param should respect numeric value from JS Function

tallytalwar:
Refer discussion https://github.com/tangrams/tangram-docs/issues/250

### DIFF
--- a/core/src/scene/styleContext.cpp
+++ b/core/src/scene/styleContext.cpp
@@ -466,6 +466,7 @@ void StyleContext::parseStyleResult(StyleParamKey _key, StyleParam::Value& _val)
                 _val = StyleParam::Width{static_cast<float>(v)};
                 break;
             }
+            case StyleParamKey::angle:
             case StyleParamKey::text_font_stroke_width:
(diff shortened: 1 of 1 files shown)


core/src/scene/styleContext.cpp
```diff
@@ -487,6 +487,11 @@ void StyleContext::parseStyleResult(StyleParamKey _key, StyleParam::Value& _val)
                 _val = static_cast<uint32_t>(duk_get_uint(m_ctx, -1));
                 break;
             }
+            case StyleParamKey::angle: {
+                std::string value = doubleToString(static_cast<double>(duk_get_number(m_ctx, -1)));
+                _val = StyleParam::parseString(_key, value);
```
matteblair: Can't we just assign the double directly to the parameter value?

tallytalwar: Ohh yes, since `auto` return will be handled by `duk_is_string`.

tallytalwar: Updated!

tallytalwar: K actually fixed now!

PR #1827, opened 2018-06-26
Refactored CMake configuration with better iOS integration

matteblair:
This is a major improvement to the organization of our build system! It's a big diff, but the changes are mostly all following one of these strategies:
- **Apply properties to CMake targets rather than entire directory subtrees.** This ensures that compile options and include directories are scoped to only the targets that require them.
- **List source files instead of globbing directories.** Collecting files with GLOB is *not* recommended in CMake scripts (see the [docs](https://cmake.org/cmake/help/v3.2/command/file.html?highlight=GLOB)). When you change the set of source files for a target you should also change the CMake file for it so that CMake knows that the build files need to be regenerated.
- **Use PUBLIC and PRIVATE keywords to declare usage requirements.** This allows CMake to correctly propagate include directories and compile options as usage requirements between targets.
- **Remove CMake logic that isn't used.** There was a lot of it.
- **Reorganize platform-specific configuration files.** Here's how this works after this change:
  - Builds always start with the CMakeLists.txt in the project root.
  - When invoking the root CMake file you must include a `TANGRAM_PLATFORM` parameter, like `-DTANGRAM_PLATFORM=linux`.
  - The root CMake looks for a file at `platforms/$TANGRAM_PLATFORM/config.cmake` and includes it if found.
  - The platform `config.cmake` file then sets any global toolchain-related variables required and configures platform-specific targets, linking them against `tangram-core`. At this point, `tangram-core` has not yet been configured, but we can still use it as a target for linking.
  - After running the platform `config.cmake` script, the root CMake adds the `core` subdirectory, whose CMakeLists.txt configures the `tangram-core` target.
- **Rewrite the iOS build system to work within an Xcode workspace.** Turns out we don't need the iOS CMake toolchain file we've been using? Instead we can just set a few global variables to adjust the SDK used by Xcode and the configuration "just works". Here's how this works now:
  - Start an iOS build with `make cmake-ios`. This generates an Xcode project for the iOS framework target in the iOS build folder and also applies some settings to the workspace file at `platforms/ios/Tangram.xcworkspace`.
  - The workspace file `platforms/ios/Tangram.xcworkspace` contains a reference to the generated framework Xcode project and to the demo app Xcode project `platforms/ios/demo/TangramDemo.xcodeproj`.
  - Through the workspace, the demo app project contains a reference to the framework output from the generated Xcode project.
  - In the workspace you can build and debug either the framework or the demo app using the typical Xcode configurations.

TODO:
- [x] Consolidate git history
- [x] Update build instructions in README files
- [x] Add comments to explain non-obvious parts of Makefile and CMake files
- [ ] Improve source file grouping for core project in Xcode

Future changes:
 - Add iOS static library target
 - Probably remove Tizen targets?
 - Just write built-in shaders as raw strings in header files - generating data headers with a script is more complicated than it needs to be


core/CMakeLists.txt
```diff
@@ -156,10 +156,10 @@ if(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
...
 endif()
 
-set_target_properties(tangram-core
-  PROPERTIES
-  CXX_STANDARD 14
-  CXX_STANDARD_REQUIRED ON
+# C++14 is required for compiling tangram-core and for using the interface headers.
+# Making this option PUBLIC declares it as a requirement for targets linking against tangram-core.
+target_compile_options(tangram-core
+  PUBLIC -std=c++14
```
tallytalwar: Can you `cxx_std_14` here. https://cmake.org/cmake/help/v3.8/prop_gbl/CMAKE_CXX_KNOWN_FEATURES.html

matteblair: That feature option isn't available until CMake 3.8. Some platforms don't have 3.8 available by default (e.g. Ubuntu 16.04).

Makefile
```diff
@@ -219,34 +184,19 @@ osx: cmake-osx
 	cmake --build ${OSX_BUILD_DIR}
 
 cmake-osx:
-	@mkdir -p ${OSX_BUILD_DIR}
-	@cd ${OSX_BUILD_DIR} && \
-	cmake ../.. ${DARWIN_CMAKE_PARAMS}
-
-OSX_BUILD = \
-	xcodebuild -target ${OSX_TARGET} -project ${OSX_XCODE_BUILD_DIR}/${OSX_XCODE_PROJ} -configuration ${BUILD_TYPE}
+	cmake -H. -B${OSX_BUILD_DIR} ${OSX_CMAKE_PARAMS}
```
tallytalwar: Nice, did not know about these un documented parameters. (https://stackoverflow.com/a/31092052)

bench/CMakeLists.txt
```diff
@@ -1,35 +1,43 @@
...
+  set(EXECUTABLE_NAME "${bench_name}.out")
 
-    add_executable(${EXECUTABLE_NAME} ${_src_file_path})
+  add_executable(${EXECUTABLE_NAME} ${_src_file_path})
 
-    target_include_directories(${EXECUTABLE_NAME}
-        PUBLIC
-        ${CMAKE_CURRENT_SOURCE_DIR}/benchmark/include)
+  target_include_directories(${EXECUTABLE_NAME}
+    PUBLIC
```
tallytalwar: This should have `PRIVATE` scope.

platforms/ios/Tangram.xcworkspace/xcshareddata/IDEWorkspaceChecks.plist
```diff
@@ -2,7 +2,7 @@
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
-	<key>NextzenApiKey</key>
```
tallytalwar: Should we also include `NextzenApiKey` variable here in the plist file?

tallytalwar: Ahh, this is just github being weird. Its actually a plist for the workspace and not the demo!

platforms/ios/Tangram.xcworkspace/xcshareddata/Tangram.xcscmblueprint
```diff
@@ -0,0 +1,65 @@
...
+      "DVTSourceControlWorkspaceBlueprintRemoteRepositorySystemKey" : "com.apple.dt.Xcode.sourcecontrol.Git",
+      "DVTSourceControlWorkspaceBlueprintRemoteRepositoryIdentifierKey" : "BEF4CD3BF6A9DF306C3773D17F039B2DC1C72614"
+    },
+    {
+      "DVTSourceControlWorkspaceBlueprintRemoteRepositoryURLKey" : "https:\/\/github.com\/tangrams\/css-color-parser-cpp",
+      "DVTSourceControlWorkspaceBlueprintRemoteRepositorySystemKey" : "com.apple.dt.Xcode.sourcecontrol.Git",
+      "DVTSourceControlWorkspaceBlueprintRemoteRepositoryIdentifierKey" : "E1DE65396865A9214CEC7108611BC4DAC82E507C"
+    }
+  ]
+}
```
tallytalwar: new line!

And Can you document some explanation here? Maybe just a reply to this comment should be sufficient.

tallytalwar: Actually for all the `Tangram.xcworkspace` files.

platforms/android/config.cmake
```diff
@@ -0,0 +1,24 @@
...
+
+add_library(tangram SHARED
+  platforms/common/platform_gl.cpp
+  platforms/android/tangram/src/main/cpp/jniExports.cpp
+  platforms/android/tangram/src/main/cpp/androidPlatform.cpp
+  platforms/android/tangram/src/main/cpp/sqlite3ndk.cpp
+)
+
+target_include_directories(tangram
+  PUBLIC
```
tallytalwar: Should have `PRIVATE` scope.

Makefile
```diff
@@ -195,15 +169,6 @@ clean-tizen-arm:
...
-	rm -rf ${IOS_FRAMEWORK_BUILD_DIR}
-
-clean-ios-framework-sim:
-	rm -rf ${IOS_FRAMEWORK_SIM_BUILD_DIR}
-
-clean-ios-framework-universal:
-	rm -rf ${IOS_FRAMEWORK_UNIVERSAL_BUILD_DIR}
-
 android: android-demo
 	@echo "run: 'adb install -r android/demo/build/outputs/apk/demo-debug.apk'"
```
tallytalwar: Not part of this review, but we should update the path here to be: `platforms/android/demo/build/outputs/apk/debug/demo-debug.apk`

tests/CMakeLists.txt
```diff
@@ -1,15 +1,18 @@
 
 add_library(platform_test
-  ${CMAKE_CURRENT_SOURCE_DIR}/src/catch.cpp)
+  ${CMAKE_CURRENT_SOURCE_DIR}/src/catch.cpp
+)
 
 target_link_libraries(platform_test
   PUBLIC
```
tallytalwar: `PRIVATE` scope?

tests/CMakeLists.txt
```diff
@@ -1,15 +1,18 @@
...
+)
 
 target_include_directories(platform_test
-    PUBLIC
-    ${CMAKE_CURRENT_SOURCE_DIR}/catch)
+  PUBLIC
+  ${CMAKE_CURRENT_SOURCE_DIR}/catch
+)
 
 file(GLOB TEST_SOURCES ${CMAKE_CURRENT_SOURCE_DIR}/unit/*.cpp)
```
tallytalwar: Probably should also not use `GLOB` here for the same reasons you mentioned.

platforms/ios/Tangram.xcworkspace/xcshareddata/Tangram.xcscmblueprint
matteblair: These are created by xcode and should never really be opened in a text editor. I don't think we should modify these manually.

Makefile
```diff
@@ -256,99 +206,55 @@ endif
...
-cmake-ios-framework:
-ifndef TANGRAM_IOS_FRAMEWORK_SLIM
-	@mkdir -p ${IOS_FRAMEWORK_BUILD_DIR}
-	@cd ${IOS_FRAMEWORK_BUILD_DIR} && \
-	cmake ../.. ${IOS_FRAMEWORK_CMAKE_PARAMS} -DBUILD_IOS_FRAMEWORK=TRUE
-endif
+cmake-ios:
+	cmake -H. -B${IOS_BUILD_DIR} ${IOS_CMAKE_PARAMS}
+	@mkdir -p platforms/ios/Tangram.xcworkspace/xcuserdata/${USER}.xcuserdatad
+	@cp platforms/ios/WorkspaceSettings.xcsettings platforms/ios/Tangram.xcworkspace/xcuserdata/${USER}.xcuserdatad/WorkspaceSettings.xcsettings
```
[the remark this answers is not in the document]
matteblair: Yeah I need to explain this in comments here. It's a long story.

platforms/ios/Tangram.xcworkspace/xcshareddata/IDEWorkspaceChecks.plist
matteblair: Wow I don't know how git saw this as a move.. just consider this file as deleted, the new one is totally unrelated.

tests/CMakeLists.txt
matteblair: Yeah good point, I'll list out the test files here.

tests/CMakeLists.txt
tallytalwar: Nopes, `PUBLIC` is good here. Catch won't be available if its private scoped.

matteblair:
@tallytalwar Check out https://github.com/tangrams/tangram-es/pull/1827/commits/71582e71c44141881d94b4a8f7e3e2d340b25efa for more information than you ever wanted about the Xcode workspace setup haha

tallytalwar:
Haha, I am always hungry for information. Checking thanks.

tallytalwar:
With respect to my previous review question of `make android` building all architectures, I remember now that we added an option to specify `ABI Targets` using command line which gets applied to the `abiFilter` in tangram's android library's `build.gradle`. And if no abi target is specified, `build.gradle` is configured to build `all` (`armeabi-v7a`, `arm64-v8a`, `x86`).

platforms/ios/DEVELOPING.md
```diff
@@ -0,0 +1,26 @@
...
+
+Unfortunately, this means that developers will need to either:
+  1. Consistently ignore these changes when checking git status, or
+  2. Frequently restore the Xcode project from git and re-enter the App ID / Developer Team changes as needed.
+
+## Builds must be executed from the Xcode workspace ##
+
+The Xcode workspace that contains both the demo app project and the framework project uses specific build location settings to ensure that build products for both the framework and demo app are placed in the right locations and are able to build together successfully. If you try to open the projects separately and build them, the build output might not be in the expected location or it might not succeed at all.
```
tallytalwar: ~I am not sure if the language above is correct based on the apple documentation (link you mentioned below)~

~Based on the documentation, it seems one can open the framework project explicitly and build it and it should be fine.~


Hmm, for clarity maybe change `build output` to `build directory`, and remove "might not succeed at all". 
> You can use the workspace’s default build directory or you can specify one. Note that if a project specifies a build directory, that directory is overridden by the build directory of whatever workspace the project is in at the time you build the project.

platforms/ios/DEVELOPING.md
```diff
@@ -0,0 +1,26 @@
...
+
+Building a project through an Xcode workspace behaves a little differently from building the project on its own (see [Apple's documentation](https://developer.apple.com/library/archive/featuredarticles/XcodeConcepts/Concept-Workspace.html#//apple_ref/doc/uid/TP40009328-CH7-SW1) for a thorough explanation). By default, a workspace overrides the build output locations of the individual projects in it, placing the output for all projects into a separate, shared build folder instead. This causes build errors for projects generated by CMake. CMake hard-codes the absolute paths of build products like library dependencies into the Xcode project. Link flags in the project have those same paths hard-coded as well, so if build outputs are relocated by an Xcode workspace the link commands will fail to find the libraries at the paths in the project.
+
+To overcome this, we use a workspace setting that keeps build products in the locations specified by their original projects. This can be set in Xcode by opening a workspace, then navigating to _File_ > _Workspace Settings_ in the menu bar, then clicking _Advanced..._, then choosing the _Legacy_ option.
```
tallytalwar: 👍

Makefile
matteblair: Guess I might as well fix that path while I'm messing around in the Makefile - will update.

platforms/ios/DEVELOPING.md
matteblair: True, I think the framework project would be fine by itself, but the demo app project could either fail to build or link to a "stale" framework. I'll update the text to reflect that.

PR #1833, opened 2018-07-07
Couple of tiny updates to Makefile (for android) and iOS readme.

tallytalwar:


platforms/ios/README.md
```diff
@@ -36,7 +36,12 @@ make ios NEXTZEN_API_KEY=yourApiKeyHere
 
 You can optionally append a `BUILD_TYPE` variable to choose the build type, for example `BUILD_TYPE=Debug`.
 
-This will generate and compile an Xcode project for the Tangram iOS framework and demo application. To run/debug the demo application, open the projects in the iOS Xcode workspace:
+This will generate and compile an Xcode project for the Tangram iOS framework and demo application. 
+
+Note: Above will show code signing errors as no developer team is set for the `TangramDemo` application project, however
```
matteblair: Maybe the readme should just say to run `make cmake-ios` and the open the xcworkspace instead, since the first run will always have code signing errors. We could even have a make target that generates and then opens the workspace.

tallytalwar:
@matteblair Does f363f7a answer what you mentioned above?

PR #1835, opened 2018-07-13
Refactor TGMapViewController into TGMapView

matteblair:
If everything worked, you won't notice any differences ;)

Providing a map as a `UIView` is more flexible than a `UIViewController` because view hierarchies often don't have or need a view controller for every view. 

Most of `TGMapViewController` was copied directly into `TGMapView`. Wherever possible I kept interfaces the same, though some structural changes are needed for client apps to adopt this change.

One major internal change from `TGMapViewController` is how the render loop is managed. Previously `TGMapViewController` subclassed `GLKViewController`, which provides a render loop that can be paused and resumed. `TGMapView` must implement this independently now. Fortunately, it is reasonably simple to us a `CADisplayLink` to synchronize render callbacks with the display refresh rate.


platforms/ios/framework/src/TGMapView.h
```diff
@@ -0,0 +1,913 @@
...
+
+ @note Changing this property will override the inferred value from the scene.
+ Enabling continuous rendering can significantly increase the energy usage of
+ an application.
+ */
+@property (assign, nonatomic) BOOL continuous;
+
+#pragma mark Gesture Responders
+
+// TODO: Is it necessary to expose these selectors as properties? - Matt
```
msmollin: No. I'd remove them :)

platforms/ios/framework/src/TGMapView.h
```diff
@@ -0,0 +1,913 @@
+//
+//  TGMapView.h
+//  TangramMap
+//
+//  Created by Matt Blair on 7/9/18.
```
msmollin: General note: should go through here and annotate the interface with nullability annotations. Would "solve" some of the "this must not be nil" things.

msmollin: NVM saw the macro. These long files are hard to follow

platforms/ios/framework/src/TGMapView.h
```diff
@@ -0,0 +1,913 @@
...
+
+/**
+ Adds a named data layer to the map.
+
+ See `TGMapData` for more details.
+
+ @param name The name of the data layer.
+ @return The map data, or nil if the data source can't be initialized.
+
+ @note You cannot create more than one data source with the same name, the same
```
msmollin: Nit: More descriptive note? Maybe like "The previous data source by the same name will be returned"

platforms/ios/framework/src/TGMarker+Internal.h
```diff
@@ -3,10 +3,11 @@
 //  TangramMap
 //
 //  Created by Karim Naaji on 4/10/17.
+//  Updated by Matt Blair on 7/13/18.
 //  Copyright (c) 2017 Mapzen. All rights reserved.
 //
 
-#import "map.h"
+#include "map.h"
```
msmollin: Why the change here? Import does additional checks to make sure things are only included once.

matteblair:
@msmollin What do you think about how `TGMapView` does nothing but wrap our `GLKView`? I'm thinking it will be convenient for adding other views like attribution or a compass, but is there significant waste to having this outer UIView that doesn't really do anything?

platforms/ios/framework/src/TGMapView.mm
```diff
@@ -1,44 +1,318 @@
...
+    self.map->render();
+
+    if (_viewComplete && [self.mapViewDelegate respondsToSelector:@selector(mapViewDidCompleteLoading:)]) {
+        [self.mapViewDelegate mapViewDidCompleteLoading:self];
+    }
+
+    if (self.mapViewDelegate && [self.mapViewDelegate respondsToSelector:@selector(mapView:didCaptureScreenshot:)]) {
+        if (self->_shouldCaptureFrame && (!self->_captureFrameWaitForViewComplete || self->_viewComplete)) {
+            UIGraphicsBeginImageContext(self.frame.size);
+            [self drawViewHierarchyInRect:self.frame afterScreenUpdates:YES];
```
msmollin: Not required for this, but an FYI there's a `snapshot` property on GLKView that might be more performant. The downside is you will lose any UIKit modifications people may have added to the map view.

platforms/ios/framework/src/TGMapView.mm
```diff
@@ -1,44 +1,318 @@
...
+@interface TGMapView () <UIGestureRecognizerDelegate, GLKViewDelegate> {
+    BOOL _shouldCaptureFrame;
+    BOOL _captureFrameWaitForViewComplete;
+    BOOL _viewComplete;
+    BOOL _viewInBackground;
 }
 
 @property (nullable, strong, nonatomic) EAGLContext* context;
-@property (assign, nonatomic) CGFloat contentScaleFactor;
+@property (nonatomic) GLKView* glView;
```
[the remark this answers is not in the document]
matteblair: Should subviews always be strongly referenced? Is there any risk of creating a cycle?

We do technically have a policy to left-align pointers: https://github.com/tangrams/tangram-es/blob/master/.clang-format#L42 But Xcode has some very strongly held opinions about this so we're not totally consistent in the iOS code :\

platforms/ios/framework/src/TGMarker.h
```diff
@@ -3,18 +3,17 @@
...
-#import <UIKit/UIKit.h>
 #import "TGExport.h"
 #import "TGGeoPoint.h"
 #import "TGGeoPolygon.h"
 #import "TGGeoPolyline.h"
 #import "TGTypes.h"
-
-@class TGMapViewController;
+#import <Foundation/Foundation.h>
+#import <UIKit/UIKit.h>
```
msmollin: Nit: UIKit imports foundation, so you don't need both imports.

platforms/ios/framework/src/TGMarker+Internal.h
msmollin: Yeah it's used in other codebases (the mixing) but I don't like it (but again, objc dev being opinionated). Also, the top comment on this answer notes performance improvements by using import: https://stackoverflow.com/a/439691

I don't know if that's still relevant in modern llvm and clang but yeah.

platforms/ios/framework/src/TGMapView.mm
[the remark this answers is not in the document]
msmollin: It won't create a retain cycle unless the glView has a strong reference back to the MapView. In fact, the way you have it coded here has it defaulted to strong anyway cuz the compiler defaults any pointer properties to strong refs.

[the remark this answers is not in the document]
msmollin: And yeah, Cocoa Style guidelines is to right align pointers, which is why Xcode is like how it is :)

msmollin:
@matteblair There are other minor benefits to wrapping, like UIViews are easier to work with in Interface Builder generally speaking. But i hold no strong (har har) opinion on it. Part of the reason for wrapping it originally was to insulate external developers from having to import the entirety of the tangram c++ codebase. With much of the refactoring you've been doing, I am unsure that's a concern anymore.

platforms/ios/framework/src/TGMapView.mm
matteblair: Yeah I had the same thought! It's conceivable that you might want a screenshot of just the map canvas without decorations. We could just make that an option, like `withDecorations:(BOOL)`.

platforms/ios/framework/src/TGMarker+Internal.h
matteblair: Interesting thread, it sounds like some C system headers might rely on `#include` behavior so I'm inclined to do that just in case. Tangram's headers all use `#pragma once` so it should still be easy for the preprocessor.

tallytalwar:
@matteblair can you also update the README to reflect these changes.

platforms/ios/framework/src/TGMapView.mm
```diff
@@ -1,44 +1,320 @@
...
+    CFTimeInterval dt = _displayLink.targetTimestamp - _displayLink.timestamp;
+    _viewComplete = self.map->update(dt);
+
+    self.map->render();
+
+    if (_viewComplete && [self.mapViewDelegate respondsToSelector:@selector(mapViewDidCompleteLoading:)]) {
+        [self.mapViewDelegate mapViewDidCompleteLoading:self];
+    }
+
+    if (self.mapViewDelegate && [self.mapViewDelegate respondsToSelector:@selector(mapView:didCaptureScreenshot:)]) {
```
matteblair: @msmollin In the conditional here, is it necessary to check the delegate for nullity first? I've read that you can just send `respondsToSelector` to a `nil` receiver and you'll just get `NO`, but I don't know whether this is a common/good practice.

matteblair:
@tallytalwar Glad you reminded me to check over the documentation - found a few things amiss. But the README itself doesn't actually mention specific classes. Might be good in a future update to add a description of the classes or at least indicate where to start reading.

platforms/ios/framework/src/TGMapView.mm
msmollin: Yeah that's fine. The nil check first is unnecessary.

platforms/ios/framework/src/TGMapView.mm
```diff
@@ -1,44 +1,336 @@
...
+    [_glView addGestureRecognizer:_longPressGestureRecognizer];
+}
+
+- (void)didMoveToWindow
+{
+    [self setupDisplayLink];
+    [super didMoveToWindow];
+}
+
+- (void)didMoveToSuperview
```
tallytalwar: Should we also be handling `didAddSubview`?

matteblair: I can't think of anything we would need to do in `TGMapView` when a subview is added.

PR #1836, opened 2018-07-16
Fixes an issue where 1D size values were erroring

tallytalwar:
- Was introduced by commit: 50f56938fc1ea84896ba8fb53e3210566fa86f62


matteblair:
I think I understand the erroring code path here, but I don't understand how this didn't break tests that use % sizes like this: https://github.com/tangrams/tangram-es/blob/master/tests/unit/stopsTests.cpp#L256-L280

tallytalwar:
`stops` do not use that code path. We do not have any coverage for that any of style parameter parsing functionality.

tallytalwar:
Yeah so the code path this is catering to is actual parsing of style parameter from the scene file. `Stop` evaluation happens much later.

tallytalwar:
A scene validator script should make sure a valid scene file is being passed to tangram. It will be hard for tangram scene parser to test for "ALL" the possible values a user can put in. I think this was the reason we never added any unit tests for scene file parsing/style parameter parsing functionality.

matteblair:
Ah you're right, it is a different code path for Stops.

Unit tests for parsing parts of the scene file isn't super difficult - it would probably be a good idea to start making those. We've had too many regressions on really hard-to-spot stuff like this.

PR #1839, opened 2018-07-18
Refactor iOS URL request interface

matteblair:
- `TGHttpHandler` was being used for `file://` URLs as well as `http://` URLs, so the name `TGURLHandler` is more appropriate (and capitalization is now consistent with Cocoa names).
- Allowing the `urlHandler` for a `TGMapView` to be changed after initialization could result in request handles from one URL handler being used with a different URL handler (start URL request -> change URL handler -> cancel original URL request -> new URL handler is asked to cancel a request that it didn't create!). The simplest solution is to set the URL handler at initialization and never change it.
- To make a custom URL handler you had to subclass `TGURLHandler`, which means that your custom  URL handler will always initialize and retain a `NSURLSession` instance, even if you don't use it. Instead, we'll now make `TGURLHandler` a protocol for classes to conform to and provide a default implementation called `TGDefaultURLHandler`.

### DIFF
--- a/platforms/ios/framework/src/TGURLHandler.h
+++ b/platforms/ios/framework/src/TGURLHandler.h
@@ -1,8 +1,9 @@
 //
-//  TGHttpHandler.h
+//  TGURLHandler.h
 //  TangramMap
 //
 //  Created by Karim Naaji on 11/23/16.
+//  Updated by Matt Blair on 7/16/18.
 //  Copyright (c) 2017 Mapzen. All rights reserved.
 //
 
@@ -10,7 +11,7 @@
 #import "TGExport.h"
 
 /**
- A network request completion callback, called when a download request of `TGHttpHandler`
+ A URL request completion callback, called when a download request of `TGURLHandler`
  completed an asynchronous request.
 */
 typedef void(^TGDownloadCompletionHandler)(NSData * _Nullable data, NSURLResponse * _Nullable response, NSError * _Nullable error);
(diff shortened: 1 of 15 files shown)


platforms/ios/framework/src/TGURLHandler.h
```diff
@@ -0,0 +1,77 @@
+//
+//  TGURLHandler.h
+//  TangramMap
+//
+//  Created by Karim Naaji on 11/23/16.
+//  Updated by Matt Blair on 7/16/18.
+//  Copyright (c) 2017 Mapzen. All rights reserved.
+//
+
```
tallytalwar: Missing `#pragma once`

matteblair: I think we _could_ add `#pragma once` and nothing bad would happen, but so far we've consistently not used it in Objective-C headers because we always `#import` them - which makes sure they're only included once.

platforms/ios/framework/src/TGURLHandler.h
```diff
@@ -0,0 +1,77 @@
...
+ A URL request completion callback, called when a download request of `TGURLHandler` completes an asynchronous request.
+*/
+typedef void(^TGDownloadCompletionHandler)(NSData * _Nullable data, NSURLResponse * _Nullable response, NSError * _Nullable error);
+
+NS_ASSUME_NONNULL_BEGIN
+
+/**
+ A URL Handler interface for creating cancellable asynchronous URL requests.
+ */
+@protocol TGURLHandler <NSObject>
```
tallytalwar: We should `TG_EXPORT` this thing.

matteblair: I had to look this up when I was setting up symbol exporting - turns out protocols are just interface declarations, there aren't actually any symbols associated with it that we need to export.

tallytalwar: Same I also looked this up after I commented. Makes sense.

platforms/ios/framework/src/iosPlatform.h
```diff
@@ -2,7 +2,7 @@
 
 #include "platform.h"
 
-#import "TGMapView.h"
+#import "TGMapView+Internal.h"
```
tallytalwar: We can do a forward declaration for `@class TGMapView`.

tallytalwar: Addressed d081d7, also a few other forward declaration cleanups.

platforms/ios/framework/src/iosPlatform.mm
```diff
@@ -147,10 +148,10 @@ void initGLExtensions() {
         return 0;
     }
 
-    TGHttpHandler* httpHandler = [mapView httpHandler];
+    id<TGURLHandler> urlHandler = mapView.urlHandler;
```
tallytalwar: After working with `protocol` and `id` in another context, I find it so much cleaner for code architecture.

PR #1845, opened 2018-07-25
Replace incbin.sh with shaders.cmake

hjanetzek:


hjanetzek:
Ok drop the second commit. cmake lists are semicolon separated strings..
https://github.com/ampl/gsl/blob/master/CMakeLists.txt#L229

core/CMakeLists.txt
```diff
@@ -180,10 +180,7 @@ foreach(_shader ${SHADER_FILES})
       OUTPUT "${SHADER_OUTPUT}"
       WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}/..
       MAIN_DEPENDENCY "${_shader}"
-      COMMAND ${PROJECT_SOURCE_DIR}/../scripts/incbin.sh
-        "${_shader}"
-        "core/generated/${_shader_name}.h"
-        "${_shader_name}")
+      COMMAND ${CMAKE_COMMAND} -Din="${_shader}" -Dout="core/generated/${_shader_name}.h" -Dname="${_shader_name}" -P ${CMAKE_CURRENT_SOURCE_DIR}/shaders.cmake)
```
tallytalwar: Can we use `SHADER_OUTPUT` instead of explicit `core/generated` here?

matteblair:
Probably also better to explicitly list the files we want to generate. That wouldn't require globbing and looping over a directory or checking file extensions, and it would enable cmake to know when a new file needs to be generated.

PR #1847, opened 2018-07-27
Add map region change events to iOS map view delegate

matteblair:
Enables delegates to respond to changes in the displayed map area.

Callbacks are missing during camera animations for now. I plan to add those when the new Camera API is added to the iOS interface.


platforms/ios/framework/src/TGMapView.mm
```diff
@@ -231,6 +233,7 @@ - (void)setupGestureRecognizers
     _rotationGestureRecognizer = [[UIRotationGestureRecognizer alloc] initWithTarget:self action:@selector(respondToRotationGesture:)];
     _shoveGestureRecognizer = [[UIPanGestureRecognizer alloc] initWithTarget:self action:@selector(respondToShoveGesture:)];
     _shoveGestureRecognizer.minimumNumberOfTouches = 2;
+    _shoveGestureRecognizer.maximumNumberOfTouches = 2;
```
tallytalwar: fyi google maps allows 2+ touches for shoving. Yes client app can override this but maybe default should allow 2+ touches for shoving. Thoughts?

platforms/ios/framework/src/TGMapViewDelegate.h
```diff
@@ -0,0 +1,116 @@
...
+ `-[TGMapView updateSceneAsync:]`
+
+ @param mapView The map view instance.
+ @param sceneID The ID of the scene that was loaded or updated.
+ @param sceneError Any error encountered while loading or updating the scene.
+ */
+- (void)mapView:(nonnull TGMapView *)mapView didLoadScene:(int)sceneID withError:(nullable NSError *)sceneError;
+
+/**
+ Called immediately before the displayed map region changes.
```
tallytalwar: Can we clarify what does change mean over here. Because adding a marker or a client data source is also a change. But I think this involves all view changes like camera updates?

platforms/ios/framework/src/TGMapViewDelegate.h
```diff
@@ -0,0 +1,116 @@
...
+
+ @param mapView The map view instance.
+ @param animated If YES the change will be animated over time, if NO the change will happen immediately.
+ */
+- (void)mapView:(nonnull TGMapView *)mapView regionWillChangeAnimated:(BOOL)animated;
+
+/**
+ Called repeatedly when the displayed map region changes due to an animation or a gesture.
```
tallytalwar: nitpick; "when the displayed map region is changing". Makes the reason for "repeatedly" more clear.

platforms/ios/framework/src/TGMapView.mm
```diff
@@ -996,6 +1018,9 @@ - (void)respondToPanGesture:(UIPanGestureRecognizer *)panRecognizer
     if ([self.gestureDelegate respondsToSelector:@selector(mapView:recognizer:didRecognizePanGesture:)]) {
         [self.gestureDelegate mapView:self recognizer:panRecognizer didRecognizePanGesture:displacement];
     }
+
+    // Reset translation to zero so that subsequent calls get relative value.
+    [panRecognizer setTranslation:CGPointZero inView:_glView];
```
tallytalwar: 👍

platforms/ios/framework/src/TGMapViewDelegate.h
matteblair: Yeah this does not apply to when the _contents_ of the map region changes, but rather when the _region itself_ changes. Maybe it would be more clear to say "before the region displayed in the map changes"?

platforms/ios/framework/src/TGMapView.mm
matteblair: Apple Maps only allows 2, so I don't think this will be unexpected to iOS users. We already enforced this anyway, just in the gesture responder rather than in the recognizer itself.

platforms/ios/framework/src/TGMapViewDelegate.h
matteblair: Oh actually it's much clearer if I just replace "changes" with "moves" - I'll try that out.

platforms/ios/framework/src/TGMapView.mm
tallytalwar: Cool, I did notice the check of "touches != 2" there. We have not touched any of gesture handling code in ages :D.

platforms/ios/framework/src/TGMapViewDelegate.h
tallytalwar: yeah "moves" sounds more apt.

PR #1848, opened 2018-07-27
Refactor HttpHandler interface for android

tallytalwar:
- HttpHandler is an interface which can be implemented by the client to
override the default implementation (DefaultHttpHandler)
- HttpHandler.Callback interface provides methods for network response handling for native code.
- Client will wrap HttpHandler.Callback in its implementation dependent callback
        - Example DefaultHttpHandler wraps Tangram.HttpHandler.Callback in okhttp3.Callback
- DefaultHttpHandler provides a default HttpHandler implementation which
clients can use if none is provided by the client app.
        - Overrides HttpHandler methods
        - Wraps HttpHandler.Callback in OkHttp3.Callback


platforms/android/tangram/src/main/java/com/mapzen/tangram/DefaultHttpHandler.java
```diff
@@ -0,0 +1,235 @@
...
+                        cb.failure(e);
+                        throw e;
+                    }
+                    else {
+                        cb.success(body.bytes());
+                    }
+                }
+            };
+            final Request.Builder builder = new Request.Builder().url(httpUrl).tag(requestHandle);
+            final CacheControl cacheControl = cachePolicy.apply(httpUrl);
```
matteblair: Our `CachePolicy` interface also uses OkHTTP types. If we want to remove OkHTTP from our interfaces then it will also need to be refactored.

tallytalwar:
Yeah I agree with most of your naming comments. Will address couple of more things here.

platforms/android/tangram/src/main/java/com/mapzen/tangram/HttpResponse.java
```diff
@@ -0,0 +1,20 @@
...
+    /**
+     * Network request failed response
+     * @param e Exception representing the failed network request
+     */
+    public void failure(IOException e);
+    /**
+     * Network request success response
+     * @param rawDataBytes raw bytes from the body of a network response
+     */
+    public void success(byte[] rawDataBytes);
```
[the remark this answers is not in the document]
matteblair: Having done a bit more research now, it's unclear to me whether it's worth trying to optimize array copies any more at this point. I think we _could_ reduce some overhead when the handler is able to stream the response body - we could pre-allocate memory for the data in native code and turn it into a direct byte buffer that the stream could then write straight into. However this seems like more complexity than its worth at this point.

platforms/android/tangram/src/main/java/com/mapzen/tangram/HttpResponse.java
```diff
@@ -0,0 +1,20 @@
+package com.mapzen.tangram;
+
+import java.io.IOException;
+
+/**
+ * {@code HttpHandler} interface implemented internally in {@link MapController#startUrlRequest(String, long)}
+ * Passes network responses from underlying implementation to be processed internally
+ */
+public interface HttpResponse {
```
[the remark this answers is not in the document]
tallytalwar: Yeah `HttpCallback` is better. But...

Since its internally implemented to "bridge" calls to the native network callbacks, but needs to be public to be accessed and called from client httpHandler implementation, how about `HttpCallbackBridge` or something on those lines?

platforms/android/tangram/src/main/java/com/mapzen/tangram/HttpResponse.java
```diff
@@ -0,0 +1,20 @@
...
+    /**
+     * Network request failed response
+     * @param e Exception representing the failed network request
+     */
+    public void failure(IOException e);
```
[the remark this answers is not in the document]
tallytalwar: I am fine with `onFailure` and `onResponse`.

But for `onCancel` maybe not so much. My idea here is Client's network callback will wrap these methods and send apt error message on a "failure". Which will then be set for us on the native side in `UrlResponse`.

We can scope out much bigger refactor where tangram handles all sorts of failure/cancel/error cases with respect to network responses. But accordingly to me these should be handled by the implementation (Okhttp, etc) and just pass in an error message.

That being said I will change these to be called `onFailure` and `onSuccess`. Where `onFailure` is any kind of failure; elaborated in the message by the client implementation and `onSuccess` is always the success case, where we did get some sensible data back with no errors.

platforms/android/tangram/src/main/java/com/mapzen/tangram/HttpResponse.java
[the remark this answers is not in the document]
tallytalwar: Agreed, though I should read more :p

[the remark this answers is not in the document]
tallytalwar: Agreed, we should benchmark to see how bad these byte array copies are, and totally try to optimize if makes sense.

platforms/android/tangram/src/main/java/com/mapzen/tangram/DefaultHttpHandler.java
```diff
@@ -0,0 +1,237 @@
...
+ */
+public class DefaultHttpHandler implements HttpHandler {
+
+    protected OkHttpClient okClient;
+    protected CachePolicy cachePolicy;
+    protected CacheControl tileCacheControl;
+
+    /**
+     * Enables TLS v1.2 when creating SSLSockets.
+     * <p/>
```
matteblair: Stray HTML tag?

platforms/android/tangram/src/main/java/com/mapzen/tangram/HttpCallbackBridge.java
```diff
@@ -0,0 +1,20 @@
+package com.mapzen.tangram;
+
+import java.io.IOException;
+
+/**
+ * {@code HttpCallbackBridge} interface implemented internally in {@link MapController#startUrlRequest(String, long)}
+ * Passes network responses from underlying implementation to be processed internally
+ */
+public interface HttpCallbackBridge {
```
matteblair: Do you think "Bridge" is helpful in this name? I saw your comment about this but I'm not sure that this implementation detail is important to the interface.

We could also consider nesting this interface in the definition of `HttpHandler` since they are so closely related. You'd end up calling it something like `HttpHandler.Callback` which seems reasonable.

platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
```diff
@@ -1363,10 +1361,12 @@ public void onSurfaceCreated(final GL10 gl, final EGLConfig config) {
     // ==================
     @Keep
     void cancelUrlRequest(final long requestHandle) {
-        if (httpHandler == null) {
-            return;
+        if (httpHandler != null) {
+            Object request = httpRequestHandles.get(requestHandle);
+            if (request != null) {
```
matteblair: Hmm personally I'd leave the request objects as-is and let the implementation deal with any nulls that it created for itself. But I'm not strongly opposed to this either.

platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
```diff
@@ -1375,31 +1375,33 @@ void startUrlRequest(@NonNull final String url, final long requestHandle) {
...
-                    throw new IOException("Unexpected null body for URL: " + url);
+            public void onResponse(final int code, final byte[] rawDataBytes, final Map<String, String> headers) {
+                // TODO: Use of returned error code and headers for better network response logging/retries, etc
+                if (code >= 200 && code < 300) {
+                    nativeOnUrlComplete(mapPointer, requestHandle, rawDataBytes, null);
                 }
                 else {
-                    final byte[] bytes = body.bytes();
-                    nativeOnUrlComplete(mapPointer, requestHandle, bytes, null);
+                    nativeOnUrlComplete(mapPointer, requestHandle, null, "Unexpected response code: " + code + " for URL: " + url);
```
matteblair: Why not pass in the body data here too?

platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
```diff
@@ -1375,31 +1375,33 @@ void startUrlRequest(@NonNull final String url, final long requestHandle) {
...
                 }
             }
+
+            @Override
+            public void onCancel() {
+                Log.d("Tangram", "");
```
matteblair: No need to log anything here. But it is crucial that we call a native url handler, even for canceled requests - otherwise we will never release the native callback objects for these requests.

platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
```diff
@@ -1375,31 +1375,33 @@ void startUrlRequest(@NonNull final String url, final long requestHandle) {
...
+            }
         };
 
-        httpHandler.onRequest(url, callback, requestHandle);
+        Object request = httpHandler.startRequest(url, callback);
+        if (request != null) {
+            httpRequestHandles.put(requestHandle, request);
```
matteblair: I don't see handles ever being removed from this mapping. Seems like we should be removing the handle for a request in every code path of the callback.

platforms/android/tangram/src/main/java/com/mapzen/tangram/DefaultHttpHandler.java
```diff
@@ -0,0 +1,230 @@
...
+                        cb.onCancel();
+                        return;
+                    }
+                    cb.onFailure(e);
+                }
+
+                @Override
+                public void onResponse(final Call call, final Response response) throws IOException {
+                    final ResponseBody body = response.body();
+                    if (body == null) {
```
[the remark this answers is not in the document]
tallytalwar: Do not remember why, probably a leftover from okhttp2, maybe 🤷‍♂️ . But will update here.

platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
```diff
@@ -1291,6 +1288,7 @@ Marker markerById(final long markerId) {
     private FontFileParser fontFileParser;
     private DisplayMetrics displayMetrics = new DisplayMetrics();
     private HttpHandler httpHandler;
+    private Map<Long, Object> httpRequestHandles;
```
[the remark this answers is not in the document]
tallytalwar: Yup, I also noticed headers could have a list of values. Refer: http://square.github.io/okhttp/3.x/okhttp/okhttp3/Headers.html#toMultimap--

platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
tallytalwar: I would like to have this safety net here.

platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
tallytalwar: In the core code, if the `UrlResponse` has an error string, we never check the validity of the data bytes. Not sure we should pass in any data if there is an error associated the response.

platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
tallytalwar: I was testing explicit cancels and hence had that log, will remove.

platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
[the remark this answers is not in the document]
matteblair: My comment wasn't about headers - are you referring to a different line?

platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
matteblair: Fair enough - saves us a bit of work by not copying anything into a native vector.

platforms/android/tangram/src/main/java/com/mapzen/tangram/DefaultHttpHandler.java
```diff
@@ -0,0 +1,230 @@
...
+                    if (call.isCanceled()) {
+                        cb.onCancel();
+                        return;
+                    }
+                    cb.onFailure(e);
+                }
+
+                @Override
+                public void onResponse(final Call call, final Response response) throws IOException {
+                    final ResponseBody body = response.body();
```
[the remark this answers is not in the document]
tallytalwar: The way its mentioned in the docs using `Try-with-resource`, requires API level 19. Will have to use and `close` the response explicitly.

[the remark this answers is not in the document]
matteblair: Oh I see - yeah we can just write the try, catch, finally blocks explicitly.

matteblair:
Ok I think the last thing to work out here is what do about request/response headers (including cache-control). The only use-cases I can think of are:
- Client app wants to customize the cache control for certain URLs
- Client app wants to insert custom headers into requests for analytics or authorization
Are there other common things we'd want to be able to do with headers?

platforms/android/tangram/src/main/java/com/mapzen/tangram/DefaultHttpHandler.java
```diff
@@ -0,0 +1,231 @@
...
+                    }
+                    cb.onFailure(e);
+                }
+
+                @Override
+                public void onResponse(final Call call, final Response response) throws IOException {
+                    try {
+                        final ResponseBody body = response.body();
+                        // TODO headers? cacheControl?
+                        cb.onResponse(response.code(), body.bytes(), response.headers().toMultimap());
```
matteblair: The control flow here should ensure that onResponse or onFailure is called. Even if an exception is thrown while reading the response we want to invoke the HttpHandler.Callback so that our native request is properly finalized.

platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
```diff
@@ -1363,10 +1359,13 @@ public void onSurfaceCreated(final GL10 gl, final EGLConfig config) {
...
     @Keep
     void cancelUrlRequest(final long requestHandle) {
-        if (httpHandler == null) {
-            return;
+        if (httpHandler != null) {
+            Object request = httpRequestHandles.get(requestHandle);
+            if (request != null) {
+                httpHandler.cancelRequest(request);
+            }
+            httpRequestHandles.remove(requestHandle);
```
matteblair: Requests and responses will most likely be occurring on different threads and I don't think `LongSparseArray` is thread-safe, so we should synchronize access to it.

platforms/android/tangram/src/main/java/com/mapzen/tangram/DefaultHttpHandler.java
matteblair: Actually this gets very interesting!
> Code that is writing the request body or reading the response body will suffer an IOException when its call is canceled

(https://github.com/square/okhttp/wiki/Calls#calls)

So maybe the `catch` path should call `onFailure`? Or should check the cancelled status of the call and call `onCancel`?

tallytalwar: Yeah probably makes sense to check for cancellation in the catch block and aptly call `onCancel` or `onFailure` otherwise.

platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
```diff
@@ -1361,45 +1359,68 @@ public void onSurfaceCreated(final GL10 gl, final EGLConfig config) {
...
+                nativeOnUrlComplete(mapPointer, requestHandle, null, null);
+                syncRemoveHttpHandle(requestHandle);
             }
         };
 
-        httpHandler.onRequest(url, callback, requestHandle);
+        Object request = httpHandler.startRequest(url, callback);
+        if (request != null) {
+            synchronized (this) {
+                httpRequestHandles.put(requestHandle, request);
```
matteblair: I think there's a subtle edge case to be careful about here. If `startRequest` completes immediately (like in `DefaultHttpHandler` when the URL can't be parsed) then the callback is executed before the request is added to the handle map. I think this the worst case is that we get some request entries in the map that never get removed, which ends not being a problem with `DefaultHttpHandler`. It might not be worth guarding against this now, but we should keep aware of it.

platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
```diff
@@ -216,9 +213,11 @@ private Bitmap capture() {
     protected MapController(@NonNull final GLSurfaceView view) {
         if (Build.VERSION.SDK_INT > 18) {
             clientTileSources = new ArrayMap<>();
-        } else {
+        }
+        else {
```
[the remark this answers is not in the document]
tallytalwar: Hmm, we have a mix of these in the android code base.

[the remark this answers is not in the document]
tallytalwar: Will update here though.

platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
tallytalwar: Good Point.
For `DefaultHttpHandler` purposes its not a problem, as the returned `request` object is `null` and won't be added to the map. But there could be cases where callbacks gets executed before returned `request` object is put in the map, which I think is kinda not trivial to solve. Will leave it for subsequent PR.

PR #1851, opened 2018-08-08
Camera API tweaks

matteblair:
- Change "ease" member in Map from a vector to a single unique_ptr, since we only have one at a time.
- Allow specifying separate padding values for each side of the view when fitting a camera to a region.
- Fix some JNI warnings in Android Studio

### DIFF
--- a/core/include/tangram/map.h
+++ b/core/include/tangram/map.h
@@ -122,7 +122,7 @@ struct CameraUpdate {
     float tilt = 0;
     float tiltBy = 0;
     std::array<LngLat,2> bounds;
-    int boundsPadding = 0;
+    std::array<int, 4> padding = {{0}};
 };
 
 class Map {
@@ -230,8 +230,9 @@ public:
     // Get the tilt angle of the view in radians; 0 corresponds to straight down
     float getTilt();
 
-    // Get the CameraPosition that encloses the bounds given by _a and _b
-    CameraPosition getEnclosingCameraPosition(LngLat a, LngLat b, int buffer);
+    // Get the CameraPosition that encloses the bounds given by _a and _b and
+    // leaves at least the given amount of padding on each side (in logical pixels).
+    CameraPosition getEnclosingCameraPosition(LngLat _a, LngLat _b, int _leftPad, int _topPad, int _rightPad, int _bottomPad);
 
     // Run flight animation to change postion and zoom  of the map
     // If _duration is 0, speed is used as factor to change the duration that is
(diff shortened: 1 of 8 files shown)


core/src/map.cpp
```diff
@@ -760,27 +760,36 @@ float Map::getTilt() {
...
+    const MapProjection& projection = view.getMapProjection();
+
+    // Convert the bounding coordinates into Mercator meters.
     glm::dvec2 aMeters = projection.LonLatToMeters(glm::dvec2(_a.longitude, _a.latitude));
     glm::dvec2 bMeters = projection.LonLatToMeters(glm::dvec2(_b.longitude, _b.latitude));
+    glm::dvec2 dMeters = glm::abs(aMeters - bMeters);
 
-    double tileSize = projection.TileSize() * impl->view.pixelScale();
-    double buffer = _buffer * 2 * impl->view.pixelScale();
+    // Calculate the inner size of the view that the bounds must fit within.
```
tallytalwar: 👍 for comments.

core/include/tangram/map.h
```diff
@@ -230,8 +230,9 @@ class Map {
     // Get the tilt angle of the view in radians; 0 corresponds to straight down
     float getTilt();
 
-    // Get the CameraPosition that encloses the bounds given by _a and _b
-    CameraPosition getEnclosingCameraPosition(LngLat a, LngLat b, int buffer);
+    // Get the CameraPosition that encloses the bounds given by _a and _b and
+    // leaves at least the given amount of padding on each side (in logical pixels).
+    CameraPosition getEnclosingCameraPosition(LngLat _a, LngLat _b, int _leftPad, int _topPad, int _rightPad, int _bottomPad);
```
tallytalwar: Will it be better (cleaner) to pass in std::array here instead of explicit 4 padding dimensions?

matteblair: I'm pretty sure I'd mess up the order of these arguments if they weren't named. The better solution for consistency is probably to define a padding struct with named members.

tallytalwar: Yeah, maybe its a small fish in big ocean, but we do lose some cache coherence by breaking the well formed memory struct for these paddings.

core/include/tangram/map.h
```diff
@@ -100,6 +100,18 @@ struct CameraPosition {
     float tilt = 0;
 };
 
+struct EdgePadding {
+    int left = 0;
```
tallytalwar: Can we have negative padding? Or rather should we allow negative padding?

matteblair: Yeah it's possible. It's not very useful but some application might need it - might as well allow it.

PR #1853, opened 2018-08-13
iOS camera position interface

matteblair:
- Adds a `TGCameraPosition` class to describe a map camera position and orientation
- Adds methods to `TGMapView` that control the map view using a `TGCameraPosition`
- Adds a `TGCoordinateBounds` struct to describe a map area with a bounding box
- Adds `-(TGCameraPosition*)[TGMapView cameraThatFitsBounds:withPadding:]` to calculate a camera position for viewing a map area
- Adds `flyTo` animation methods to `TGMapView`
- Removes animation methods for individual camera parameters

Future work:
- Camera positions use `CLLocationCoordinate2D` to represent LngLats because it's the standard type for this data on iOS. However other parts of `TGMapView` use `TGGeoPoint` for this purpose. We should standardize on `CLLocationCoordinate2D` as the type used to represent LngLats in iOS interfaces.
- `TGCameraPosition` uses `zoom` to specify the scale of the visible map area, whereas the equivalent MapKit type uses "altitude". If it helps with migration we could provide a function to convert between zoom level and effective altitude.


platforms/ios/framework/src/TGCameraPosition.mm
```diff
@@ -0,0 +1,90 @@
...
+    Tangram::CameraPosition camera;
+    camera.longitude = _center.longitude;
+    camera.latitude = _center.latitude;
+    camera.zoom = _zoom;
+    camera.rotation = (-M_PI / 180.0) * _bearing;
+    camera.tilt = (M_PI / 180.0) * _pitch;
+    return camera;
+}
+
+- (id)copyWithZone:(NSZone *)zone
```
tallytalwar: I do not know about `NSZone`, but based on the document (https://developer.apple.com/documentation/foundation/nszone?language=objc), its not recommended to be used.

matteblair: It's just what has to be implemented to satisfy the `NSCopying` protocol, which is typical of data classes (https://developer.apple.com/documentation/foundation/nscopying?language=objc)

platforms/ios/framework/src/TGMapView.h
```diff
@@ -206,28 +210,44 @@ TG_EXPORT
 @property (nonatomic) TGCameraType cameraType;
 
 /**
- TODO
+ Move the map camera to a new position with an easing animation.
+
+ @param cameraPosition The new camera position
+ @param duration The animation duration in seconds
+ @param easeType The type of easing animation
```
tallytalwar: good to add a link to `TGEaseType` in the docs.

tallytalwar:
👍

PR #1856, opened 2018-08-15
Add getters+setters for enabling map gestures on Android

matteblair:
This creates a simple way to enable or disable some or all of the gestures that can be used to change the map view. 

On iOS this can be achieved by accessing the gesture recognizers directly like:
```objective-c
mapView.rotationGestureRecognizer.enabled = NO;
```

On Android this now leaves us with some potentially redundant interfaces. The gesture responder interfaces can be used to pre-empt map view changes cause by gestures, but this is no longer necessary. We can now potentially simplify these interfaces by removing return values from the methods, or even remove these gesture responders from the public interface.


tallytalwar:
Rebasing and merging.

PR #1859, opened 2018-08-17
Treat warnings as errors in iOS projects

matteblair:
This should catch problems like the missing method implementation that I just had to fix, as well as other potential bugs.

This only applies to the Tangram iOS framework and iOS demo projects, not the core library. The core library produces many warnings right now (mostly about integer precision loss) and I didn't want to tackle all of those at once.


tallytalwar:
I am going to rebase and merge this.

tallytalwar:
Actually as discussed #1861 ios static library builds are failing on CI. Blocking merging this until we have a fix for this.

PR #1860, opened 2018-08-20
Simplified antimeridian wrapping

matteblair:
This is a new way of handling the map looping past 180° longitude. Rather than creating distinct tile addresses for the areas past the 180th meridian, we now use the same tile addresses and translate them to the appropriate location when area is visible past the 180th meridian. The view position is now kept within the canonical longitude range as well. Converting from LngLat to screen position also accounts for this behavior now, so that converting from screen position to LngLat and then back to screen position always produces the original value (this was not true before).

### DIFF
--- a/core/src/view/view.cpp
+++ b/core/src/view/view.cpp
@@ -180,13 +180,17 @@ float View::getMaxPitch() const {
 
 }
 
-
 void View::setPosition(double _x, double _y) {
-
-    m_pos.x = _x;
-    m_pos.y = _y;
+    // Wrap horizontal position around the 180th meridian, which corresponds to +/- HALF_CIRCUMFERENCE meters.
+    m_pos.x = _x - std::round(_x / MapProjection::CIRCUMFERENCE) * MapProjection::CIRCUMFERENCE;
+    // Clamp vertical position to the span of the map, which is +/- HALF_CIRCUMFERENCE meters.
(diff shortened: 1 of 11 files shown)


matteblair:
This branch was failing with a linking error for me locally too, I wanted to test whether a clean build on another machine would fix it :\

Looks like the iOS build script isn't correctly piping build errors into the output - I'll look into that too.

core/src/view/view.cpp
```diff
@@ -585,3 +592,14 @@ void View::getVisibleTiles(const std::function<void(TileID)>& _tileCb) const {
...
+// [1]
+// The maximum visible span horizontally is the span covered by 2^z - 2 tiles. We consider one tile to be
+// effectively not visible because at the 180th meridian it will be drawn on one side or the other, not half
+// on both sides. Tile coverage is calculated from the floor() of our zoom value but here we operate on the
+// continuous zoom value, so we remove one more tile to under-approximate the coverage. Algebraically we get:
+// (span of view in meters) = (view diameter in pixels) * (earth circumference) / ((tile size in pixels) * 2^z)
+// If we substitute the desired view span at the minimum zoom:
+// (span of view in meters) = (earth circumference) * (2^z - 2) / 2^z
+// We can solve for the minimum zoom:
+// z = log2((view diameter in pixels) / (tile size in pixels) + 2)
```
tallytalwar: Nicely explained.

PR #1861, opened 2018-08-20
Fix iOS build status in CircleCI

matteblair:


matteblair:
☝️That little green checkmark next to the last commit is the problem I'm trying to fix. The iOS build is clearly broken, but the CircleCI script doesn't detect the failure.

I suspect it's because we're piping Xcode's output through xcpretty. This has been an issue before, though I thought I'd fixed it the last time I changed this part of the makefile: https://github.com/tangrams/tangram-es/commit/32b95786dc6e6e2c96ff21f61262f7d1734100fb

matteblair:
☝️ The failing status for this commit means that the CI script is now correctly detecting failures. Now I can drop the previous commit.

matteblair:
As expected - the failing iOS builds are now being detected as CI failures.

And now, for my last feat, I will vanish the errors entirely!

platforms/ios/demo/TangramDemo.xcodeproj/project.pbxproj
```diff
@@ -411,7 +415,7 @@
 			buildSettings = {
 				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
 				CODE_SIGN_STYLE = Automatic;
-				DEVELOPMENT_TEAM = "";
+				DEVELOPMENT_TEAM = G436CJ835H;
```
tallytalwar: oops, I would prefer dev team string empty here. I know xcode added it automatically.

matteblair: Gah! Pay no attention to the DEVELOPMENT_TEAM behind the curtain!

PR #1862, opened 2018-08-21
Instantiate MapController asyncly

tallytalwar:
Reason:
- MapController instantiation results in loading of native libraries on the main thread
- MapController.init calls were also performed on the main thread which caused system font files to be read on the main thread.

Following map initialization api from google maps sdk, this PR provides a way to instantiate and initialize MapController on a async thread thus preventing any file read/native system libraries to be loaded on the main thread. With this approach user of the tangram android sdk can provide implementation for `MapReadyCallBack.onMapReady` which will be invoked when `MapController` is instantiated, initialized and ready to be used.

TODO:
- [x] With this approach `SceneLoadListener` can be perceived as a map state event, and can be handled aptly.

### DIFF
--- a/platforms/android/tangram/src/main/java/com/mapzen/tangram/MapView.java
+++ b/platforms/android/tangram/src/main/java/com/mapzen/tangram/MapView.java
@@ -77,8 +77,7 @@ public class MapView extends FrameLayout {
                         return mapController;
                     }
                 }
-                MapController controller = getMapInstance();
-                controller.init(ctx, handler);
+                MapController controller = getMapInstance(ctx, handler);
 
                 return controller;
             }
@@ -105,8 +104,8 @@ public class MapView extends FrameLayout {
     }
 
     @NonNull
-    protected MapController getMapInstance() {
-        return new MapController(glSurfaceView);
+    protected MapController getMapInstance(@NonNull final Context context, @Nullable final HttpHandler handler) {
+        return new MapController(glSurfaceView, context, handler);
     }
 
     protected void configureGLSurfaceView() {
--- a/platforms/android/tangram/src/main/java/com/mapzen/tangram/MapView.java
+++ b/platforms/android/tangram/src/main/java/com/mapzen/tangram/MapView.java
@@ -56,19 +56,26 @@ public class MapView extends FrameLayout {
         @Override
         protected MapController doInBackground(Void... voids) {
             MapView view = mapViewRef.get();
+            if (view == null) {
+                return null;
+            }
             return view.mapInitInBackground(httpHandler);
         }
 
         @Override
         protected void onPostExecute(MapController controller) {
             MapView view = mapViewRef.get();
-            view.onMapInitOnUIThread(controller, mapReadyCallback);
+            if (view != null) {
+                view.onMapInitOnUIThread(controller, mapReadyCallback);
+            }
         }
 
         @Override
         protected void onCancelled(MapController controller) {
             MapView view = mapViewRef.get();
-            view.onMapInitCancelled(controller);
+            if (view != null) {
+                view.onMapInitCancelled(controller);
+            }
         }
     }
 
--- a/platforms/android/tangram/src/main/java/com/mapzen/tangram/MapView.java
+++ b/platforms/android/tangram/src/main/java/com/mapzen/tangram/MapView.java
@@ -112,39 +112,6 @@ public class MapView extends FrameLayout {
 
     }
 
-    /**
-     * Construct a {@code MapController}; may only be called from the UI thread
-     * Map instance uses {@link DefaultHttpHandler} for retrieving remote map resources
-     * @param listener The listener to receive to receive scene load events;
-     * the callback will be made on the UI thread
-     */
-    @NonNull
-    public MapController getMap(@Nullable final MapController.SceneLoadListener listener) {
-        return getMap(listener, null);
-    }
-
-    /**
-     * Construct a {@code MapController}; may only be called from the UI thread
-     * @param listener The listener to receive to receive scene load events;
-     * @param handler Set the client implemented {@link HttpHandler} for retrieving remote map resources
-     *                when null {@link DefaultHttpHandler} is used
-     * the callback will be made on the UI thread
-     */
-    @NonNull
-    public MapController getMap(@Nullable final MapController.SceneLoadListener listener, @Nullable final HttpHandler handler) {
-        disposeMapReadyTask();
-        if (mapController != null) {
-            return mapController;
-        }
-        mapController = getMapInstance();
-
-        mapController.setSceneLoadListener(listener);
-        mapController.init(glSurfaceView.getContext(), handler);
-        mapController.postInit();
-
-        return mapController;
-    }
-
     @NonNull
     protected MapController getMapInstance() {
         return new MapController(glSurfaceView);
--- a/platforms/android/tangram/src/main/java/com/mapzen/tangram/MapView.java
+++ b/platforms/android/tangram/src/main/java/com/mapzen/tangram/MapView.java
@@ -100,7 +100,7 @@ public class MapView extends FrameLayout {
             if (mapController == null) {
                 mapController = controller;
                 configureGLSurfaceView();
-                mapController.postInit();
+                mapController.UIThreadInit();
             }
             if (callback != null) {
                 callback.onMapReady(mapController);
--- a/platforms/android/tangram/src/main/java/com/mapzen/tangram/MapView.java
+++ b/platforms/android/tangram/src/main/java/com/mapzen/tangram/MapView.java
@@ -59,33 +59,87 @@ public class MapView extends FrameLayout {
         protected MapController doInBackground(Void... voids) {
             MapView view = mapViewRef.get();
             Context ctx = contextRef.get();
-            synchronized (view.lock) {
-                if (view.mapController != null) {
-                    return view.mapController;
-                }
-            }
-            return view.getMapInstance(ctx, httpHandler);
+            return view.mapInitInBackground(ctx, httpHandler);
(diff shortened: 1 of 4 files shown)


matteblair:
I think there is a problem here with the GLSurfaceView lifecycle. When the MapView is created it immediately creates a GLSurfaceView as a subview. As the GLSurfaceView is setting itself up it will try to invoke the callbacks on its GLSurfaceView.Renderer whether a Renderer has been set or not. So if the GLSurfaceView reaches those callbacks before the AsyncTask finishes instantiating the MapController then the MapController will miss those callbacks entirely. This is even more clearly a problem if an Activity doesn't immediately call getMapAsync, but waits for a later point to initialize the map instead. 

This problem might be possible with our current code for this too, but I haven't seen it happen until testing out this change.

matteblair:
We may also want a code path that handles multiple calls to getMapAsync. For instance if an Activity calls getMapAsync a second time, after the MapController has already been created, we can immediately return the same MapController.

tallytalwar:
Yes, also I am missing a call to actually get `mapController` instance from `MapView` instance. Will also add that.

tallytalwar:
> We may also want a code path that handles multiple calls to getMapAsync. For instance if an Activity calls getMapAsync a second time, after the MapController has already been created, we can immediately return the same MapController.

Actually currently I have it start initiating another mapController instead of returning a prior configured mapController. But it does make sense to not do this and instead return a mapController if already set.

platforms/android/tangram/src/main/java/com/mapzen/tangram/MapView.java
```diff
@@ -18,45 +19,130 @@
...
      *                when null {@link DefaultHttpHandler} is used
-     * the callback will be made on the UI thread
+     * @return An instance of {@link MapController} if previously created else, null
      */
-    @NonNull
-    public MapController getMap(@Nullable final MapController.SceneLoadListener listener, @Nullable final HttpHandler handler) {
-        if (mapController != null) {
-            return mapController;
+    @Nullable
+    public MapController getMapAsync(@Nullable final MapReadyCallback readyCallback,
```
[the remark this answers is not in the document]
matteblair: Also, on the early return path we probably don't want to invoke the `MapReadyCallback` directly in this method.  In my experience it can be confusing or disruptive to have a supposedly asynchronous method invoke its completion handler before the original synchronous call returns. Instead we could allow the async task to run and just skip most of the work when the `MapController` is already initialized.

platforms/android/tangram/src/main/java/com/mapzen/tangram/MapView.java
```diff
@@ -18,45 +19,130 @@
...
+        disposeMapReadyTask();
+        if (glSurfaceView == null) {
+            glSurfaceView = new GLSurfaceView(getContext());
         }
-        mapController = getMapInstance();
-        mapController.setSceneLoadListener(listener);
-        mapController.init(handler);
+        final Context ctx = glSurfaceView.getContext();
+
+        getMapAsyncTask = new AsyncTask<Void, Void, MapController>() {
```
matteblair: Android Studio shows a warning here about creating an anonymous inner class here - it seems like this could lead to leaks. We should probably make an explicit inner class for this instead.

platforms/android/tangram/src/main/java/com/mapzen/tangram/MapView.java
```diff
@@ -18,45 +19,130 @@
...
 
-        return mapController;
+    /**
+     * Returns the mapController previously created using
+     * {@link MapView#getMapAsync(MapReadyCallback)} or {@link MapView#getMapAsync(MapReadyCallback, MapController.SceneLoadListener)} or
+     * {@link MapView#getMapAsync(MapReadyCallback, MapController.SceneLoadListener, HttpHandler)}
+     * @return An instance of {@link MapController} if previously created else, null
+     */
+    @Nullable
+    public MapController getMapController() {
```
matteblair: I don't think I understand the need for this method. It seems like an application could keep its own nullable reference to a `MapController` for this purpose, right?

matteblair:
I got curious about exactly how and when static initializer blocks like the one in `MapController` get called - i.e. how and when the class gets "initialized". Aside from constructing an instance of the class, it turns out that accessing a static member or method of the class also initializes the class. We don't have any reason for accessing static portions of `MapController` at the moment, but it's easy to imagine adding some. 

If we actually care about precisely when we call `System.loadLibrary` then we should probably call them in a more concrete execution path than static initializer blocks. We could track whether native libraries are loaded using a static member of `MapController` and load them once in the constructor (though this requires some extra synchronization). Or we could load the native libraries in a static initializer for a class that only `MapController` can construct (the runtime automatically synchronizes static initializers for us).

platforms/android/tangram/src/main/java/com/mapzen/tangram/MapView.java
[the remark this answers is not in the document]
tallytalwar: Yeah will update the client should be able to keep a reference of the instantiated `MapController` through the `onMapReady` interface.

platforms/android/tangram/src/main/java/com/mapzen/tangram/MapView.java
tallytalwar: Yeah, updated this.

platforms/android/tangram/src/main/java/com/mapzen/tangram/MapView.java
tallytalwar: Ahh, I see the issue here. 
Reference: https://stackoverflow.com/a/46166223 and https://www.androiddesignpatterns.com/2013/01/inner-class-handler-memory-leak.html

platforms/android/tangram/src/main/java/com/mapzen/tangram/MapView.java
```diff
@@ -11,72 +12,190 @@
...
      */
-    @NonNull
-    public MapController getMap(@Nullable final MapController.SceneLoadListener listener, @Nullable final HttpHandler handler) {
-        if (mapController != null) {
-            return mapController;
-        }
-        mapController = getMapInstance();
-        mapController.setSceneLoadListener(listener);
-        mapController.init(handler);
+    public void getMapAsync(@Nullable final MapReadyCallback readyCallback,
```
matteblair: Should the `MapReadyCallback` really be `Nullable`? I can't think of a case where you wouldn't want to get a reference to the `MapController`.

platforms/android/tangram/src/main/java/com/mapzen/tangram/NativeLibraryLoader.java
```diff
@@ -0,0 +1,16 @@
+package com.mapzen.tangram;
+
+public class NativeLibraryLoader {
```
matteblair: No need to make this `public`, I think package-private is better.

platforms/android/tangram/src/main/java/com/mapzen/tangram/NativeLibraryLoader.java
```diff
@@ -0,0 +1,16 @@
...
+    public static final boolean sNativeLibraryLoaded = loadLibrary();
```
matteblair: This is .. possibly too clever. It works as long as we're very careful to only reference this class in one spot, but it makes me nervous.

platforms/android/tangram/src/main/java/com/mapzen/tangram/NativeLibraryLoader.java
```diff
@@ -0,0 +1,16 @@
...
+
+    private static synchronized boolean loadLibrary() {
+        try {
+            System.loadLibrary("c++_shared");
+            System.loadLibrary("tangram");
+            return true;
+        } catch (final Exception e) {
```
matteblair: This should probably be more specific than "any exception". The expected exceptions here would be `SecurityException` or `UnsatisfiedLinkError` (https://docs.oracle.com/javase/7/docs/api/java/lang/System.html#loadLibrary(java.lang.String))

platforms/android/tangram/src/main/java/com/mapzen/tangram/MapView.java
```diff
@@ -11,72 +12,190 @@
...
+    protected MapAsyncTask getMapAsyncTask;
+
+    private final Object lock = new Object();
+
+    /**
+     * MapReadyCallback interface
+     * onMapReady gets invoked on the ui thread when {@link MapController} is instantiated and ready to be used
+     */
+    public interface MapReadyCallback {
+        void onMapReady(@Nullable MapController mapController);
```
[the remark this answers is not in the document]
tallytalwar: Yeah I made this `@Nullable` when I added the native library erroring piece in. I was debating on a "panicky map rendering library" vs "giving control of what to do to the client app". 

I decided on the latter. What do you think?

platforms/android/tangram/src/main/java/com/mapzen/tangram/MapView.java
tallytalwar: Yeah makes sense will update.

platforms/android/tangram/src/main/java/com/mapzen/tangram/MapView.java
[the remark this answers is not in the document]
matteblair: Yeah maybe this is a reasonable way to handle the error. I guess we could also just not invoke the `MapReadyCallback` if we can't load native libraries. Not sure that's better.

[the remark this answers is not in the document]
tallytalwar: Also added another case for this when `view` from `WeakReference` is null, InBackground will return a `null` MapController.

PR #1863, opened 2018-08-22
Use Foundation and CoreLocation types in iOS SDK

matteblair:
- Replaces all uses of TGGeoPoint with CLLocationCoordinate2D
- Replaces TGHelpers class with free functions in TGTypes+Internal.h
- Creates TGGeometry header for geometry types and functions
- Links CoreLocation framework for symbols related to CLLocationCoordinate2D
- Renames some functions for consistency with argument types


platforms/ios/framework/src/TGGeoPolygon.h
```diff
@@ -24,39 +25,39 @@ TG_EXPORT
...
- Returns a pointer to a sequence of `TGGeoPoint` with a total array length equal to `-[TGGeoPolygon count:]`.
+ Returns a pointer to a sequence of coordinates with a total array length equal to `-[TGGeoPolygon count:]`.
 
  @return a pointer to the list of polygons coordinates
  */
-- (TGGeoPoint*)coordinates;
+- (CLLocationCoordinate2D *)coordinates;
 
 /**
  The list of polygon `rings` contained in this geometry.
```
tallytalwar: This is was a few lines below, is there a reason you did not use system type (`NSInteger`) for the ` - (int *)rings;` below?

Issue #1778, opened 2018-03-07
sceneReadyCallback returns null error when there is an error


dgngulcan:
**TO REPRODUCE THE ISSUE, FOLLOW THESE STEPS:**
load a scene with an invalid URL

**RESULT:**
On console, exception is printed 
`D/OkHttp: java.io.IOException: Unexpected response code: Response{protocol=http/1.1, code=404, message=Not Found...`
 
sceneReadyCallback returns a null `Error` object with an integer sceneId

**EXPECTED RESULT:**
since the URL gives 404, I would expect an error of `MapController.Error` with `sceneReadyCallback`

**ENVIRONMENT:**
Android

polar:
I can confirm this issue.

matteblair:
Yep this error reporting facility is currently a little bit .. inadequate haha

I believe this error parameter was added specifically for scene updates - which reload the scene and invoke the scene ready callback. The current interface isn't well-suited for capturing all the things that could go wrong when loading a scene. To make things trickier, scene loading happens in native code so we can't just use exceptions. Our strategy so far has been to just write those kinds of errors to the log.

PR #1865, opened 2018-08-29
Fix for issue 1864. Segfault reading bad scene file.

polar:
The Importer was reading a sceneNode from YAML::Load as a Yaml Map without checking for it being a Map first, causing the error.

### DIFF
--- a/core/src/scene/importer.cpp
+++ b/core/src/scene/importer.cpp
@@ -144,13 +144,15 @@ std::vector<Url> Importer::getResolvedImportUrls(const Node& sceneNode, const Ur
         base = getBaseUrlForZipArchive(baseUrl);
     }
 
-    if (const Node& import = sceneNode["import"]) {
-        if (import.IsScalar()) {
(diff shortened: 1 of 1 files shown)


polar:
Fair enough.  But as another issue, I'd like to see an error like this propagated up to delivering a SceneError for such things, like the import pointer not existing or resolvable, parsing errors, etc. 
https://github.com/tangrams/tangram-es/issues/1778

tallytalwar:
Yeah we will address #1778 also. Thanks

polar:
I probably can investigate and work on this the #1778 issue, as I may have some time allocated to do so.

tallytalwar:
Thanks @polar 

We love outside contributions :D.

matteblair:
I'm good with this fix for the crash - just gonna rebase on master so we can merge it.

tallytalwar:
Thanks @polar. This got merged with #1866

polar:
Thanks!

PR #1869, opened 2018-09-07
Only send a dummy onPan when other gestures are detected

tallytalwar:
- Make sure singleTap and pan gestures are not detected simultaneously


tallytalwar:
Closing this, including the fix in #1852

PR #1870, opened 2018-09-11
Simplify Android gesture configuration

matteblair:
I spotted some ways that the gesture handling code for Android could be simplified and improved. This changes the interface a little bit but it sure cleans up MapController!

TODO:

- [ ] Document how to override gesture recognizers in touch input and forward events to MapController
- [ ] Document (or annotate?) requirements to call some methods on the main thread
- [ ] Move gesture listener implementations from public TouchInput interface to private nested class

### DIFF
--- a/platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
+++ b/platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
@@ -255,16 +255,16 @@ public class MapController implements Renderer {
         touchInput = new TouchInput(mapView.getContext());
         mapView.setOnTouchListener(touchInput);
 
-        setPanResponder(null);
-        setScaleResponder(null);
-        setRotateResponder(null);
-        setShoveResponder(null);
+        touchInput.setPanResponder(getPanResponder());
+        touchInput.setScaleResponder(getScaleResponder());
+        touchInput.setRotateResponder(getRotateResponder());
+        touchInput.setShoveResponder(getShoveResponder());
 
-        touchInput.setSimultaneousDetectionAllowed(Gestures.SHOVE, Gestures.ROTATE, false);
-        touchInput.setSimultaneousDetectionAllowed(Gestures.ROTATE, Gestures.SHOVE, false);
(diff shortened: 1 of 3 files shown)


platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
```diff
@@ -772,330 +772,139 @@ public void setRenderMode(@IntRange(from=0,to=1) final int renderMode) {
...
-            }
-        });
+    public TouchInput getTouchInput() {
+        return touchInput;
     }
 
     /**
-     * Set a responder for pan gestures
-     * @param responder PanResponder to call; if onPan returns true, normal panning behavior will not occur
+     * Get a responder for pan gestures
```
tallytalwar: Even though the above documentation for `TouchInput` mention that these responders provide the default implementation. I would still prefer the documentation for these explicitly mention "default - map movement behavior".

platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
```diff
@@ -1267,49 +1076,7 @@ public void removeAllMarkers() {
...
-            public void onRegionDidChange(final boolean animated) {
-                uiThreadHandler.post(new Runnable() {
-                    @Override
-                    public void run() {
-                        listener.onRegionDidChange(animated);
-                    }
-                });
-            }
-        };
+        mapChangeListener = listener;
```
tallytalwar: Shouldn't tangram library not make sure there are posted on the ui thread, or make it flexible for client's to aptly handle this.

Open to either but make sense to properly document this.

matteblair: Currently we invoke the mapChangeListener from two places:
- gesture listeners, which are always invoked on the UI thread
- camera updates, which the client application invokes
Handling client invocations of camera updates off the UI thread would be tricky. If we post the event to the UI thread and then immediately execute the camera update then we can't be sure that the event was executed before the camera changed. It seems like we're stuck with a choice between making sure the event is on the UI thread and making sure the event happens before the camera change.

PR #1872, opened 2018-09-12
Dispose all listener and callbacks

tallytalwar:
- dispose all listener and callbacks associated with mapController
        - prevent potential leaks in client implementations
- [x] dispose app context from `TouchInput`

### DIFF
--- a/platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
+++ b/platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
@@ -288,7 +288,8 @@ public class MapController implements Renderer {
             }
         });
 
-        // set all mapController listener references to null
+        // Dispose all listener and callbacks associated with mapController
+        // This will help prevent leaks of references from the client code, possibly used in these listener/callbacks
         touchInput = null;
         mapChangeListener = null;
         featurePickListener = null;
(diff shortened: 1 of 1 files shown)


platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
```diff
@@ -287,6 +287,16 @@ public void run() {
                 markers.clear();
             }
         });
+
+        // set all mapController listener references to null
```
matteblair: I think it's pretty clear _what_ this does, but it will be useful to explain _why_ we do it.

tallytalwar: Sure will expand on that.

matteblair:
Hm actually I just thought of a potential error - @tallytalwar  tell me if this makes sense:
- MapController enters onDrawFrame on the GL thread, checks that a listener member is not null, and is about to call a method on the listener member (like here: https://github.com/tangrams/tangram-es/blob/master/platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java#L1610)
- MapController gets disposed on UI thread, listener members are set to null
- MapController calls the method on the listener member which has just become null - NPE!

It's a very specific scenario but I don't think it's impossible. There's a simple solution too - we can grab the listener as a final local variable before we check them or call them.

tallytalwar:
Hmm.. We can guard against this by adding these `null` settings for listeners between a sync block?

matteblair:
According to https://docs.oracle.com/javase/tutorial/essential/concurrency/atomic.html
> Reads and writes are atomic for reference variables and for most primitive variables (all types except long and double).
So if we read a reference to the listener member and store it in a local var, there should be no need to synchronize.

PR #1874, opened 2018-09-16
Create a YamlUtil namespace of conversion functions

matteblair:
Allows us to share common conversion functions that
- check node types
- check conversion validity
- avoid slow paths in YamlCpp

We should prefer these functions to `YAML::convert<T>::decode` (which is slow to parse numbers) and `node.as<T>` (which is slow to parse numbers _and_ throws conversion exceptions).

### DIFF
--- a/core/src/scene/sceneLoader.cpp
+++ b/core/src/scene/sceneLoader.cpp
@@ -666,8 +666,8 @@ void SceneLoader::loadTexture(const std::shared_ptr<Platform>& platform, const s
 
     float density = 1.f;
     if (Node d = textureConfig["density"]) {
-        double val;
-        if (YamlUtil::getDouble(d, val, false)) { density = val; }
+        float val;
+        if (YamlUtil::getFloat(d, val)) { density = val; }
     }
 
     std::unique_ptr<SpriteAtlas> atlas;
@@ -807,6 +807,7 @@ void SceneLoader::loadStyleProps(const std::shared_ptr<Platform>& platform, Styl
         if (auto polylineStyle = dynamic_cast<PolylineStyle*>(&style)) {
             if (dashNode.IsSequence()) {
                 std::vector<float> dashValues;
+                dashValues.reserve(dashNode.size());
                 for (const auto& dashValue : dashNode) {
                     float floatValue;
                     if (YamlUtil::getFloat(dashValue, floatValue)) {
@@ -1794,6 +1795,7 @@ void SceneLoader::loadLayer(const std::pair<Node, Node>& layer, const std::share
             if (data_layer.IsScalar()) {
                 collections.push_back(data_layer.Scalar());
             } else if (data_layer.IsSequence()) {
+                collections.reserve(data_layer.size());
                 for (const auto& entry : data_layer) {
                     if (entry.IsScalar()) {
                         collections.push_back(entry.Scalar());
--- a/core/src/scene/sceneLoader.cpp
+++ b/core/src/scene/sceneLoader.cpp
@@ -304,7 +304,7 @@ bool SceneLoader::applyConfig(const std::shared_ptr<Platform>& _platform, const 
 
     Node animated = config["scene"]["animated"];
     if (animated) {
-        _scene->animated(animated.as<bool>());
+        _scene->animated(YamlUtil::getBoolOrDefault(animated, false));
     }
 
     for (auto& style : _scene->styles()) {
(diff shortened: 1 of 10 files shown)


core/src/util/yamlUtil.cpp
```diff
@@ -0,0 +1,109 @@
...
+    return sstream.str();
+}
+
+
+bool getInt(const YAML::Node& node, int& result, bool allowTrailingJunk) {
+    if (node.IsScalar()) {
+        auto& str = node.Scalar();
+        char* position = nullptr;
+        constexpr int base = 10;
+        long value = std::strtol(str.data(), &position, base);
```
tallytalwar: We might have discussed this a while back, but std methods do not handle European number format where `,`->`.` and vice-versa!

core/src/scene/sceneLoader.cpp
```diff
@@ -661,7 +667,7 @@ void SceneLoader::loadTexture(const std::shared_ptr<Platform>& platform, const s
     float density = 1.f;
     if (Node d = textureConfig["density"]) {
         double val;
-        if (getDouble(d, val)) { density = val; }
+        if (YamlUtil::getDouble(d, val, false)) { density = val; }
```
tallytalwar: Though implicit cast works, but probably makes sense to be explicit and verbose here and use `YamlUtil::getFloat`

core/src/scene/sceneLoader.cpp
```diff
@@ -781,24 +788,30 @@ void SceneLoader::loadStyleProps(const std::shared_ptr<Platform>& platform, Styl
...
+        }
     }
 
     if (Node dashNode = styleNode["dash"]) {
         if (auto polylineStyle = dynamic_cast<PolylineStyle*>(&style)) {
             if (dashNode.IsSequence()) {
                 std::vector<float> dashValues;
-                for (auto dashValue : dashNode) {
-                    dashValues.push_back(dashValue.as<float>());
+                for (const auto& dashValue : dashNode) {
```
tallytalwar: We can do a reserve for `dashValues` based on the size of the sequence.

core/src/util/yamlUtil.cpp
tallytalwar: Or do we not have to worry about this for integers? ..

matteblair: RIGHT! I totally forgot about this. If I can remember now, I think this results in floating point values like `1.234`(a float) being interpreted _in European locales_ as `1234` (an int).

I think rather than struggling to correctly distinguish between floats and ints, it'll be more practical to just parse all numbers as floats or doubles and round if we need to.

matteblair:
I'm adding some tests for this and changing the handling of "int" values. Stay tuned.

matteblair:
@tallytalwar Updated with a different `getInt` implementation that parses a double, then rounds to an int. We can't use this to distinguish whether a number is "typed" as an int, but we shouldn't need to anywhere in the scene config.

I also updated Catch to a newer version in order to use the `GENERATOR` macro to make test code smaller.

core/src/util/yamlUtil.cpp
```diff
@@ -66,10 +66,15 @@ int getIntOrDefault(const YAML::Node& node, int defaultValue, bool allowTrailing
 }
 
 bool getFloat(const YAML::Node& node, float& result, bool allowTrailingJunk) {
-    double doubleValue;
-    if (getDouble(node, doubleValue, allowTrailingJunk)) {
-        result = static_cast<float>(doubleValue);
-        return true;
+    if (node.IsScalar()) {
```
tallytalwar: Probably do an early return if scalar is empty?

core/src/util/yamlUtil.cpp
```diff
@@ -82,9 +87,11 @@ float getFloatOrDefault(const YAML::Node& node, float defaultValue, bool allowTr
 bool getDouble(const YAML::Node& node, double& result, bool allowTrailingJunk) {
     if (node.IsScalar()) {
         const std::string& scalar = node.Scalar();
+        int size = static_cast<int>(scalar.size());
```
tallytalwar: Early return when scalar is empty.

tests/unit/yamlUtilTests.cpp
```diff
@@ -0,0 +1,134 @@
...
+        CHECK(resultDouble == initialDouble);
+        CHECK(YamlUtil::getDoubleOrDefault(node, defaultDouble) == defaultDouble);
+
+        CHECK_FALSE(YamlUtil::getBool(node, resultBool));
+        CHECK(resultBool == initialBool);
+        CHECK(YamlUtil::getBoolOrDefault(node, defaultBool) == defaultBool);
+    }
+
+    SECTION("String Node") {
+        YAML::Node node("I saw 3 ships go sailing by");
```
tallytalwar: 🌊 🚢 🚢 🚢 🌊

tallytalwar:
New changes and unit tests look good. There are couple of nitpicks I added to the new updates and couple of suggestions unanswered from before.

matteblair:
Ok I think everything is resolved now. When CI passes I'll merge this (choosing not to squash here so that the Catch update remains a separate commit).

PR #1878, opened 2018-09-22
Pause/Resume GLSurfaceView

hjanetzek:
The parameters to initialize MapController changed slightly to create and set glSurfaceView in onMapInitOnUIThread so that onPause/Resume cannot fail due to GLSurfaceView Renderer not being set.

### DIFF
--- a/platforms/android/tangram/src/main/java/com/mapzen/tangram/MapView.java
+++ b/platforms/android/tangram/src/main/java/com/mapzen/tangram/MapView.java
@@ -23,8 +23,6 @@ public class MapView extends FrameLayout {
     protected MapController mapController;
     protected MapAsyncTask getMapAsyncTask;
 
-    private final Object lock = new Object();
-
     /**
      * MapReadyCallback interface
      * onMapReady gets invoked on the ui thread when {@link MapController} is instantiated and ready to be used
@@ -84,15 +82,12 @@ public class MapView extends FrameLayout {
      * @return new or previously initialized {@link MapController}
      */
     protected MapController mapInitInBackground() {
-        synchronized (lock) {
-            if (!NativeLibraryLoader.sNativeLibraryLoaded) {
-                android.util.Log.e("Tangram", "Unable to initialize MapController, because failed to initialize native libraries");
-                return null;
-            }
-            if (mapController != null) {
-                return mapController;
-            }
+
+        if (!NativeLibraryLoader.sNativeLibraryLoaded) {
+            android.util.Log.e("Tangram", "Unable to initialize MapController: Failed to initialize native libraries");
+            return null;
         }
+
         return new MapController(this.getContext());
     }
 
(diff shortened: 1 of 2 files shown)


platforms/android/tangram/src/main/java/com/mapzen/tangram/MapView.java
```diff
@@ -170,18 +171,12 @@ public void getMapAsync(@NonNull final MapReadyCallback readyCallback,
...
-    }
 
-    @NonNull
-    protected MapController getMapInstance(@Nullable final HttpHandler handler) {
-        return new MapController(glSurfaceView, handler);
+        executeMapAsyncTask(readyCallback, handler);
     }
 
     protected void configureGLSurfaceView() {
+        glSurfaceView = new GLSurfaceView(getContext());
```
tallytalwar: Does it make sense to only init a glSurfaceView once, like I had it before?

tallytalwar: Maybe I am wrong, but the reason I had `GLSurfaceView` created before initializing `MapController` was to make sure `Context` is passed post initialization for `GLSurfaceView` to account for any MapView context changes happening because of `GLSurfaceView`.

hjanetzek:
@tallytalwar @matteblair I've changed configureGLSurfaceView to createGLSurfaceView with the last push. Feel free to merge when you are OK with it

platforms/android/tangram/src/main/java/com/mapzen/tangram/MapView.java
```diff
@@ -94,20 +93,23 @@ protected MapController mapInitInBackground(@Nullable final HttpHandler handler)
...
      * @param controller {@link MapController} created in a background thread
+     * @param handler {@link HttpHandler} required for network handling
      * @param callback {@link MapReadyCallback}
      */
-    protected void onMapInitOnUIThread(@Nullable final MapController controller, @NonNull final MapReadyCallback callback) {
+    protected void onMapInitOnUIThread(@Nullable final MapController controller, @Nullable final HttpHandler handler,
+                                       @NonNull final MapReadyCallback callback) {
         synchronized (lock) {
             if (mapController == null && controller != null) {
+                glSurfaceView = createGLSurfaceView();
```
tallytalwar: One nitpick we discussed during our meeting, it will be good as a sanity check to make sure glSurfaceView has not been previously created, though its probably not possible.

hjanetzek: I'm not sure where this would helps. 

For sanity I would propose to not allow getMapAsync to be called twice since the MapController could be created once or twice in that case currently.. One moment I'll push a fix to ensure MapController can only be created once

tallytalwar: fb0bc17 looks good to me.

hjanetzek: .. no this solution is not really good either. Lets move this to another PR

tallytalwar: :/ what did I miss in my review!

hjanetzek: That change should improves things - nothing missed I hope :) To me it's not clear why we support to call getMapAsync more than once

matteblair: This process is surprisingly tricky. I think the glSurfaceView initialization has to be matched one-to-one with the map controller initialization because we can't set a renderer on the glSurfaceView more than once (GLSurfaceView throws if we try to).

hjanetzek:
Yeah.. pushed a change that should properly follow the getMapAsync description. I would prefer to have a createMapAsync which throws when called a second time + getMap returning MapController (which blocks until the async task is done)

PR #1880, opened 2018-09-22
Fix unintended callback

hjanetzek:


platforms/android/tangram/src/main/java/com/mapzen/tangram/networking/DefaultHttpHandler.java
```diff
@@ -93,14 +93,14 @@ public void onResponse(final Call call, final Response response) {
                 if (body != null) {
                     try {
                         data = body.bytes();
+                        cb.onResponse(response.code(), data);
                     } catch (final IOException e) {
-                        Log.e("Tangram", "Error reading bytes from response body.", e);
+                        Log.e("Tangram", "Error reading bytes from response body:" + e.getMessage());
```
tallytalwar: Ahh both `onResponse` with `null` data and `onFailure` was being called. 👍

tallytalwar:
Rebase and can be merged.

PR #1881, opened 2018-09-22
Fix android on url response

hjanetzek:


platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
```diff
@@ -1671,39 +1685,56 @@ void cancelUrlRequest(final long requestHandle) {
 
     @Keep
     void startUrlRequest(@NonNull final String url, final long requestHandle) {
-        if (httpHandler == null) {
+        // FIXME
+        // This is still does not ensure that handler.startRequest is not
+        // executed after MapController.dispose() when startUrlRequest is
+        // callend from worker threads. At least the result is ignored
+
+        HttpHandler handler = httpHandler;
```
tallytalwar: This should be synchronized!

platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
```diff
@@ -1650,7 +1664,7 @@ public void onSurfaceCreated(final GL10 gl, final EGLConfig config) {
     // ==================
 
     private void syncRemoveHttpHandle(final long requestHandle) {
-        synchronized(this) {
+        synchronized(httpRequestHandles) {
```
matteblair: Android Studio warns about synchronizing on a non-final object. Might as well declare `httpRequestHandles` as `final`.

platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
```diff
@@ -1650,7 +1664,7 @@ public void onSurfaceCreated(final GL10 gl, final EGLConfig config) {
     // ==================
 
     private void syncRemoveHttpHandle(final long requestHandle) {
```
matteblair: Not a big deal, but it might be clearer to just perform this synchronized remove in-line so that the calling code is more obvious.

PR #1882, opened 2018-09-26
Improve thread safety and optimize JNI usage in Android callbacks

matteblair:
- Refactor camera animation callback implementation to only modify the callback reference on the UI thread
- Move listener interfaces into separate files
- Refactor JNI implementations of listener callbacks to remove need for JNI global references


platforms/android/tangram/src/main/cpp/jniExports.cpp
```diff
@@ -22,6 +22,10 @@ std::vector<Tangram::SceneUpdate> unpackSceneUpdates(JNIEnv* jniEnv, jobjectArra
 
 extern "C" {
 
+    JNIEXPORT jint JNI_OnLoad(JavaVM* vm, void* reserved) {
```
tallytalwar: https://docs.oracle.com/javase/8/docs/technotes/guides/jni/spec/invocation.html#JNJI_OnLoad 👍

platforms/android/tangram/src/main/cpp/androidPlatform.cpp
```diff
@@ -79,12 +86,13 @@ void AndroidPlatform::setupJniEnv(JNIEnv* jniEnv) {
...
-    if (hashmapClass) {
-        jniEnv->DeleteGlobalRef(hashmapClass);
-    }
+    // We need a reference to the class object later to invoke the constructor. FindClass produces a
+    // local reference that may not be valid later, so create a global reference to the class.
     hashmapClass = (jclass)jniEnv->NewGlobalRef(jniEnv->FindClass("java/util/HashMap"));
     hashmapInitMID = jniEnv->GetMethodID(hashmapClass, "<init>", "()V");
     hashmapPutMID = jniEnv->GetMethodID(hashmapClass, "put", "(Ljava/lang/Object;Ljava/lang/Object;)Ljava/lang/Object;");
+
+    return JNI_VERSION_1_6;
```
tallytalwar: Should probably document the link here: https://docs.oracle.com/javase/8/docs/technotes/guides/jni/spec/invocation.html#JNJI_OnLoad

The return value has special logic.

platforms/android/tangram/src/main/cpp/androidPlatform.cpp
```diff
@@ -63,8 +63,15 @@ void AndroidPlatform::bindJniEnvToThread(JNIEnv* jniEnv) {
     jniEnv->GetJavaVM(&jvm);
 }
 
-void AndroidPlatform::setupJniEnv(JNIEnv* jniEnv) {
-    bindJniEnvToThread(jniEnv);
+jint AndroidPlatform::jniOnLoad(JavaVM* javaVM) {
```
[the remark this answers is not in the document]
tallytalwar: Or does these get GC'ed automatically? Still it would be prudent to delete these.

platforms/android/tangram/src/main/cpp/androidPlatform.cpp
```diff
@@ -385,65 +359,52 @@ void setCurrentThreadPriority(int priority) {
...
     if (labelPickResult) {
         auto properties = labelPickResult->touchItem.properties;
 
-        position[0] = labelPickResult->touchItem.position[0];
-        position[1] = labelPickResult->touchItem.position[1];
+        x = labelPickResult->touchItem.position[0];
+        y = labelPickResult->touchItem.position[1];
 
-        jobject hashmap = jniEnv->NewObject(hashmapClass, hashmapInitMID);
+        hashmap = jniEnv->NewObject(hashmapClass, hashmapInitMID);
```
tallytalwar: Just FYI, I was curious on if the allocations here are reflected aptly in the profiler, seems to be the case and gets deallocated aptly during onLabelPick implementation.

platforms/android/tangram/src/main/cpp/androidPlatform.cpp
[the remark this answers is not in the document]
matteblair: Updated this to delete the global ref on the hasmap class in OnUnload. The tangramClass ref is handled by the dispose call and would be difficult to delete from a free function like this because it's a member of the AndroidPlatform class.

PR #1883, opened 2018-09-26
Map changing events and map change state management

tallytalwar:
- Provides access to platform code to check if camera is in an animating state.
- Adds logic to use this camera animating state to aptly invoke map changing MapChangeListener interface methods for both iOS and android platforms.
- [Android] Adds MapChangeStateManager to aptly manage map change states (`will update`, `will update animated`, `is updating`, `did update`, `did update animated`).
- [x] Implement the above mentioned StateManager for iOS
- [x] implement state change notifier for iOS

This also fixes (3ec875c) a regression which was causing make `DID_CHANGE` to be called repeated irrespective of a cameraAnimationCallback being set or not. (https://github.com/tangrams/tangram-es/pull/1853/commits/0908a17cd8a556b57a075f77ada21dfaf784863c)

### DIFF
--- a/platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
+++ b/platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
@@ -92,12 +92,9 @@ public class MapController implements Renderer {
     }
 
     public enum MapRegionChangeState {
-        MAP_REGION_NO_CHANGE,
-        MAP_REGION_WILL_CHANGE,
-        MAP_REGION_WILL_CHANGE_ANIMATED,
-        MAP_REGION_IS_CHANGING,
-        MAP_REGION_DID_CHANGE,
-        MAP_REGION_DID_CHANGE_ANIMATED,
+        IDLE,
+        JUMPING,
+        ANIMATING,
     }
 
     /**
@@ -460,9 +457,12 @@ public class MapController implements Renderer {
      */
     public void updateCameraPosition(@NonNull final CameraUpdate update, final int duration, @NonNull final EaseType ease, @Nullable final CameraAnimationCallback cb) {
         checkPointer(mapPointer);
-        final boolean animated = (duration > 0);
-        onRegionWillChange(animated);
-        setPendingCameraAnimationCallback(cb, animated);
+        if (duration > 0) {
+            setMapRegionState(MapRegionChangeState.ANIMATING);
+        } else {
+            setMapRegionState(MapRegionChangeState.JUMPING);
+        }
+        setPendingCameraAnimationCallback(cb);
         final float seconds = duration / 1000.f;
         nativeUpdateCameraPosition(mapPointer, update.set, update.longitude, update.latitude, update.zoom,
                 update.zoomBy, update.rotation, update.rotationBy, update.tilt, update.tiltBy,
@@ -503,19 +503,20 @@ public class MapController implements Renderer {
 
     private void flyToCameraPosition(@NonNull final CameraPosition position, final int duration, @Nullable final CameraAnimationCallback callback, final float speed) {
         checkPointer(mapPointer);
-        onRegionWillChange(true);
-        setPendingCameraAnimationCallback(callback, true);
+        // TODO: Make sense to have mark animating for flyTo irrespective of duration/speed?
+        setMapRegionState(MapRegionChangeState.ANIMATING);
+        setPendingCameraAnimationCallback(callback);
         final float seconds = duration / 1000.f;
         nativeFlyTo(mapPointer, position.longitude, position.latitude, position.zoom, seconds, speed);
     }
 
-    private void setPendingCameraAnimationCallback(final CameraAnimationCallback callback, final Boolean animated) {
+    private void setPendingCameraAnimationCallback(final CameraAnimationCallback callback) {
         synchronized (cameraAnimationCallbackLock) {
             // Wrap the callback to run corresponding map change events.
             pendingCameraAnimationCallback = new CameraAnimationCallback() {
                 @Override
                 public void onFinish() {
-                    onRegionDidChange(animated);
+                    setMapRegionState(MapRegionChangeState.IDLE);
                     if (callback != null) {
                         callback.onFinish();
                     }
@@ -524,7 +525,7 @@ public class MapController implements Renderer {
                 @Override
                 public void onCancel() {
                     // Possible camera update was cancelled in between, so should account for this map change
-                    onRegionDidChange(animated);
+                    setMapRegionState(MapRegionChangeState.IDLE);
                     if (callback != null) {
                         callback.onCancel();
                     }
@@ -798,25 +799,26 @@ public class MapController implements Renderer {
         return new TouchInput.PanResponder() {
             @Override
             public boolean onPanBegin() {
-                onRegionWillChange(true);
+                setMapRegionState(MapRegionChangeState.ANIMATING);
                 return true;
             }
 
             @Override
             public boolean onPan(final float startX, final float startY, final float endX, final float endY) {
-                onRegionIsChanging();
+                setMapRegionState(MapRegionChangeState.ANIMATING);
                 nativeHandlePanGesture(mapPointer, startX, startY, endX, endY);
                 return true;
             }
 
             @Override
             public boolean onPanEnd() {
-                onRegionDidChange(true);
+                setMapRegionState(MapRegionChangeState.IDLE);
                 return true;
             }
 
             @Override
             public boolean onFling(final float posX, final float posY, final float velocityX, final float velocityY) {
+                setMapRegionState(MapRegionChangeState.ANIMATING);
                 nativeHandleFlingGesture(mapPointer, posX, posY, velocityX, velocityY);
                 return true;
             }
@@ -824,7 +826,6 @@ public class MapController implements Renderer {
             @Override
             public boolean onCancelFling() {
                 cancelCameraAnimation();
-                onRegionDidChange(true);
                 return true;
             }
         };
@@ -837,20 +838,20 @@ public class MapController implements Renderer {
         return new TouchInput.RotateResponder() {
             @Override
             public boolean onRotateBegin() {
-                onRegionWillChange(true);
+                setMapRegionState(MapRegionChangeState.ANIMATING);
                 return true;
             }
 
             @Override
             public boolean onRotate(final float x, final float y, final float rotation) {
-                onRegionIsChanging();
+                setMapRegionState(MapRegionChangeState.ANIMATING);
                 nativeHandleRotateGesture(mapPointer, x, y, rotation);
                 return true;
             }
 
             @Override
             public boolean onRotateEnd() {
-                onRegionDidChange(true);
+                setMapRegionState(MapRegionChangeState.IDLE);
                 return true;
             }
         };
@@ -863,20 +864,20 @@ public class MapController implements Renderer {
         return new TouchInput.ScaleResponder() {
             @Override
             public boolean onScaleBegin() {
-                onRegionWillChange(true);
+                setMapRegionState(MapRegionChangeState.ANIMATING);
                 return true;
             }
 
             @Override
             public boolean onScale(final float x, final float y, final float scale, final float velocity) {
-                onRegionIsChanging();
+                setMapRegionState(MapRegionChangeState.ANIMATING);
                 nativeHandlePinchGesture(mapPointer, x, y, scale, velocity);
                 return true;
             }
 
             @Override
             public boolean onScaleEnd() {
-                onRegionDidChange(true);
+                setMapRegionState(MapRegionChangeState.IDLE);
                 return true;
             }
         };
@@ -889,20 +890,20 @@ public class MapController implements Renderer {
         return new TouchInput.ShoveResponder() {
             @Override
             public boolean onShoveBegin() {
-                onRegionWillChange(true);
+                setMapRegionState(MapRegionChangeState.ANIMATING);
                 return true;
             }
 
             @Override
             public boolean onShove(final float distance) {
-                onRegionIsChanging();
+                setMapRegionState(MapRegionChangeState.ANIMATING);
                 nativeHandleShoveGesture(mapPointer, distance);
                 return true;
             }
 
             @Override
             public boolean onShoveEnd() {
-                onRegionDidChange(true);
+                setMapRegionState(MapRegionChangeState.IDLE);
                 return true;
             }
         };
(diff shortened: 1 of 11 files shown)


platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
```diff
@@ -1089,24 +1088,66 @@ public void removeAllMarkers() {
...
 
-    //Convenience member functions
-    private void onRegionWillChange(boolean animated) {
-        if (mapChangeListener != null) {
-            mapChangeListener.onRegionWillChange(animated);
+    private static class MapRegionChangeStateManager {
+
+        enum MapRegionChangeState {
+            MAP_REGION_NO_CHANGE,
+            MAP_REGION_WILL_CHANGE,
```
matteblair: Do you think it would be easier to think of change events as transitions rather than states? The states you have here seem to me like transitions between "idle", "animating", and "jumping".

tallytalwar: Hmm, it might be actually, will give it a try.

tallytalwar: Gave this some more thought, I think it is simpler and more logical to use transitions instead of the `idle`, `animating` and `jumping` states.
Transitions seems to match exactly with the map events we are catering to here, also can help reduce some logic, example:
- One should not repeated call `mapChangeListener.onRegionWillChange(animated)` is the transition is already `WILL_CHANGE_ANIMTED` or `WILL_CHANGE`. I do not think this can be aptly mapped with the states.

tallytalwar: I can however change the enum name to `MapRegionChangeTransitions`?

matteblair: It was easy enough to implement on top of your changes so I wrote out the logic as I've been thinking of it: https://github.com/tangrams/tangram-es/commit/15e1859cb1ddbef181b8cf8b5275bbb9a9eb7d92

tallytalwar: Hmm I see, will see if I can extend that logic. Example if I start panning  the map when map was animating say due to a flyTo, or if some event explicitly sets a position (basically state set to jumping) when map is animating. Such state management logic will have to be added to this.

matteblair: Not claiming that this is objectively better in any way, I just wanted to see how it would work.

tallytalwar: Ohh totally, I actually like lesser enums (states vs transitions) getting us to the same place. So will be a cleaner approach.

tallytalwar: @matteblair adopted States here: 53fb10f3ed0e1d64a4a1d33a617556c9a5e3f000

platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
```diff
@@ -1089,24 +1098,56 @@ public void removeAllMarkers() {
...
     public void setMapChangeListener(@Nullable final MapChangeListener listener) {
         mapChangeListener = listener;
+        mapRegionChangeStateManager = new MapRegionChangeStateManager(mapChangeListener);
     }
 
-    //Convenience member functions
-    private void onRegionWillChange(boolean animated) {
-        if (mapChangeListener != null) {
-            mapChangeListener.onRegionWillChange(animated);
+    private static class MapRegionChangeStateManager {
```
matteblair: Why create an inner class here? Seems to create clutter and I'm not sure what's gained from it.

platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
```diff
@@ -1089,24 +1098,56 @@ public void removeAllMarkers() {
...
+            IDLE,
+            JUMPING,
+            ANIMATING,
         }
-    }
 
-    private void onRegionDidChange(boolean animated) {
-        if (mapChangeListener != null) {
-            mapChangeListener.onRegionDidChange(animated);
+        WeakReference<MapChangeListener> mapChangeListenerWeakReference;
```
matteblair: Why is this a weak reference?

platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
```diff
@@ -1089,24 +1098,56 @@ public void removeAllMarkers() {
...
+                        if (state == MapRegionChangeState.JUMPING) {
+                            mapChangeListener.onRegionWillChange(false);
+                        } else if (state == MapRegionChangeState.ANIMATING){
+                            mapChangeListener.onRegionWillChange(true);
+                        }
+                        break;
+                    case JUMPING:
+                        if (state == MapRegionChangeState.IDLE) {
+                            mapChangeListener.onRegionDidChange(false);
+                        } else if (state == MapRegionChangeState.ANIMATING) { // e.g. panning to fling
```
matteblair: I'm not sure that `JUMPING` -> `ANIMATING` is a valid transition. This would mean we produce "is changing" calls with no preceding "will change animated", which seems wrong. Also, panning currently sets the state to `ANIMATING` (though maybe it _should_ be `JUMPING`).

platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
```diff
@@ -1089,24 +1098,56 @@ public void removeAllMarkers() {
...
+                            mapChangeListener.onRegionIsChanging();
+                        }
+                        break;
+                    case ANIMATING:
+                        if (state == MapRegionChangeState.IDLE) {
+                            mapChangeListener.onRegionDidChange(true);
+                        } else if (state == MapRegionChangeState.ANIMATING) {
+                            mapChangeListener.onRegionIsChanging();
+                        } else { // ?? ANIMATING -> JUMPING?? Possible?
```
matteblair: I'd prefer `else if (state == MapRegionChangeState.JUMPING)` just to be clearer, and in case we add more states.

matteblair: I don't _think_ this transition is ever triggered - animation should be canceled and set the state to `IDLE` before it becomes `JUMPING`. I guess we could find out for sure by putting a breakpoint here or throwing an exception.

platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
```diff
@@ -1380,9 +1422,14 @@ public void onDrawFrame(final GL10 gl) {
...
         boolean viewComplete;
+        boolean isCameraEasing;
         synchronized(this) {
             viewComplete = nativeUpdate(mapPointer, delta);
-            nativeRender(mapPointer);
+            isCameraEasing = nativeRender(mapPointer);
+        }
+
+        if (isCameraEasing) {
+            mapRegionChangeStateManager.setMapRegionState(MapRegionChangeStateManager.MapRegionChangeState.ANIMATING);
```
matteblair: This executes a listener callback on the GL thread, we should run it on the UI thread instead.

platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
tallytalwar: Yeah I was not able to trigger this at all. Wanted to keep it in there for discussion point of view.

platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
tallytalwar: I wanted to encapsulate the logic and not make it part of MapController itself. Since its internal, maybe I can add a package private external class for the same.

platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
tallytalwar: To me it didn't make sense for MapRegionChangeManager to hold a strong reference to MapChangeListener. Also during MapController init, we need to initialize MapRegionChangeManager with a `null`!

platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
matteblair: There's not much to encapsulate. I think the worst that we could do is accidentally set the state directly instead of using the method - and actually we can do that anyway even with the inner class.

platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
matteblair: Also not an issue if we remove the inner class..

tallytalwar: Inner class is gone => weak reference is gone!

platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
tallytalwar: I feel these gesture responders should be `ANIMATING`.

tallytalwar: `JUMPING` state should imply any change which results in sudden change in map view:
eg. a cameraUpdate call with duration set to 0!

platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
tallytalwar: Going to remove this. Do not think it makes sense to respect this state transition of `ANIMATING`->`JUMPING`. The transitions of `DID_ANIMATING`, `WILL_CHANGE` implying an `IDLE` state in between will always happen.

platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
```diff
@@ -85,6 +85,12 @@
         SELECTION_BUFFER,
     }
 
+    public enum MapRegionChangeState {
```
matteblair: Think this enum should be private? No need for it to be visible outside this class.

platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
```diff
@@ -1380,9 +1406,19 @@ public void onDrawFrame(final GL10 gl) {
...
         boolean viewComplete;
+        boolean isCameraEasing;
         synchronized(this) {
             viewComplete = nativeUpdate(mapPointer, delta);
-            nativeRender(mapPointer);
+            isCameraEasing = nativeRender(mapPointer);
+        }
+
+        if (isCameraEasing) {
+            uiThreadHandler.post(new Runnable() {
```
matteblair: Small thing: we can make one Runnable and post it repeatedly to avoid excess GC churn.

tallytalwar: Yeah totally, I feel I have caused this comment at least a couple of times now! Lesson Learnt!

tallytalwar:
I am going to add ios work to the same PR. So do not merge.

tallytalwar:
Ok this is good for final review.

Updated with following since last discussion on this:
- fixed fling map region state change behavior on android
- Added map state change logic for iOS
- fixed a regression with cameraAnimationListener calling "DID_CHANGE" too aggressively
- Added state change notifier for a client to explicit set map region change states.

platforms/ios/framework/src/TGMapView.mm
```diff
@@ -38,6 +38,15 @@ inline CLLocationDirection convertRotationRadiansToBearingDegrees(float rotation
     return TGDegreesFromRadians(-rotation);
 }
 
+/**
+ Map region change states
+ */
+typedef NS_ENUM(NSInteger, TGMapRegionChangeStates) {
```
matteblair: Conventionally, enum names are singular - it makes a little more sense when using it as a type name.

platforms/ios/framework/src/TGMapView.mm
```diff
@@ -38,6 +38,15 @@ inline CLLocationDirection convertRotationRadiansToBearingDegrees(float rotation
...
+    IDLE = 0,
```
matteblair: Enum values should use our "TG" prefix because unlike Java everything is in the global namespace :\

Conventionally they would be something like `TGMapRegionIdle`.

platforms/ios/demo/src/MapViewController.m
```diff
@@ -150,18 +165,32 @@ - (void)mapView:(TGMapView *)view recognizer:(UIGestureRecognizer *)recognizer d
...
+        NSLog(@"Animation completed %d", !canceled);
+    }];
+}
+
+- (void)mapView:(TGMapView *)view recognizer:(UIGestureRecognizer *)recognizer didRecognizeDoubleTapGesture:(CGPoint)location {
+    CLLocationCoordinate2D coordinates = [view coordinateFromViewPosition:location];
+    TGCameraPosition* camera = [view cameraPosition];
+    camera.center = CLLocationCoordinate2DMake(coordinates.latitude, coordinates.longitude);
+    camera.zoom += 1;
+    [view setCameraPosition:camera withDuration:5000 easeType:TGEaseTypeCubic callback: ^(BOOL canceled){
```
matteblair: This animation is 5000 seconds long, probably not what you intended 😄

matteblair:
I'm not seeing map events during pan-flinging on iOS. Not sure why yet - are you observing this as well?

platforms/ios/demo/src/MapViewController.m
tallytalwar: Nopes, I purposely did 5000, its very soulful to take a slow route from sf to new delhi :P.

tallytalwar:
> I'm not seeing map events during pan-flinging on iOS. Not sure why yet - are you observing this as well?

Interesting I was able to see flinging show as animating, I specifically tested for that. Will check again, I might have refactored something later!

platforms/ios/framework/src/TGMapView.mm
tallytalwar: Yeah I always forget this!!!!! Will update.

platforms/ios/demo/src/MapViewController.m
tallytalwar: Also this is from a drop commit.

platforms/ios/demo/src/MapViewController.h
```diff
@@ -12,12 +12,16 @@
 
 - (void)mapView:(nonnull TGMapView *)mapView didLoadScene:(int)sceneID withError:(nullable NSError *)sceneError;
 - (void)mapViewDidCompleteLoading:(nonnull TGMapView *)mapView;
+- (void)mapViewRegionIsChanging:(TGMapView *)mapView;
```
msmollin: Nit: Unnecessary to define these in the header as its implied by the `TGMapViewDelegate` protocol conformance. It doesn't hurt anything but it's unnecessary.

platforms/ios/framework/src/TGMapView.mm
msmollin: Unless you mark it as static, then its scoped to the file, but lets not do that :)

platforms/ios/framework/src/TGMapView.h
```diff
@@ -399,6 +399,26 @@ TG_EXPORT
  */
 @property (strong, nonatomic) UILongPressGestureRecognizer* longPressGestureRecognizer;
 
+#pragma mark Map region change state notifiers
+
+/**
+ Client can use these to trigger explicit `TGMapRegionChangeStates` changes
+ Notifies that client is going to begin map region animation
+ */
+- (void)notifyGestureDidBegin;
```
msmollin: So, the general iOS guideline for callbacks is
`will` / `is` / `did` and you want to pass along the source, so normally the map view. For example:
`- (void)mapView:(TGMapView*)mapview willChangeRegion;`
`- (void)mapView:(TGMapView*)mapview regionChanged;`
`- (void)mapView:(TGMapView*)mapview didFinishChangingRegion;`

You might also want to think about, similar to gesture recognizers, something like:
`- (BOOL)shouldMapview:(TGMapView*)mapview allowRegionChange`

Another thought: I would say the naming here is a little confusing from an iOS perspective. Notification is generally done via delegate callbacks, NSNotificationCenter, or KVO. Function callbacks *do* happen, like in `viewDidLoad` and other lifecycle-ish methods, but generally it's preferable to go with one of the other 3.

Lastly, if you plan on keeping these, I would add comments here about subclassing requirements that developers *must* call super, or internal functionality will break if you don't.

tallytalwar:
@matteblair flinging on both platforms should behave smoothly now.

platforms/ios/framework/src/TGMapView.h
msmollin: Discussed this in person with @tallytalwar and understand the decisions here now :)

platforms/ios/framework/src/TGMapView.mm
```diff
@@ -1022,15 +1050,15 @@ - (void)respondToPanGesture:(UIPanGestureRecognizer *)panRecognizer
 
     switch (panRecognizer.state) {
         case UIGestureRecognizerStateBegan:
-            [self regionWillChangeAnimated:YES];
+            [self setMapRegionChangeState:TGMapRegionAnimating];
```
matteblair: This initially seemed like a good way to categorize these events, but it turns out that MapKit actually calls `regionWillChangeAnimated: false` at the start of panning and other gestures. To match this behavior we'll need to modify our state machine a little. I think we can just add a transition from "jumping" to itself, and call `regionIsChanging` on that transition. Then for gesture events like this we'll set the state to "jumping" repeatedly.

tallytalwar: Yeah, works great. Good suggestion.

tallytalwar:
Ok Updated. All good here now?

PR #1884, opened 2018-09-27
Changed name to match arguments

polar:
The CameraUpdateFactory has newLatLngBounds with arguments of LngLat, so changing it to newLngLatBounds is prudent.


matteblair:
Merged in #1903 :)

PR #1885, opened 2018-09-27
Add zoom stops for scene.background.color.

polar:
I had a requirement to alter the background color based on zoom level. 
This pull request adds a zoom stop specification for scene.background.color and applies it
to the rendering of the map.
```
scene:
   background:
      color: [[1,blue],[5,orange],[15,black]]
```
The colors are interpolated between the zoom stops. 
Its a pretty nifty effect.

You can still use the default, such as a color in name, hash, or rgb sequence.

### DIFF
--- a/core/src/map.cpp
+++ b/core/src/map.cpp
@@ -540,10 +540,17 @@ void Map::render() {
         impl->selectionQueries.clear();
     }
 
+    // Get background color for frame based on zoom level, if there are stops
+    auto background = impl->scene->background();
+    if (impl->scene->backgroundStops().frames.size() > 0) {
+        background = impl->scene->backgroundStops().evalColor(getZoom());
+    }
+
     // Setup default framebuffer for a new frame
     glm::vec2 viewport(impl->view.getWidth(), impl->view.getHeight());
+
     FrameBuffer::apply(impl->renderState, impl->renderState.defaultFrameBuffer(),
-                       viewport, impl->scene->background().toColorF());
+                       viewport, background.toColorF());
 
     if (drawSelectionBuffer) {
         impl->selectionBuffer->drawDebug(impl->renderState, viewport);
--- a/core/src/scene/scene.h
+++ b/core/src/scene/scene.h
@@ -2,6 +2,7 @@
 
 #include "map.h"
 #include "platform.h"
+#include "stops.h"
 #include "util/color.h"
 #include "util/url.h"
 #include "util/yamlPath.h"
@@ -32,7 +33,6 @@ class Style;
 class Texture;
 class TileSource;
 class ZipArchive;
-struct Stops;
 
 // Delimiter used in sceneloader for style params and layer-sublayer naming
 const std::string DELIMITER = ":";
@@ -88,6 +88,7 @@ public:
     auto& functions() { return m_jsFunctions; }
     auto& stops() { return m_stops; }
     auto& background() { return m_background; }
+    auto& backgroundStops() { return m_backgroundStops; }
     auto& fontContext() { return m_fontContext; }
     auto& globalRefs() { return m_globalRefs; }
     auto& featureSelection() { return m_featureSelection; }
@@ -192,6 +193,7 @@ private:
     std::list<Stops> m_stops;
 
     Color m_background;
+    Stops m_backgroundStops;
 
     std::shared_ptr<FontContext> m_fontContext;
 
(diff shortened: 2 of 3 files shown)


nvkelso:
Neat!

polar:
Yeah, it can be pretty trippy. :)

bcamper:
If we merge this, please add a corresponding issue in Tangram JS.

bcamper:
One behavioral note though: this looks like it will smoothly interpolate and update the background color in between zooms (per frame); is this correct? That's straightforward implementation-wise, and can provide some nice styling effects.

However, it's different from how all other styling color interpolation with zoom stops works, where the interpolation is only done on integer zoom levels (corresponding to when new tiles are built). We could choose to restrict the background color to interpolate on integer zooms to match.

Either is possible, but we should just consider if it would be confusing or undesirable if we have different interpolation behavior for different color properties. (Side note, we DO have variable zoom interpolation for camera properties for `focal_length`/`fov`, but they require that by nature to achieve a proper continuous zoom, and don't conflict with other similar properties.)

cc @nvkelso @sensescape @meetar

meetar:
Maybe an option for stepwise interpolation? I agree it would be odd if it was the only color changing between the integer zooms…

matteblair:
Definitely not trivial to make arbitrary properties interpolate continuously between zooms, but this specific case is pretty easy to interpolate - as the diff shows. 

I just discovered that in Tangram ES you can make this effectively step-wise by writing your stops like:
```
color: [[1, blue], [5, blue], [5, orange], [15, orange], [15, black]]
```
Not sure if this is advisable though because the color values at exactly zoom 5 and zoom 15 are not very well-defined.

bcamper:
Yep, not suggesting pursuing making other properties interpolate
continuously!

I just think that there's potential liability in having inconsistent
behavior only for this color parameter.

brianLeeson:
@bcamper
I am inquiring on the path forward for this PR. Are you looking for this request to be modified to use integer zoom levels before it is accepted? What external action, if any, needs to happen before this can be merged? Thank you.

polar:
So, where else is color interpreted for zoom levels? Is it for foreground color, such as text color? And why is it only for integer zoom levels, when we travel zoom levels in a decimal fashion? I was under the impression that you could use decimals in a zoom stops specification, such as [[2,blue], [3.25, orange]].  Maybe, I'm wrong?

In any case, my only requirement was to have different colors for certain zoom levels, since a single color for all zoom levels is quite limiting. I only did the color interpolation because the function was there, and I didn't have to write one. :)  

The effect of the pull request implementation is that when the zooms are in between numbers (not necessarily integers), when smoothly zooming through the levels, the color change isn't very abrupt, or should I say, it doesn't have to be abrupt. As @matteblair mentions above, you can specify it to be abrupt. If zoom level specification went to the level of only integers, this restriction may leave the user with startling changes. Especially true when doing large swaths of area such as background color, as opposed to a text foreground color.  I think that the foreground/background distinction may be relevant here.

Regardless, of how it goes, I wouldn't like to see the changes only restricted to integer values, because then at least, you can still specify somewhat of a gradient with decimal zoom stops.

bcamper:
Color is interpreted for zoom levels for properties such as `color` values for feature rendering for polygons, lines, points, etc., and also for `font` (and font `stroke`) color properties. Those are all only evaluated at integer zooms, because they are sampled when new tile geometry is built. The color values that are computed for the zoom get baked into the GL buffer data for the geometry (which is why they are currently fixed for the entire zoom range at each integer level).

I believe you can specify decimal values inside zoom stops, and they will be taken into account accurately when the interpolation is sampled -- but, that sampling will still only occur at each integer zoom change, if that makes sense.

My comment on the smooth interpolation here was simply that I think it would be confusing to have *different* color interpolation behavior for this one property. It's not because I think it's better! :) More that I think consistency may be more important. Open to more feedback though.

While it would be *possible* to add capability to interpolate those color values through fractional zooms, it's not trivial because it involves new vertex attributes and/or shader logic; this was @matteblair's point. So it's not something we're actively pursuing, though I think it is definitely worth keeping in mind for future functionality.

polar:
Ah, so it's an efficiency issue. I'm not too familiar with the GL shading logic, and now I assume that the continuous non-integer feature incurs a big hit. Is there anything I can do to make it less so?

matteblair:
Updated this with the `master` branch and modified the logic to evaluate background color stops at integer zooms. Using a single background color over an integer zoom level is likely to be more useful for designing a background to match the styling of tiles at each zoom level. For gradual changes in color the instant change isn't too jarring.

PR #1887, opened 2018-09-28
Ensure memory used by texture data is returned to system after GL upload

matteblair:
Calling `clear()` on the data buffer in `Texture` doesn't actually free the memory, it just changes the `size` value. `shrink_to_fit()` will request that the buffer is actually freed. (https://en.cppreference.com/w/cpp/container/vector/shrink_to_fit)

In practice this can reduce runtime memory usage by 10-15 MB

### DIFF
--- a/core/src/gl/texture.cpp
+++ b/core/src/gl/texture.cpp
@@ -213,6 +213,8 @@ void Texture::update(RenderState& rs, GLuint _textureUnit) {
     update(rs, _textureUnit, data);
 
     m_data.clear();
+    // Free the allocated buffer by swapping with an empty vector.
+    std::vector<GLuint>().swap(m_data);
 }
 
 void Texture::update(RenderState& rs, GLuint _textureUnit, const GLuint* data) {


tallytalwar:
Another recommended approach is to swap the vector with an empty vector.

tallytalwar:
We are doing the swap with empty vector at other places in the code.

matteblair:
Yeah the swap-with-empty-vector trick is the only way I know to _definitely_ free the buffer. I tried both and there doesn't seem to be a difference in behavior (at least on iOS), so I'll go with "ask nicely" for now :)

tallytalwar:
Even though I know we are doing swapping with empty vector at a few places, maybe we can do more and might be missing a few.

matteblair:
Looks like the swap trick does produce better assembly (https://godbolt.org/z/EPxR7O). I'll update.

tallytalwar:
Rebased, merging when circle is done.

PR #1888, opened 2018-09-28
Cleanup old camera setters and getters

tallytalwar:
removed setPosition/getPosition from core code, osx/linux demo apps, rpi code and ios sdks


tallytalwar:
Closing this as off now. It might be still good to keep these around, at least for the next release cycle.

PR #1890, opened 2018-10-02
Configure demo app Info plist with Nextzen API key from env

matteblair:
This resolves two issues with the current iOS demo build process:
 - Since the Nextzen API key used for tiles is set as an environment variable for the app, it is missing when run from outside of Xcode (e.g. by clicking on the app icon on the phone). Therefore, the demo app aborts when run from outside of Xcode.
 - All builds of the demo app were using a single API key, since it was unintentionally included in the demo build schemes XD

CMake provides a convenient method to "configure" files: it replaces any `${VARS}` with their current value in the CMake script and copies the result to the output location. This makes is easy to grab the user's `NEXTZEN_API_KEY` from their shell environment and bake it into the Info.plist used by the demo app.


tallytalwar:
Rebased, will merge when circle has done its job.

PR #1894, opened 2018-10-08
Parse StyleParam values directly from YAML nodes

matteblair:
Previously, style parameters were parsed from the scene file by first serializing the YAML node of the parameter into a string, then passing that string to StyleParam to parse it. This resulted in a lot of cumbersome and fragile string parsing.

This change enables StyleParam to parse values directly from YAML nodes, removing the need to encode and decode YAML structures into strings and allowing re-use of YAML conversion utilities.

### DIFF
--- a/core/src/scene/stops.cpp
+++ b/core/src/scene/stops.cpp
@@ -93,7 +93,10 @@ auto Stops::Sizes(const YAML::Node& _node, UnitSet _units) -> Stops {
         if (StyleParam::parseSizeUnitPair(_frameNode.Scalar(), _result)) {
             if ( !_units.contains(_result.unit) ) {
                 LOGW("Size StyleParam can only take in pixel, %% or auto values in: %s", Dump(_node).c_str());
-                return false;
+                _result.unit = Unit::pixel;
+            }
+            if (_result.unit == Unit::none) {
+                _result.unit = Unit::pixel;
             }
         } else {
             LOGW("could not parse node %s\n", Dump(_frameNode).c_str());
--- a/core/src/scene/stops.cpp
+++ b/core/src/scene/stops.cpp
@@ -163,7 +163,7 @@ auto Stops::Offsets(const YAML::Node& _node, UnitSet _units) -> Stops {
             for (const auto& sequenceNode : frameNode[1]) {
                 StyleParam::ValueUnitPair width;
                 width.unit = Unit::pixel; // default to pixel
-                if (StyleParam::parseValueUnitPair(sequenceNode.Scalar(), 0, width)) {
+                if (sequenceNode.IsScalar() && StyleParam::parseValueUnitPair(sequenceNode.Scalar(), width)) {
                     widths.push_back(width);
                     if (lastUnit != width.unit) {
                         LOGW("Mixed units not allowed for stop values", Dump(frameNode[1]).c_str());
--- a/core/src/scene/stops.cpp
+++ b/core/src/scene/stops.cpp
@@ -76,7 +76,7 @@ auto Stops::FontSize(const YAML::Node& _node) -> Stops {
     return stops;
 }
 
-auto Stops::Sizes(const YAML::Node& _node, uint8_t _units) -> Stops {
+auto Stops::Sizes(const YAML::Node& _node, UnitSet _units) -> Stops {
     Stops stops;
 
     // mixed dim stops not allowed for sizes (except when 1D stop uses %)
@@ -91,7 +91,7 @@ auto Stops::Sizes(const YAML::Node& _node, uint8_t _units) -> Stops {
 
     auto constructFrame = [&](const auto& _frameNode, StyleParam::ValueUnitPair& _result) -> bool {
         if (StyleParam::parseSizeUnitPair(_frameNode.Scalar(), 0, _result)) {
-            if ( !(_units & _result.unit) ) {
+            if ( !_units.contains(_result.unit) ) {
                 LOGW("Size StyleParam can only take in pixel, %% or auto values in: %s", Dump(_node).c_str());
                 return false;
             }
@@ -140,7 +140,7 @@ auto Stops::Sizes(const YAML::Node& _node, uint8_t _units) -> Stops {
     return stops;
 }
 
-auto Stops::Offsets(const YAML::Node& _node, uint8_t _units) -> Stops {
+auto Stops::Offsets(const YAML::Node& _node, UnitSet _units) -> Stops {
     Stops stops;
     if (!_node.IsSequence()) {
         return stops;
@@ -174,8 +174,7 @@ auto Stops::Offsets(const YAML::Node& _node, uint8_t _units) -> Stops {
                 }
             }
             if (widths.size() == 2) {
-                if ( !(widths[0].unit & _units ) &&
-                        !(widths[1].unit & _units) ) {
+                if ( !_units.contains(widths[0].unit) && !_units.contains(widths[1].unit) ) {
                     LOGW("Non-pixel unit not allowed for multidimensionnal stop values");
                 }
                 stops.frames.emplace_back(key, glm::vec2(widths[0].value, widths[1].value));
@@ -186,7 +185,7 @@ auto Stops::Offsets(const YAML::Node& _node, uint8_t _units) -> Stops {
     return stops;
 }
 
-auto Stops::Widths(const YAML::Node& _node, const MapProjection& _projection, uint8_t _units) -> Stops {
+auto Stops::Widths(const YAML::Node& _node, const MapProjection& _projection, UnitSet _units) -> Stops {
     Stops stops;
     if (!_node.IsSequence()) { return stops; }
 
@@ -212,7 +211,7 @@ auto Stops::Widths(const YAML::Node& _node, const MapProjection& _projection, ui
 
         if (StyleParam::parseValueUnitPair(frameNode[1].Scalar(), start, width)) {
 
-            if (! (width.unit & _units) ) {
+            if (! _units.contains(width.unit) ) {
                 LOGW("Invalid unit is being used for stop %s", Dump(frameNode[1]).c_str());
             }
 
--- a/core/src/scene/stops.cpp
+++ b/core/src/scene/stops.cpp
@@ -90,7 +90,7 @@ auto Stops::Sizes(const YAML::Node& _node, UnitSet _units) -> Stops {
     float lastKey = 0;
 
     auto constructFrame = [&](const auto& _frameNode, StyleParam::ValueUnitPair& _result) -> bool {
-        if (StyleParam::parseSizeUnitPair(_frameNode.Scalar(), 0, _result)) {
+        if (StyleParam::parseSizeUnitPair(_frameNode.Scalar(), _result)) {
             if ( !_units.contains(_result.unit) ) {
                 LOGW("Size StyleParam can only take in pixel, %% or auto values in: %s", Dump(_node).c_str());
                 return false;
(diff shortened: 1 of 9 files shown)


core/src/scene/styleParam.cpp
```diff
@@ -592,106 +545,125 @@ bool StyleParam::parseTime(const std::string &_value, float &_time) {
...
-        StyleParam::ValueUnitPair v;
-        offset = StyleParam::parseValueUnitPair(_value, offset, v);
-        if (offset < 0) { return i; }
-
-        if ( !(v.unit & allowedUnits) ) {
-            return 0;
-        }
-        _vec.units[i] = v.unit;
-        _vec.value[i] = v.value;
+    if (node.IsSequence() && node.size() >= 2) {
```
tallytalwar: it looks like you have it parsing for just the first 2 units if size is a sequence. Maybe it makes sense to also warn if size sequence is greater than 2 and still allow the 1st 2 values to be parsed for size.

core/src/scene/styleParam.cpp
```diff
@@ -592,106 +545,125 @@ bool StyleParam::parseTime(const std::string &_value, float &_time) {
...
+            success = false;
+            break;
+        }
+        const auto& nodeElement = node[i];
+        if (!nodeElement.IsScalar()) {
+            success = false;
+            continue;
+        }
+        StyleParam::ValueUnitPair result;
+        success |= StyleParam::parseValueUnitPair(nodeElement.Scalar(), result);
```
tallytalwar: I am not sure if this logic is correct. If for one of the values in the sequence, `parseValueUnitPair` returns false, `success` being set to `true` above, will not be aptly set to `false`. Also probably it makes sense to `continue` if `parseValueUnitPair` for a value in the sequence returns false.

core/src/scene/stops.cpp
```diff
@@ -90,10 +90,13 @@ auto Stops::Sizes(const YAML::Node& _node, uint8_t _units) -> Stops {
     float lastKey = 0;
 
     auto constructFrame = [&](const auto& _frameNode, StyleParam::ValueUnitPair& _result) -> bool {
-        if (StyleParam::parseSizeUnitPair(_frameNode.Scalar(), 0, _result)) {
-            if ( !(_units & _result.unit) ) {
+        if (StyleParam::parseSizeUnitPair(_frameNode.Scalar(), _result)) {
+            if ( !_units.contains(_result.unit) ) {
                 LOGW("Size StyleParam can only take in pixel, %% or auto values in: %s", Dump(_node).c_str());
-                return false;
+                _result.unit = Unit::pixel;
```
tallytalwar: Does it make sense to correct an erroneous unit value provided by the user? Maybe just have the warning like before.

core/src/scene/styleParam.cpp
```diff
@@ -592,106 +545,125 @@ bool StyleParam::parseTime(const std::string &_value, float &_time) {
...
-        if ( !(v.unit & allowedUnits) ) {
-            return 0;
-        }
-        _vec.units[i] = v.unit;
-        _vec.value[i] = v.value;
+    if (node.IsSequence() && node.size() >= 2) {
+        const auto& nodeFirst = node[0];
+        const auto& nodeSecond = node[1];
+        success |= nodeFirst.IsScalar() && parseSizeUnitPair(nodeFirst.Scalar(), result.x);
+        success |= nodeSecond.IsScalar() && parseSizeUnitPair(nodeSecond.Scalar(), result.y);
```
matteblair: Think these should also be `&=`, and `success` should start as `true`.

core/src/scene/styleParam.cpp
matteblair: Sure, could warn if there are more than 2 elements in the sequence.

core/src/util/extrude.cpp
```diff
@@ -1,61 +1,38 @@
...
-    // Get a float if possible & advance pos to the end of the number
-    const int length = _str.length();
-    int count = 0;
-    int offset = 0;
-
-    first = ff::stof(str, length, &count);
-    if (count == 0) {
-        // No numbers found, return zero extrusion
-        return Extrude(0, 0);
+    if (node.IsSequence() && node.size() >= 2) {
```
tallytalwar: Good to warn if sequence size is >2.

core/src/util/extrude.cpp
```diff
@@ -1,61 +1,38 @@
...
-    }
-    while (str[offset] == ' ') { offset++; }
-    if (str[offset++] != ',') {
-        return Extrude(0, first);
-    }
-
-    // Get a float if possible & advance pos to the end of the number
-    second = ff::stof(str + offset, length - offset, &count);
-    if (count == 0) {
+    if (YamlUtil::getFloat(node, first)) {
```
tallytalwar: Will be simpler to just check if `first` is > 0, and extrude from 0 to `first`.

core/src/scene/styleParam.cpp
```diff
@@ -152,62 +120,70 @@ StyleParamKey StyleParam::getKey(const std::string& _key) {
...
-        if (_value == "false") {
-            return std::numeric_limits<uint32_t>::max();
-        }
-        if (_value == "true") {
-            return uint32_t(15); // DEFAULT
-        }
-        if (parseInt(_value, textWrap) > 0 && textWrap > 0) {
-             return static_cast<uint32_t>(textWrap);
+        bool useWrap = false;
+        int wrapLength = 15; // Default wrap length.
```
tallytalwar: Can you add a const or a define for this?

core/src/scene/styleParam.cpp
```diff
@@ -152,62 +120,70 @@ StyleParamKey StyleParam::getKey(const std::string& _key) {
...
-        return vec.value;
+        LOGW("Invalid offset parameter '%s'.", Dump(node).c_str());
+        return Value();
     }
     case StyleParamKey::text_buffer:
     case StyleParamKey::buffer: {
-        UnitVec<glm::vec2> vec;
-        if (!parseVec2(_value, allowedUnits, vec)) {
-            LOGW("Invalid buffer parameter '%s'.", _value.c_str());
+        ValueUnitPair scalar;
```
tallytalwar: This block of close for `text_buffer`/`buffer` is same as `text_offset`/`offset`. Maybe just club these together.

core/src/scene/styleParam.cpp
```diff
@@ -218,36 +194,42 @@ StyleParam::Value StyleParam::parseString(StyleParamKey key, const std::string&
...
-                anchors.anchor[anchors.count++] = labelAnchor;
-            } else {
-                LOG("Invalid anchor %s", anchor.c_str());
+        LabelProperty::Anchors anchorSet;
+        LabelProperty::Anchor anchor;
+        if (node.IsScalar() && LabelProperty::anchor(node.Scalar(), anchor)) {
+            anchorSet.anchor[anchorSet.count++] = anchor;
+        } else if (node.IsSequence()) {
+            for (const auto& subNode : node) {
+                if (anchorSet.count >= LabelProperty::max_anchors) { break; }
```
tallytalwar: `anchorSet.count` can never become > `LabelProperty::max_anchors`. Equality check should suffice here.

core/src/scene/styleParam.cpp
```diff
@@ -294,11 +282,14 @@ StyleParam::Value StyleParam::parseString(StyleParamKey key, const std::string&
     case StyleParamKey::outline_visible:
     case StyleParamKey::collide:
     case StyleParamKey::text_optional:
-    case StyleParamKey::text_collide:
-        if (_value == "true") { return true; }
-        if (_value == "false") { return false; }
-        LOGW("Invalid boolean value %s for key %s", _value.c_str(), StyleParam::keyName(key).c_str());
+    case StyleParamKey::text_collide: {
+        bool result = false;
+        if (YAML::convert<bool>::decode(node, result)) {
```
tallytalwar: I vaguely remember discussion around, `true`/`false` and `yes`/`no` and other boolean yaml values.

Can you confirm if `YAML::convert<bool>::decode(node, result)` will only handle `true`/`false` values?

core/src/scene/styleParam.cpp
```diff
@@ -307,105 +298,115 @@ StyleParam::Value StyleParam::parseString(StyleParamKey key, const std::string&
...
+        if (node.IsScalar() && parseValueUnitPair(node.Scalar(), result)) {
+            if (result.unit == Unit::none) {
+                result.unit = Unit::pixel;
             }
         }
-
-        return Width(placementSpacing);
+        if (result.unit != Unit::pixel) {
+            LOGW("Invalid placement spacing value '%s'", Dump(node).c_str());
+            result.value = 80.f;
```
tallytalwar: I think I initially added the `80` default value here. Should have added a const or a define for it. Can you take care of it here?

core/src/scene/stops.cpp
matteblair: Tangram JS seems to use the default unit if the suffix is not recognized. I'll keep this the same.

core/src/util/extrude.cpp
matteblair: Don't understand this comment - can you be more specific?

tallytalwar:
```
if (first > 0) {
    return Extrude(0, first);
}
```

Instead of getting `float` value from the yaml node again, already have the `first` value parsed from the yaml node above on line 23.

matteblair: The block before this is for cases where `node` is a sequence, but this block is for scalars. They have to be accessed differently.

tallytalwar: Ohh yeah, missed that. All good.

core/src/scene/styleParam.cpp
matteblair: It accepts only `true`, `True`, `TRUE`, `false`, `False`, and `FALSE`. Will update to use the `YamlUtil` version of this, for consistency.

core/src/scene/styleParam.cpp
matteblair: Combined these blocks, seems to work just fine.

core/src/scene/styleParam.cpp
matteblair: Just being paranoid, I don't think there's a downside to using `>=`.

core/src/scene/styleParam.cpp
```diff
@@ -307,105 +298,115 @@ StyleParam::Value StyleParam::parseString(StyleParamKey key, const std::string&
...
-                LOGW("Invalid unit provided for placement spacing");
+        ValueUnitPair result;
+        if (node.IsScalar() && parseValueUnitPair(node.Scalar(), result)) {
+            if (result.unit == Unit::none) {
+                result.unit = Unit::pixel;
             }
         }
-
-        return Width(placementSpacing);
+        if (result.unit != Unit::pixel) {
```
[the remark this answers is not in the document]
matteblair: Same as above, to match Tangram JS behavior we'll use the default unit if we don't recognize the suffix.

core/src/scene/styleParam.cpp
matteblair: On second thought, I'd rather err on the side of simplicity here and not warn for something harmless like this.

core/src/util/extrude.cpp
matteblair: Again, erring on the side of simplicity and not warning here.

core/src/scene/styleParam.cpp
tallytalwar: @matteblair can you also add couple of unit tests to cover this logic. These things can be nasty to debug if broken, which unit tests can help avoid.

matteblair:
Added some basic tests, looks like it needs another approval now.

Issue #1896, opened 2018-10-10
Marker.setBitmap seems to be very slow


westnordost:
So I have been looking into the cold start performance of my app with the Android profiler and noticed that a big portion seems to be due to Tangram-ES:

![performance](https://user-images.githubusercontent.com/4661658/46744235-04fd8e80-ccab-11e8-8c6a-8ffe0985f6e8.png)
**How to read this chart?** It's a timeline, the first row is the seconds that passed, the Y-axis drills down on which method was on the call stack at that time. Of course, measuring the performance this way somewhat slows down the app, so it wouldn't really be 5.5 seconds just to set two marker bitmaps.

**What bitmap is set?**
direction marker (96x96 dp): https://github.com/westnordost/StreetComplete/blob/master/app/src/main/res/drawable-xxhdpi/location_direction.png
accuracy marker (256x256 dp): https://github.com/westnordost/StreetComplete/blob/master/app/src/main/res/drawable/accuracy_circle.xml

**How to measure start-up performance?** Set a breakpoint on the first line of the `onCreate` method of your application class, start the app in debug, wait for the breakproint to trigger, go to Android Studio's Profiler tab, click on CPU, click on Record, go back to the Debug tab and continue. Then, in the record, you can move and zoom in/out on what method executed when and how long.

**Device**: Sony Xperia Z1 Compact
**Android Version**: 5.1.1
**Tangram Version**: 0.9.6

westnordost:
Also, according to the profiling record, note that the time tangram-ES spends off the main thread to asynchronously initialize the map (`getMapAsync`) seems to be smaller than the time `getMapAsync` needs on the main thread.

westnordost:
I am not sure if the Android profiler also shows functions in native code, my guess is not. However, already the Java code could be improved. For a 256x256 dp bitmap, I see already around 589824 unnecessary divisions and 589824 unnecessary modulo operations for a screen with a high screen density (xxhdpi = x3).

matteblair:
We could just move these Bitmap calculations into native code - should be noticeably faster in c++ where we can skip array bounds checks and such.

Or we could just use `GLUtils.texImage2D` directly on the Bitmap https://developer.android.com/reference/android/opengl/GLUtils.html#texImage2D(int,%20int,%20android.graphics.Bitmap,%20int)

westnordost:
[...]

PR #1897, opened 2018-10-10
Performance improvement on setting marker bitmap

westnordost:
Replaced divisions and modulos with two counters. See #1896

### DIFF
--- a/platforms/android/tangram/src/main/java/com/mapzen/tangram/Marker.java
+++ b/platforms/android/tangram/src/main/java/com/mapzen/tangram/Marker.java
@@ -219,16 +219,18 @@ public class Marker {
         bitmap.getPixels(argb, 0, width, 0, 0, width, height);
 
         final int[] abgr = new int[width * height];
-        int row, col;
+        int row = 0, col = 0;
         for (int i = 0; i < argb.length; i++) {
-            col = i % width;
-            row = i / width;
             final int pix = argb[i];
             final int pb = (pix >> 16) & 0xff;
             final int pr = (pix << 16) & 0x00ff0000;
             final int pix1 = (pix & 0xff00ff00) | pr | pb;
             final int flippedIndex = (height - 1 - row) * width + col;
             abgr[flippedIndex] = pix1;
+            if(++col >= width) {
+                ++row;
+                col = 0;
+            }
         }
 
         return map.setMarkerBitmap(markerId, width, height, abgr);


westnordost:
I did not test it. I assume it will have *some* impact, but I do not know how much.

platforms/android/tangram/src/main/java/com/mapzen/tangram/Marker.java
```diff
@@ -219,16 +219,18 @@ private boolean setBitmap(@NonNull final Bitmap bitmap) {
         bitmap.getPixels(argb, 0, width, 0, 0, width, height);
 
         final int[] abgr = new int[width * height];
-        int row, col;
+        int row = 0, col = 0;
         for (int i = 0; i < argb.length; i++) {
-            col = i % width;
-            row = i / width;
             final int pix = argb[i];
```
karimnaaji: You can try to use `bitmap.getPixel` in order to remove initial alloc on line `218`, and maybe `bitmap.flip(1f, -1f, pixel_x, pixel_y)` to remove index flipping.

westnordost: Hmm, there is no mention of `bitmap.flip` in the [Android documentation for Bitmap](https://developer.android.com/reference/android/graphics/Bitmap).

matteblair: Can you elaborate? Not sure what these methods are

karimnaaji: I thought `flip` was a thing but it's not you're right @westnordost , for the first one it's https://developer.android.com/reference/android/graphics/Bitmap.html#getPixel(int,%20int).

rwrx: I have tested this change on around 180 markers and drawing them with this change took around 35 seconds and original code took around 30 seconds.

westnordost: I claim there is something odd with your test setup. I haven't seen any CPU architecture where an integer-division is faster than doing an increment.

Did you execute both in debug mode? Is the project on Github?

westnordost: I wrote a test myself now. The new code is about 16-22% faster.

See measuring code: https://gist.github.com/westnordost/dc4d8cf93a70c0a68853f7c2a545a8a4

Much less than I expected, to be honest. So even with this merged, the performance issue observed in #1896 would still be visible somewhat. I think @matteblair 's idea to either move this "flipping"+"convert argb to abgr" to the native code or feed OpenGL the argb directly (if I understood correctly) is better. However, I am not deep enough in this project to do this, I can only offer this PR. Even if you plan to move it to native code some time later, it still can't hurt to merge this here, for the meantime and since it is already a small improvement.

rwrx: I am sorry, I have done something wrong in the test. Unfortunately I cannot show actual code but basically it is showing markers 180 on map and setting drawable to marker by calling `Marker.setDrawable()` which is calling `Marker.SetBitmap()`. I have rerun test several times and now I am getting same results for both versions - 30 seconds. So I think that in real-world use case this change has negligible performance advantage. Maybe it has something to do with Android JVM optimizations for repeated instructions, JIT, caching. I am not sure, but I have these results.

matteblair:
See https://github.com/tangrams/tangram-es/pull/1907

westnordost:
Ok cool, then this is obsolete :-) 

Note though that it looks like that for loop is not the thing that takes the most time.

See https://github.com/tangrams/tangram-es/issues/1896#issuecomment-429534423
Perhaps @rwrx would like to compare this PR with the current code in his test setup.

rwrx:
Hi @westnordost I have tried the new code in PR, but it seems that calling `AndroidBitmap_lockPixels` inside `Java_com_mapzen_tangram_MapController_nativeMarkerSetBitmap` always return 0 so no Marker is drawn on the map..

PR #1898, opened 2018-10-10
Refactor screenPositionToLngLat into View and validate matrices

matteblair:
Moved `screenPositionToLngLat` logic into `View`, renamed methods for consistency, and fixed an issue where calls to `screenPositionToLngLat` could use out-of-date view matrices.

### DIFF
--- a/core/src/view/view.cpp
+++ b/core/src/view/view.cpp
@@ -457,14 +457,29 @@ void View::updateMatrices() {
 
 }
 
-glm::vec2 View::lonLatToScreenPosition(double lon, double lat, bool& clipped) const {
-    glm::dvec2 absoluteMeters = m_projection->LonLatToMeters({lon, lat});
+glm::vec2 View::lngLatToScreenPosition(double lng, double lat, bool& clipped) {
+
+    if (m_dirtyMatrices) { updateMatrices(); } // Need the view matrices to be up-to-date
+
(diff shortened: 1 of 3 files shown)


core/src/view/view.cpp
```diff
@@ -467,6 +470,18 @@ glm::vec2 View::lonLatToScreenPosition(double lon, double lat, bool& clipped) co
...
 
+LngLat View::screenPositionToLngLat(float x, float y, bool& _intersection) {
+
+    if (m_dirtyMatrices) { updateMatrices(); } // Need the view matrices to be up-to-date
+
+    double dx = x, dy = y;
+    double distance = screenToGroundPlaneInternal(dx, dy);
+    _intersection = (distance >= 0);
+    glm::dvec2 meters(dx + m_pos.x, dy + m_pos.y);
+    glm::dvec2 lngLat = m_projection->MetersToLonLat(meters);
```
tallytalwar: MetersToLngLat for consistency, since you are already on it :D.
Yes, Projection headers will need to change :|

matteblair: Already working on obliterating the MapProjection class in another branch - I'll save it for that ;)

tallytalwar: Don't think anyone touched projection classes since I added it in 2014, very beginning of tangram-es :D

PR #1905, opened 2018-10-19
Update iOS gesture recognizers to more closely match Apple Maps and others

msmollin:
Two small changes here for the recognizer setup on iOS.

First, the current iOS gesture recognizers are a little strict in how panning is handled versus how users tend to interact with maps. The general expectation is to be able to pinch/zoom AND pan at the same time, which allows the user to alter the viewport slightly but generally feels much better than the current setup. We still want the shove gesture recognizer to be exclusive though, which matches how apple maps works.

For developers, I moved `UIGestureRecognizerDelegate` conformance into the public header so that the compiler will allow implementing devs to subclass `TGMapView` and override `shouldRecognizeSimultaneouslyWithGestureRecognizer:` functionality. This allows devs to call the `super` implementation, versus having the conformance privately declared prohibits this entirely in Swift, and is frustrating to work around in objective-c.

### DIFF
--- a/platforms/ios/framework/src/TGMapView.mm
+++ b/platforms/ios/framework/src/TGMapView.mm
@@ -38,7 +38,7 @@ inline CLLocationDirection convertRotationRadiansToBearingDegrees(float rotation
     return TGDegreesFromRadians(-rotation);
 }
 
-@interface TGMapView () <UIGestureRecognizerDelegate, GLKViewDelegate> {
+@interface TGMapView () <GLKViewDelegate> {
     BOOL _shouldCaptureFrame;
     BOOL _captureFrameWaitForViewComplete;
     BOOL _viewComplete;
@@ -245,7 +245,7 @@ inline CLLocationDirection convertRotationRadiansToBearingDegrees(float rotation
     [_tapGestureRecognizer requireGestureRecognizerToFail:_doubleTapGestureRecognizer];
 
     _panGestureRecognizer = [[UIPanGestureRecognizer alloc] initWithTarget:self action:@selector(respondToPanGesture:)];
-    _panGestureRecognizer.maximumNumberOfTouches = 1;
+    _panGestureRecognizer.maximumNumberOfTouches = 2;
 
     _pinchGestureRecognizer = [[UIPinchGestureRecognizer alloc] initWithTarget:self action:@selector(respondToPinchGesture:)];
     _rotationGestureRecognizer = [[UIRotationGestureRecognizer alloc] initWithTarget:self action:@selector(respondToRotationGesture:)];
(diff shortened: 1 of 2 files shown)


platforms/ios/framework/src/TGMapView.mm
```diff
@@ -245,7 +245,7 @@ - (void)setupGestureRecognizers
     [_tapGestureRecognizer requireGestureRecognizerToFail:_doubleTapGestureRecognizer];
 
     _panGestureRecognizer = [[UIPanGestureRecognizer alloc] initWithTarget:self action:@selector(respondToPanGesture:)];
-    _panGestureRecognizer.maximumNumberOfTouches = 1;
+    _panGestureRecognizer.maximumNumberOfTouches = 2;
```
matteblair: Seems like Apple Maps and Google Maps on iOS don't even set a `maximumNumberOfTouches` for panning - that or the maximum is greater than 5 fingers

matteblair:
Seems to be working for the case of panning/scaling/rotating with additional touches after an initial single touch - which feels really smooth and nice. However if I start with two touches then only the tilt gesture is recognized, no matter which way the touches move.

I wonder if we can use some additional criteria for simultaneous recognition that takes into account the vertical motion required for a tilt gesture.

msmollin:
Thats true, but also the way Apple & Google Maps work on iOS. Is there something else you think we should be doing?

msmollin:
And, in my testing, if you start out with two fingers and start a pinch gesture, then everything works correctly.

matteblair:
Try putting down two or more touches at the same time and moving them horizontally. On Apple/Google Maps this pans the map. On Tangram in this branch it doesn't pan but seems to start a tilt gesture with zero tilt.

matteblair:
It seems like the thing I'm describing is maybe very difficult and possibly requires a customized gesture recognizer. The change here is definitely an improvement over the current behavior, so let's go ahead with it.

platforms/ios/framework/src/TGMapView.mm
```diff
@@ -962,11 +962,8 @@ - (void)touchesBegan:(NSSet *)touches withEvent:(UIEvent *)event {
 - (BOOL)gestureRecognizer:(UIGestureRecognizer *)gestureRecognizer shouldRecognizeSimultaneouslyWithGestureRecognizer:(UIGestureRecognizer *)otherGestureRecognizer
 {
     // make shove gesture exclusive
-    if ([gestureRecognizer isKindOfClass:[UIPanGestureRecognizer class]]) {
-        return [gestureRecognizer numberOfTouches] != 2;
-    }
-    if ([otherGestureRecognizer isKindOfClass:[UIPanGestureRecognizer class]]) {
-        return [otherGestureRecognizer numberOfTouches] != 2;
+    if ([gestureRecognizer isEqual:_shoveGestureRecognizer] || [otherGestureRecognizer isEqual:_shoveGestureRecognizer]) {
```
matteblair: `isEqual` seems to work but wouldn't pointer comparison (`==`) do what we want here as well?

msmollin: isEqual: defaults to pointer comparison from `NSObject` when not implemented IIRC, which is why I used it here.

PR #1907, opened 2018-10-19
Move Android Marker Bitmap conversion into native code

matteblair:
This should reduce some of the CPU cost of applying Bitmaps to Markers on Android.

### DIFF
--- a/platforms/android/tangram/src/main/java/com/mapzen/tangram/Marker.java
+++ b/platforms/android/tangram/src/main/java/com/mapzen/tangram/Marker.java
@@ -214,10 +214,7 @@ public class Marker {
         final int density = context.getResources().getDisplayMetrics().densityDpi;
         final int width = bitmap.getScaledWidth(density);
         final int height = bitmap.getScaledHeight(density);
-        final int[] argb = new int[width * height];
-        bitmap.getPixels(argb, 0, width, 0, 0, width, height);
-
-        return map.setMarkerBitmap(markerId, width, height, argb);
+        return map.setMarkerBitmap(markerId, width, height, bitmap);
     }
 
     void invalidate() {
--- a/platforms/android/tangram/src/main/java/com/mapzen/tangram/Marker.java
+++ b/platforms/android/tangram/src/main/java/com/mapzen/tangram/Marker.java
@@ -91,7 +91,7 @@ public class Marker {
      * <li>{ style: 'polygons', color: '#06a6d4', width: 5px, order: 2000 }</li>
      * </ul>
      *
-     * @param styleStr the style string
+     * @param styleString the style string
      * @return whether the style was successfully set
      */
     public boolean setStylingFromString(final String styleString) {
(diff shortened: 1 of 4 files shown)


matteblair:
I only did simple testing of this with:
```diff
diff --git a/platforms/android/demo/src/main/java/com/mapzen/tangram/android/MainActivity.java b/platforms/android/demo/src/main/java/com/mapzen/tangram/android/MainActivity.java
index 6be68fd2..25cf50eb 100644
--- a/platforms/android/demo/src/main/java/com/mapzen/tangram/android/MainActivity.java
+++ b/platforms/android/demo/src/main/java/com/mapzen/tangram/android/MainActivity.java
@@ -235,7 +235,8 @@ public class MainActivity extends AppCompatActivity implements MapController.Sce
             markers.addPolyline(line, props);
 
             Marker p = map.addMarker();
-            p.setStylingFromPath(pointStylingPath);
+            p.setStylingFromString("{ style: 'points', color: 'white', size: [50px, 50px], order: 2000, collide: false }");
+            p.setDrawable(R.drawable.ic_launcher);
             p.setPoint(tappedPoint);
             pointMarkers.add(p);
         }
```

platforms/android/tangram/src/main/java/com/mapzen/tangram/Marker.java
```diff
@@ -214,24 +214,10 @@ private boolean setBitmap(@NonNull final Bitmap bitmap) {
         final int density = context.getResources().getDisplayMetrics().densityDpi;
         final int width = bitmap.getScaledWidth(density);
         final int height = bitmap.getScaledHeight(density);
-
         final int[] argb = new int[width * height];
```
karimnaaji: It's also possible access the data in place in native code a9b2954f867a46348b6f1c014c871877365b5cac.

matteblair:
@karimnaaji Nice find, the JNI Bitmap functions should be much more efficient. 

One catch is that the Bitmap colors from the NDK method are alpha premultiplied, so we have to undo that for each pixel as well (there are options to create the Bitmap in Java without alpha premultiplied, but they aren't available in API 16 which is our current min version). 

It also seems like the pixels could be returned in formats other than RGBA8888 but I haven't observed this yet in my (brief) testing. If that happens during normal use then we'll have to add code to perform the appropriate conversion or go back to the old getPixels method.

tallytalwar:
While having a chat with @karimnaaji yesterday, he mentioned it might be further beneficial to provide a cache for reusing bitmaps and save of the conversion process. He noticed this while playing around with StreetComplete and most of the marker drawables are the same, and we can totally cache aptly converted bitmap data and get some more perf kicks.

A possible approach might be to map drawableId to bitmap, and in native code, bitmap to `pixelOutput`.

matteblair:
Yeah I mean the real performance killer here is that Tangram just blindly re-creates the bitmap data and a GL texture each time you set the bitmap on a marker, even if you've set it on another marker before. Saving and re-using the bitmaps is the right thing to do, but it's more involved. I made the PR because this change is easy and pretty safe, even though it isn't perfect.

tallytalwar:
Totally, we should definitely work on further optimizing this. But this is good and thanks to @rwrx testing we see it gets some meaty (3x) jump.

tallytalwar:
Going to mark it approved from my end. @hjanetzek mentioned he has a commit saving on couple of multiplications. Will wait for him to push and merge then.

tallytalwar:
@hjanetzek can you bring EGL config commits to another PR? Good to isolate these since these tasks.
Once done I am going to merge this PR.

hjanetzek:
@tallytalwar sorry, the last push to this branch was not intended. git-gui missed that I switched branches outside and pushed these commits.. cherry picked to.. oh already merged :)

PR #1913, opened 2018-10-26
Android view holder abstraction

tallytalwar:
PR defines a ViewHolder interface, which is used by the MapController to hold either GLSurfaceView or a client provided SurfaceView implementation for map rendering purposes.

The client application has an option to either use GLSurfaceView or other SurfaceView implementations for rendering by specifying a specific GLViewHolderFactory: GLSurfaceViewHolderFactory or appropriate client implemented SurfaceViewHolderFactory.

GLSurfaceViewHolderFactory implement GLViewHolderFactory interface and are responsible to build GLSurfaceViewHolder, which holds a reference to GLSurfaceView. This can also be used as a reference implementation for clients to provide their own SurfaceView implementation.

GLSurfaceViewHolder is just a wrapper around GLSurfaceView and forwards most of the ViewHolder interface calls to the GLSurfaceView instance. 

However if client code wants to implement its own GLViewHolder implementation using a explicit SurfaceView, they will have to manage EGLContext and Render Thread behavior. Another helpful example explaining this can be found [here]( https://github.com/google/grafika/blob/master/app/src/main/java/com/android/grafika/TextureViewGLActivity.java).

### DIFF
--- a/platforms/android/tangram/src/main/java/com/mapzen/tangram/MapView.java
+++ b/platforms/android/tangram/src/main/java/com/mapzen/tangram/MapView.java
@@ -7,8 +7,12 @@ import android.os.Bundle;
 import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
 import android.util.AttributeSet;
+import android.view.MotionEvent;
 import android.widget.FrameLayout;
 
+import com.mapzen.tangram.viewholder.GLSurfaceViewHolderFactory;
+import com.mapzen.tangram.viewholder.GLViewHolder;
+import com.mapzen.tangram.viewholder.GLViewHolderFactory;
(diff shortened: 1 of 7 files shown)


platforms/android/tangram/src/main/java/com/mapzen/tangram/MapView.java
```diff
@@ -117,7 +117,7 @@ protected void onMapInitOnUIThread(@Nullable final MapController controller, @Nu
         }
 
         if (controller != null) {
-            viewHolder = createViewHolder(viewHolderFactory);
+            viewHolder = viewHolderFactory.build(getContext());
```
tallytalwar: 👍

PR #1915, opened 2018-10-30
Add extensible GUI to desktop demo app

matteblair:
This brings in the _Dear Imgui_ library as a dependency of the GLFW desktop demo app and uses it to render a custom GUI for controlling and visualizing aspects of Tangram. 

_Dear Imgui_, or just _Imgui_, is a richly featured, widely used, well loved library made for building GUIs for tools: https://github.com/ocornut/imgui _Imgui_ itself doesn't do any rendering on the GPU, but there already exist example implementations with OpenGL 3 and GLFW that were easily adaptable to our Tangram demo app. The integration of _Imgui_ here consists of the library's main source and header files, the implementation files (`imgui_impl_*`), some deletions from the OSX `main.mm`, and some changes in `glfwApp.cpp`.

Here is the current state of the custom GUI built for Tangram so far:
screen shot 2018-10-29 at 6 32 09 pm

What _Imgui_ lacks in beauty it makes up for in utility. The GUI is super easy to extend. For example, here's all the code required to create the "Reload Scene" button:
```cpp
if (ImGui::Button("Reload Scene")) {
    loadSceneFile();
}
```

_Imgui_ is a pretty hefty chunk of code to add, but it is only used in the desktop app so there will be no size or performance impacts on the mobile SDKs. I think _Imgui_ will more than pull its weight once we start building tools with it.


matteblair:
Seeing some intermittent CircleCI failures, possibly OOM errors. Will work on this in a separate PR.

PR #1919, opened 2018-10-30
Try to limit Ninja concurrency in Android CircleCI builds

matteblair:
https://cmake.org/cmake/help/v3.12/envvar/CMAKE_BUILD_PARALLEL_LEVEL.html#envvar:CMAKE_BUILD_PARALLEL_LEVEL


matteblair:
Doesn't seem to work as expected.

Issue #1925, opened 2018-11-05
Marker.setDrawable does not handle screen density.


polar:
1. Have a phone with a screen density of 3.374 (Motorola Moto4)
2. Create a Bitmap Drawable with a bitmap at the same target density (540Dpi).
    Say bitmap is 540x540 pixels. 
3. Marker.setDrawable(bitmap).

Expected Result:

Marker should be about an inch high and wide.

Actual Result:

Marker is significantly larger than the screen.

This is an issue especially for Text drawing functions. Scaling the drawn bitmap down to an acceptable size is ugly. I want to be able to draw a bitmap at the current screen resolution and get it displayed at that resolution.

I find this a problem with that `nativeSetMarkerBitmap()` does not take into account a density parameter, which would eventually be handed off to the `Texture` constructor. 

I solved this problem by adding a `float density` parameter to the following calls:

    JNIEXPORT jboolean JNICALL Java_com_mapzen_tangram_MapController_nativeMarkerSetBitmap(JNIEnv* jniEnv, jobject obj, jlong mapPtr, jlong markerID, jobject jbitmap, jfloat density)
    ---\calls/----
    bool Map::markerSetBitmap(MarkerID _marker, int _width, int _height, const unsigned int* _data, float _density)
    ---\calls/----
    bool MarkerManager::setBitmap(MarkerID markerID, int width, int height, const unsigned int* bitmapData, float density) {
         ...\calls/...
          auto texture = std::make_unique<Texture>(width, height, options, false, density)
         ...
    }
  
It is invoked by `Marker.java` using:

    public boolean setDrawable(final int drawableId) {
        DisplayMetrics metrics = context.getResources().getDisplayMetrics();
        BitmapFactory.Options options = new BitmapFactory.Options();
        options.inTargetDensity = metrics.densityDpi;
        Bitmap bitmap = BitmapFactory.decodeResource(context.getResources(), drawableId, options);
        return setBitmap(bitmap, metrics.density);
    }
    public boolean setDrawable(@NonNull final Drawable drawable) {
        DisplayMetrics metrics = context.getResources().getDisplayMetrics();

        float screenDensity = metrics.density;
        int density = context.getResources().getDisplayMetrics().densityDpi;
        BitmapDrawable bitmapDrawable = (BitmapDrawable) drawable;
        bitmapDrawable.setTargetDensity(density);
        Bitmap bitmap = bitmapDrawable.getBitmap();
        bitmap.setDensity(density);
[...]

PR #1927, opened 2018-11-07
Adding density parameter to Marker.setBitmap.

polar:
This pull requests addresses the following issue: https://github.com/tangrams/tangram-es/issues/1925 
Previously, bitmaps were not transferred to the Texture object with a density parameter, which lead to scaled up bitmaps on the map when the device density is higher than 1.

### DIFF
--- a/core/include/tangram/map.h
+++ b/core/include/tangram/map.h
@@ -324,7 +324,7 @@ public:
     // Set a bitmap to use as the image for a point marker; _data is a buffer of RGBA pixel data with
     // length of _width * _height; pixels are in row-major order beginning from the bottom-left of the
     // image; returns true if the marker ID was found and successfully updated, otherwise returns false.
-    bool markerSetBitmap(MarkerID _marker, int _width, int _height, const unsigned int* _data);
+    bool markerSetBitmap(MarkerID _marker, int _width, int _height, const unsigned int* _data, float density);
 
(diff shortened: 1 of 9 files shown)


polar:
I kind of new that this PR would break the iOS build. I haven't set up a Mac to try it out yet, and I'm not too familiar with the iPhone since a few years ago. It looks like TGMarker.mm needs to add a density parameter to `tangramInstance.setMarkerBitmap(...., density);`  
I am unclear of what `density` should be. Should it be `UIScreen.scale`?

matteblair:
I can check out the iOS implementation and see how to make that match. 

For the Android changes, I'm now setting up a repro case to verify that the changes fix what we expect. @polar How were you creating the BitmapDrawables that ended up with the wrong size?

polar:
@matteblair Yes, I was creating the BitmapDrawables.

polar:
I just tried it, and sadly the density is not fixed.

matteblair:
Hey sorry, I dropped the ball on this PR - didn't want to delay the already-late release :\

I'll take another run at this to figure out the root issue and I expect we'll be able to put the change in a patch release - it shouldn't be a breaking change.

matteblair:
Hey thanks again for finding the cause of this issue, the fix should be applied in https://github.com/tangrams/tangram-es/pull/1976, let me know if you see other problems!

PR #1928, opened 2018-11-07
Store UrlRequestHandle with TileTask

hjanetzek:
- Bugfix: Also cancel raster tiles

### DIFF
--- a/core/include/tangram/tile/tileTask.h
+++ b/core/include/tangram/tile/tileTask.h
@@ -123,8 +123,9 @@ public:
     std::shared_ptr<std::vector<char>> rawTileData;
 
     bool dataFromCache = false;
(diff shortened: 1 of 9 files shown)


core/include/tangram/tile/tileTask.h
```diff
@@ -122,6 +123,8 @@ class BinaryTileTask : public TileTask {
     std::shared_ptr<std::vector<char>> rawTileData;
 
     bool dataFromCache = false;
+
+    UrlRequestHandle urlRequestHandle = -1;
```
matteblair: I think it's better to keep this type opaque and not do logic on its value. We can simply add a boolean here to determine whether the request is active.

PR #1929, opened 2018-11-08
Fix tile cancellation

hjanetzek:


core/src/tile/tileManager.cpp
```diff
@@ -407,9 +410,13 @@ void TileManager::updateTileSet(TileSet& _tileSet, const ViewState& _view) {
             if (entry.getProxyCounter() > 0) {
                 if (entry.isReady()) {
                     m_tiles.push_back(entry.tile);
-                } else if (curTileId.z < maxZoom) {
-                    // Cancel loading
```
hjanetzek: This had no effect, since ...

core/src/tile/tileManager.cpp
```diff
@@ -423,13 +430,9 @@ void TileManager::updateTileSet(TileSet& _tileSet, const ViewState& _view) {
         auto it = tiles.find(removeTiles.back());
         removeTiles.pop_back();
 
-        if ((it != tiles.end()) &&
-            (!it->second.isVisible()) &&
-            (it->second.getProxyCounter() <= 0  ||
-             it->first.z >= maxZoom)) {
```
hjanetzek: ... this check will always be false for `(curTileId.z < maxZoom)` above

hjanetzek:
@tallytalwar tested the weak_ptr and it drops old tiletasks when a new scene got loaded -- so it probably also solves the ios RasterSource issue.

core/src/tile/tileManager.cpp
```diff
@@ -348,27 +349,29 @@ void TileManager::updateTileSet(TileSet& _tileSet, const ViewState& _view) {
...
+                auto sourceGeneration = entry.tile->sourceGeneration();
+                if (!entry.isInProgress() && (sourceGeneration < generation)) {
                     // Tile needs update - enqueue for loading
                     entry.task = _tileSet.source->createTask(visTileId);
                     enqueueTask(_tileSet, visTileId, _view);
                 }
             } else if (entry.needsLoading()) {
                 // Not yet available - enqueue for loading
+                if (!entry.task) {
+                    entry.task = _tileSet.source->createTask(visTileId);
```
tallytalwar: Can we put a comment here, mentioning how this is possible?
Basically, an entry's task could be null either when its ready (implying tile is loaded) or when its cleared (cancelled from loading and cleared, or removed from tileSet completely).

Here it seems to be because of https://github.com/tangrams/tangram-es/blob/4de407f92284649cadc4def9c5544c4a0ece0cb8/core/src/tile/tileManager.cpp#L418

core/src/tile/tileManager.cpp
```diff
@@ -407,9 +410,13 @@ void TileManager::updateTileSet(TileSet& _tileSet, const ViewState& _view) {
...
                 if (entry.isReady()) {
                     m_tiles.push_back(entry.tile);
-                } else if (curTileId.z < maxZoom) {
-                    // Cancel loading
-                    removeTiles.push_back(curTileId);
+                } else if (entry.isInProgress()) {
+                    if (curTileId.z >= maxZoom || curTileId.z <= minZoom) {
+                        // Cancel tile loading but keep tile entry for referencing
+                        // this tiles proxy tiles.
+                        _tileSet.source->cancelLoadingTile(*entry.task);
```
tallytalwar: Benefit of having a proxy tile is only when it is actually available to be rendered, since it is not in this `else if` branch, I think we can just cancel loading and remove this in progress proxyTile, in favor of spending resources for the actual visible tileset?

core/src/tile/tileManager.cpp
tallytalwar: If https://github.com/tangrams/tangram-es/pull/1929/files#r232113259 is acknowledged then I do not think we need to create a task here, since the entry will itself be removed.

core/src/tile/tileManager.cpp
tallytalwar: Ohh I think I remember this now, even though this tile is loading, this tile's proxy tiles will also be removed if we remove this tile, hence we can not remove this tile!

core/src/tile/tileManager.cpp
```diff
@@ -583,19 +586,18 @@ bool TileManager::updateProxyTile(TileSet& _tileSet, TileEntry& _tile,
...
 
                 if (entry.isReady()) {
                     m_tiles.push_back(entry.tile);
                 }
-
-                // Note: No need to check the cache: When the tile is in
-                // tileSet it would have already been fetched from cache
                 return true;
             }
+            // Note: No need to check the cache: When the tile is in
```
tallytalwar: Not necessary now! (Also rename "tile" to "tileEntry" here!)

A tileEntry could be sitting in the tileSet even without any task/tile, as per the change here: https://github.com/tangrams/tangram-es/blob/4de407f92284649cadc4def9c5544c4a0ece0cb8/core/src/tile/tileManager.cpp#L413-L420

Though we still should have the early return here, but update the comment.

core/src/data/rasterSource.cpp
```diff
@@ -28,12 +31,9 @@ class RasterTileTask : public BinaryTileTask {
     }
 
     void process(TileBuilder& _tileBuilder) override {
-
-        auto source = reinterpret_cast<RasterSource*>(m_source.get());
-
         if (!m_texture) {
             // Decode texture data
-            m_texture = source->createTexture(*rawTileData);
+            m_texture = rasterSource()->createTexture(*rawTileData);
```
tallytalwar: This could segfault here,  `rasterSource()` can return a `nullptr` if the original `m_source` has been destroyed.

hjanetzek: Good point!

core/src/data/tileSource.cpp
```diff
@@ -98,7 +98,7 @@ void TileSource::loadTileData(std::shared_ptr<TileTask> _task, TileTaskCb _cb) {
     }
 
     for (auto& subTask : _task->subTasks()) {
-        subTask->source().loadTileData(subTask, _cb);
+        subTask->source()->loadTileData(subTask, _cb);
```
tallytalwar: Same, check for validity of the source here. Though I think it might be unlikely that subtask source is null and main task's source is not, in which case this will never be called, maybe put an assert in that case?

core/src/tile/tileWorker.cpp
```diff
@@ -76,7 +76,7 @@ void TileWorker::run(Worker* instance) {
                     if (a->isProxy() != b->isProxy()) {
                         return !a->isProxy();
                     }
-                    if (a->source().id() == b->source().id() &&
+                    if (a->sourceId() == b->sourceId() &&
```
tallytalwar: We can save on saving sourceId in tileTask, and instead modify this lambda to give priority to the worker containing a valid not-deleted source.

core/src/data/tileSource.cpp
hjanetzek: This is called within `Map::update()` only for TileSources of the current Scene - so this should be safe.

core/src/tile/tileWorker.cpp
hjanetzek: Not sure what you see to improve here

core/src/tile/tileTask.cpp
```diff
@@ -32,10 +33,11 @@ void TileTask::setTile(std::unique_ptr<Tile>&& _tile) {
 
 void TileTask::process(TileBuilder& _tileBuilder) {
 
-    auto tileData = m_source->parse(*this);
+    auto source = m_source.lock();
+    auto tileData = source->parse(*this);
```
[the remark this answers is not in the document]
hjanetzek: yup

core/src/tile/tileWorker.cpp
tallytalwar: Why to check for sourceId for a source which might have been deleted?

hjanetzek: It's probably rare that a source is deleted here. The intention was to not lock (increment/decrement ref counters) the source in a loop generally. As there are only few items compare it would be ok in this case. We could do it either way

PR #1930, opened 2018-11-08
Texture cleanup

matteblair:
- TextureOptions now includes `generateMipmaps` and `density` parameters.
- TextureOptions has default values.
- Texture constructor doesn't require width and height.
- Filtering, Wrapping, and PixelFormat are now typed enums (I often forget which options belong to which parameter).
- Texture loads image data from pointer and size instead of vector (this saves us from copying the data in some cases where the vector type doesn't match)
- Reorganized and renamed Texture methods for consistency (use `get` for getters, `setData` -> `setPixelData`, `setDirty` -> `setRowsDirty`).
- Applied suggestions from clang-tidy.

### DIFF
--- a/core/src/gl/texture.cpp
+++ b/core/src/gl/texture.cpp
@@ -21,19 +21,17 @@
 
 namespace Tangram {
 
-Texture::Texture(unsigned int _width, unsigned int _height, TextureOptions _options,
-        bool _generateMipmaps, float _density)
-    : m_options(_options), m_generateMipmaps(_generateMipmaps) {
+Texture::Texture(unsigned int _width, unsigned int _height, TextureOptions _options)
+    : m_options(_options) {
 
-    m_invDensity = 1.f/_density;
     m_glHandle = 0;
     m_shouldResize = false;
     m_target = GL_TEXTURE_2D;
     resize(_width, _height);
 }
 
-Texture::Texture(const std::vector<char>& _data, TextureOptions options, bool generateMipmaps, float density)
-    : Texture(0u, 0u, options, generateMipmaps, density) {
+Texture::Texture(const std::vector<char>& _data, TextureOptions options)
+    : Texture(0u, 0u, options) {
 
     loadImageFromMemory(_data);
 }
@@ -93,7 +91,6 @@ Texture& Texture::operator=(Texture&& _other) {
     m_width = _other.m_width;
     m_height = _other.m_height;
     m_target = _other.m_target;
-    m_generateMipmaps = _other.m_generateMipmaps;
     m_rs = _other.m_rs;
 
     return *this;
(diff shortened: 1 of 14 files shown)


core/src/style/pointStyleBuilder.cpp
```diff
@@ -310,15 +310,15 @@ bool PointStyleBuilder::evalSizeParam(const DrawRule& _rule, Parameters& _params
     glm::vec2 spriteSize(NAN);
 
     if (_texture) {
-        spriteSize = glm::vec2{_texture->getWidth(), _texture->getHeight()} * _texture->invDensity();
+        spriteSize = glm::vec2{_texture->getWidth(), _texture->getHeight()} / _texture->getDensity();
```
tallytalwar: Feel this will be called enough times, that it makes sense to save some cycles and not do the division here.

core/src/gl/texture.cpp
```diff
@@ -42,13 +35,10 @@ Texture::~Texture() {
...
-bool Texture::loadImageFromMemory(const std::vector<char>& _data) {
+bool Texture::loadImageFromMemory(const uint8_t* data, size_t length) {
     unsigned char* pixels = nullptr;
-    int width, height, comp;
-
-    if (_data.size() != 0) {
-        pixels = stbi_load_from_memory(reinterpret_cast<const stbi_uc*>(_data.data()), _data.size(), &width, &height, &comp, STBI_rgb_alpha);
-    }
+    int width = 0, height = 0, comp = 0;
+    pixels = stbi_load_from_memory(data, length, &width, &height, &comp, STBI_rgb_alpha);
```
tallytalwar: Can we not skip this call if the length passed is 0?

core/src/gl/texture.cpp
```diff
@@ -73,120 +63,100 @@ bool Texture::loadImageFromMemory(const std::vector<char>& _data) {
...
     m_height = _other.m_height;
-    m_target = _other.m_target;
-    m_generateMipmaps = _other.m_generateMipmaps;
     m_rs = _other.m_rs;
 
     return *this;
 }
 
-void Texture::setData(const GLuint* _data, unsigned int _dataSize) {
+void Texture::setPixelData(const GLuint *data, size_t length) {
```
tallytalwar: nit: styling, `GLuint*` instead of `GLuint *`

matteblair: Forgot to update my CLion style settings haha

core/src/gl/texture.cpp
matteblair: I feel it's not worth adding a branch here for that case, since it should not occur in normal use.

core/src/style/pointStyleBuilder.cpp
matteblair: TextureOptions now stores a `displayScale` property instead of density so it can be applied with multiplication.

PR #1931, opened 2018-11-09
Using fallback tile when tile is not available in DataSource

### LINKED ISSUE #1553: Fallback to lower zoom levels

rwrx:
I have created new pull request for based on previous https://github.com/tangrams/tangram-es/pull/1626 because I didn't created separate branch for it. I am also attaching comment from previous pull request:

I have added fallback to tiles from lower zoom levels when tiles for higher zoom levels are not available. I have tested this on different repo on Android device with mbtiles Data Source. Sometimes it is not able to get tiles when using `MemoryCacheDataSource` before `MBTilesDataSource`, however I haven't found why. Maybe you will see it immediately why in some cases it does not work as I was also unable to attach native debugger to Android device.

This also closes #1553

### DIFF
--- a/core/include/tangram/data/tileSource.h
+++ b/core/include/tangram/data/tileSource.h
@@ -55,6 +55,8 @@ public:
 
         virtual bool loadTileData(std::shared_ptr<TileTask> _task, TileTaskCb _cb) = 0;
 
+        virtual TileID getFallbackTileID(const TileID& _tileID, int32_t _maxZoom, int32_t _zoomBias) = 0;
+
         /* Stops any running I/O tasks pertaining to @_tile */
         virtual void cancelLoadingTile(TileTask& _task) {
             if (next) { next->cancelLoadingTile(_task); }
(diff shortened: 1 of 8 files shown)


rwrx:
I will make a new pull request for this from latest source.

PR #1932, opened 2018-11-09
Add top-center and bottom-center as anchors

rwrx:
Add top-center and bottom-center as possible values for anchor style param


matteblair:
Are these equivalent to the current `bottom` and `top` anchor positions?

rwrx:
@matteblair thank you for your question. Actually from the documentation of anchor https://tangrams.readthedocs.io/en/latest/Syntax-Reference/draw/#anchor it wasn't clear to me how it works. I basically needed to be able to set anchor that will be at bottom in y-axis and at center in x-axis. I have revisited this and I was able to get this working by setting `[top, center]` as value for `anchor`. I am not still sure why this works, I thought that it should be `[bottom, center]`, but it works for me. I will close this pull request.

PR #1935, opened 2018-11-15
[wip] Big scene refactor

hjanetzek:
Goals
- Enable tile fetching as soon as possible, not waiting for the Scene to be completely loaded
- In general these changes should improve parallelization of sceneloading
- Move all things that hold data referencing Scene into Scene so that it's simple to free all data of an old Scene - not having multiple instances (TileWorker, TileBuilder, MarkerManager, ..) still holding references to it.

This adds lots of timing logs - We should test it on some more devices and different scenes to see where is more room for improvement.

### DIFF
--- a/bench/CMakeLists.txt
+++ b/bench/CMakeLists.txt
@@ -7,6 +7,7 @@ target_compile_options(benchmark PRIVATE -O3 -DNDEBUG)
 set(BENCH_SOURCES
   src/builders.cpp
   src/tileLoading.cpp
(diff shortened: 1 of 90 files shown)


tallytalwar:
Like the `LOGTime` here.. good to see numbers for everything!

hjanetzek:
Ok I have pretty much all refactoring done that was on my list. Now it's cleanups and more profiling. Please have a try and let's discuss things on Tuesday

PR #1936, opened 2018-11-16
Fix tile load ordering

hjanetzek:
Please check if this could have unintended side-effects. Now tile load in the appropriate  order - not just at the equator :)


hjanetzek:
Just checked 'center' is only used here so there are no more side-effects!

PR #1939, opened 2018-11-19
Update alfons

hjanetzek:
This should use InputSource move-data constructor on ios


tallytalwar:
@hjanetzek this is good. Can you first make a PR in alfons and bring it in alfons:master before we update submodule in tangram-es?

hjanetzek:
@tallytalwar I've pushed the commit to alfons master - so the submodule ref is still valid. It just seems not to be possible in github gui ;)

tallytalwar:
We should also think about updating harfbuzz, I looked at it last year, and there were a few goodies added to it which could simplify some of alfons code. Something to think about.

PR #1942, opened 2018-11-19
JavaScript context abstraction

matteblair:
Create a JavaScript context abstraction that can be implemented either by Duktape or JavaScriptCore, based on a build-time option.

This results in a much cleaner, more readable implementation of StyleContext and opens the possibility of using JavaScript evaluation in more places (like scene import logic!).

iOS and macOS are configured to use JavaScriptCore and all other platforms use Duktape. 

Performance with Duktape isn't meaningfully different, but with JSCore there is actually a small performance decrease for most function evaluation. This is due to some annoying behavior in JSCore that requires frequent internal context locking. This happens when properties of the context's `feature` object are accessed via our native callback function, so it occurs in almost every function call. It also happens when we create JavaScript string objects, but we can alleviate this by creating a cache of string references to re-use. 

When using the JavaScriptCore implementation, our release artifacts become roughly __700kb smaller per architecture__. This is nice improvement, and I think it justifies further effort to optimize function performance with JavaScriptCore.


bcamper:
Interested to hear how using JavaScriptCore affects performance!

bcamper:
Oh you just addressed that in the description :) Nice that it makes the build smaller. I wonder about a V8 option in the future?

matteblair:
Would love to try V8 too, it just isn't provided as a platform library the way JSCore is on iOS. We could try linking to it dynamically, but it definitely isn't as straightforward.

hjanetzek:
Some more introspection - with current bubble-wrap-style.zip:
valgrind --tool=callgrind ./bench/benchTileBuilder.out --benchmark_filter=DUK/JSC
Look for StyleContext::evalFilter/Style 45% vs 42% of the tile build time is spend in javascript context:

- JSCore
jscore-build-tile
jscore-call

- Duktape
duk-build-tile
duk-call

hjanetzek:
In benchStyleContext I've removed the feature building part from benchTileBuilder, so that only filter- and evalStyleParam functions are called.

Duktape:
```------------------------------------------------------------------------------------
Benchmark                                             Time           CPU Iterations
------------------------------------------------------------------------------------
JSTileStyleFnFixture/TileStyleFnBench          11268834 ns   11203091 ns         62
JSTileStyleFnFixture/TileStyleFnBench          11209765 ns   11144296 ns         62
JSTileStyleFnFixture/TileStyleFnBench          11231128 ns   11164314 ns         62
JSTileStyleFnFixture/TileStyleFnBench          11214786 ns   11151790 ns         62
JSTileStyleFnFixture/TileStyleFnBench_mean     11231128 ns   11165873 ns         62
JSTileStyleFnFixture/TileStyleFnBench_median   11222957 ns   11158052 ns         62
JSTileStyleFnFixture/TileStyleFnBench_stddev      26740 ns      26150 ns         62
```
JScore
```------------------------------------------------------------------------------------
Benchmark                                             Time           CPU Iterations
------------------------------------------------------------------------------------
JSTileStyleFnFixture/TileStyleFnBench          22536629 ns   22460711 ns         31
JSTileStyleFnFixture/TileStyleFnBench          22445475 ns   22304805 ns         31
JSTileStyleFnFixture/TileStyleFnBench          22348890 ns   22268187 ns         31
JSTileStyleFnFixture/TileStyleFnBench          22415510 ns   22313816 ns         31
JSTileStyleFnFixture/TileStyleFnBench_mean     22436626 ns   22336880 ns         31
JSTileStyleFnFixture/TileStyleFnBench_median   22430493 ns   22309310 ns         31
JSTileStyleFnFixture/TileStyleFnBench_stddev      77937 ns      84880 ns         31
```

hjanetzek:
So evaluating filters+functions and getting StyleParams from js takes less that a 1/4 of the tile build time (for this particular tile with bubble-wrap) with Duktape.

core/CMakeLists.txt
```diff
@@ -126,10 +126,17 @@ target_include_directories(tangram-core
 
 # Add JavaScript implementation.
 if(TANGRAM_JSCORE_ENABLED)
+
+  target_compile_definitions(tangram-core
+    PRIVATE
```
matteblair: Using a private compile definition here means that the benchmark targets will never get `TANGRAM_USE_JSCORE=1`

hjanetzek: Yeah so I've added it to benchmarks. It's not really of public interest though :)

core/CMakeLists.txt
```diff
@@ -124,10 +124,48 @@ target_include_directories(tangram-core
...
+    link_directories(${ICU_LIBRARY_DIRS})
+    target_link_libraries(tangram-core
+      PRIVATE
+      ${ICU_LIBRARIES}
+      "${PROJECT_SOURCE_DIR}/deps/jscore/lib/libJavaScriptCore.a"
+      "${PROJECT_SOURCE_DIR}/deps/jscore/lib/libWTF.a"
+      "${PROJECT_SOURCE_DIR}/deps/jscore/lib/libbmalloc.a")
+    target_include_directories(tangram-core
+      PRIVATE
+      "${PROJECT_SOURCE_DIR}/deps/jscore/include")
```
matteblair: @hjanetzek I take it this is for you to test JSCore on linux - Is there any way this can be simplified or moved so that it clutters the file a bit less? Can we `find_package` on an installed JavaScriptCore package instead?

core/CMakeLists.txt
hjanetzek: I was thinking of making this private for a libtangram-stylecontext to have less to rebuild when switching js implementations.

core/src/tile/tileBuilder.h
```diff
@@ -29,14 +29,17 @@ class TileBuilder {
 
     const Scene& scene() const { return *m_scene; }
 
+    // For testing
+    TileBuilder(std::shared_ptr<Scene> _scene, StyleContext* _styleContext);
+
```
hjanetzek: This could be removed now

hjanetzek:
@matteblair - looking at the Circle logs it's probably not my fault that ios does not build

core/CMakeLists.txt
matteblair: Agree that a public flag isn't ideal. 

I can see the logic in making a stylecontext library, but I think switching JS engines is going to be pretty infrequent in practice.

core/CMakeLists.txt
matteblair: Something like https://packages.ubuntu.com/bionic/libjavascriptcoregtk-4.0-dev ? I would try it myself but I'm not on my Ubuntu machine.

matteblair: But yeah also it's probably not going to be frequently used and requires providing other files anyway, so maybe the build configuration could be kept externally too.

matteblair:
@hjanetzek I'll take a look at the iOS build failure, thanks for the cleanups.

The result of this so far is that building with JSCore creates substantially smaller binaries, but at the cost of roughly 2x slower dynamic styling evaluation at runtime. In my view this is not a good trade. I'm in favor of leaving JSCore disabled until we can either improve its styling evaluation performance or reduce the frequency with which we execute JS for styling.

matteblair:
@tallytalwar You may be interested to look at the new types and file structure here. If you have other feedback or concerns we can address that in a follow-up PR, but I'm going to merge this one now so we can move ahead with other optimization work.

PR #1943, opened 2018-11-20
Texture refactoring

hjanetzek:
- Update stb_image to v2.19: now has fast vertical flip!
- Use flip_vertically_on_load provided by stb_image: their implementation looks better. Maybe could be optimized when the row is written directly to the intended location during decoding.
- Add Texture::movePixelData(): use stb_image data directly, no copy
- Make GlyphTexture subclass of Texture / Texture refactor 
   - Use Texture::m_buffer directly instead passing own ptr on upload
   - Remove special Texture::upload() used by GlyphTexture
   - Make Texture non-copyable/moveable: It was just made moveable for GlyphTexture. Though now that Texture holds no more raw pointers we could easily enable move again without having to bother that the custom implementation breaks when changing Texture members.
- Auto-update Texture on bind(): Nicer api: see old comment for update()..
- Simplify RenderState::texture/Unit(): They always go together. That's what we have the state cache for!
- Update getter naming: couldYouGiveMeTheBufferPlease()
- Save some bytes when PixelFormat is not RGBA: Let stb_image decode the requested number of channels instance of always decoding to RGBA. 
- Move stb_image implementation to separate file: Faster texture.cpp recompile

### DIFF
--- a/core/src/gl/texture.cpp
+++ b/core/src/gl/texture.cpp
@@ -58,6 +58,8 @@ bool Texture::loadImageFromMemory(const uint8_t* data, size_t length) {
         setPixelData(1, 1, bpp(), pixel, bpp());
         return false;
     }
+    m_bufferSize = width * height * bpp();
+
     resize(width, height);
     return true;
 }
--- a/core/src/gl/texture.cpp
+++ b/core/src/gl/texture.cpp
@@ -26,12 +26,16 @@ Texture::Texture(const uint8_t* data, size_t length, TextureOptions options)
 }
 
 Texture::~Texture() {
+    freeBufferData();
+
     if (m_rs) {
         m_rs->queueTextureDeletion(m_glHandle);
     }
 }
 
 bool Texture::loadImageFromMemory(const uint8_t* data, size_t length) {
+    freeBufferData();
+
     // stbi_load_from_memory loads the image as a series of scanlines starting
     // from the top-left corner of the image. This flips the output such that
     // the data begins at the bottom-left corner, as required for our OpenGL
@@ -42,13 +46,15 @@ bool Texture::loadImageFromMemory(const uint8_t* data, size_t length) {
     int channelsInFile = 0;
     int channelsRequested = bpp();
 
-    unsigned char* pixels = stbi_load_from_memory(data, static_cast<int>(length),
-                                                  &width, &height, &channelsInFile,
-                                                  channelsRequested);
+    m_buffer = stbi_load_from_memory(data, static_cast<int>(length),
+                                     &width, &height, &channelsInFile,
+                                     channelsRequested);
 
-    if (!pixels) {
+    m_bytesPerPixel = channelsRequested;
+
+    if (!m_buffer) {
         LOGE("Could not load image data: %dx%d bpp:%d/%d",
-             width, height, channelsInFile, channelsRequested);
+             m_width, m_height, channelsInFile, channelsRequested);
 
         // Default inconsistent texture data is set to a 1*1 pixel texture
         // This reduces inconsistent behavior when texture failed loading
@@ -57,30 +63,7 @@ bool Texture::loadImageFromMemory(const uint8_t* data, size_t length) {
         setPixelData(1, 1, bpp(), pixel, bpp());
         return false;
     }
-
-    bool ok = movePixelData(width, height, channelsRequested, pixels,
-                            width * height * channelsRequested);
-    if (!ok) {
-        LOGE("Could not load image data: %dx%d bpp:%d/%d",
-             width, height, channelsInFile, channelsRequested);
-    }
-    return ok;
-}
-
-bool Texture::movePixelData(int _width, int _height, int _bytesPerPixel,
-                            GLubyte* _data, size_t _length) {
-
-    if (!sanityCheck(_width, _height, _bytesPerPixel, _length)) {
-        free(_data);
-        return false;
-    }
-
-    m_buffer.reset(_data);
-    m_bufferSize = _length;
-    m_bytesPerPixel = _bytesPerPixel;
-
-    resize(_width, _height);
-    setRowsDirty(0, m_height);
+    resize(width, height);
     return true;
 }
 
@@ -91,17 +74,21 @@ bool Texture::setPixelData(int _width, int _height, int _bytesPerPixel,
         return false;
     }
 
-    m_buffer.reset();
-    m_bufferSize = 0;
+    if (m_bufferSize != _length) {
+        freeBufferData();
+    }
 
-    auto buffer = reinterpret_cast<GLubyte*>(std::malloc(_length));
-    if (!buffer) {
+    if (!m_buffer) {
+        m_buffer = reinterpret_cast<GLubyte*>(std::malloc(_length));
+    }
+
+    if (!m_buffer) {
         LOGE("Could not allocate texture: Out of memory!");
         return false;
     }
-    std::memcpy(buffer, _data, _length);
 
-    m_buffer.reset(buffer);
+    std::memcpy(m_buffer, _data, _length);
+
     m_bufferSize = _length;
     m_bytesPerPixel = _bytesPerPixel;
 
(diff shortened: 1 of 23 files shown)


hjanetzek:
@matteblair ping!

core/src/gl/texture.h
```diff
@@ -56,48 +56,42 @@ class Texture {
...
     /* Resize the texture */
     void resize(int width, int height);
 
-    void bind(RenderState& rs, GLuint _unit);
+    bool bind(RenderState& rs, GLuint _unit);
 
     /* Width and Height texture getters */
-    int getWidth() const { return m_width; }
-    int getHeight() const { return m_height; }
+    int width() const { return m_width; }
```
matteblair: lol I _just_ renamed these to include "get" https://github.com/tangrams/tangram-es/pull/1930

What's your thought process here?

hjanetzek: I thought we decided a while ago to get rid of these get-prefix for basic getters and this was old leftovers - at least it appeared to me like it and was always on my list to get this naming consistent for Texture. It's a simple command like request to get a property it already holds. When there is no work involved in getting things this is not needed to mention. E.g. a surgeon doesn't ask his assistant if he could be so kind to give him a scalpel - it's just 'scalpel' everybody knows in this context what is wanted. If more work is involved I would agree that is inappropriate when the surgeon to just called 'pizza!' while operating.

matteblair:
Reviewing this today, thanks for being patient :grin:

core/src/gl/texture.h
tallytalwar: 💉

core/src/gl/texture.h
```diff
@@ -56,48 +56,42 @@ class Texture {
...
-    /* Perform texture updates, should be called at least once and after adding data or resizing */
-    virtual void update(RenderState& rs, GLuint _textureSlot);
-
-    virtual void update(RenderState& rs, GLuint _textureSlot, const GLuint* data);
-
     /* Resize the texture */
     void resize(int width, int height);
 
-    void bind(RenderState& rs, GLuint _unit);
+    bool bind(RenderState& rs, GLuint _unit);
```
matteblair: Could use comments for return value.

core/src/gl/texture.h
```diff
@@ -107,9 +106,19 @@ class Texture {
...
+    void setBufferData(GLubyte* buffer, size_t size) {
+        if (m_buffer == buffer) { return; }
+        std::free(m_buffer);
+        m_buffer = buffer;
+    }
+
     TextureOptions m_options;
 
-    std::unique_ptr<GLubyte> m_buffer;
+    GLubyte* m_buffer = nullptr;
```
matteblair: So unique_ptr doesn't free the pointer it holds? I assumed the default deleter would call free - does it just destruct the pointer itself?

hjanetzek: unique_ptr default deleter is calling delete on the malloc'd pointer which Valgrind was so nice to point out.

hjanetzek: Oh what about clang static analyzer? Any experiences with it? Would probably be good to have it in our before merging bigger changes checklist as well

tallytalwar: yeah static analyzer should have caught something like this.

matteblair: I was curious so I ran the xcode static analyzer with the old version of this - no warning about the leak!

matteblair: To be a little more careful, and maybe help ourselves in other situations, we could make a struct to wrap malloced memory that frees it in a destructor.

tallytalwar: Yeah RAII construct will definitely be better here. Something like this: https://embeddedartistry.com/blog/2017/2/23/c-smart-pointers-with-aligned-mallocfree

tallytalwar: Another more concise of what the above link shows: https://gist.github.com/evakili/a1b9fd65c496046dcd24

core/src/gl/texture.cpp
```diff
@@ -8,226 +8,154 @@
...
+    // from the top-left corner of the image. This flips the output such that
+    // the data begins at the bottom-left corner, as required for our OpenGL
+    // texture coordinates.
+    stbi_set_flip_vertically_on_load(true);
+
+    int width = 0, height = 0;
+    int channelsInFile = 0;
+    int channelsRequested = bpp();
+
+    m_buffer = stbi_load_from_memory(data, static_cast<int>(length),
```
matteblair: Does `m_bufferSize` ever get set to match this data?

hjanetzek: Oh forgot to push the fixup I had..

core/src/gl/texture.h
hjanetzek: Yeah using a custom unique_ptr this way sounds good. We can add it now or later.

core/src/gl/texture.h
```diff
@@ -56,48 +56,42 @@ class Texture {
...
+    Texture& operator=(Texture&& _other) = delete;
 
     virtual ~Texture();
 
     bool loadImageFromMemory(const uint8_t* data, size_t length);
 
     /* Sets texture pixel data */
-    void setPixelData(const GLuint* data, size_t length);
+    bool movePixelData(int _width, int _height, int _bytesPerPixel, GLubyte* _data, size_t _length);
+    bool setPixelData(int _width, int _height, int _bytesPerPixel, const GLubyte* _data, size_t _length);
```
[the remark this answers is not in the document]
hjanetzek: Removed the move()

core/src/gl/texture.cpp
```diff
@@ -142,92 +156,85 @@ void Texture::setSpriteAtlas(std::unique_ptr<Tangram::SpriteAtlas> sprites) {
...
+    GL::texParameteri(TEXTURE_TARGET, GL_TEXTURE_WRAP_T,
+                      static_cast<GLint>(m_options.wrapT));
 
-    m_rs = &rs;
+    m_rs = &_rs;
 }
 
 bool Texture::isValid() const {
-    return m_glHandle != 0;
+    return (m_glHandle != 0) || bool(m_buffer);
```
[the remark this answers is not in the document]
hjanetzek: Removed

core/src/gl/texture.h
hjanetzek: Done

core/src/gl/texture.cpp
hjanetzek: Fixed

core/src/gl/texture.cpp
```diff
@@ -38,62 +32,82 @@ Texture::~Texture() {
...
-    m_data.clear();
+    auto buffer = reinterpret_cast<GLubyte*>(std::malloc(_length));
+    if (!buffer) {
+        LOGE("Could not allocate texture: Out of memory!");
+        return false;
+    }
+    std::memcpy(buffer, _data, _length);
 
-    m_data.insert(m_data.begin(), data, data + length);
+    m_buffer.reset(buffer);
```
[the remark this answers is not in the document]
hjanetzek: There's no more movePixelData

core/src/gl/texture.cpp
```diff
@@ -38,62 +32,82 @@ Texture::~Texture() {
...
     }
-    // Default inconsistent texture data is set to a 1*1 pixel texture
-    // This reduces inconsistent behavior when texture failed loading
-    // texture data but a Tangram style shader requires a shader sampler
-    GLuint blackPixel = 0x0000ff;
 
-    setPixelData(&blackPixel, 1);
+    m_buffer.reset(_data);
+    m_bufferSize = _length;
+    m_bytesPerPixel = _bytesPerPixel;
```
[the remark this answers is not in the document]
hjanetzek: yup

core/src/util/stbImage.cpp
```diff
@@ -0,0 +1,8 @@
+// Enable only JPEG, PNG, GIF, TGA and PSD
+#define STBI_NO_BMP
+#define STBI_NO_HDR
+#define STBI_NO_PIC
+#define STBI_NO_PNM
+
+#define STB_IMAGE_IMPLEMENTATION
+#include "stb_image.h"
```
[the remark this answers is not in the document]
hjanetzek: ok

tallytalwar:
Rebase and merge please. Thanks

PR #1945, opened 2018-11-20
Random cleanup and bugfix series

hjanetzek:
This PR is open for anyone to push to.


core/src/view/view.cpp
```diff
@@ -53,7 +53,7 @@ ViewState View::state() const {
 
     return {
         m_changed,
-        glm::dvec2(m_pos.x, -m_pos.y),
+        glm::dvec2(m_pos.x, m_pos.y),
```
tallytalwar: haha, nice catch! Was incorrect for a while it seems.

bench/src/template.cpp
```diff
@@ -0,0 +1,27 @@
+#include "benchmark/benchmark_api.h"
+#include "benchmark/benchmark.h"
+
+
+using namespace Tangram;
+
+class TemplateFixture : public benchmark::Fixture {
```
tallytalwar: 👍

bench/src/template.cpp
```diff
@@ -0,0 +1,27 @@
...
+public:
+    // std::unique_ptr<Template> ctx;
+
+    void SetUp() override {
+        ctx = std::make_unique<TestContext>();
```
tallytalwar: nit; 
This is a template and is not compiled still make sense to have it clean. Just comment out lines using `ctx`.

core/include/tangram/log.h
```diff
@@ -15,7 +15,20 @@
  * LOGS: Screen log, no LOG_LEVEL
  */
 
-#define __FILENAME__ (strrchr(__FILE__, '/') ? strrchr(__FILE__, '/') + 1 : __FILE__)
+//#define __FILENAME__ (strrchr(__FILE__, '/') ? strrchr(__FILE__, '/') + 1 : __FILE__)
+// From: https://blog.galowicz.de/2016/02/20/short_file_macro/
```
tallytalwar: 👍 for reference!

core/src/map.cpp
```diff
@@ -454,7 +454,7 @@ bool Map::update(float _dt) {
     bool tilesLoading = impl->tileManager.hasLoadingTiles();
     bool labelsNeedUpdate = impl->labels.needUpdate();
 
-    if (viewChanged || tilesChanged || tilesLoading || labelsNeedUpdate || impl->sceneLoadTasks > 0) {
+    if (viewChanged || tilesLoading || labelsNeedUpdate || impl->sceneLoadTasks > 0) {
```
tallytalwar: I am actually not sure about this.
It is possible one of the visible tile is ready which causes `m_tileSetChanged` to be set to `true` (https://github.com/tangrams/tangram-es/blob/1e16d40622633fcaca853999760f856e315fe2d2/core/src/tile/tileManager.cpp#L318). But this does not mean view is complete, as its possible other visible tiles are still loading.

core/src/scene/filters.h
```diff
@@ -52,14 +52,15 @@ struct Filter {
         uint32_t id;
     };
     using Data = variant<none_type,
+                         Equality,
```
tallytalwar: @hjanetzek did you find perf benefit for these? If so can you explain me why one will see perf benefit in variant order?

core/src/scene/sceneLayer.cpp
```diff
@@ -1,9 +1,12 @@
...
 #include <algorithm>
+#include <type_traits>
 
 namespace Tangram {
 
+static_assert(std::is_move_constructible<SceneLayer>::value, "check");
+
 SceneLayer::SceneLayer(std::string _name, Filter _filter,
                        std::vector<DrawRuleData> _rules,
                        std::vector<SceneLayer> _sublayers,
```
tallytalwar: Should be a rvalue reference here for `_sublayers`?

core/src/map.cpp
hjanetzek: See the 'tilesLoading' right next to it - that's what tells if tiles are still loading :)

I've came across this when checking the logs big-scene-refactoring on ios that it didn't show 'view complete' when the view was visibly complete.

hjanetzek:
@tallytalwar I've added 'CMake options' and 'update benchmark module' to this PR

core/src/scene/filters.h
hjanetzek: Dropped this commit

core/src/scene/sceneLayer.cpp
matteblair: I believe the vector's contents are still moved at each step (i.e. the backing array is not copied) and that's what matters here. Stepping through the debugger seems to confirm this. If we were more paranoid we could make SceneLayer non-copyable, but I don't think this case is common enough to warrant that.

matteblair:
> 2. Tests on macOS seem to be broken, an assertion fails in some harfbuzz code.

This is occurring on master as well, seems like we just haven't run the full set of tests on macOS in a while. Whatever the cause, this isn't an issue with this PR.

CMakeLists.txt
```diff
@@ -6,6 +6,11 @@ project(tangram)
 option(TANGRAM_USE_SYSTEM_FONT_LIBS "Use system libraries Freetype, ICU and Harfbuzz via pkgconfig" OFF)
 option(TANGRAM_USE_SYSTEM_GLFW_LIBS "Use system libraries for GLFW3 via pkgconfig" OFF)
 
+option(TANGRAM_BUILD_TESTS "Build unit tests" OFF)
+option(TANGRAM_BUNDLE_TESTS "Compile all tests into a single binary" ON)
+option(TANGRAM_BUILD_BENCHMARKS "Build benchmarks" OFF)
+option(TANGRAM_DEV_MODE "For developers only: Don't omit the frame pointer" OFF)
```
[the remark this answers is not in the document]
hjanetzek: I thought it would be good to have a flag to mark things only used during development - like our debug layers, TextDisplay, DebugPrimitives,..

hjanetzek:
Thanks!

PR #1946, opened 2018-11-20
[wip] Big scene refactor

hjanetzek:
Goals
- Enable tile fetching as soon as possible, not waiting for the Scene to be completely loaded
- In general these changes should improve parallelization of sceneloading
- Move all things that hold data referencing Scene into Scene so that it's simple to free all data of an old Scene - not having multiple instances (TileWorker, TileBuilder, MarkerManager, ..) still holding references to it.

This adds lots of timing logs - We should test it on some more devices and different scenes to see where is more room for improvement.


hjanetzek:
Timing for bubble-wrap on android:
https://gist.github.com/hjanetzek/c8b04a580784d14abd9b2dfa50605042

PR #1947, opened 2018-11-20
Platform work

hjanetzek:
- Add Platform::shutdown() to allow stopping workers before going to delete Map. 
- Keep track of all url requests in Platform to allow synchronous cancellation on shutdown. This also ensures that no UrlCallbacks come in while Map is destructing itself on all platforms.

- Linux: 
  - Don't let http requests block reading local files: Since there is a fixed amount of UrlClient workers all of them can be busy waiting for network responses. This blocks reading local files until network calls completed.
- Android:
  - Call into java from less threads, enqueues async calls on single thread. TileWorker don't need to attach/detach their thread for each tile just to call requestRender()
  - Load files asynchronously
  - ￼Use WeakGlobalRef for android MapController instance

### DIFF
--- a/core/include/tangram/platform.h
+++ b/core/include/tangram/platform.h
@@ -97,16 +97,13 @@ protected:
     // It's purpose is to be able to cancel UrlRequests
     using UrlRequestId = uint64_t;
 
-    // TODO is this safe on ios/osx that task identifier are never 0?
-    // Wouldn't hur too much if we don't cancel one request though
-    static constexpr UrlRequestId UrlRequestNotCancelable = 0;
-
     // To be called by implementations to pass UrlResponse
-    void onUrlResponse(UrlRequestHandle _request, UrlResponse&& _response);
+    void onUrlResponse(const UrlRequestHandle _request, UrlResponse&& _response);
 
-    virtual void urlRequestCanceled(UrlRequestId _id) = 0;
+    virtual void cancelUrlRequestImpl(const UrlRequestId _id) = 0;
 
-    virtual UrlRequestId startUrlRequest(Url _url, UrlRequestHandle _request) = 0;
+    // Return true when UrlRequestId has been set (i.e. when request is async and can be canceled)
+    virtual bool startUrlRequestImpl(const Url& _url, const UrlRequestHandle _request, UrlRequestId& _id) = 0;
 
     static bool bytesFromFileSystem(const char* _path, std::function<char*(size_t)> _allocator);
 
@@ -118,6 +115,7 @@ protected:
     struct UrlRequestEntry {
         UrlCallback callback;
         UrlRequestId id;
+        bool cancelable;
     };
     std::unordered_map<UrlRequestHandle, UrlRequestEntry> m_urlCallbacks;
     std::atomic_uint_fast64_t m_urlRequestCount = {0};
--- a/core/include/tangram/platform.h
+++ b/core/include/tangram/platform.h
@@ -98,12 +98,12 @@ protected:
     using UrlRequestId = uint64_t;
 
     // To be called by implementations to pass UrlResponse
-    void onUrlResponse(const UrlRequestHandle _request, UrlResponse&& _response);
+    void onUrlResponse(UrlRequestHandle _request, UrlResponse&& _response);
 
-    virtual void cancelUrlRequestImpl(const UrlRequestId _id) = 0;
+    virtual void cancelUrlRequestImpl(UrlRequestId _id) = 0;
 
     // Return true when UrlRequestId has been set (i.e. when request is async and can be canceled)
-    virtual bool startUrlRequestImpl(const Url& _url, const UrlRequestHandle _request, UrlRequestId& _id) = 0;
+    virtual bool startUrlRequestImpl(const Url& _url, UrlRequestHandle _request, UrlRequestId& _id) = 0;
 
     static bool bytesFromFileSystem(const char* _path, std::function<char*(size_t)> _allocator);
 
--- a/core/include/tangram/platform.h
+++ b/core/include/tangram/platform.h
@@ -2,16 +2,26 @@
 
 #include "util/url.h"
 
+#include <atomic>
 #include <functional>
+#include <mutex>
 #include <string>
+#include <unordered_map>
 #include <vector>
 
 namespace Tangram {
 
-
-// Identifier type for URL requests. This handle is interpreted differently for
-// each platform type, so do not perform any application logic with its value.
+// Handle for URL requests.
+// This is the handle which Platform uses to identify an UrlRequest.
 using UrlRequestHandle = uint64_t;
+
+// Platform specific id for URL requests. This id is interpreted differently for
+// each platform type, so do not perform any application logic with its value.
+//
+// It's purpose is to be able to cancel UrlRequests
+using UrlRequestId = uint64_t;
+
+constexpr UrlRequestId UrlRequestNotCancelable = 0;
 
 // Result of a URL request. If the request could not be completed or if the
 // host returned an HTTP status code >= 400, a non-null error string will be
@@ -61,7 +71,9 @@ public:
     Platform();
     virtual ~Platform();
 
-    virtual void shutdown() = 0;
+    // Subclasses must call Platform::shutdown() when overriding shutdown
+    virtual void shutdown();
+
     // Request that a new frame be rendered by the windowing system
     virtual void requestRender() const = 0;
 
@@ -75,12 +87,12 @@ public:
     // finished, the callback _callback will be run with the data or error that
     // was retrieved from the URL _url. The callback may run on a different
     // thread than the original call to startUrlRequest.
-    virtual UrlRequestHandle startUrlRequest(Url _url, UrlCallback _callback) = 0;
+    UrlRequestHandle startUrlRequest(Url _url, UrlCallback&& _callback);
 
     // Stop retrieving data from a URL that was previously requested. When a
     // request is canceled its callback will still be run, but the response
     // will have an error string and the data may not be complete.
-    virtual void cancelUrlRequest(UrlRequestHandle _request) = 0;
+    void cancelUrlRequest(UrlRequestHandle _request);
 
     virtual FontSourceHandle systemFont(const std::string& _name, const std::string& _weight, const std::string& _face) const;
 
@@ -88,12 +100,26 @@ public:
 
 protected:
 
+    // To be called by implementations to pass UrlResponse
+    void onUrlResponse(UrlRequestHandle _request, UrlResponse&& _response);
+
+    virtual void urlRequestCanceled(UrlRequestId _id) = 0;
+
+    virtual UrlRequestId startUrlRequest(Url _url, UrlRequestHandle _request) = 0;
+
     static bool bytesFromFileSystem(const char* _path, std::function<char*(size_t)> _allocator);
 
-private:
+    std::atomic<bool> m_shutdown{false};
 
     bool m_continuousRendering;
 
+    std::mutex m_callbackMutex;
+    struct UrlRequestEntry {
+        UrlCallback callback;
+        UrlRequestId id;
+    };
+    std::unordered_map<UrlRequestHandle, UrlRequestEntry> m_urlCallbacks;
+    std::atomic_uint_fast64_t m_urlRequestCount = {0};
 };
 
 } // namespace Tangram
--- a/core/include/tangram/platform.h
+++ b/core/include/tangram/platform.h
@@ -61,6 +61,7 @@ public:
     Platform();
     virtual ~Platform();
 
+    virtual void shutdown() = 0;
     // Request that a new frame be rendered by the windowing system
     virtual void requestRender() const = 0;
 
--- a/core/include/tangram/platform.h
+++ b/core/include/tangram/platform.h
@@ -15,14 +15,6 @@ namespace Tangram {
 // This is the handle which Platform uses to identify an UrlRequest.
 using UrlRequestHandle = uint64_t;
 
-// Platform specific id for URL requests. This id is interpreted differently for
-// each platform type, so do not perform any application logic with its value.
-//
-// It's purpose is to be able to cancel UrlRequests
-using UrlRequestId = uint64_t;
-
-constexpr UrlRequestId UrlRequestNotCancelable = 0;
-
 // Result of a URL request. If the request could not be completed or if the
 // host returned an HTTP status code >= 400, a non-null error string will be
 // present. This error string is only valid in the scope of the UrlCallback
@@ -99,6 +91,15 @@ public:
     virtual std::vector<FontSourceHandle> systemFontFallbacksHandle() const;
 
 protected:
+    // Platform implementation specific id for URL requests. This id is
+    // interpreted differently for each platform type, so do not perform any
+    // application logic with its value.
+    // It's purpose is to be able to cancel UrlRequests
+    using UrlRequestId = uint64_t;
+
+    // TODO is this safe on ios/osx that task identifier are never 0?
+    // Wouldn't hur too much if we don't cancel one request though
+    static constexpr UrlRequestId UrlRequestNotCancelable = 0;
 
     // To be called by implementations to pass UrlResponse
     void onUrlResponse(UrlRequestHandle _request, UrlResponse&& _response);
--- a/core/src/platform.cpp
+++ b/core/src/platform.cpp
@@ -95,7 +95,7 @@ UrlRequestHandle Platform::startUrlRequest(Url _url, UrlCallback&& _callback) {
     entry->id = startUrlRequest(_url, handle);
 
     if (entry->id == UrlRequestNotCancelable) {
-        LOGW("Cannot cancel request for handle %d! %s", handle, _url.path().c_str());
+        LOGW("Cannot cancel request for handle %llu! %s", handle, _url.path().c_str());
     }
 
     return handle;
--- a/core/src/platform.cpp
+++ b/core/src/platform.cpp
@@ -94,10 +94,6 @@ UrlRequestHandle Platform::startUrlRequest(Url _url, UrlCallback&& _callback) {
     // Start Platform specific url request
     entry->id = startUrlRequest(_url, handle);
 
-    if (entry->id == UrlRequestNotCancelable) {
-        LOGW("Cannot cancel request for handle %llu! %s", handle, _url.path().c_str());
-    }
-
     return handle;
 }
 
--- a/core/src/platform.cpp
+++ b/core/src/platform.cpp
@@ -5,6 +5,7 @@
 #include <string>
 #include <cassert>
 
+constexpr char const* shutdown_message = "Shutting down";
 constexpr char const* cancel_message = "Request canceled";
 
 namespace Tangram {
@@ -55,12 +56,11 @@ void Platform::shutdown() {
     {
         std::lock_guard<std::mutex> lock(m_callbackMutex);
 
-        UrlResponse response;
-        response.error = "Shutting down";
-
         for (auto& entry : m_urlCallbacks) {
             auto& request = entry.second;
             if (request.callback) {
+                UrlResponse response;
+                response.error = shutdown_message;
                 request.callback(std::move(response));
             }
             if (request.id) {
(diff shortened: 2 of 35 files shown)


platforms/android/tangram/src/main/cpp/androidPlatform.cpp
```diff
@@ -355,20 +361,21 @@ void AndroidPlatform::onUrlComplete(JNIEnv* _jniEnv, jlong _jRequestHandle, jbyt
...
-    UrlCallback callback;
-    {
-        std::lock_guard<std::mutex> lock(m_callbackMutex);
-        UrlRequestHandle requestHandle = static_cast<UrlRequestHandle>(_jRequestHandle);
-        auto it = m_callbacks.find(requestHandle);
-        if (it != m_callbacks.end()) {
-            callback = std::move(it->second);
-            m_callbacks.erase(it);
+
+    m_fileWorker.enqueue([this, _jRequestHandle, r = std::move(response)]() mutable {
```
matteblair: What's the purpose of sending this to another thread? Moving heavy callbacks off of the networking threads? A comment explaining this would be helpful since it might not always be obvious.

core/include/tangram/platform.h
```diff
@@ -61,6 +61,7 @@ class Platform {
     Platform();
     virtual ~Platform();
 
+    virtual void shutdown() = 0;
```
matteblair: It looks like this isn't called anywhere, am I missing something?

platforms/linux/src/linuxPlatform.cpp
```diff
@@ -25,57 +25,88 @@ void logMsg(const char* fmt, ...) {
...
+    // Create FontSourceHandle from the found list of fallback fonts
     std::vector<FontSourceHandle> handles;
     handles.reserve(fallbackFonts.size());
 
-    for (auto& path : fallbackFonts) {
-        handles.emplace_back(Url(path));
-    }
+    std::transform(std::begin(fallbackFonts), std::end(fallbackFonts),
+                   std::back_inserter(handles),
+                   [](auto& path) { return FontSourceHandle(Url(path)); });
```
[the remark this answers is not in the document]
hjanetzek: It transforms fallback fonts begin to end items by calling the lambda and assigning the result to the back_inserter for 'handles'.

Ok I remove this and my cpplint for hinting where to use cpp algorithms :)

platforms/android/tangram/src/main/cpp/androidPlatform.cpp
hjanetzek: > Don't let http requests block reading local files: Since there is a fixed amount of UrlClient workers all of them can be busy waiting for network responses. This blocks reading local files until network calls completed.

hjanetzek: Whoops wrong section..
Yeah it's done for moving decoding of scenes images from Okhttp threads.

core/include/tangram/platform.h
hjanetzek: Yes I've required this during scene-refactor development. It has been used in glfwApp. I'll remove the shutdown part from this PR.

platforms/android/tangram/src/main/cpp/androidPlatform.cpp
```diff
@@ -291,18 +296,17 @@ std::vector<char> AndroidPlatform::bytesFromFile(const Url& url) const {
 
 UrlRequestHandle AndroidPlatform::startUrlRequest(Url _url, UrlCallback _callback) {
 
-    JniThreadBinding jniEnv(jvm);
-
     // Get the current value of the request counter and add one, atomically.
     UrlRequestHandle requestHandle = m_urlRequestCount++;
+    if (!_callback) { return requestHandle; }
 
     // If the requested URL does not use HTTP or HTTPS, retrieve it synchronously.
```
tallytalwar: Update the comment here to reflect async behavior for file loading.

platforms/android/tangram/src/main/cpp/androidPlatform.cpp
```diff
@@ -313,26 +318,28 @@ UrlRequestHandle AndroidPlatform::startUrlRequest(Url _url, UrlCallback _callbac
...
     }
 
-    jlong jRequestHandle = static_cast<jlong>(requestHandle);
+    m_jniWorker.enqueue([=](JNIEnv *jniEnv) {
+        jlong jRequestHandle = static_cast<jlong>(requestHandle);
 
-    // Check that it's safe to convert the UrlRequestHandle to a jlong and back.
-    assert(requestHandle == static_cast<UrlRequestHandle>(jRequestHandle));
+        // Check that it's safe to convert the UrlRequestHandle to a jlong and back... cmon :P
+        assert(requestHandle == static_cast<UrlRequestHandle>(jRequestHandle));
```
tallytalwar: I am fine removing this assert!

matteblair: I kind of like this here just as a reminder to the programmer, I don't think it does any harm.

hjanetzek: I think we can make it a static_assert. jlong is defined as 64bit currently and uint64_t should also stick to 64bit for a while.

hjanetzek: And yeah it safes us in case we should change our UrlRequestHandle definition

core/include/tangram/platform.h
hjanetzek: @matteblair I recalled what shutdown was good for. It's now called first in `~Map()`. Does the comment there make sense? It can probably be handled individually for each platform but at least for glfwApp it seemed appropriate to this way.

platforms/android/tangram/src/main/cpp/androidPlatform.cpp
```diff
@@ -319,8 +319,12 @@ UrlRequestHandle AndroidPlatform::startUrlRequest(Url _url, UrlCallback _callbac
     m_jniWorker.enqueue([=](JNIEnv *jniEnv) {
         jlong jRequestHandle = static_cast<jlong>(requestHandle);
 
-        // Check that it's safe to convert the UrlRequestHandle to a jlong and back... cmon :P
-        assert(requestHandle == static_cast<UrlRequestHandle>(jRequestHandle));
+        // Make sure no one changed UrlRequestHandle from being uint64_t,
+        // so that it's safe to convert to jlong and back.
+        static_assert(sizeof(jlong) == sizeof(UrlRequestHandle), "Who changed UrlRequestHandle?!");
```
tallytalwar: 👍

hjanetzek:
@matteblair @tallytalwar would be good if we can get this shutdown logic also implemented for ios/osx as it should fix the app-not-responsive crashes.

core/src/text/fontContext.cpp
```diff
@@ -296,20 +296,6 @@ void FontContext::releaseFonts() {
...
     // Unload Freetype and Harfbuzz resources for all font faces
     m_alfons.unload();
-
-    // Release system font fallbacks input source data from default fonts, since
-    // those are 'weak' resources (would be automatically reloaded by alfons from
-    // its URI or source callback.
-    for (auto& font : m_font) {
-        if (!font) { continue; }
-        for (auto& face : font->faces()) {
-            alfons::InputSource& fontSource = face->descriptor().source;
```
tallytalwar: Comment says you removed `clearData` from `alfons::InputSource`, should there be a commit to update alfons submodule here?

core/src/platform.cpp
```diff
@@ -46,4 +47,92 @@ std::vector<FontSourceHandle> Platform::systemFontFallbacksHandle() const {
...
+        return 0;
+    }
+
+    UrlRequestHandle handle= ++m_urlRequestCount;
+
+    // Need to do this in advance in case startUrlRequest calls back synchronously..
+    UrlRequestEntry* entry = nullptr;
+    {
+        std::lock_guard<std::mutex> lock(m_callbackMutex);
+        auto it = m_urlCallbacks.emplace(handle, UrlRequestEntry{std::move(_callback), 0});
```
tallytalwar: Instead of using `0`, use the `constexpr` defined above: `UrlRequestNotCancelable`

core/src/platform.cpp
```diff
@@ -46,4 +47,92 @@ std::vector<FontSourceHandle> Platform::systemFontFallbacksHandle() const {
...
+        entry = &it.first->second;
+    }
+
+    // Start Platform specific url request
+    entry->id = startUrlRequest(_url, handle);
+
+    if (entry->id == UrlRequestNotCancelable) {
```
tallytalwar: Why to have this as a warning??

platforms/android/tangram/src/main/cpp/androidPlatform.cpp
```diff
@@ -310,37 +294,20 @@ std::vector<char> AndroidPlatform::bytesFromFile(const Url& url) const {
...
+             onUrlResponse(_handle, std::move(response));
         });
-        return requestHandle;
-    }
-
-    // Store our callback, associated with the request handle.
-    {
-        std::lock_guard<std::mutex> lock(m_callbackMutex);
-        m_callbacks[requestHandle] = _callback;
+        return 0;
```
tallytalwar: Again use `UrlRequestNotCancelable`.

platforms/common/urlClient.h
```diff
@@ -23,16 +23,18 @@ class UrlClient {
     UrlClient(Options options);
     ~UrlClient();
 
-    UrlRequestHandle addRequest(const std::string& url, UrlCallback onComplete);
+    using Callback = std::function<void(UrlResponse&&)>;
```
tallytalwar: Why to create another alias here? Do not think `Callback` serves a different purpose than the `UrlCallback` defined in `platform.h`.

core/src/platform.cpp
tallytalwar: Instead of using `0`, use the `constexpr` defined above: `UrlRequestNotCancelable`

platforms/android/tangram/src/main/java/com/mapzen/tangram/MapRenderer.java
```diff
@@ -0,0 +1,169 @@
...
+
+        synchronized (map) {
+            nativeSetupGL(mapPointer);
+        }
+    }
+
+    public void captureFrame(MapController.FrameCaptureCallback callback, boolean waitForCompleteView) {
+        frameCaptureCallback = callback;
+        frameCaptureAwaitCompleteView = waitForCompleteView;
+
```
tallytalwar: Should we not be calling `requestRender` here?

platforms/android/tangram/src/main/cpp/jniExports.cpp
```diff
@@ -7,6 +7,10 @@
 
 using namespace Tangram;
 
+#define FUNC(CLASS, NAME) JNICALL Java_com_mapzen_tangram_ ## CLASS ## _ ## NAME
+#define MapController(NAME) FUNC(MapController, NAME)
+#define MapRenderer(NAME) FUNC(MapRenderer, NAME)
```
tallytalwar: 👍

platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
```diff
@@ -1247,15 +1247,6 @@ boolean setMarkerDrawOrder(final long markerId, final int drawOrder) {
     // Networking methods
     // ==================
 
-    void cancelAllNetworkRequests() {
```
tallytalwar: shutdown takes care of this!

platforms/android/tangram/src/main/cpp/androidPlatform.h
```diff
@@ -60,10 +57,20 @@ class AndroidPlatform : public Platform {
     jobject m_tangramInstance;
     AAssetManager* m_assetManager;
 
+    mutable JniWorker m_jniWorker;
+    AsyncWorker m_fileWorker;
+};
+
+class AndroidMap : public Map {
```
tallytalwar: 👍

platforms/android/tangram/src/main/java/com/mapzen/tangram/MapView.java
```diff
@@ -108,7 +108,9 @@ public boolean getMapAsync(@NonNull final MapReadyCallback readyCallback,
 
         loadLibraryTask = loadNativeLibraryAsync(new NativeLibraryLoadCb() {
             @Override
-            public void onLibraryReadyAsync(Boolean ok) {}
+            public void onLibraryReadyAsync(Boolean ok) {
+                //FontFileParser.init();
```
tallytalwar: We can remove a few of these commented code!

core/src/platform.cpp
```diff
@@ -46,4 +47,92 @@ std::vector<FontSourceHandle> Platform::systemFontFallbacksHandle() const {
...
+        LOGW("Cannot cancel request for handle %llu! %s", handle, _url.path().c_str());
```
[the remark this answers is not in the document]
tallytalwar: In that case, can I comment be put here? Or maybe an assert?

platforms/android/tangram/src/main/cpp/jniExports.cpp
```diff
@@ -21,619 +22,609 @@ std::vector<Tangram::SceneUpdate> unpackSceneUpdates(JNIEnv* jniEnv, jobjectArra
     return sceneUpdates;
 }
 
-extern "C" {
+#define EXTERN_C_BEGIN extern "C" {
+#define EXTERN_C_END }
```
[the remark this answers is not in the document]
tallytalwar: Probably to remove indentation from extern's braces. I am fine with either, but probably better to stick with pure `extern`!

platforms/linux/src/linuxPlatform.cpp
```diff
@@ -25,57 +25,88 @@ void logMsg(const char* fmt, ...) {
...
 }
 
 UrlRequestHandle LinuxPlatform::startUrlRequest(Url _url, UrlCallback _callback) {
-    return m_urlClient.addRequest(_url.string(), _callback);
+    if (m_shutdown) { return 0; }
+    if (_url.hasHttpScheme()) {
+        return m_urlClient->addRequest(_url.string(),
+                                       [this, cb = _callback](UrlResponse&& r) {
+                                           cb(std::move(r));
+                                           requestRender();
```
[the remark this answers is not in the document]
hjanetzek: Removed now - It was needed on the path to 'synchronous url request cancel on shutdown': When it was not safe to assume that platform is still alive when the url callback runs one could not call `platform->requestRender()` from a callback. (currently we worked around this by looping with `requestRender()` in `Map::update()` to check whether scene resources are ready)

platforms/linux/src/linuxPlatform.cpp
```diff
@@ -25,57 +25,88 @@ void logMsg(const char* fmt, ...) {
...
+             UrlResponse response;
+             auto allocator = [&](size_t size) {
+                 response.content.resize(size);
+                 return response.content.data();
+             };
+
+             Platform::bytesFromFileSystem(path.c_str(), allocator);
+             _callback(std::move(response));
+        });
+        return std::numeric_limits<uint64_t>::max();
```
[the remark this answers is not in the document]
hjanetzek: gone

platforms/android/tangram/src/main/cpp/jniExports.cpp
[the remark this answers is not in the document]
hjanetzek: Ok removed - with 'Follow code indent' option IDEA does not insist on indenting `extern "C"`

core/include/tangram/platform.h
```diff
@@ -74,25 +79,48 @@ class Platform {
...
+    // TODO is this safe on ios/osx that task identifier are never 0?
+    // Wouldn't hur too much if we don't cancel one request though
+    static constexpr UrlRequestId UrlRequestNotCancelable = 0;
+
+    // To be called by implementations to pass UrlResponse
+    void onUrlResponse(UrlRequestHandle _request, UrlResponse&& _response);
+
+    virtual void urlRequestCanceled(UrlRequestId _id) = 0;
+
+    virtual UrlRequestId startUrlRequest(Url _url, UrlRequestHandle _request) = 0;
```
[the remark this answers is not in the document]
hjanetzek: Ok

core/include/tangram/platform.h
```diff
@@ -74,25 +79,48 @@ class Platform {
...
+    void cancelUrlRequest(UrlRequestHandle _request);
 
     virtual FontSourceHandle systemFont(const std::string& _name, const std::string& _weight, const std::string& _face) const;
 
     virtual std::vector<FontSourceHandle> systemFontFallbacksHandle() const;
 
 protected:
+    // Platform implementation specific id for URL requests. This id is
+    // interpreted differently for each platform type, so do not perform any
+    // application logic with its value.
```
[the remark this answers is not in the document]
hjanetzek: I was also not completely happy with this. Working on better version now that properly distinguishes between (non-)async and (non-)cancelable task handling in `startUrlRequestImpl()`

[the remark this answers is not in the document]
hjanetzek: So that `RequestId` can be an arbitrary value (though must unique for async tasks)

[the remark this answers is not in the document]
hjanetzek: Btw can the cases where `iOSPlatform::startUrlRequest()` returns 0 happen unless an app uses our library in wrong ways?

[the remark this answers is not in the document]
matteblair: That's correct, it will only return 0 in cases of user error.

core/src/text/fontContext.cpp
hjanetzek: Updating alfons is not needed for this PR

hjanetzek:
@matteblair @tallytalwar please check again

core/src/platform.cpp
```diff
@@ -79,8 +80,7 @@ UrlRequestHandle Platform::startUrlRequest(Url _url, UrlCallback&& _callback) {
         UrlResponse response;
         response.error = shutdown_message;
         _callback(std::move(response));
-
-        return UrlRequestNotCancelable;
+        return 0;
```
tallytalwar: `0` is still a special case, can we either document it or use a constexpr for it?

tallytalwar: I would suggest to use `constexpr` and use that at all the places below, where intended.

core/include/tangram/platform.h
```diff
@@ -74,25 +79,48 @@ class Platform {
...
+    // It's purpose is to be able to cancel UrlRequests
+    using UrlRequestId = uint64_t;
+
+    // TODO is this safe on ios/osx that task identifier are never 0?
+    // Wouldn't hur too much if we don't cancel one request though
+    static constexpr UrlRequestId UrlRequestNotCancelable = 0;
```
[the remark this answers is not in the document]
hjanetzek: We also have `lower_case_underscore` const/expr in many places. I would prefer reserving all-caps for defines

core/src/platform.cpp
hjanetzek: It has the very special meaning: "Platform is shutting down - we have already handled the request synchronously. You don't need a handle since there is nothing left to handle". 

We could also return an incremented urlRequestCount but that would not change anything

hjanetzek: Ok the `_request ==  0` check in cancelUrlRequest is a bit superfluous..

hjanetzek: Hmm yeah maybe 0 should have the meaning `request_done` for synchronous handled requests, so that callers know not to add these handles into their queue or something

matteblair:
I'm going to push a fix for `OSXPlatform` and some other nominal changes, then I think we'll be good enough to merge.

tallytalwar:
Sounds good to me.
@matteblair can you document your intended changes please, maybe in an issue?

core/include/tangram/platform.h
```diff
@@ -98,12 +98,12 @@ class Platform {
...
     // To be called by implementations to pass UrlResponse
-    void onUrlResponse(const UrlRequestHandle _request, UrlResponse&& _response);
+    void onUrlResponse(UrlRequestHandle _request, UrlResponse&& _response);
 
-    virtual void cancelUrlRequestImpl(const UrlRequestId _id) = 0;
+    virtual void cancelUrlRequestImpl(UrlRequestId _id) = 0;
 
     // Return true when UrlRequestId has been set (i.e. when request is async and can be canceled)
-    virtual bool startUrlRequestImpl(const Url& _url, const UrlRequestHandle _request, UrlRequestId& _id) = 0;
+    virtual bool startUrlRequestImpl(const Url& _url, UrlRequestHandle _request, UrlRequestId& _id) = 0;
```
hjanetzek: I've kind of liked the idea to communicate that if one wants to change `_request` handle one is doing something wrong. Not needed for me though.

matteblair: I see what you mean, but I find that it looks too much like a typo (e.g. you actually meant to make it a `const &`) and `clang-tidy` warns about it.

PR #1952, opened 2018-11-28
Move scene data

hjanetzek:
With the added YAML::Load(const char* data, length) we can avoid a superfluous copy from vector<char> to std::string.


hjanetzek:
This also updates yaml-cpp. I'll tag the yaml commit before merging this so that the ref wont get lost

hjanetzek:
Now it should work!

PR #1955, opened 2018-11-29
Add TANGRAM_MBTILES_DATASOURCE and TANGRAM_USE_SYSTEM_SQLITE_LIBS build options

hjanetzek:
This PR allows Tangram to be build without MBTiles support. This enables users of our library to not include the large SQLite dependency when they don't need it. It's also handy to turn off for faster build times during development.


matteblair:
This requires some modifications in order for Android and iOS builds to work with the MBTiles option off. I've got it mostly done, will push it soon.

In other news, turning off MBTiles/SQlite on Android reduces the release binary size by __2-4 MB per architecture__!

matteblair:
@hjanetzek If my updates look good to you then go ahead and squash or merge. I think we might have slightly different style preferences in CMake - if I haven't formatted thing to your liking, it's nothing personal ;)

hjanetzek:
@matteblair LGTM. Was just not used to `target_sources`

PR #1957, opened 2018-11-30
Rebase javascriptcontext

hjanetzek:
And add linux build options. @matteblair may I push this to 'javascriptcontext' so that it's easier for me to test changes on this branch?


matteblair:
@hjanetzek Yeah you can push these changes to `javascriptcontext`, it'll be good to be benchmark both engines. Thanks!

PR #1964, opened 2018-12-05
Fix getEnclosingCameraPosition for very small bounds

### LINKED ISSUE #1961: 0.10: Bugs with getEnclosingCameraPosition (Android)
I am trying to use the new method `MapController.getEnclosingCameraPosition`, unfortunately, it is not always working as expected. 

I observed the following issues with this method:
1. it sets the tilt always to 0
2. it sets the rotation always to 0
3. it (tries to set) the zoom to something above the previously set maximumZoomLevel. Not sure if this is intended or not
4. **it often does not position the center correctly**. See this video as an example: https://www.westnordost.de/misc/getEnclosingCameraPosition.mp4

Regarding 4, I have not been able to find out any regularity for when the function works correctly and when not. The method I use in my code is this (shortened for easier read):
```java
	private void zoomAndMoveToContain(ElementGeometry geometry)
	{
		BoundingBox objectBounds = geometry.getBounds();
		LngLat sw= TangramConst.toLngLat(objectBounds.getMin());
		LngLat ne= TangramConst.toLngLat(objectBounds.getMax());
		CameraPosition position = controller.getEnclosingCameraPosition(sw, ne, padding);
		controller.updateCameraPosition(CameraUpdateFactory.newCameraPosition(position), 500);
	}
```
`padding` is a Rect with `top=0, left=0, bottom=600, right=0` (in portrait mode)

One thing I can definitely say is that the given padding is ignored when getting the enclosing camera position for point features (`sw==nw`). Perhaps the bug has something to do with point 3? (On zoom level ∞, the lnglat center position would need to be offsetted only by  degrees to meet the padding.)

matteblair:
If the bounding points passed into `getEnclosingCameraPosition` are very close together or are equivalent, then the resulting zoom of the output is larger than the maximum applicable zoom of the map view - resulting in incorrect placement of the view.

To simplify the code a little I moved some functionality into `MapProjection` and wrote tests to check its correctness. 

I also addressed a similar problem in `setCameraPositionEased` that could lead to incorrect or unexpected interpolation. 

Resolves #1961

### DIFF
--- a/core/src/map.cpp
+++ b/core/src/map.cpp
@@ -656,7 +656,7 @@ void Map::setCameraPositionEased(const CameraPosition& _camera, float _duration,
     e.end.pos = MapProjection::lngLatToProjectedMeters({lonEnd, latEnd});
 
     e.start.zoom = getZoom();
-    e.end.zoom = _camera.zoom;
+    e.end.zoom = glm::clamp(_camera.zoom, getMinZoom(), getMaxZoom());
 
     float radiansStart = getRotation();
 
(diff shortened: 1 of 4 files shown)


matteblair:
Ping @hjanetzek

PR #1965, opened 2018-12-05
IOS build dir config

hjanetzek:
This allows setting a custom IOS_BUILD_DIR in Makefile. Very handy for developing on linux and osx at the same time.


hjanetzek:
Yay!

Makefile
```diff
@@ -206,12 +206,12 @@ endif
 	@cd platforms/ios && \
 	jazzy --config jazzy.yml
 
+
+
```
matteblair: nit: extra space

hjanetzek:
Ok no hurry for me - still figuring out how to build things for ios. So it's good that this part works for me :)

I think we could keep the default files with relative build dirs in the repo. This way it's easy to see when our changes should be committed. 
Only when invoking `BUILD_ROOT=/... make ios-xcode` the extra configure_files step is needed.

matteblair:
That might be a good compromise. The `xcuserdata` file will always need to be configured, but the `xcworkspacedata` and `pbxproj` files only nee to be configured if a custom build directory is set, and in that scenario we can document that Xcode project settings won't be preserved.

PR #1969, opened 2018-12-10
Fix segv in TextStyleBuilder

hjanetzek:
- Early return on !params.text caused segv due to missing params.font
  for boundary labels
- Add safety check that TextStyle::Param::font exists

### DIFF
--- a/core/src/style/textStyleBuilder.cpp
+++ b/core/src/style/textStyleBuilder.cpp
@@ -254,6 +254,7 @@ bool TextStyleBuilder::addFeature(const Feature& _feat, const DrawRule& _rule) {
     if (!checkRule(_rule)) { return false; }
 
     TextStyle::Parameters params = applyRule(_rule, _feat.props, false);
+    if (!params.font) { return false; }
 
     Label::Type labelType;
     if (_feat.geometryType == GeometryType::lines) {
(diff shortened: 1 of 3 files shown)


core/src/style/textStyleBuilder.cpp
```diff
@@ -254,6 +254,7 @@ bool TextStyleBuilder::addFeature(const Feature& _feat, const DrawRule& _rule) {
     if (!checkRule(_rule)) { return false; }
 
     TextStyle::Parameters params = applyRule(_rule, _feat.props, false);
+    if (!params.font) { return false; }
```
tallytalwar: We can also do an early return in `applyRule` when `getFont` returns nothing!

tallytalwar: Ahh.. already being done 👍

core/src/text/fontContext.cpp
```diff
@@ -368,17 +373,13 @@ std::shared_ptr<alfons::Font> FontContext::getFont(const std::string& _family, c
...
-
-        // Add fallbacks from default font.
-        if (m_font[sizeIndex]) {
-            font->addFaces(*m_font[sizeIndex]);
-        }
+        LOGD("Using fallback font for Family: %s, Style: %s, Weight: %s, Size %f",
+             _family.c_str(), _style.c_str(), _weight.c_str(), _size);
+    }
+    // Add fallbacks from default font.
+    if (m_font[sizeIndex]) {
```
tallytalwar: Why add fallback font faces when its not required?

tallytalwar:
> Early return on !params.text caused segv due to missing params.font
for boundary labels
Instead of removing the early return when no text parameter is specified, why not set the initial value of `font` to `nullptr` and check before usage in case of boundary labels, maybe the stack trace for this segv you saw can explain this further.

core/src/text/fontContext.cpp
tallytalwar: Ahh yes missed that. thanks for pointing.

matteblair:
Can we resolve any questions here and get this merged?

PR #1972, opened 2018-12-10
Don't dereference null fonts in FontContext::releaseFonts

### LINKED ISSUE #1968: Crash in Map::OnMemoryWarning

this crash can be easily reproduced in iOS/XCode by using
1.  "Simulate Memory Warnings" option in Simulator
2. An application that uses TGMapView
3. add observer for : UIApplicationDidReceiveMemoryWarningNotification in TGMapView.
3. Invoke "Simulate Memory Warnings" in transition to Map view controller (that uses TGMapView).

**RESULT:**
Runtime crash:
Crashed: com.apple.main-thread
0  HereSDKMapKit                  0x1051d50f4 Tangram::FontContext::releaseFonts() + 28
1  CoreFoundation                 0x21858e928 __CFNOTIFICATIONCENTER_IS_CALLING_OUT_TO_AN_OBSERVER__ + 20
2  CoreFoundation                 0x21858e8f4 ___CFXRegistrationPost_block_invoke + 64
3  CoreFoundation                 0x21858dde4 _CFXRegistrationPost + 392
4  CoreFoundation                 0x21858da90 ___CFXNotificationPost_block_invoke + 96
5  CoreFoundation                 0x218505d70 -[_CFXNotificationRegistrar find:object:observer:enumerator:] + 1404
6  CoreFoundation                 0x21858d51c _CFXNotificationPost + 696
7  Foundation                     0x218f9dc10 -[NSNotificationCenter postNotificationName:object:userInfo:] 

**EXPECTED RESULT:**
Memory should be released without any stability issues

**ENVIRONMENT:**
- XCode 9.4
- Simulator : IPhone 8
- Used iOS Demo + adding memorywarning observer in TGMapView

#1967

matteblair:
Resolves #1968 

Credit to @Developers-Here-Mobility for finding the crash here and the fix. I've just made the same fix with a slightly smaller diff.

### DIFF
--- a/core/src/text/fontContext.cpp
+++ b/core/src/text/fontContext.cpp
@@ -294,6 +294,7 @@ void FontContext::releaseFonts() {
     // those are 'weak' resources (would be automatically reloaded by alfons from
     // its URI or source callback.
     for (auto& font : m_font) {
+        if (!font) { continue; }
         for (auto& face : font->faces()) {
             alfons::InputSource& fontSource = face->descriptor().source;
 


core/src/text/fontContext.cpp
```diff
@@ -294,6 +294,7 @@ void FontContext::releaseFonts() {
     // those are 'weak' resources (would be automatically reloaded by alfons from
     // its URI or source callback.
     for (auto& font : m_font) {
+        if (!font) { continue; }
```
tallytalwar: We can do slightly better actually. If even a single font face is set, then all 3 sizes are set, so no need to call the for loop.
Another thing is to maintain a state which gets set in `loadFonts` and not call `releaseFonts` if this state is not set. Slightly cleaner and logical in my opinion.

matteblair: No performance concern here. Anything more abstracted is gonna start making assumptions about code elsewhere and I'd rather just keep it dumb.

PR #1975, opened 2018-12-10
Disable tile loading for datasources with layer disabled

### LINKED ISSUE #1973: Tiles are downloaded from sources in layers that are disabled.
I have a situation where I have a disabled layer, or disabled a layer during a scene update, and the layer's data source tiles are getting downloaded. 

For example a Scene containing:

```
layers:
  weather_radar:
    data: { source: WeatherRadar }
    enabled: false
sources:
  WeatherRadar:
    type: 'Raster'
    url: 'https://tiles.com/weatherimage?x={x}&y={y}&z={z}'
    tile_size: 256
```
results in calls to ```'https://tiles.com/weatherimage``` for tiles.

It is expected to not make any calls as the source is exclusive to that layer.

tallytalwar:
Fixes #1973

cc. @polar


core/src/scene/sceneLoader.cpp
```diff
@@ -1800,6 +1800,12 @@ void SceneLoader::loadLayer(const std::pair<Node, Node>& layer, const std::share
     }
 
     auto sublayer = loadSublayer(layer.second, name, scene);
+    if (!sublayer.enabled()) {
+        auto dataSource = scene->getTileSource(source);
+        if (dataSource) {
+            dataSource->generateGeometry(dataSource->generateGeometry() | false);
```
matteblair: This doesn't make sense to me. Did you mean to use the bitwise OR here? Isn't `(x OR false) == x`?

tallytalwar: oops vim messed up. "OR", will update.

tallytalwar: `(x OR false) == x`

Is what I am looking for. `generateGeometry` is set to `false` before. But if there are multiple layers in the scene using a datasource, `generateGeometry` to be set to `false` only if all these layers are disabled.

matteblair: `dataSource->generateGeometry(dataSource->generateGeometry() | false);`
Is equivalent to:
`dataSource->generateGeometry(dataSource->generateGeometry());`
I think something is missing here.

tallytalwar: I want it to be a `||`. Was a typo.

tallytalwar:
@matteblair PTAL

matteblair:
Do you have a scene configuration that tests this?

Since a regression of this feature would be very hard to notice in regular operation, we may want some kind of test for this (or at least some healthy comments explaining the necessity of this logic).

tallytalwar:
Agreed, will add comments and see if I can come up with a unit test for this.

tallytalwar:
Yeah I tested it yesterday using the following diff:

```diff
diff --git a/scenes/scene.yaml b/scenes/scene.yaml
index f949f8964..5a242549e 100644
--- a/scenes/scene.yaml
+++ b/scenes/scene.yaml
@@ -96,6 +96,12 @@ styles:
             interactive: true

 sources:
+    stamen-terrain:
+        type: Raster
+        #url: http://a.tile.stamen.com/terrain-background/{z}/{x}/{y}.jpg
+        url: http://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png
+        url_subdomains: [a, b, c]
+
     osm:
         type: MVT
         url: https://tile.nextzen.org/tilezen/vector/v1/256/all/{z}/{x}/{y}.mvt
@@ -104,6 +110,13 @@ sources:
             api_key: global.sdk_api_key

 layers:
+    terrain:
+        data: { source: stamen-terrain }
+        draw:
+            raster:
+                #color: [0.8, 0.8, 0.8]
+                order: 0 # draw on bottom
+
     touch:
         data: { source: touch }
         line:
```

scenes/scene-with-disabled-stamen-raster.yaml
```diff
@@ -0,0 +1,721 @@
...
+            outline:
+                color: [.8, .8, .8]
+                width: 1px
+            interactive: true
+
+sources:
+    stamen-terrain:
+        type: Raster
+        #url: http://a.tile.stamen.com/terrain-background/{z}/{x}/{y}.jpg
+        url: http://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png
```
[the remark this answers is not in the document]
matteblair: I'm still not sure how this will alert us to incorrect behavior. Unless you want to run this with breakpoints in the networking code or something?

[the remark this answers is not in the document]
tallytalwar: Fare point, I will just put a bad url in there instead of legit stamen url.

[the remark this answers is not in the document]
matteblair: Less importantly, the source doesn't seem to be using the stamen URL and the origin of the tiles isn't actually relevant to the test.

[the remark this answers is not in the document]
tallytalwar: Haha true!

PR #1976, opened 2018-12-10
Apply pixel density to Marker bitmaps

matteblair:
Similar fix as https://github.com/tangrams/tangram-es/pull/1927 just with a smaller diff.

Resolves https://github.com/tangrams/tangram-es/issues/1925


matteblair:
@polar If you get a chance to try this branch can you confirm that your issue is resolved? This _should_ have the same effect as the PR you submitted.

polar:
This SEGFaults the setBitmap function. Will file new issue.

PR #1977, opened 2018-12-11
TileManager removes tile sources that are not enabled.

### LINKED ISSUE #1973: Tiles are downloaded from sources in layers that are disabled.

polar:
Made TileManager remove tile sources that are not referenced by enabled layers.

Fixes #1973 

I'm not sure if this is the way to go, but it solved my problem, and with Scene Updates as well.
Looks like the TileManager adds/removes tile sources on setScene (which happens on updateScene).
So, I made it remove tile sources that are not enabled by the scene.

### DIFF
--- a/core/src/map.cpp
+++ b/core/src/map.cpp
@@ -168,7 +168,7 @@ void Map::Impl::setScene(std::shared_ptr<Scene>& _scene) {
     }
 
     inputHandler.setView(view);
-    tileManager.setTileSources(_scene->tileSources());
+    tileManager.setTileSourcesFromScene(_scene);
     tileWorker.setScene(_scene);
     markerManager.setScene(_scene);
 
--- a/core/src/tile/tileManager.cpp
+++ b/core/src/tile/tileManager.cpp
@@ -3,6 +3,7 @@
 #include "data/tileSource.h"
 #include "map.h"
 #include "platform.h"
+#include "scene/dataLayer.h"
 #include "tile/tile.h"
 #include "tile/tileCache.h"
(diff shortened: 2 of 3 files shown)


tallytalwar:
@polar did you refer #1975?

tallytalwar:
Just merged #1975, @polar can you verify if it addressed your issue?

Thanks

tallytalwar:
If you refer the [documentation](https://tangrams.readthedocs.io/en/latest/Syntax-Reference/layers/#data), the difference between a [DataLayer](https://github.com/tangrams/tangram-es/blob/master/core/src/scene/dataLayer.h) and [SceneLayer](https://github.com/tangrams/tangram-es/blob/master/core/src/scene/sceneLayer.h) is the inclusion of the `data`(source) parameter. Refer the inheritance logic here!


Since #1975 resolves the underlying issue for you and since that code change is more succinct and appropriate, I am going to close this PR. Kindly let us know if you see other issues wrt this issue.
Thanks

PR #1978, opened 2018-12-12
Optimize ShaderSource::applySourceBlocks()

hjanetzek:
- 3x faster style building than with std::getline on ios

### DIFF
--- a/core/src/gl/shaderSource.cpp
+++ b/core/src/gl/shaderSource.cpp
@@ -68,15 +68,12 @@ std::string ShaderSource::applySourceBlocks(const std::string& _source, bool _fr
         if (end == std::string::npos) {
             end = _source.length();
         }
-        size_t length = end - pos;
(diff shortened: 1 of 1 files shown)


core/src/gl/shaderSource.cpp
```diff
@@ -59,13 +59,25 @@ std::string ShaderSource::applySourceBlocks(const std::string& _source, bool _fr
...
+        size_t pos = end;
+        end = _source.find('\n', pos);
+        if (end == std::string::npos) {
+            end = _source.length();
+        }
+        size_t length = end - pos;
+        if (length > 0 && _source[end-1] == '\r') {
+            length -= 1;
+        }
+        if (length == 0) {
```
tallytalwar: nit but move this above before calculating length.
```suggestion
        if (end == pos) {
        ...
        }
```

PR #1981, opened 2018-12-14
Fix Bitmap density for Markers added at startup

### LINKED ISSUE #1979: MarkerManager::setBitmap SEG Faults.
@matteblair The #1927 pull request, I believe, is throwing a SIGSEGV:

> 2018-12-12 15:52:20.391 6058-6058/com.xxx A/libc: Fatal signal 11 (SIGSEGV), code 1, fault addr 0x244 in tid 6058 (.xxx), pid 6058 (.xxx)
> 2018-12-12 15:52:20.506 6555-6555/? A/DEBUG:     #00 pc 0000000000184edc  /data/app/com.xxxx-q3UwEvgNM8DHzEegeRUpUw==/lib/arm64/libtangram.so (Tangram::MarkerManager::setBitmap(unsigned int, int, int, unsigned int const*)+144)
> 2018-12-12 15:52:20.506 6555-6555/? A/DEBUG:     #01 pc 000000000012b98c  /data/app/com.xxx-q3UwEvgNM8DHzEegeRUpUw==/lib/arm64/libtangram.so (Tangram::Map::markerSetBitmap(unsigned int, int, int, unsigned int const*)+24)
> 2018-12-12 15:52:20.506 6555-6555/? A/DEBUG:     #02 pc 000000000011df48  /data/app/com.xxx-q3UwEvgNM8DHzEegeRUpUw==/lib/arm64/libtangram.so (Java_com_mapzen_tangram_MapController_nativeMarkerSetBitmap+724)
> 

I'm not sure because the `m_scene` may not be set. I'm not sure, but it seems that we should probably have a 
`
if (!m_scene) { return false; }
`
call in setBitMap. 

I've actually done this, and it solves the problem.

matteblair:
Fixes crash when marker is created before scene finishes loading

This should fix #1979 but I need to test it concretely before merging.

### DIFF
--- a/core/include/tangram/map.h
+++ b/core/include/tangram/map.h
@@ -324,7 +324,7 @@ public:
     // Set a bitmap to use as the image for a point marker; _data is a buffer of RGBA pixel data with
     // length of _width * _height; pixels are in row-major order beginning from the bottom-left of the
     // image; returns true if the marker ID was found and successfully updated, otherwise returns false.
-    bool markerSetBitmap(MarkerID _marker, int _width, int _height, const unsigned int* _data);
+    bool markerSetBitmap(MarkerID _marker, int _width, int _height, const unsigned int* _data, float _density = 1.f);
 
     // Set the geometry of a marker to a point at the given coordinates; markers can have their
     // geometry set multiple times with possibly different geometry types; returns true if the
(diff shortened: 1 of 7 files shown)


matteblair:
My initial solution caused inconsistent behavior for Markers added in `onMapReady` before a scene is loaded. At that point the map rendering surface hasn't received its size, so the map itself doesn't have the correct density to apply to bitmaps. The solution is to pass the density from the Java app context all the way down to the `MarkerManager`.

PR #1983, opened 2018-12-14
Fix zero ease duration

hjanetzek:


platforms/ios/framework/src/TGMapView.mm
```diff
@@ -857,10 +857,11 @@ - (void)setCameraPosition:(TGCameraPosition *)cameraPosition
     Tangram::EaseType ease = TGConvertTGEaseTypeToCoreEaseType(easeType);
     if (duration > 0) {
         [self setMapRegionChangeState:TGMapRegionAnimating];
+        self.map->setCameraPositionEased(camera, duration, ease);
     } else {
         [self setMapRegionChangeState:TGMapRegionJumping];
+        self.map->setCameraPosition(camera);
```
matteblair: In this `else` branch the callback doesn't get executed after the position change. It gets set to the `cameraAnimationCallback` property below, but that won't get executed until the _next_ camera change. I think for this case we want to just execute the callback synchronously and _not_ set the `cameraAnimationCallback`.

platforms/ios/framework/src/TGMapView.mm
```diff
@@ -857,11 +857,10 @@ - (void)setCameraPosition:(TGCameraPosition *)cameraPosition
     Tangram::EaseType ease = TGConvertTGEaseTypeToCoreEaseType(easeType);
     if (duration > 0) {
         [self setMapRegionChangeState:TGMapRegionAnimating];
-        self.map->setCameraPositionEased(camera, duration, ease);
     } else {
         [self setMapRegionChangeState:TGMapRegionJumping];
-        self.map->setCameraPosition(camera);
     }
+    self.map->setCameraPositionEased(camera, duration, ease);
```
tallytalwar: 👍

PR #1984, opened 2018-12-15
Update Android build tools and configuration

matteblair:
- Upgrade Android Gradle plugin from 3.0.1 to 3.2.1
- Upgrade Gradle from 4.4 to 4.6
- Replace deprecated 'compile' declaration with 'implementation'
- Remove 'clang' toolchain CMake argument, it's now the default

We've been down this road before: https://github.com/tangrams/tangram-es/pull/1818 

The newest version of the Gradle plugin still omits the 'res' folder from the AAR, but now I think the issues with Buck have been resolved.

### DIFF
--- a/platforms/android/build.gradle
+++ b/platforms/android/build.gradle
@@ -5,7 +5,7 @@ allprojects {
       jcenter()
     }
     dependencies {
-      classpath 'com.android.tools.build:gradle:3.0.1'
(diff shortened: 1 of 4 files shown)


tallytalwar:
> but now I think the issues with Buck have been resolved.
Do you have any reference for this?

matteblair:
https://github.com/facebook/buck/issues/1055#issuecomment-417149150

PR #1985, opened 2018-12-17
fixup benchmark

hjanetzek:
Hope this works


tallytalwar:
You pushed before me :p

hjanetzek:
@matteblair don't if it works for you: I find it comfortable to run benchmarks from the build like:
`ninja benchStyleBuillder.out && bench/benchStyleBuilder.out` which works well when the resources are copied to bin_dir/res

PR #1986, opened 2018-12-18
Make sure correct gl context is set before rendering

tallytalwar:
It is possible there are multiple instances of MapView active, in such scenarios we have to make sure EAGLContext is aptly set with the appropriate context before calling any gl calls.

Also dealloc needs to set apt EAGLContext because map deallocation requires cleaning of gl resources and hence requires context to be aptly set.

This PR addresses these concerns by making sure a validate and invalidate gl context calls are aptly placed before calling any gl function calls (within map->draw or other places in the native code).

### DIFF
--- a/platforms/ios/framework/src/TGMapView.mm
+++ b/platforms/ios/framework/src/TGMapView.mm
@@ -104,17 +104,15 @@ typedef NS_ENUM(NSInteger, TGMapRegionChangeState) {
 
 - (void)dealloc
 {
+    [self validateContext];
     if (_map) {
         delete _map;
-    }
-
-    if ([EAGLContext currentContext] == _context) {
-        [EAGLContext setCurrentContext:nil];
     }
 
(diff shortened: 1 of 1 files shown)


platforms/ios/framework/src/TGMapView.mm
```diff
@@ -369,6 +381,8 @@ - (void)glkView:(GLKView *)view drawInRect:(CGRect)rect
         return;
     }
 
+    [self validateContext];
```
matteblair: This check is unnecessary, I believe. According to the GLKView docs the context is set before this delegate method is called: https://developer.apple.com/documentation/glkit/glkviewdelegate/1615595-glkview?language=objc

tallytalwar: Yup, that documentation is good. The fix is still holding its ground.

PR #1987, opened 2018-12-18
Optimize shader source

hjanetzek:
Further optimize applySourceBlocks() ￼
- write everything between #pragma: tangram in one batch
- preallocate shader source out with appropriate size

Don't add Light and Material shader blocks to Point- and TextStyles￼
- there is just no #pragma for them, so unnecessary to add these blocks
- now buildShaderSource usually overallocates nothing and always less than 1k in my tests

Cleanup: Move setupRaster function into Style::build()
- This 'setup' function was a bit out-of-line with the other 'setup' functions
  which do setup the style for rendering a frame - If we want to keep
  this as a separate function addRasterShaderBlocks would fit better

@matteblair I've added your commit just to avoid solving some merge conflicts again

### DIFF
--- a/core/src/gl/shaderSource.cpp
+++ b/core/src/gl/shaderSource.cpp
@@ -116,6 +116,7 @@ std::string ShaderSource::applySourceBlocks(const std::string& _source, bool _fr
         }
     }
 
+    // printf("overalloc'd: %d\n",  shaderLength - int(out.length()));
     // for (auto& block : m_sourceBlocks) {
     //     if (pragmas.find(block.first) == pragmas.end()) {
     //         logMsg("Warning: expected pragma '%s' in shader source\n",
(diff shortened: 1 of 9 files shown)


core/src/gl/shaderSource.cpp
```diff
@@ -42,45 +43,53 @@ void ShaderSource::addSourceBlock(const std::string& _tagName, const std::string
...
+    }
+
+    size_t shaderLength = out.length() + _source.length();
+    for (auto& block : m_sourceBlocks) {
+        for (auto& s : block.second) {
+            shaderLength += s.length() + 1;
+        }
+        shaderLength += 1;
     }
+    out.reserve(shaderLength);
```
tallytalwar: 👍

PR #1989, opened 2018-12-20
Use CMake source_group function to organize tangram-core files in Xcode

matteblair:
Requires listing all headers in tangram-core, but now Xcode can index header contents and match them with their cpp counterparts!

### DIFF
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -4,96 +4,194 @@ project(tangram-core)
 add_subdirectory(deps)
 
 add_library(tangram-core
+  include/tangram/log.h
+  include/tangram/map.h
+  include/tangram/platform.h
+  include/tangram/tangram.h
+  include/tangram/data/clientGeoJsonSource.h
+  include/tangram/data/properties.h
+  include/tangram/data/propertyItem.h
+  include/tangram/data/tileSource.h
+  include/tangram/tile/tileID.h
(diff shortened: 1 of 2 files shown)


matteblair:
Ugh this `source_group` syntax actually requires CMake 3.8 and Android Studio only has 3.6 (https://cmake.org/cmake/help/v3.8/command/source_group.html). Gonna see if I can run this conditionally or find some other way to specify source groups.

tallytalwar:
Yeah `3.8` add more (`source_group(TREE <root> [PREFIX <prefix>] [FILES <src>...])`) to `source_group` which seems to be introduced in `3.0`. https://cmake.org/cmake/help/v3.0/command/source_group.html

core/CMakeLists.txt
```diff
@@ -231,3 +326,7 @@ foreach(_shader ${SHADER_FILES})
 endforeach()
 add_custom_target(shader-headers DEPENDS ${SHADER_OUTPUT_FILES})
 add_dependencies(tangram-core shader-headers)
+
+# Create source groups for Xcode.
+get_target_property(TANGRAM_CORE_SOURCES tangram-core SOURCES)
+source_group(TREE ${PROJECT_SOURCE_DIR} FILES ${TANGRAM_CORE_SOURCES})
```
tallytalwar: 👍

PR #1991, opened 2018-12-20
Fix RasterSource cache

hjanetzek:
In createTexture() Textures that may not be added to the
  cache later (in getRaster) held a deleter which removed
  their tileId from the cache, leading to invalid rasters. 

Instead now it clears the cache by checking whether the cache
  is the only one holding references to a texture in getRaster()

`Style::draw()`:
  - fix: when tile->rasters are empty uniforms should be set accordingly
  - cleanup: Rasters must be valid - asserted in ´RasterTileTask::complete()´ 
 
Edit: The flickering of hillshade happened from a combination of wrongly invalidated Rasters
and not unsetting the style uniforms.

Edit: The RasterSource cache can still be improved, ~currently it does not remove entries for emptyTexture~. I'll work on this in a follow up PR.

### DIFF
--- a/core/src/data/rasterSource.cpp
+++ b/core/src/data/rasterSource.cpp
@@ -2,6 +2,7 @@
 #include "data/propertyItem.h"
 #include "data/tileData.h"
 #include "tile/tile.h"
+#include "tile/tileBuilder.h"
 #include "tile/tileTask.h"
 #include "util/mapProjection.h"
 #include "log.h"
@@ -10,8 +11,12 @@ namespace Tangram {
 
 class RasterTileTask : public BinaryTileTask {
 public:
-    RasterTileTask(TileID& _tileId, std::shared_ptr<TileSource> _source, int _subTask)
-        : BinaryTileTask(_tileId, _source, _subTask) {}
+
+    const bool subTask = false;
+
+    RasterTileTask(TileID& _tileId, std::shared_ptr<TileSource> _source, bool _subTask)
+        : BinaryTileTask(_tileId, _source),
+          subTask(_subTask) {}
 
     std::shared_ptr<Texture> m_texture;
 
@@ -24,7 +29,7 @@ public:
     }
 
     bool isReady() const override {
-        if (!isSubTask()) {
+        if (!subTask) {
             return bool(m_tile);
         } else {
             return bool(m_texture);
@@ -41,8 +46,9 @@ public:
         }
 
         // Create tile geometries
-        if (!isSubTask()) {
-            BinaryTileTask::process(_tileBuilder);
+        if (!subTask) {
+            m_tile = _tileBuilder.build(m_tileId, *(source->m_tileData), *source);
+            m_ready = true;
         }
     }
 
@@ -50,7 +56,7 @@ public:
         auto source = rasterSource();
         if (!source) { return; }
 
-        auto raster = source->addRaster(*this);
+        auto raster = source->addRaster(m_tileId, m_texture);
         assert(raster.isValid());
 
         m_tile->rasters().push_back(std::move(raster));
@@ -65,7 +71,7 @@ public:
         auto source = rasterSource();
         if (!source) { return; }
 
-        auto raster = source->addRaster(*this);
+        auto raster = source->addRaster(m_tileId, m_texture);
         assert(raster.isValid());
 
         _mainTask.tile()->rasters().push_back(std::move(raster));
@@ -79,6 +85,25 @@ RasterSource::RasterSource(const std::string& _name, std::unique_ptr<DataSource>
       m_texOptions(_options) {
 
     m_emptyTexture = std::make_shared<Texture>(m_texOptions);
+
+    GLubyte pixel[4] = { 0, 0, 0, 0 };
+    auto bpp = _options.bytesPerPixel();
+    m_emptyTexture->setPixelData(1, 1, bpp, pixel, bpp);
+
+    Feature rasterFeature;
+    rasterFeature.geometryType = GeometryType::polygons;
+    rasterFeature.polygons = { { {
+                {0.0f, 0.0f},
+                {1.0f, 0.0f},
+                {1.0f, 1.0f},
+                {0.0f, 1.0f},
+                {0.0f, 0.0f}
+            } } };
+    rasterFeature.props = Properties();
+
+    m_tileData = std::make_shared<TileData>();
+    m_tileData->layers.emplace_back("");
+    m_tileData->layers.back().features.push_back(rasterFeature);
 }
 
 std::shared_ptr<Texture> RasterSource::createTexture(TileID _tile, const std::vector<char>& _rawTileData) {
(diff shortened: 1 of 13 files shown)


core/src/data/rasterSource.cpp
```diff
@@ -157,20 +150,29 @@ std::shared_ptr<TileTask> RasterSource::createTask(TileID _tileId, int _subTask)
...
 
-    auto& task = static_cast<const RasterTileTask&>(_task);
-    m_textures->emplace(id, task.m_texture);
+    auto entry = m_textures.emplace(id, _task.m_texture);
+    std::shared_ptr<Texture> texture = entry.first->second;
+    LOGD("Cache: add %d/%d/%d - reused: %d", id.x, id.y, id.z, !entry.second);
+
+    // Remove textures that are only held by cache
+    for(auto it = m_textures.begin(), end = m_textures.end(); it != end; ) {
+        if (it->second.use_count() == 1) {
```
tallytalwar: Ahh, I should have caught this issue before (https://github.com/tangrams/tangram-es/pull/1944/files#diff-10249f514974ffe886f51c3c334bf004L188)!

tallytalwar: It seems counter intuitive to remove raster textures corresponding to other tileIDs when `getRaster` is called to emplace a raster texture for given tileID (rasterTask).

To me the old code (refer link above) was straight forward in removing all raster textures associated with a tile when it gets deleted.

core/src/data/rasterSource.cpp
```diff
@@ -77,100 +91,127 @@ RasterSource::RasterSource(const std::string& _name, std::unique_ptr<DataSource>
...
     }
 
-    auto& task = static_cast<const RasterTileTask&>(_task);
-    m_textures->emplace(id, task.m_texture);
+    texture = std::shared_ptr<Texture>(_texture.release(),
+                                       [c = std::weak_ptr<Cache>(m_textures), id](auto t) {
+                                           if (auto cache = c.lock()) {
+                                               cache->erase(id);
+                                               LOG("%d - remove %s", cache->size(), id.toString().c_str());
+                                           }
```
hjanetzek: cleanup

core/src/data/rasterSource.cpp
```diff
@@ -77,100 +91,127 @@ RasterSource::RasterSource(const std::string& _name, std::unique_ptr<DataSource>
...
+                {1.0f, 0.0f},
+                {1.0f, 1.0f},
+                {0.0f, 1.0f},
+                {0.0f, 0.0f}
+            } } };
+    rasterFeature.props = Properties();
+
+    m_tileData = std::make_shared<TileData>();
+    m_tileData->layers.emplace_back("");
+    m_tileData->layers.back().features.push_back(rasterFeature);
```
hjanetzek: Needs only to be initialized when `generateGeometry`

core/src/data/tileSource.cpp
```diff
@@ -122,9 +115,12 @@ void TileSource::cancelLoadingTile(TileTask& _task) {
 }
 
 void TileSource::addRasterSource(std::shared_ptr<TileSource> _rasterSource) {
-    /*
-     * We limit the parent source by any attached raster source's min/max.
-     */
+    auto rasterSource = dynamic_cast<RasterSource*>(_rasterSource.get());
+    if (!rasterSource) {
+        assert(false);
```
matteblair: This seems like it would be triggered if a user assigns a data source of the wrong type to a `rasters` property in the scene file. We shouldn't crash the app on scene file errors!

core/src/data/rasterSource.cpp
```diff
@@ -2,58 +2,78 @@
...
+            if (!texture) {
+                raster = std::make_unique<Raster>(m_tileId, source->emptyTexture());
+            }
         }
 
         // Create tile geometries
-        if (!isSubTask()) {
-            BinaryTileTask::process(_tileBuilder);
+        if (!subTask) {
+            m_tile = _tileBuilder.build(m_tileId, *(source->m_tileData), *source);
```
tallytalwar: I am not sure this is correct, when this task is coming from a primary "RasterSource", we will need to call `process` to get tile data.

tallytalwar: Probably this is the reason for the crash @matteblair noticed.

hjanetzek: Look at `TileTask::process`: it calls parse to get the `TileData` then runs `tileBuilder.builder()` with it. RasterSource provides always the same `TiledData` for a simple rectangle mesh - so we can skip the `TileSource::parse()` indirection here.
`source->m_tileData` should valid when RasterSource canGenerateGeometry.... ok hold on I know where the bug is

tallytalwar: Agreed on the redirection, and looks like you fixed this specific bug here: ed544d6.
ed544d6, does fix the tiledata being null here.

core/src/data/rasterSource.cpp
tallytalwar: Ok this fixes the crash, but we do not need to set this `Feature` for all `RasterSources`. As your comment mentioned we only need this when a raster source is actually a primary tile source.

hjanetzek: Yeah I would say negligible overhead for now - RasterSource needs some serious work anyway :)

core/src/data/rasterSource.cpp
```diff
@@ -77,100 +91,133 @@ RasterSource::RasterSource(const std::string& _name, std::unique_ptr<DataSource>
...
+    }
 
-}
+    auto rasterTask = createRasterTask(subTileID, true);
+
+    _task.subTasks().push_back(rasterTask);
 
-std::shared_ptr<TileTask> RasterSource::createTask(TileID _tileId, int _subTask) {
-    auto task = std::make_shared<RasterTileTask>(_tileId, shared_from_this(), _subTask);
+    // Needed? Do we support recursive raster-source inclusion?
```
tallytalwar: We could have a normal raster source on a raw raster/image tile source. So possible to have at least 1 more level.
Can you remove this comment, and maybe add above reasoning on when can we have raster source dependent on another raster source?

tallytalwar: Ahh... I see in this case `TileTask` is our 1st raster source. That said, we can remove this, do not know of any use case for the same.

tallytalwar: cc. @bcamper

core/src/data/rasterSource.cpp
tallytalwar: Thanks for fixing this! :D

core/src/data/rasterSource.cpp
```diff
@@ -77,100 +91,133 @@ RasterSource::RasterSource(const std::string& _name, std::unique_ptr<DataSource>
...
+    auto texture = textureEntry.lock();
+    if (texture) {
+        LOGD("%d - drop duplicate %s", m_textures->size(), id.toString().c_str());
+        // The same texture has been loaded in the meantime: Reuse it and drop _texture..
+        return texture;
     }
 
-    auto& task = static_cast<const RasterTileTask&>(_task);
-    m_textures->emplace(id, task.m_texture);
+    texture = std::shared_ptr<Texture>(_texture.release(),
```
tallytalwar: This is good!!

core/src/data/rasterSource.cpp
```diff
@@ -2,58 +2,78 @@
...
 class RasterTileTask : public BinaryTileTask {
 public:
-    RasterTileTask(TileID& _tileId, std::shared_ptr<TileSource> _source, int _subTask)
-        : BinaryTileTask(_tileId, _source, _subTask) {}
 
-    std::shared_ptr<Texture> m_texture;
+    const bool subTask = false;
+
+    std::unique_ptr<Texture> texture;
+    std::unique_ptr<Raster> raster;
```
tallytalwar: Looks like we can simplify this is only used to determine if the texture associated is "empty". Can we simplify this logic by just having a bool determining if the texture associated with the raster task is "empty"?

tallytalwar: Also does it make sense to add these "empty" textures to tile's rasters?

- Though to answer this question, might open a can of worms. I would suggest to document it and leave it for a subsequent PR.

matteblair:
Oops - we need to clean up the log statements from this branch!

PR #1992, opened 2018-12-20
Fix: old bug in clearTileSets() / TileManager shutdown

hjanetzek:
- cancel fetching of TileTasks in progress
- no more warnings about callback for deleted DataSource

### DIFF
--- a/core/src/tile/tileManager.cpp
+++ b/core/src/tile/tileManager.cpp
@@ -133,7 +133,19 @@ struct TileManager::TileEntry {
 TileManager::TileSet::TileSet(std::shared_ptr<TileSource> _source, bool _clientSource) :
     source(_source), clientTileSource(_clientSource) {}
 
-TileManager::TileSet::~TileSet() {}
+TileManager::TileSet::~TileSet() {
+    cancelTasks();
+}
+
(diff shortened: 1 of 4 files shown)


core/src/tile/tileManager.cpp
```diff
@@ -239,6 +254,14 @@ void TileManager::clearTileSets(bool clearSourceCaches) {
 void TileManager::clearTileSet(int32_t _sourceId) {
     for (auto& tileSet : m_tileSets) {
         if (tileSet.source->id() != _sourceId) { continue; }
+
+        for (auto& tile : tileSet.tiles) {
+            auto& entry = tile.second;
+            if (entry.isInProgress()) {
```
tallytalwar: Instead of duplicating the code here, directly use `cancelTasks` as done in `clearTileSets`.

hjanetzek: Right - I've added this commit with this cleanup to platform-work PR already. Is that ok or do you prefer to merge them separately?

tallytalwar: Nopes thats good. Already checked that. Thanks.

tallytalwar:
Looks like these got incorporated in #1947. So closing.

PR #1993, opened 2018-12-26
Expose CADisplayLink property

msmollin:
@matteblair / @tallytalwar / @karimnaaji PTAL!

This exposes a read-only access to the underlying CADisplayLink property that the iOS Tangram-es SDK uses for synchronizing drawing to display refresh.

*Reasoning*

Currently, when an animation of some kind is defined in a stylesheet, tangram-es will render that animation continuously regardless of whether or not the `continuous` property is set to `NO`. Tangram currently correctly handles pausing/unpausing rendering for OS lifecycle events, but there are other times when developers might want to pause / unpause rendering, for example when a modal view is displayed over top of the map, or the need to respond to one of the iOS view controller lifecycle events such as `viewWillDisappear:`. 

In some of these scenarios, we could utilize a scene update to disable the animations, but that causes a re-render to occur and will cause the map to flicker twice, once when disabling and once when re-enabling.

### DIFF
--- a/platforms/ios/framework/src/TGMapView.h
+++ b/platforms/ios/framework/src/TGMapView.h
@@ -349,6 +349,14 @@ TG_EXPORT
 @property (assign, nonatomic) NSInteger preferredFramesPerSecond;
 
 /**
+ The CADisplayLink used by the renderer. Exposed mostly for the ability to pause/unpause rendering without having to initiate a scene update, as animations defined within the stylesheet force a continuous rendering regardless of the `continuous` setting defined in code.
+ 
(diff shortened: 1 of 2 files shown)


platforms/ios/framework/src/TGMapView.h
```diff
@@ -348,6 +348,14 @@ TG_EXPORT
  */
 @property (assign, nonatomic) NSInteger preferredFramesPerSecond;
 
+/**
+ The CADisplayLink used by the renderer. Exposed mostly for the ability to pause/unpause rendering without having to initiate a scene update, as animations defined within the stylesheet force a continuous rendering regardless of the `continuous` setting defined in code.
```
tallytalwar: Thanks @msmollin for the comment here!

matteblair:
Starting to catch up on github after the holidays. I see that this change just got merged, but I think we should find another solution for the original problem. As evidenced by the number of caveats in the comments here, this is likely to be confusing for developers to use. It's also not likely to be a stable interface, since the `CADisplayLink` property is an implementation detail that can (and probably will) be replaced as we update the engine.

What if instead we expose a `paused` boolean property that both pauses the display link and allows Tangram's internal logic to respect the user's request? Does that satisfy the original use case?

PR #1994, opened 2019-01-09
UrlClient use curl multi api

hjanetzek:


platforms/common/urlClient.cpp
```diff
@@ -17,186 +22,376 @@ struct CurlGlobals {
...
     }
 
-    // Next check the active request list.
-    for (auto& task : m_tasks) {
-        if (task.request.id == id) {
-            task.request.canceled = true;
+    // Check requests that are already active.
+    {
+        std::lock_guard<std::mutex> lock(m_requestMutex);
+        auto it = std::find_if(std::begin(m_tasks), std::end(m_tasks),
```
matteblair: nit: `m_tasks.begin()` seems simpler and clearer here.

hjanetzek: Somehow I find the free functions easier to parse - maybe it's just the syntax highlight  for `std::` that takes my focus to begin/end.

PR #1996, opened 2019-01-10
View's zoom should ALWAYS respect zoom constrain

tallytalwar:
- Previous a client could set a zoom (< constrained min zoom) which gets updated in the next update cycle. But any call to `getZoom` before this update cycle would result in unconstrained zoom value to be passed to client code.
- Only calculate constrained min zoom when view's aspect ratio changes
- Provide a setter for view constrained (removed parameterized view.update() method)

FYI @msmollin

### DIFF
--- a/core/src/view/view.cpp
+++ b/core/src/view/view.cpp
@@ -38,6 +38,7 @@ void View::setPixelScale(float _pixelsPerPoint) {
     m_pixelScale = _pixelsPerPoint;
     m_dirtyMatrices = true;
     m_dirtyTiles = true;
+    m_dirtyConstraint = true;
 
 }
 
@@ -126,7 +127,7 @@ float View::getFocalLength() const {
 
 void View::setMinZoom(float minZoom) {
 
-    m_minZoom = (m_constrainToWorldBounds) ? std::max(minZoom, m_constrainZoom) : std::max(minZoom, 0.f);
+    m_minZoom = std::max(minZoom, 0.f);
     m_maxZoom = std::max(minZoom, m_maxZoom);
     // Set the current zoom again to validate it.
     setZoom(m_zoom);
@@ -165,7 +166,11 @@ float View::getMaxPitch() const {
 }
 
 void View::setConstrainToWorldBounds(bool constrainToWorldBounds) {
+    // Reset view constraints
+    m_constrainMinZoom = 0.f;
+    m_dirtyConstraint = true;
     m_constrainToWorldBounds = constrainToWorldBounds;
+    setZoom(m_zoom);
 }
 
 void View::setPosition(double _x, double _y) {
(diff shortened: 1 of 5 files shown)


core/src/view/view.h
```diff
@@ -258,7 +262,9 @@ class View {
 
     bool m_dirtyMatrices;
     bool m_dirtyTiles;
+    bool m_dirtyConstraint;
     bool m_changed;
+    bool m_constrainToWorldBounds = true;
```
tallytalwar: Probably makes sense to put this in a bitset, will add another commit.

matteblair: Not worth the complexity - we only have one instance of `View`

core/src/view/view.h
```diff
@@ -241,6 +244,7 @@ class View {
     float m_pitch = 0.f;
 
     float m_zoom = 0.f;
+    float m_constrainZoom = 0.f;
```
matteblair: Can the name indicate that this is a lower bound on the zoom?

core/src/view/view.cpp
```diff
@@ -125,7 +126,7 @@ float View::getFocalLength() const {
 
 void View::setMinZoom(float minZoom) {
 
-    m_minZoom = std::max(minZoom, 0.f);
+    m_minZoom = (m_constrainToWorldBounds) ? std::max(minZoom, m_constrainZoom) : std::max(minZoom, 0.f);
```
matteblair: Something seems inconsistent here - if you set a minZoom with constrainToWorldBounds=true and then set constrainToWorldBounds=false, the minZoom is still the constrained zoom.

core/src/view/view.cpp
```diff
@@ -69,6 +69,7 @@ void View::setSize(int _width, int _height) {
     m_aspect = (float)m_vpWidth/ (float)m_vpHeight;
     m_dirtyMatrices = true;
     m_dirtyTiles = true;
+    m_dirtyConstraint = true;
```
matteblair: Looks like this is never set to false?

core/src/view/view.cpp
matteblair: I think it will be more straightforward to conditionally apply `m_minZoom` or `m_constrainZoom` directly in `setZoom`.

core/src/view/view.cpp
```diff
@@ -228,18 +233,16 @@ LngLat View::getCenterCoordinates() const {
...
-    if (_constrainToWorldBounds) {
+    if (m_constrainToWorldBounds && m_dirtyConstraint) {
         // Approximate the view diameter in pixels by taking the maximum dimension.
         double viewDiameterPixels = std::fmax(getWidth(), getHeight()) / pixelScale();
         // Approximate the minimum zoom that keeps with view span within the drawable projection area. [1]
-        double minZoom = std::log(viewDiameterPixels / MapProjection::tileSize() + 2) / std::log(2);
-        if (m_zoom < minZoom) {
-            m_zoom = static_cast<float>(minZoom);
-        }
+        m_constrainZoom = std::log(viewDiameterPixels / MapProjection::tileSize() + 1.5f) / std::log(2);
```
matteblair: The constant `2` here is changed without explanation, and the new value allows for some (albeit rare) cases where the map shows blank areas.

core/src/view/view.cpp
matteblair: The calculated value also depends on `pixelScale` so logically we'd need to reset the flag in that setter too.

core/src/view/view.cpp
tallytalwar: Ohh I was playing with the values to understand the equation. Will update.

core/src/view/view.h
tallytalwar: I meant all the bools now, as View States or something, to save "memory". But again tiny saving, will ignore for now.

core/src/view/view.cpp
tallytalwar: And accidentally pushed this.

tallytalwar:
> I think the simplest way to implement this will be to just calculate and apply the constrained min zoom directly in setZoom. It isn't all that much work to do and we'd have less state to worry about.

I agree on moving this calculation to `setZoom`, but I also feel we can save on some cycles by only calculating the constrainMinZoom when it is required, so I am going to keep the `dirty` check in here. Pushing updates..

tallytalwar:
@matteblair PTAL. Updated with cleanups. Thanks

matteblair:
Superseded by https://github.com/tangrams/tangram-es/pull/1998

PR #1998, opened 2019-01-12
Apply view constraint immediately when setting position or zoom

matteblair:
See also https://github.com/tangrams/tangram-es/pull/1996

Previously a View could have a position/zoom assigned that violated the world bounds constraint until the subsequent `update()` call. This change ensures that the position and zoom of the view always satisfy the constraint preference.

### DIFF
--- a/core/src/view/view.cpp
+++ b/core/src/view/view.cpp
@@ -38,6 +38,7 @@ void View::setPixelScale(float _pixelsPerPoint) {
     m_pixelScale = _pixelsPerPoint;
     m_dirtyMatrices = true;
     m_dirtyTiles = true;
+    m_dirtyConstraint = true;
 
 }
 
@@ -69,6 +70,7 @@ void View::setSize(int _width, int _height) {
     m_aspect = (float)m_vpWidth/ (float)m_vpHeight;
     m_dirtyMatrices = true;
     m_dirtyTiles = true;
+    m_dirtyConstraint = true;
 
     // Screen space orthographic projection matrix, top left origin, y pointing down
     m_orthoViewport = glm::ortho(0.f, (float)m_vpWidth, (float)m_vpHeight, 0.f, -1.f, 1.f);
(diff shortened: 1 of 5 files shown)


core/src/view/view.cpp
```diff
@@ -228,23 +243,23 @@ LngLat View::getCenterCoordinates() const {
     return center;
 }
 
-void View::update(bool _constrainToWorldBounds) {
+void View::applyConstraint() {
+    // Approximate the view diameter in pixels by taking the maximum dimension.
+    double viewDiameterPixels = std::fmax(getWidth(), getHeight()) / pixelScale();
+    // Approximate the minimum zoom that keeps with view span within the drawable projection area. [1]
+    double minZoom = std::log(viewDiameterPixels / MapProjection::tileSize() + 2) / std::log(2);
```
tallytalwar: I was trying to avoid calling these every time zoom or position is set and only call `applyConstraint` when either the size gets updated or pixelScale changes. (Yes I missed `setPosition` in my PR :| ).

core/src/view/view.cpp
```diff
@@ -179,7 +182,12 @@ void View::setPosition(double _x, double _y) {
     m_pos.y = glm::clamp(_y, -MapProjection::EARTH_HALF_CIRCUMFERENCE_METERS, MapProjection::EARTH_HALF_CIRCUMFERENCE_METERS);
     m_dirtyTiles = true;
     if (m_constrainToWorldBounds) {
-        applyConstraint();
+        if (m_dirtyConstraint) {
```
tallytalwar: `setPosition` is something which will be called quite often, and it does not make sense to spend cpu cycles on calculating the min constrained zoom every single time when we know it only changes with viewport size or with pixelScale.

https://godbolt.org/z/gHZ3mp
Does not mimic the entire thing here, but shows the assembly of what operations we will be doing every time setPosition and setZoom is called.

PR #2011, opened 2019-01-19
floating point values for priority style param

### LINKED ISSUE #2010: Support float values on priority property
Sister issue to https://github.com/tangrams/tangram/issues/692 on JS side.

Currently the `priority` property of point and text draw styles only supports ints, but Tangram JS support floats. (Both support functions).

Tilezen is adding a new `collision_rank` property in v1.7 that we'd like to use along with `min_zoom` to return a float.

This comment is the tl;dr:

- https://github.com/tilezen/vector-datasource/issues/988#issuecomment-452470193

Specifically, this should allow a function like (which returns a float):

> priority = feature.properties.min_zoom + (1 - 1 / feature.properties.collision_rank) * 0.1

tallytalwar:
- floating point values for priority will be used soon once
https://github.com/tilezen/vector-datasource/issues/988 lands in
tilzen tiles

resolves #2010

### DIFF
--- a/core/src/scene/styleContext.cpp
+++ b/core/src/scene/styleContext.cpp
@@ -313,6 +313,8 @@ bool StyleContext::evalStyle(FunctionID _id, StyleParamKey _key, StyleParam::Val
                 break;
             }
             case StyleParamKey::angle:
+            case StyleParamKey::priority:
+            case StyleParamKey::text_priority:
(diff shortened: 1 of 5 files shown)


tallytalwar:
Sure, I will add random float priority values right now. But it might make sense to update these when tiles have the collision property set per feature.

PR #2016, opened 2019-01-22
Fix duplicate, non-cyclic scene imports

matteblair:
The current scene import process can behave incorrectly when a scene file is imported more than once in the recursive import graph.

TODO:
- [x] Add a test to demonstrate the existing problem with imports - test should fail currently
- [x] Apply a fix to the scene merge process and check that test passes

### DIFF
--- a/core/src/scene/importer.cpp
+++ b/core/src/scene/importer.cpp
@@ -75,6 +75,9 @@ Node Importer::applySceneImports(Platform& platform) {
         });
     }
 
+    if (m_importedScenes.size() == 1) {
+        return m_importedScenes.begin()->second;
+    }
     Node root;
 
     LOGD("Processing scene import Stack:");
@@ -212,7 +215,7 @@ void Importer::mergeMapFields(Node& target, const Node& import) {
                  Dump(target).c_str(), Dump(import).c_str());
         }
 
-        target = import;
+        target = YAML::Clone(import);
 
     } else {
         for (const auto& entry : import) {
--- a/core/src/scene/importer.cpp
+++ b/core/src/scene/importer.cpp
@@ -5,6 +5,7 @@
 #include "scene/sceneLoader.h"
 #include "util/zipArchive.h"
 
+#include <algorithm>
 #include <atomic>
 #include <cassert>
 #include <condition_variable>
@@ -56,10 +57,8 @@ Node Importer::applySceneImports(Platform& platform) {
                 continue;
             }
 
-            if (m_importedScenes.find(nextUrlToImport) != m_importedScenes.end()) {
-                // This scene URL has already been imported, we're done!
-                continue;
-            }
+            // Mark Url as going-to-be-imported to prevent duplicate work.
+            m_importedScenes.emplace(nextUrlToImport, std::make_pair<Node, std::vector<Url>>(Node(), {}));
         }
 
         activeDownloads++;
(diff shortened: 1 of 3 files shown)


tallytalwar:
Root cause for this is, because how `YAML::Node` references are being modified during `mergeMapNode` during recursive scene importing, the original imported scene gets modified, and hence when its imported again for some other subsequent import ("duplicates") we get incorrect node values.

tallytalwar:
I am fine with the `Quick Fix` commit here. The suggestion I mentioned on slack might be more involved, and this quick fix is simple enough and for any production usage we should anyhow be using a pre-imported scene.

hjanetzek:
@tallytalwar hold on :)

matteblair:
@hjanetzek Still looking over your new change - but the "quick fix" of cloning the merged nodes is still being applied and if I remove that the behavior goes back to being broken. Isn't the goal to not require the cloning?

hjanetzek:
@matteblair hmm works here and seems to work on CI as well

matteblair:
@hjanetzek it fails in an example case - check slack

core/src/scene/importer.h
```diff
@@ -60,7 +61,7 @@ class Importer {
     // Imported scenes must be parsed into YAML nodes to find further imports.
     // The parsed scenes are stored in a map with their URLs to be merged once
     // all imports are found and parsed.
-    std::unordered_map<Url, Node> m_importedScenes;
+    std::unordered_map<Url, std::pair<Node, std::vector<Url>>> m_importedScenes = {};
```
tallytalwar: Can you update the documentation here?

matteblair: Also, a struct instead of a pair would make this much easier to read.

core/src/scene/importer.cpp
```diff
@@ -56,10 +57,8 @@ Node Importer::applySceneImports(Platform& platform) {
                 continue;
             }
 
-            if (m_importedScenes.find(nextUrlToImport) != m_importedScenes.end()) {
-                // This scene URL has already been imported, we're done!
-                continue;
-            }
+            // Mark Url as going-to-be-imported to prevent duplicate work.
+            m_importedScenes.emplace(nextUrlToImport, std::make_pair<Node, std::vector<Url>>(Node(), {}));
```
tallytalwar: nit:
`The element may be constructed even if there already is an element with the key in the container, in which case the newly constructed element will be destroyed immediately.` - https://en.cppreference.com/w/cpp/container/unordered_map/emplace

Might still make sense to check for the existence of the key before calling emplace to avoid this construction/destruction!

hjanetzek: The same url cannot be added twice to `m_sceneQueue`, see `addSceneYaml` - so I would think this is safe

core/src/scene/importer.cpp
```diff
@@ -162,43 +162,37 @@ std::vector<Url> Importer::getResolvedImportUrls(const Node& sceneNode, const Ur
...
-    }
-
-    sceneStack.push_back(sceneUrl);
+    // Insert self to handle self-imports cycles
+    imported.insert(sceneUrl);
 
-    auto sceneNode = m_importedScenes[sceneUrl];
+    auto& entry = m_importedScenes[sceneUrl];
+    auto& sceneNode = entry.first;
+    auto& imports = entry.second;
```
tallytalwar: nit: `imports` and `imported` can easily be messed up in future refactor. I would maybe rename `imports` to `currentSceneImports` or something on those line.

matteblair:
This looks like a good solution! I'm gonna push some small commits to clarify the logic a little.

PR #2018, opened 2019-01-28
Ability to provide speed instead of duration for flyTo animation on iOS

tallytalwar:
This exposes the functionality of specifying a speed with the flyTo animation instead of an explicitly specified duration.

In the internal logic, if a valid explicit duration is not specified, then the duration is calculated with "distance" to be travelled and "speed" (which defaults to 1.0).

Exposing this API provides iOS clients to specify flyTo speed which will "Scale" the flyTo animations appropriately.

NOTE: This is already exposed in android!


platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
```diff
@@ -480,7 +480,11 @@ public void flyToCameraPosition(@NonNull final CameraPosition position, @Nullabl
 
     private void flyToCameraPosition(@NonNull final CameraPosition position, final int duration, @Nullable final CameraAnimationCallback callback, final float speed) {
         checkPointer(mapPointer);
-        // TODO: Make sense to have mark animating for flyTo irrespective of duration/speed?
+        if (duration == 0) {
+            setMapRegionState(MapRegionChangeState.JUMPING);
+        } else {
+            setMapRegionState(MapRegionChangeState.ANIMATING);
+        }
         setMapRegionState(MapRegionChangeState.ANIMATING);
```
matteblair: Probably want to remove this line since you added the branches above.

platforms/ios/framework/src/TGMapView.h
```diff
@@ -225,7 +225,7 @@ TG_EXPORT
 /**
  Move the map camera to a new position with an animation that pans and zooms in a smooth arc.
 
- The animation duration is calculated based on the distance to the new camera position.
+ The animation duration is calculated based on the distance to the new camera position assuming a speed of 1.0 (mercator meters per second)
```
matteblair: Not quite accurate - `speed` is a unit-less scaling factor on the calculated duration. No need to mention speed in this context.

platforms/ios/framework/src/TGMapView.h
```diff
@@ -236,13 +236,29 @@ TG_EXPORT
 /**
  Move the map camera to a new position with an animation that pans and zooms in a smooth arc.
 
+ The animation duration is explicitly specified.
```
matteblair: Not needed, I'd remove.

platforms/ios/framework/src/TGMapView.h
```diff
@@ -236,13 +236,29 @@ TG_EXPORT
...
+
  @param cameraPosition The new camera position
+ @param duration Explicitly specified duration for the camera animation to finish.
```
matteblair: "Duration of the animation in seconds" will do.

platforms/ios/framework/src/TGMapView.h
```diff
@@ -236,13 +236,29 @@ TG_EXPORT
...
  @param callback A callback to execute when the animation completes
  */
 - (void)flyToCameraPosition:(TGCameraPosition *)cameraPosition
                withDuration:(NSTimeInterval)duration
                    callback:(nullable void (^)(BOOL canceled))callback;
 
+/**
+ Move the map camera to a new position with an animation that pans and zooms in a smooth arc.
+
+ The animation duration is calculated based on the distance to the new camera position and the specified speed (mercator meters per second)
```
matteblair: Speed is unit-less. I'd just say "specified speed scaling factor".

platforms/ios/framework/src/TGMapView.mm
```diff
@@ -1220,6 +1220,21 @@ - (void)notifyGestureDidEnd
...
 #pragma mark Internal Logic
 
+- (void)flyToCameraPosition:(TGCameraPosition *)cameraPosition
+               withDuration:(NSTimeInterval)duration
+                      speed:(CGFloat)speed
+                   callback:(void (^)(BOOL))callback
+{
+    Tangram::CameraPosition camera = [cameraPosition convertToCoreCamera];
+    if (duration == 0) {
+        [self setMapRegionChangeState:TGMapRegionJumping];
```
matteblair: Feels weird to bifurcate this state behavior based on duration. Can't think of any practical problems with it though.

platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
tallytalwar: Yeah will clean.

platforms/ios/framework/src/TGMapView.mm
tallytalwar: The practical affect is actually "JUMPING" so I think it does make sense.

PR #2022, opened 2019-01-29
Pin Jazzy gem to version that works on CircleCI

matteblair:
Gonna see if this fixes the iOS build job.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -53,7 +53,7 @@ jobs:
       - checkout
       - run: git submodule update --init
       # Install dependencies.
-      - run: sudo gem install jazzy --no-document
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -39,6 +39,7 @@ jobs:
       - checkout
       - run: git submodule update --init
       # Install dependencies.
+      - run: sudo gem install jazzy --no-document --version 0.9.4 # Just for testing, remove me!
```
tallytalwar: Can you remove this for build stage? I think thats what you also have in comments.

matteblair: That's the plan, just needed to put it in so that the PR would actually run this step.

Issue #2019, opened 2019-01-29
Add zoom_offset data source parameter for down-sampling data


nvkelso:
Tangram ES sister issue to the new Tangram JS functionality in https://github.com/tangrams/tangram/pull/685.

From the JS release notes:

- In some circumstances, the default tiles for a source may be either too large (in file size) or too detailed (in feature density or coordinate precision) than needed for the current application. The new `zoom_offset` data source parameter allows the user to request lower zooms for each zoom level, reducing tile size and visual detail.
- For example, a `zoom_offset: 1` will down-sample the tile data (for both vector and raster sources) by one level, e.g. at zoom 12, zoom 11 tiles will be loaded instead.
- Note: sources with `tile_size: 512` already do this implicitly, loading one level lower than the typical Web Mercator "view" zoom level; any `zoom_offset` for these data source is applied in addition to the tile size adjustment, e.g. `zoom_offset: 1` with `tile_size: 512` will request tiles from two zoom levels earlier (but the 512px tiles will have been designed to not show any loss in visual detail from the first zoom subtraction).
- An example showing Mapzen terrain normal tiles with `zoom_offset: 1` (as compared to full detail), automatically generalizing the terrain while also lowering network bandwidth:

![image](https://user-images.githubusercontent.com/16733/51883976-53669400-2353-11e9-9241-9de27eea6f6d.gif)

tallytalwar:
Addressed by #2026

PR #2026, opened 2019-02-02
Add support for zoom_offset for data sources

tallytalwar:
Addresses #2019 

- Zoom offset behave exactly like zoom bias which gets calculated using
tile size. This can be useful when there are tilesources (unlike tilezen
tiles) which do not provide multiple sized tiles, in such senarios it
can be helpful to use zoom_offset to provide down sampled tiles,
specially when using dependent raster tiles.

- This also fixes the behavior when tile_size for a dependent raster source is greater than tile_size for the parent vector source.

TODO:
- [ ] It is possible in certain scenarios, different vector tiles, end up using the same raster tile. It is quite wasteful to make duplicate url requests for these "same" raster tiles. We might be able to rearchitect some of subTask management code to share subTasks between tasks, and once a "subtasks" is completed, it trickles this to all its master tasks. (cc. @hjanetzek for thoughts on this).

### DIFF
--- a/scenes/raster-terrain-offset.yaml
+++ b/scenes/raster-terrain-offset.yaml
@@ -1,0 +1,33 @@
+global:
+    sdk_api_key: ''
+    sdk_landcover_api_key: ''
+
+sources:
+  nextzen:
+    type: MVT
+    url: https://tile.nextzen.org/tilezen/vector/v1/256/all/{z}/{x}/{y}.mvt
+    url_params:
+        api_key: global.sdk_api_key
+    rasters: ['terrain']
+  terrain:
+    type: Raster
+    url: https://d3gy1yxv8mxm7h.cloudfront.net/tilezen/landcover/v1/512/all/{z}/{x}/{y}.png
+    url_params:
+        api_key: global.sdk_landcover_api_key
+    zoom_offset: 1
+
+styles:
+  raster-test:
+    base: raster
+    shaders:
+      blocks:
+        color: 
+          "color.rgba = sampleRasterAtPixel(0, vec2(currentRasterPixel(0)));"
+
+layers:
+  earth:
+    data: { source: nextzen }
+    draw:
+      raster-test:
+        order: 1
+        color: white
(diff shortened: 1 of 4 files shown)


tallytalwar:
If everyone is fine with it, I can work on the todo in a subsequent PR so that we can get this feature in and unblock carto team(s) who plan to use this.

tallytalwar:
Marking this for review. Will try to bring the "todo" in a subsequent PR.

scenes/raster-terrain-offset.yaml
```diff
@@ -0,0 +1,33 @@
...
+sources:
+  nextzen:
+    type: MVT
+    url: https://tile.nextzen.org/tilezen/vector/v1/256/all/{z}/{x}/{y}.mvt
+    url_params:
+        api_key: global.sdk_api_key
+    rasters: ['terrain']
+  terrain:
+    type: Raster
+    url: https://d3gy1yxv8mxm7h.cloudfront.net/tilezen/landcover/v1/512/all/{z}/{x}/{y}.png
```
matteblair: This seems like it might not age well. Maybe use a Nextzen terrain source or something instead?

scenes/raster-terrain-offset.yaml
```diff
@@ -0,0 +1,33 @@
...
+        api_key: global.sdk_landcover_api_key
+    zoom_offset: 1
+
+styles:
+  raster-test:
+    base: raster
+    shaders:
+      blocks:
+        color: 
+          "color.rgba = sampleRasterAtPixel(0, vec2(currentRasterPixel(0)));"
```
matteblair: Why use a custom color function here instead of the default raster behavior?

scenes/raster-terrain-offset.yaml
tallytalwar: Sure, I was testing the `sampleRasterAtPixel` fix with this change. Will revert to use nextzen terrain.

scenes/raster-terrain-offset.yaml
tallytalwar: I was just testing `sampleRasterAtPixel` fix. Hence this change. Will remove.

PR #2027, opened 2019-02-04
Set text_source to to empty string when js function fails to evaluate

tallytalwar:
- jsFunction evaluation can result in an "Undefined" value when it fails
to evaluate. Example when trying to return a feature property which is
not present, "name:short". In such scenarios we do not want the
text_source to fall back to "name" feature property.
- NOTE: feature filtering in the style sheet should be written such that
this above described "Undefined" JS value never happens, but still
tangram-es should be more graceful and lenient about this.


hjanetzek:
LGTM. Using 'Undefined' to suppress a label is actually a smart way to do things when one would otherwise use one js-function for filtering and one for text-soure.

PR #2030, opened 2019-02-11
Add TextureViewHolder implementation as a reference for client to use for a custom GLViewHolder/GLViewHolderFactory

tallytalwar:
- Now tangram android interface can use TextureView as a canvas for map
gl rendering.
- This also provides a reference for clients to provide their own
GLViewHolder (example if they want share a single SurfaceView between
multiple Android App Views).

### DIFF
--- a/platforms/android/tangram/src/main/java/com/mapzen/tangram/viewholder/GLViewHolder.java
+++ b/platforms/android/tangram/src/main/java/com/mapzen/tangram/viewholder/GLViewHolder.java
@@ -2,7 +2,6 @@ package com.mapzen.tangram.viewholder;
 
 import android.opengl.GLSurfaceView;
 import android.support.annotation.NonNull;
-import android.view.TextureView;
 import android.view.View;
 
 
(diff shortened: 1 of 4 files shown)


platforms/android/tangram/src/main/java/com/mapzen/tangram/viewholder/GLViewHolder.java
```diff
@@ -67,7 +68,8 @@
 
     /**
      * Can be used by the client to get access to the underlying view to pass any view controls
-     * @return {@link GLSurfaceView} or Client provided implementation for a view held by {@link GLViewHolder}
+     * @return {@link GLSurfaceView} or {@link TextureView} or Client provided implementation for
```
matteblair: No need to call out (or import) specific implementations, it's just a `GLSurfaceView`

platforms/android/tangram/src/main/java/com/mapzen/tangram/viewholder/GLViewHolderFactory.java
```diff
@@ -5,8 +5,8 @@
 
 public interface GLViewHolderFactory {
     /**
-     * Responsible to create an instance of {@link GLSurfaceViewHolder} or client provided implementation
-     * for {@link GLViewHolder}
+     * Responsible to create an instance of {@link GLSurfaceViewHolder} or {@link TextureViewHolder} or
```
matteblair: Again, it's just a `GLSurfaceView`, no need to call out specific implementations.

PR #2035, opened 2019-02-13
Windows (MinGW)

karimnaaji:
Supersedes PR https://github.com/tangrams/tangram-es/pull/1812 (thanks to @tehKaiN for the initial work on this):
- Add platform layer for windows platform.
- Add initial AppVeyor configuration (https://ci.appveyor.com/project/karimnaaji/tangram-es, can be greatly improved by using a docker container having preinstalled dependencies, this initial work builds z-lib and curl from source which isn't the best, Conan doesn't have dependencies built for MinGW).
- curl multi API support hasn't been added yet, still TODO for that PR (added in https://github.com/tangrams/tangram-es/pull/1994)


platforms/windows/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+From the zlib source directory:
+``` bat
+cmake -H. -Bbuild -G "MinGW Makefiles" -DCMAKE_INSTALL_PREFIX=c:/cmake-install
+cmake --build build
+cmake --build build --target install
+```
+From the curl source directory:
+``` bat
+cmake -H. -Bbuild -G "MinGW Makefiles" -DCMAKE_INSTALL_PREFIX=c:/cmake-install -DCMAKE_USE_WINSSL=ON
```
matteblair: I added `-DBUILD_TESTING=OFF` here - on my machine building tests was much slower and eventually failed due to missing perl.

platforms/windows/README.md
```diff
@@ -0,0 +1,56 @@
...
+cmake --build build
+cmake --build build --target install
+```
+
+### Build Tangram
+
+In the root directory type `mingw32-make windows`. You can append `DEBUG=1` or `RELEASE=1` to choose the build type.
```
matteblair: On my machine CMake was unable to find my libcurl build unless I also specified `-DCMAKE_INSTALL_PREFIX="C:/cmake-install"` for the tangram build.

karimnaaji:
After the installation of MinGW you should be adding the appropriate path to the MinGW libraries to your PATH. An issue like this is most likely the runtime .dlls not available in your PATH. Did you make that step?

Dependency walker is a great tool to manage and resolve those kind of issues.

platforms/windows/README.md
karimnaaji: Most likely a similar issue here. Make sure that the PATH is up to date for the required dlls.

platforms/windows/README.md
karimnaaji: Yes that's unnecessary, build options updated in https://github.com/tangrams/tangram-es/pull/2035/commits/b27b18cb491e03e5c51b5ebe1bfff8090fdf3560 and https://github.com/tangrams/tangram-es/pull/2035/commits/6efd940d34cbce96258e8352d13360dc37d4e8f7 (Also adding `-DBUILD_CURL_EXE=OFF`).

matteblair:
Success! I got the exe to launch after updating my PATH to include my mingw64 bin folder and my "cmake install" bin folder. Might be worth noting that in the README for troubleshooting.

The next issue was that the file path to the default scene was incorrect at launch:
```
TANGRAM C:\Users\Matt\Documents\tangram-es\platforms\windows\src\main.cpp:25: Base URL: file://C:\Users\Matt\Documents\tangram-es\build\windows
TANGRAM C:\Users\Matt\Documents\tangram-es\core\src\map.cpp:244: Loading scene file (async): file://C:\Users\Matt\Documents\tangram-es\build\windows/scene.yaml
WARNING C:\Users\Matt\Documents\tangram-es\core\src\platform.cpp:29: Failed to read file at path: /scene.yaml
```
This was the output from launching either from Command Prompt or by clicking the exe in File Explorer. We might need to adjust the file path logic in WindowsPlatform - I don't think our Url class is able to reformat Windows-style paths.

However, once I corrected the file path manually the scene loaded and rendered! 🎉

tehKaiN:
@matteblair @karimnaaji  I've hacked around it in one of my commits, In main.cpp

``` cpp
    // Resolve the input path against the current directory.
    Url baseUrl("file:///");
    char pathBuffer[PATH_MAX] = {0};
    if (_getcwd(pathBuffer, PATH_MAX) != nullptr) {
        // URL handler apparently needs slash at end of it
        strcat(pathBuffer, "/");
        baseUrl = Url("file:///" + std::string(pathBuffer)).resolved(baseUrl);
    }
```

IIRC this fixed path problem for me. Of course proper fix would be better but perhaps it'd guide you to the root of the cause.

charlenni:
Any progress in this PR?

matteblair:
I've been sadly neglecting this PR for a long time. I had been attempting to create a "native" port that doesn't require MinGW, but it's clear that having a port of some kind is still really valuable! I'll attempt to fix the conflicts in this PR and test it again over the next week.

PR #2040, opened 2019-02-15
Map update-state flags

hjanetzek:
Instead of returning kind of arbitrary bool from `Map::update()` and `Map::render()` this returns a bitset with all the state flags from `Map::update()`

### DIFF
--- a/core/include/tangram/map.h
+++ b/core/include/tangram/map.h
@@ -109,15 +109,15 @@ struct EdgePadding {
 
 struct CameraUpdate {
     enum Flags {
-        SET_LNGLAT =      1 << 0,
-        SET_ZOOM =        1 << 1,
-        SET_ZOOM_BY =     1 << 2,
-        SET_ROTATION =    1 << 3,
-        SET_ROTATION_BY = 1 << 4,
-        SET_TILT =        1 << 5,
-        SET_TILT_BY =     1 << 6,
-        SET_BOUNDS =      1 << 7,
-        SET_CAMERA =      1 << 8,
+        set_lnglat =      1 << 0,
+        set_zoom =        1 << 1,
+        set_zoom_by =     1 << 2,
+        set_rotation =    1 << 3,
+        set_rotation_by = 1 << 4,
+        set_tilt =        1 << 5,
+        set_tilt_by =     1 << 6,
+        set_bounds =      1 << 7,
+        set_camera =      1 << 8,
     };
     int set = 0;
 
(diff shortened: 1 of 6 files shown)


tallytalwar:
Nice.. code looks clean. Will give it a try on all platforms and merge.

core/include/tangram/map.h
```diff
@@ -185,13 +202,14 @@ class Map {
     // Resize the map view to a new width and height (in pixels)
     void resize(int _newWidth, int _newHeight);
 
-    // Update the map state with the time interval since the last update, returns
+    // Update the map state with the time interval since the last update.
+    //returns
     // true when the current view is completely loaded (all tiles are available and
     // no animation in progress)
-    bool update(float _dt);
```
hjanetzek: Need to update docs

matteblair:
This change makes the macOS app completely unresponsive - not sure why yet.

hjanetzek:
Seeing this on linux as well. gdb attached shows `visibleTiles = std::set with 31416 elements` Checking further tomorrow

matteblair:
Yep, I'm seeing a flood of tile load tasks on macOS - seems to only happen when the scene tries to prefetch tiles.

hjanetzek:
Oh yes.. with the `view->update()` in `Map::update()` while scene is loading (and prefetching) `view` could have been modified concurrently.

matteblair:
Looks like it's fixed now - thanks!

platforms/android/tangram/src/main/java/com/mapzen/tangram/MapRenderer.java
```diff
@@ -34,29 +41,41 @@ public void onDrawFrame(final GL10 gl) {
...
+            if (!isPrevCameraEasing) {
+                uiThreadHandler.post(setMapRegionAnimatingRunnable);
+            }
         } else if (isPrevCameraEasing) {
             uiThreadHandler.post(setMapRegionIdleRunnable);
         }
 
-        boolean viewComplete = mapViewComplete && !isPrevMapViewComplete;
+        if (isAnimating) {
+            map.requestRender();
```
matteblair: It seems simpler to apply the re-rendering logic in the core, rather than duplicating it for each of the platforms. Why this change?

hjanetzek: It's mostly beneficial on Android to safe one native-to-java call. Though to me this seems also is more clean - or otherwise name it `updateAndRequestRerenderAsNeeded()`

matteblair: Fair enough!

PR #2046, opened 2019-02-27
Exclusive and Priority for SceneLayer

### LINKED ISSUE #2036: More layer matching control with priority and exclusive
Tangram ES sister issue to [PR #705](https://github.com/tangrams/tangram/pull/705).

What @bcamper said over there:

This PR introduces two new keywords -- `priority` and `exclusive` -- for scene `layers`, to enable greater control over feature filter matching and styling composition.

**Background**
Tangram layer matching has always been inclusive, meaning a feature can match multiple layers (at multiple levels of the layer style tree), which are then merged to determine the final rendering parameters. Composing complex sub-layer trees can be used to capture both feature and zoom-dependent styling along orthogonal dimensions, for example varying road line widths and colors by separate criteria.

However, two in many ways simpler behaviors have _not_ been easily expressed with the current layer syntax:

* The priority with which layers are matched has always been implicit. While layer depth takes priority (with an inheritance model), when layers at the same level match (siblings), the tie-breaker has been alphabetical, with the last lexically sorted layer name winning. This is cryptic at best and a significant hindrance to easy authoring where layer priority is important (see Bubble Wrap [shield matching](https://github.com/tangrams/bubble-wrap/blob/gh-pages/bubble-wrap-style.yaml#L996)).
* Since multiple sibling layers will match by default, there has been no way to express simple _exclusive_ relationships, for example an ordered list of filter conditions, where the first matching filter should be used, and the others discarded. While it has often been possible to work around this by carefully building filters to be functionally exclusive (for example, defining a filter A, then defining a filter B that negates all of A's conditions, in addition to including its own), this is also an unwieldy approach, and does not serve all cases (such as when the filters have natural overlaps that can't easily be separated in a way that avoids a feature from matching more than once).

**New Syntax**
The `priority` and `exclusive` keywords address these cases, and can be used both separately and together.

* `priority`: allows for the explicit definition of the matching order for sibling layers (alphabetical by layer name is still used as the default / fallback).
* `exclusive`: marking a layer as `exclusive` ensures that when that layer matches, _no other_ sibling layers (and their cascading sub-layers, etc.) will; when multiple `exclusive` layers match, the higher priority layer wins.

For example, `exclusive` could express a single `if-else` filter condition:

```
layers:
  if:
    filter: ...
    exclusive: true
  else:
    ...
```
When used together, `priority` and `exclusive` can be used for chained `if-elseif-else` filter cases, to ensure that only one of an ordered set of conditions matches:

```
layers:
  layerA:                 # if matches layerA...
    filter: ...
    priority: 1
    exclusive: true
    draw: ...
  layerB:                 # else if matches layerB...
    filter: ...
    priority: 2
    exclusive: true
    draw: ...
  layerC:                 # else default to layerC...
    priority: 3
    exclusive: true
    draw: ...
```
Separate from "flow control" cases, `priority` can simply be used on its own to disambiguate the precedence of multiple layers without reliance on naming conventions.

**Implementation**
The implementation is a fairly narrow impact on the layer matching logic:

* When determining layer matching and merging order:
  
  * Layers with an explicit `priority` value are matched first, with values sorted from lowest to highest (e.g. `priority: 1` is matched before `priority: 2`).
  * Layers without an explicit `priority` are matched after, and continue to be sorted lexically.
* Matching operation:
  
  * If a layer marked `exclusive` matches, all further matching at that level in the layer tree ceases and the matching layer is used (matching continues for that `exclusive` layer's child/descendant layers).
  * In the case that multiple layers are marked `exclusive`, the first one (according to the order described above) is used.

tallytalwar:
* Sort SceneLayers such that all exclusive layers come before
  non-exclusive layers and in order of increasing priority

TODO:

* [x]  Add unit test covering `priority` and `exclusive` properties.
  
Resolves #2036


matteblair:
After a year in limbo, this PR is finally ready!

This now implements the `exclusive` and `priority` parameters for scene layers as described in https://github.com/tangrams/tangram-es/issues/2036. I also added tests for this feature and refactored a lot of the surrounding code to simplify logic and improve clarity. 

The scene layer sorting in this implementation looks a bit different than that in the Tangram JS implementation (https://github.com/tangrams/tangram/pull/705). This is because we have a different way of traversing the layer tree when evaluating rules for a feature. Basically, when draw rules are being merged we keep the first value that is set for a parameter, unless the new one is from a greater depth. So, for sibling layers to follow the rules of `exclusive` and `priority`, we just sort them by exclusivity first, then by priority, and finally by fully-qualified name (in reverse-alphabetical order).

matteblair:
This is looking good to go!

The one thing I've found from starting to use this feature is that I frequently encounter layers where every sublayer can or should be `exclusive: true`. This is common when a set of sublayers categorize features by a single tag like `kind`. It would be convenient in these cases to have a shorthand like `sublayers_exclusive: true` to write once on the parent layer in these cases.

PR #2001, opened 2019-01-15
Scene refactor continued

hjanetzek:

PR #2047, opened 2019-03-04
Style texture refactor

hjanetzek:
Small refactor to not have to lookup style uniform textures for each draw.


matteblair:
Looks like this has some overlap with https://github.com/tangrams/tangram-es/pull/2001 - are you still working on that one too?

PR #2048, opened 2019-03-08
Fix osx benchmark and test

tallytalwar:
- were missing c++14 build flag!
- Includes a Tangram cmake variable to control cxx_standard

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -37,6 +37,11 @@ endif()
 
 include(cmake/utils.cmake)
 
+if(NOT TANGRAM_CXX_STANDARD)
+    message(STATUS "CXX STANDARD NOT SET. Setting to c++14")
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -37,6 +37,11 @@ endif()
 
 include(cmake/utils.cmake)
 
+if(NOT TANGRAM_CXX_STANDARD)
+    message(STATUS "CXX STANDARD NOT SET. Setting to c++14")
```
matteblair: Not sure this is necessary, wouldn't want to make someone think they've improperly configured the build because of a message like this.

matteblair:
This fix was included in https://github.com/tangrams/tangram-es/pull/2095 - turns out there were other issues preventing tests from running on macOS but those are now resolved too. Closing this PR!

Issue #2050, opened 2019-03-14
Local reference table overflow when more than 512 tiles were tried to be loaded, WITH FIX ATTACHED


vilinet:
I am just tried to use the latest map, browsing tiles, and the errors just pops out
Issue came with this commit: https://github.com/tangrams/tangram-es/commit/795347c713c7f349b3cb466af3b86e529034813a?diff=split

Solution:
in androidPlatform.Cpp:308, 
jniEnv->DeleteLocalRef(jUrl) is missing to remove the local reference.

Locally i tested it and works perfectly after this line addition.

vilinet:
Here is the patch file for latest master, as i dont have write to create branch: 
https://pastebin.com/S8J5SpNZ

nvkelso:
Fix in https://github.com/tangrams/tangram-es/pull/2051.

PR #2051, opened 2019-03-16
Fix JNI reference overflow in AndroidPlatform::startUrlRequest

matteblair:
Resolves https://github.com/tangrams/tangram-es/issues/2050

The issue and fix were both reported by @vilinet - thanks!


tallytalwar:
Thanks @vilinet.

PR #2076, opened 2019-07-10
Fix errors installing Jazzy in CircleCI builds

matteblair:
WIP

Update: Seems to be working with the latest release of Jazzy and a newer macOS VM on Circle.


tallytalwar:
Do you think its just the version?

PR #2077, opened 2019-07-22
modify dash:exchange width and heigh of m_texture

dragoon87:
i think:
m_texture->setPixelData(pixels.size(), 1, sizeof(GLuint),
is wrong, i can not render dash
in old version,it is:
m_texture->resize(1, pixels.size());
m_texture->setPixelData(pixels.data(), pixels.size());


matteblair:
You're totally right! Thanks for the PR, I'll try to merge fixes faster next time 👍

Issue #2074, opened 2019-06-19
make linux fails with gcc 7.4


keltinuous:
Attempted linux make (master branch) on clean Ubuntu 18.04.2 LTS (Bionic) install. Fail after warnings about old gcc versions and with errors regarding glm. 

Error example:

error: class 'glm::tvec4<float, (glm::precision)4>' does not have any field named 'data'

Assumed same issue as (https://github.com/tangrams/tangram-es/commit/3d5b37c68d687f59cebf7e64e39803b4f03d3f47)

Mimicked solution by duplicating lines 286 and 287 with an increment to GNUC_MINOR in core/deps/glm/glm/simd/platform.h

286    #	elif (__GNUC__ == 7) && (__GNUC_MINOR__ == 3)
287    #		define GLM_COMPILER (GLM_COMPILER_GCC72)
288    #	elif (__GNUC__ == 7) && (__GNUC_MINOR__ == 4)
289    #		define GLM_COMPILER (GLM_COMPILER_GCC72)
290    #	elif (__GNUC__ >= 8)
291    #		define GLM_COMPILER (GLM_COMPILER_GCC80)

make succeeded following this change.

PR #2091, opened 2019-10-05
GLM update

matteblair:
Updates GLM library files to release 0.9.9.6 (https://github.com/g-truc/glm/releases/tag/0.9.9.6)

Resolves https://github.com/tangrams/tangram-es/issues/2074


matteblair:
Updating GLM seems to be causing crashes and incorrect behavior, even after restoring the default initialization behavior that was removed in version 0.9.9.0. This will need to be tracked down and fixed eventually if we want to ever update GLM (which we probably should).

But for now, it is simple enough to work around the immediate issue of missing build support for GCC 7.4 instead. Closing this PR in favor of #2092

PR #2093, opened 2019-10-05
Publish Catch test results on CircleCI

matteblair:
Should make it a little easier to interpret test failures on CI.


tallytalwar:
Good one @matteblair

PR #2103, opened 2019-11-18
Android pick callbacks revision

matteblair:
This clears up some confusion and ambiguity in the feature/label/marker picking interfaces in the Tangram Android SDK. 

- Updated documentation to clarify that pick calls will always result in a callback (when a listener is set). Success of the pick is expressed in the result object passed to the callback.
- `onFeaturePick` etc. have been renamed to `onFeaturePickComplete` to both distinguish the updated method signatures and to emphasize that the pick is _complete_, but not necessarily _successful_.
- Callback methods now take a single result object as their argument, which will be non-null if the pick call resulted in a feature/label/marker and will be null otherwise.

### DIFF
--- a/platforms/android/demo/src/main/java/com/mapzen/tangram/android/MainActivity.java
+++ b/platforms/android/demo/src/main/java/com/mapzen/tangram/android/MainActivity.java
@@ -11,6 +11,7 @@ import android.widget.Toast;
 import com.mapzen.tangram.CameraPosition;
 import com.mapzen.tangram.CameraUpdateFactory;
 import com.mapzen.tangram.FeaturePickListener;
+import com.mapzen.tangram.FeaturePickResult;
 import com.mapzen.tangram.LabelPickListener;
 import com.mapzen.tangram.LabelPickResult;
 import com.mapzen.tangram.LngLat;
@@ -161,6 +162,8 @@ public class MainActivity extends AppCompatActivity implements MapController.Sce
         map.updateCameraPosition(CameraUpdateFactory.newLngLatZoom(new LngLat(-74.00976419448854, 40.70532700869127), 16));
 
         mapData = map.addDataLayer("touch");
+
+        map.setPickRadius(10);
     }
 
     @Override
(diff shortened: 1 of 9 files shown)


tallytalwar:
@matteblair this requires some more old context for me to understand and refresh my memory on the API usages. I am going to get to this during the holidays, sorry for the delay.

westnordost:
This is great! I was about to look up how this worked internally and report that the feature/marker/label pick listener is not called when the result is empty if that would have been the case. The way you implemented it - that the listener is always called - makes it feasible to extend the MapController to be used with Kotlin coroutines. Simplified example:

```kotlin
suspend fun MapController.awaitPickMarker(posX: Float, posY: Float) = 
    suspendCoroutine<MarkerPickResult?> { cont ->
        c.setMarkerPickListener { markerPickResult ->
            cont.resume(markerPickResult)
        }
        c.pickMarker(posX, posY)
    }
```

This enables to write code like this, which is cleaner than juggling with the callbacks

```kotlin
suspend fun clickAt(clickPosition: PointF) {
    val pickResult = controller.awaitPickMarker(clickPosition.x, clickPosition.y)
    if (pickResult == null) {
        toast("Hmm, no marker here...")
    } else {
        val x = pickResult.coordinates.longitude
        val y = pickResult.coordinates.latitude
        toast("Found a marker at $x, $y!")
    }
}
```

westnordost:
Ah, just a comment on the interface. The actual implementation to enable the mentioned suspend pickMarker function in a convenience wrapper is this (0.11.2):

```kotlin

private val pickLabelContinuations = mutableMapOf<PointF, Continuation<LabelPickResult?>>()

init { // constructor
    mapController.setLabelPickListener { labelPickResult: LabelPickResult?, positionX, positionY ->
        val p = PointF(positionX, positionY)
        pickLabelContinuations.remove(p)?.resume(labelPickResult)
    }
}

suspend fun pickLabel(posX: Float, posY: Float): LabelPickResult? = suspendCoroutine { cont ->
    pickLabelContinuations[PointF(posX, posY)] = cont
    c.pickLabel(posX, posY)
}
```

**Why this implementation?**
As the picking of a marker is an async call and thus it is not defined when tangram will call the `LabelPickListener` with the result, it must be assumed that there have been several calls to `pickLabel` before the first returned and that the listener is called with the results not exactly in the same order as the `pickLabel` calls were made. (At least that is the assumption one needs to make unless he has knowledge about the implementation details.)
Thus, the label pick callbacks need to be correlated to the label pick calls. The implementation above does this by checking if the click position in the callback is the same as the click position in the call.

However, since `positionX` and `positionY` are both `0` if there was no label at the clicked position (and these two parameters will be gone after `master` with this PR is released anyway), it is not possible to correlate the callback with the call.

westnordost:
I could also open a new issue if you think the information should result in one of the proposed solutions (or another) being done  and it will get lost in this closed PR.

matteblair:
The callbacks are performed in the same order that the queries are submitted (i.e. they are first-in-first-out). We can document this since it appears to be helpful for some use cases.

PR #2106, opened 2019-11-25
Fix iOS framework bundle identifier

matteblair:
I discovered that our current build process creates a Framework whose plist is missing a "Product Bundle Identifier" (or, `CFBundleIdentifier`) key. This creates errors when you try to install an app that is using the framework downstream (e.g. through CocoaPods).

This fix ensures that the bundle identifier is assigned in the framework. I'll make a minor release after merging this to propagate the fix to CocoaPods.


tallytalwar:
Good Catch @matteblair

PR #2111, opened 2019-11-29
Add support for multiple mbtiles

rwrx:
Add support for multiple mbtiles files through urls_mbtiles. This is only my little hack in order to support mutltiple mbtiles not only one as a data source. If this is not appropriate and you also want to add this feature I can modify it as you will want.

### DIFF
--- a/core/src/data/mbtilesDataSource.cpp
+++ b/core/src/data/mbtilesDataSource.cpp
@@ -115,10 +115,11 @@ struct MBTilesQueries {
 
 };
 
-MBTilesDataSource::MBTilesDataSource(Platform& _platform, std::string _name, std::string _path,
-                                     std::string _mime, bool _cache, bool _offlineFallback)
+MBTilesDataSource::MBTilesDataSource(Platform& _platform, std::string _name,
+                                     std::vector<std::string> _paths, std::string _mime,
+                                     bool _cache, bool _offlineFallback)
     : m_name(_name),
-      m_path(_path),
+      m_paths(_paths),
       m_mime(_mime),
       m_cacheMode(_cache),
       m_offlineMode(_offlineFallback),
@@ -143,7 +144,7 @@ bool MBTilesDataSource::loadTileData(std::shared_ptr<TileTask> _task, TileTaskCb
         return loadNextSource(_task, _cb);
     }
 
-    if (!m_db) { return false; }
+    if (m_dbs.empty()) { return false; }
 
     if (_task->rawSource == this->level) {
 
@@ -184,7 +185,7 @@ bool MBTilesDataSource::loadTileData(std::shared_ptr<TileTask> _task, TileTaskCb
 bool MBTilesDataSource::loadNextSource(std::shared_ptr<TileTask> _task, TileTaskCb _cb) {
     if (!next) { return false; }
 
-    if (!m_db) {
+    if (m_dbs.empty()) {
         return next->loadTileData(_task, _cb);
     }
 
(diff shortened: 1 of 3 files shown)


rwrx:
Reply to comment https://github.com/tangrams/tangram-es/pull/2119#issuecomment-566767128 example of using this parameter is (on Android platform):

source:
    mapzen:
        urls_mbtiles: ["/storage/emulated/0/offline_maps/belgium.mbtiles", "/storage/emulated/0/offline_maps/netherlands.mbtiles", "/storage/emulated/0/offline_maps/germany.mbtiles"]

in my app I allow users to download map data stored in mbtiles format. They can download mbtiles for every country in the world. Firstly I have tried after downloading mbtiles to merge them into one mbtiles file on Android device and it was very very slow - counts in minutes. Then I added support for using mutltiple mbtiles and when trying to show a tile, check which mbtiles file contains this tile and use it from that mbtiles file.

I just needed this functionality and thought that it could be useful also for other developers which want to use Tangram-ES for offline maps stored in mbtiles.

matteblair:
Aha, thanks for sharing your use case! So this data source is a set of local archives that can be used offline. This does seem like a reasonable scenario for us to support. I'll take a look at the changes and see how we can best enable this.

core/src/data/mbtilesDataSource.cpp
```diff
@@ -232,77 +233,86 @@ bool MBTilesDataSource::loadNextSource(std::shared_ptr<TileTask> _task, TileTask
...
-            // TODO better description
-            LOGE("Cannot cache to 'externally created' MBTiles database");
-            // Run in non-caching mode
-            m_cacheMode = false;
+        } catch (std::exception& e) {
+            LOGE("Unable to open SQLite database: %s - %s", path.c_str(), e.what());
+            if (!m_dbs.empty()) {
+                m_dbs[m_dbs.size() - 1].reset();
+            }
             return;
```
tallytalwar: Should you be doing a `continue` here to try the other set of mbtile path?

core/src/data/mbtilesDataSource.cpp
```diff
@@ -424,51 +434,60 @@ void MBTilesDataSource::initSchema(SQLite::Database& db, std::string _name, std:
...
+
+            if (stmt.executeStep()) {
+                SQLite::Column column = stmt.getColumn(0);
+                const int length = column.getBytes();
 
-        if (stmt.executeStep()) {
-            SQLite::Column column = stmt.getColumn(0);
-            const char* blob = (const char*) column.getBlob();
-            const int length = column.getBytes();
+                if (length > largestLength) {
```
tallytalwar: Can you add a comment about this check please?

rwrx: If multiple mbtiles contain the same tile, it will get tile which has more data.

rwrx:
Ok, I have tried to address your requests and also done some additional cleanup and fixes.

rwrx:
I am concerned about name of actual YAML paramter name of `urls_mbtiles` if this is correct, or if it should be somehow done together withing `url` parameter.

core/src/data/mbtilesDataSource.cpp
tallytalwar: Sounds great. Can you put this in the code comment please, will be helpful for anyone looking at the code in future. Thanks

matteblair:
@rwrx I've been thinking about that as well - specifically whether we should have a different way of configuring these kinds of "offline archive" data sources, since they need to behave somewhat differently than "online" tile sources. Might try out some ideas this weekend but I'll be traveling for holidays soon so apologies if I'm slow to respond!

core/src/data/mbtilesDataSource.cpp
rwrx: Ok, I have added it.

rwrx:
@matteblair I was thinking about extending `url` parameter to accept array of urls, but I wasn't sure if it will be consistent semantically, so I rather created new parameter.

core/src/data/mbtilesDataSource.cpp
tallytalwar: Thanks @rwrx

PR #2113, opened 2019-11-29
Tile fallback

rwrx:
Add fallback to first available lower zoom tile if tile is not available. Currently it does not work properly due to proxy tiles.


rwrx:
@matteblair I have checked this in my app using mbtiles and it seems to be working properly despite proxy tiles. Could I ask you to check it too when you will have some time if it is correct? Oh, I have also accidentally merged master into this branch. Is it somehow possible to remove commit 
cfebe82 from here or should I rather create new clean pull request for this?

PR #2118, opened 2019-12-15
Fix ClientDataSource losing features after scene change

### LINKED ISSUE #2117: ClientDataSources removed after sceneupdate
Hi all

I am currently facing issue regarding scene update. Whenever I add a clientsource and call loadSceneAsync to update a color, clientsource gets removed. I think this is due to the fact that loadSceneAsync makes a new scene and removes older tiles. Can you tell me what should I do to retain these clientsources.

Add any clientsource (In my case I have drawn a line)
Call scene update through loadSceneAsync 

**RESULT:**

Clientsource gets removed

**EXPECTED RESULT:**

Clientsource should not be removed no matter how many times I update the scene

matteblair:
When the current scene is changed, the `TileManager` instance used by the old scene runs `cancelTileTasks()` to stop pending asynchronous work. The implementation of this method also called `clearData()` on all of the managed `TileSource`s. `ClientDataSource::clearData()` not only removes the tiles generated from its stored features, but also removes the features themselves. This should definitely not happen when the scene changes. ~It's not clear why `clearData()` was used here at all, so I've simply removed the call.~

The behavior of `clearData()` (which is to clear caches) was mistakenly conflated with the need to clear the set of features added to a `ClientDataSource`. These two behaviors are both necessary, but should be separate. So I've given `ClientDataSource` a new method called `clearFeatures()` that removes added features, and made its `clearData()` implementation fall back to that of the base class. So now when a scene is reloaded and `clearData()` is invoked, the `ClientDataSource` features won't be cleared. But a client can explicitly clear the features through `MapData$clear`. 

This issue demonstrates how messy some of the `TileSource` abstraction is. A better, more complete solution would be to refactor these interfaces, but that will take more time.

Resolves #2117

### DIFF
--- a/core/include/tangram/data/clientDataSource.h
+++ b/core/include/tangram/data/clientDataSource.h
@@ -55,6 +55,9 @@ public:
     void addPolygonFeature(Properties&& properties, PolygonBuilder
         && polygon);
 
+    // Remove all feature data.
+    void clearFeatures();
+
     // Transform added feature data into tiles.
     void generateTiles();
 
@@ -62,7 +65,6 @@ public:
     std::shared_ptr<TileTask> createTask(TileID _tileId) override;
 
     void cancelLoadingTile(TileTask& _task) override {};
-    void clearData() override;
 
 protected:
 
--- a/core/src/data/clientDataSource.cpp
+++ b/core/src/data/clientDataSource.cpp
@@ -212,15 +212,12 @@ void ClientDataSource::loadTileData(std::shared_ptr<TileTask> _task, TileTaskCb 
     TileSource::loadTileData(_task, _cb);
 }
 
-void ClientDataSource::clearData() {
+void ClientDataSource::clearFeatures() {
 
     std::lock_guard<std::mutex> lock(m_mutexStore);
 
     m_store->features.clear();
     m_store->properties.clear();
-    m_store->tiles.reset();
-
-    m_generation++;
 }
 
 void ClientDataSource::addData(const std::string& _data) {
--- a/platforms/android/tangram/src/main/cpp/jniExports.cpp
+++ b/platforms/android/tangram/src/main/cpp/jniExports.cpp
@@ -529,7 +529,7 @@ void MapController(SetDefaultBackgroundColor)(JNIEnv* jnienv, jobject obj, jlong
     map->setDefaultBackgroundColor(r, g, b);
 }
 
-jlong MapController(AddTileSource)(JNIEnv* jniEnv, jobject obj, jlong mapPtr,
+jlong MapController(AddClientDataSource)(JNIEnv* jniEnv, jobject obj, jlong mapPtr,
                                    jstring name, jboolean generateCentroid) {
     auto_map(mapPtr);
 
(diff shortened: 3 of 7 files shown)


tallytalwar:
Thanks @matteblair for getting on this.

This is an interesting behavior. I thought and confirmed that `clearData` was also responsible for updating the `generation` of the data source (which is then used in the tileManager to determine when to update the old source's tileSet.. and also in tileWorker to determine which tile task wins in priority). 

I think this is going to cause issues during a scene update which is actually updating the datasource itself, and tileManager will continue to use old source (loaded) tiles, instead of throwing these away.

It might make sense to handle this in `ClientDataSource::clearData` overridden function, and have another function to do a clear of the features at appropriate scene update call (need to think more about when that would make sense, when a scene update changes the appropriate clientdatasource, maybe?)

matteblair:
I've reverted the original change and fixed this behavior in a different way now, so as to avoid unforeseen side effects with other data source types. 

As @tallytalwar suggested, the `ClientDataSource::clearData` implementation was a necessary behavior to have available, but is distinct from `TileSource::clearData` which it overrides. Now `ClientDataSource` uses the base implementation for `clearData` (which doesn't remove previously added features) and has a separate `clearFeatures` method (which explicitly removes added features).

While I was at it, I also managed to clean up some of the glue code between `ClientDataSource` and `MapData`.

tallytalwar:
@matteblair looks great to me. Just a couple of clarifications:
1. Can you update the description of the PR here, since we are still doing a `clearData` call when a scene is updated. So as it stands, clientdatasource will still be all "cleaned" up. If we have to give such control to the user, we can probably provide a flag with scene update, which will result in a call to `clearFeature` from `clientDataSource::clearData`. I am not sure if this is what @sheikh-iffi was running into, if so this PR won't solve it.

```
void ClientDataSource::clearData() {
     if (do_not_clear_data_but_feature_on_a_scene_update) {
          clearFeatures();
          return;
     }


    std::lock_guard<std::mutex> lock(m_mutexStore);
    std::lock_guard<std::mutex> lock(m_mutexStore);


    m_store->features.clear();
    m_store->features.clear();
    m_store->properties.clear();
    m_store->properties.clear();
    m_store->tiles.reset();

    m_generation++;
}
```

2. I see you removed all `mapPtr` `null` checks, is it because these methods are syncronized with MapController, such checks are immaterial?

tallytalwar:
> This issue demonstrates how messy some of the TileSource abstraction is. A better, more complete solution would be to refactor these interfaces, but that will take more time.

In full agreement. I find it over-complicated, but I think this was added when we refactored some work for mbtiles. Also in agreement that its a much bigger task.

matteblair:
> 1. Can you update the description of the PR here

Done - hopefully that makes the new behavior clearer. The original issue of `ClientDataSource` losing features when a scene is changed is still resolved.

> 2. I see you removed all mapPtr null checks

I don't think I did? I removed some unused methods in `MapController` and changed some `MapData` methods to no longer call into `MapController` at all (which removed the need to check some nulls).

matteblair:
Added a few more small-ish related changes. Curiously the message for https://github.com/tangrams/tangram-es/pull/2118/commits/c9b570d8baf42ffe06a398941f2ad0cc0ae02260 was supposed to be `MapData$clear also clears cached tiles` but I guess the terminal tried to expand `$clear` into a variable? 😆

tallytalwar:
Thanks for clearing all this up @matteblair. Merging.

PR #2119, opened 2019-12-17
Tile fallback

rwrx:
I have checked this in my app using mbtiles and it seems to be working properly despite proxy tiles are used. Could I ask you to check it too when you will have some time if it is working properly with mbtiles? From my testing everything worked, but maybe there is something that I am not aware of.

### DIFF
--- a/core/include/tangram/data/clientDataSource.h
+++ b/core/include/tangram/data/clientDataSource.h
@@ -50,10 +50,16 @@ public:
 
     void addPointFeature(Properties&& properties, LngLat coordinates);
 
-    void addPolylineFeature(Properties&& properties, PolylineBuilder&& polyline);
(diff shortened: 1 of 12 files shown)


matteblair:
This seems to be two separate changes: 1) "Add support for multiple mbtiles files through urls_mbtiles" and 2) "Add tile fallback". If these changes are unrelated, it would be best to submit them as two separate PRs, so that we can test and discuss them independently.

For the "multiple mbtiles" change, can you give an example use case and show how the new parameter would be used in YAML?

For the "tile fallback" change, can you describe:
    - The use case you are aiming to support
    - The current behavior in this use case, and how it doesn't meet your needs
    - The new behavior you've implemented

Thanks!

rwrx:
@matteblair thank you a lot for the answer. I am sorry but I have must created rather a new pull request #2120 and I added comments as you requested.

PR #2120, opened 2019-12-18
Tile fallback

rwrx:
Add support for getting tile for lower zoom if tile for higher zoom is not available.

Use case and current state:
User has mbtiles tile source which for some areas has tile data let's say only for zoom 10. Then he zooms to zoom 15 and the current state is that on the map he does not see anything. In this case it would be better to see a tile from zoom 10 rather than nothing.

New behavior:
When user want to show some location on the map and tile is not available in mbtiles, it would search tile parents until mbtiles contain some parent which does exist in mbtiles and then return `TileID` of that parent.

I have implemented it for `MBTilesDataSource` and `MemoryCacheDataSource`.

### DIFF
--- a/core/include/tangram/data/clientDataSource.h
+++ b/core/include/tangram/data/clientDataSource.h
@@ -50,10 +50,16 @@ public:
 
     void addPointFeature(Properties&& properties, LngLat coordinates);
 
-    void addPolylineFeature(Properties&& properties, PolylineBuilder&& polyline);
+    uint64_t addPolylineFeature(Properties&& properties, PolylineBuilder&& polyline);
 
     void addPolygonFeature(Properties&& properties, PolygonBuilder
         && polygon);
+
+    void updatePolylineFeature(uint64_t id, const Coordinates& coordinates);
+
+    void updatePolylineFeature(uint64_t id, const Properties&& properties);
+
+    void removePolylineFeature(uint64_t id);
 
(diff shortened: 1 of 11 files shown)


bcamper:
Is it intended to be an MBTiles-datasource-only feature?

matteblair:
This seems related to the `max_zoom` parameter for data sources - e.g. if you know that your tile source only has tiles up to zoom 10 then you can specify `max_zoom: 10` in the data source configuration and tiles will be "overzoomed" beyond zoom 10. However if the maximum zoom level of your tile source is not constant, then this isn't a good solution.

rwrx:
@bcamper I have currently implemented it for `MBTilesDataSource` and `MemoryCacheDataSource`.

@matteblair I know about `max_zoom` parameter but in my case the tiles which mbtiles contains are does not have constant `z` coordinate. Here and there may be tiles with another maximum `z` coordinate. Do you mean that this tile fallback is not a good solution in this case or your proposed solution of using `max_zoom`?

matteblair:
Oh my comment wasn't clear - I meant that if you don't have a constant maximum z across your data source, then `max_zoom` isn't a good solution. Some kind of "fallback" behavior like this makes more sense in that case.

rwrx:
So I will be waiting for your review of this pull request. I can also do any changes/updates to it,

PR #2142, opened 2020-02-01
Fix label collider tile scale

rwrx:
While I was testing tile fallbacking feature I discovered that when tile fallback (there is a pull request for tile fallback here: https://github.com/tangrams/tangram-es/pull/2120) is from zoom 16 to zoom 0 there just occurred an int 32 overflow in allocation https://github.com/tangrams/isect2d/blob/2e1a75cee09d9949900e926c61c86505b09205b2/include/isect2d.h#L60  . So while trying it and profiling memory usage on Android device I came out with a solution to not process labels when tile scale difference is larger or equal to 6 so it won't much consume RAM and also it won't crash on bad alloc exception. I am not sure if this proposed solution is a right one.


tallytalwar:
@hjanetzek or @karimnaaji would remember the code here the most. But I can give it a try. Thanks @rwrx

Issue #2138, opened 2020-01-27
Support for $meters_per_pixel


westnordost:
The Tangram Docs mention a parameter called <a href="https://tangrams.readthedocs.io/en/latest/Syntax-Reference/yaml/#meters_per_pixel">meters_per_pixel</a>. However, trying to use it currently results in this error message:

`EvalFilterFn: ReferenceError: identifier '$meters_per_pixel' undefined`

So it looks like it is not supported (yet).

Issue #1121, opened 2016-11-30
Feature Request: Support meters as size unit for points


westnordost:
I am trying to show the accuracy radius of a location around the user. Here, I ran into some problems. I think part of it is a bug, part of it is a missing feature probably.

The relevant portion of the style file:
```
styles:
    location-radius:
        base: points
        blend: overlay
        blend_order: 2
layers:
    streetcomplete_location:
        data: { source: streetcomplete_location }
        draw:
            location-radius:
                collide: false
                color: '#44fd6c3e'
                size: function() { return feature.radius }
```
and the code
```
locationLayer.clear();
LngLat pos = new LngLat(location.getLongitude(), location.getLatitude());
int accuracy = (int)(Math.ceil(location.getAccuracy()));

Map<String, String> props = new HashMap<>();
props.put("type", "point");
props.put("radius", ""+accuracy+"m");
locationLayer.addPoint(pos, props);
```

However, this gives i.e. the error:
`styleParam.cpp:184: Invalid size parameter '22m'`

If I set a static size in the style definition, I get no error, but it doesn't seem to show at all.

So my guess is that meters as a size unit for points is simply not implemented as there hasn't been a use case for this (until now :-P). At least it is not documented. That Tangram-ES does not report an error if you write `size: 50m` seems to be a bug then.
In any case, this ticket shall be a feature request to support meters as a unit for points.

Plan B for the use case I have in mind would be to use a circle-polygon for the radius, but I find that rather ugly.

westnordost:
Perhaps another use case would be tree-nodes.

Guillaume227:
@westnordost : one other posssible workaround would be to use a circular marker together with a js function for the size. That function would back out the size in pixel from the zoom level so it remains constant in meters. I haven't tested that yet.

westnordost:
Jup, I did that now, looks pretty nice and feels smooth. I call updateAccuracy whenever the zoom changes.

```
private void updateAccuracy()
{
	if(accuracyMarker != null && lastLocation != null)
	{
		if(lastZoom == controller.getZoom()) return;
		LngLat pos = new LngLat(lastLocation.getLongitude(), lastLocation.getLatitude());
		float size = meters2Pixels(pos, lastLocation.getAccuracy());
		accuracyMarker.setStyling("{ style: 'points', color: 'white', size: ["+size+"px, "+size+"px], order: 2000, collide: false }");
		lastZoom = controller.getZoom();
	}
}

[...]

PR #2144, opened 2020-02-03
Add $meters_per_pixel keyword for filters and JS functions

matteblair:
- Add the $meters_per_pixel keyword, determine its value from the current zoom in StyleContext, and make it accessible to filter expressions and JS functions.
- Clean up the logic for keywords in StyleContext - setting arbitrary keywords is no longer supported.
- Updated tests for changes to StyleContext and added a case using $meters_per_pixel in a JS function.

Resolves https://github.com/tangrams/tangram-es/issues/2138

### DIFF
--- a/bench/src/benchStyleContext.cpp
+++ b/bench/src/benchStyleContext.cpp
@@ -104,7 +104,7 @@ struct JSTileStyleFnFixture : public benchmark::Fixture {
     void SetUp(const ::benchmark::State& state) override {
         globalSetup();
         ctx.initFunctions(*scene);
-        ctx.setKeywordZoom(10);
+        ctx.setZoom(10);
     }
     void TearDown(const ::benchmark::State& state) override {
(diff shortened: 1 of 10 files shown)


westnordost:
Very cool! :-D

Could it be that with this change, you also changed unintentionally or not the `$zoom` keyword from an int value to a float value?

matteblair:
The variable that provides the value of `$zoom` was changed from an `int` to a `double` to reduce casting, but it still holds an integer value (because filters and styling functions are only applied at integer zoom levels).

westnordost:
Oh, that means that meters_per_pixel only changes per integer zoom level?

matteblair:
Yes. Filters and styling functions are evaluated on features once, using the zoom level of the tile that contains them.

westnordost:
Oh, too bad, so the use case I had in mind for this will not work (#1121)

PR #2149, opened 2020-02-24
Fix subclassing DefaultHttpHandler in Kotlin

### LINKED ISSUE #2145: Architectural issue of DefaultHttpHandler + Kotlin
I want to notify you of a compatibility issue with the architecture of `DefaultHttpHandler` and Kotlin

Let's take for example this subclass of `DefaultHttpHandler`:

```kotlin
class CachingHttpHandler(
    private val cacheDirectory: File?,
    private val maxCacheSizeInBytes: Long = 16 * 1024 * 1024
) : DefaultHttpHandler() {

    override fun configureClient(builder: OkHttpClient.Builder) {
        if (cacheDirectory?.exists() == true) {
            builder.cache(Cache(cacheDirectory, maxCacheSizeInBytes))
        }
    }
}

```

Believe it or not, `cacheDirectory` is always null and even `maxCacheSizeInBytes` is always 0 in `configureClient`, no matter what was passed as constructor parameter. Why is this the case? Because in Kotlin, first always the constructor of the parent class is executed, after that the constructor of the child class. In the constructor of the child class implicitly the passed parameters are put into the private fields.
Since the parent class calls `configureClient` in its own constructor, this call happens before these values are initialized.
This is a problematic language design of Kotlin at this point, but you can solve this issue for Kotlin users by not calling `configureClient` in the constructor but lazily just before the first request is made.

matteblair:
Resolves #2145 

As suggested in the reported issue, this change makes `DefaultHttpHandler` instantiate its `OkHttpClient` lazily during the first call to `startRequest`, so that a subclass in Kotlin will be fully initialized when `configureClient` is called.

I also considered other ways of resolving this, such as replacing the override-able `configureClient` and `configureRequest` methods with delegate objects. The lazy-instantiation approach has the smallest impact on existing user code, so I've opted for this until it shows problems.

### DIFF
--- a/platforms/android/tangram/src/main/java/com/mapzen/tangram/networking/DefaultHttpHandler.java
+++ b/platforms/android/tangram/src/main/java/com/mapzen/tangram/networking/DefaultHttpHandler.java
@@ -29,61 +29,14 @@ public class DefaultHttpHandler implements HttpHandler {
 
     private OkHttpClient okClient;
 
-    /**
-     * Construct an {@code DefaultHttpHandler} with default options.
-     */
-    public DefaultHttpHandler() {
-        final OkHttpClient.Builder builder = new OkHttpClient.Builder()
-                .followRedirects(true)
-                .followSslRedirects(true)
-                .connectTimeout(10, TimeUnit.SECONDS)
-                .writeTimeout(10, TimeUnit.SECONDS)
-                .readTimeout(30, TimeUnit.SECONDS);
-
-
-        builder.socketFactory(new SocketFactory() {
-            SocketFactory factory = SocketFactory.getDefault();
-
-            @Override
-            public Socket createSocket() throws IOException {
-                Socket s = factory.createSocket();
-                try {
-                    //Log.d("Tangram", "Patching socket nodelay: " + s.getTcpNoDelay());
-                    s.setTcpNoDelay(true);
-                } catch (SocketException e) {
-                    // empty
-                }
-                return s;
-            }
-
-            @Override
-            public Socket createSocket(String s, int i) throws IOException, UnknownHostException {
-                return null;
-            }
-
-            @Override
-            public Socket createSocket(String s, int i, InetAddress inetAddress, int i1) throws IOException, UnknownHostException {
-                return null;
-            }
-
-            @Override
-            public Socket createSocket(InetAddress inetAddress, int i) throws IOException {
-                return null;
-            }
-
-            @Override
-            public Socket createSocket(InetAddress inetAddress, int i, InetAddress inetAddress1, int i1) throws IOException {
-                return null;
-            }
-        });
-
-        configureClient(builder);
-
-        okClient = builder.build();
-    }
-
     @Override
     public Object startRequest(@NonNull final String url, @NonNull final HttpHandler.Callback cb) {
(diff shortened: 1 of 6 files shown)


platforms/android/tangram/src/main/java/com/mapzen/tangram/networking/DefaultHttpHandler.java
```diff
@@ -29,61 +29,14 @@
...
-        okClient = builder.build();
-    }
-
     @Override
     public Object startRequest(@NonNull final String url, @NonNull final HttpHandler.Callback cb) {
+        // Build the OkHttpClient lazily. If we run configureClient() in the constructor then
+        // subclasses in Kotlin won't yet be initialized.
+        if (okClient == null) {
+            okClient = buildClient();
+        }
```
westnordost:
```suggestion
        if (okClient == null) {
            synchronized {
                if (okClient == null) {
                    okClient = buildClient();
                }
            }
        }
```
(or synchronize on a lock object)

If an instance of this class should be usable by several threads at once (i.e. a threadpool), the lazy initialization would need to be threadsafe.

matteblair: Yeah good point - this class is definitely used by multiple threads within Tangram. I've updated the change.

matteblair:
Finally getting back to this! I appreciate the patience 🙏 

After thinking on the point about multiple threads using this class I feel that lazy initialization is not a good solution. First because it requires synchronization, as mentioned in the comment. Second because the initialization is likely to be performed on the render thread, where we definitely don't want to possibly block.

Instead, I've now just added a `DefaultHttpHandler` constructor that takes a `OkHttpHandler.Builder` argument and a static `GetClientBuilder()` method that returns the same builder object that would have been passed into `configureClient()`. Now `configureClient()` is deprecated, since this constructor does the same job. 

The Kotlin implementation will require a small change to use this, but overall this solution is much simpler and more maintainable.

platforms/android/tangram/src/main/java/com/mapzen/tangram/networking/DefaultHttpHandler.java
```diff
@@ -22,63 +22,40 @@
...
 
     private OkHttpClient okClient;
 
     /**
-     * Construct an {@code DefaultHttpHandler} with default options.
+     * @return OkHttp client builder with recommended settings for Tangram.
      */
-    public DefaultHttpHandler() {
-        final OkHttpClient.Builder builder = new OkHttpClient.Builder()
+    public static OkHttpClient.Builder GetClientBuilder() {
```
westnordost: As far as I know, the naming convention for Java is that methods should never start with an uppercase letter (also not static methods). I suggest to call the method `createClientBuilder` (or `createDefaultClientBuilder`)

westnordost:
Looks good

platforms/android/tangram/src/main/java/com/mapzen/tangram/networking/DefaultHttpHandler.java
matteblair: Oops! Good catch - I've been writing too much C# recently

PR #2166, opened 2020-06-04
try to make sentences simpler and less confusing

matkoniecz:
"targeting mobile and embedded device" - it was unclear whatever it referred to Tangram or Tangram ES

hopefully I deconfused it in a correct way


matteblair:
I see now how one could misunderstand that phrasing. Complete sentences are generally better for clarity though.

What about revising the second sentence to just say "It is a counterpart to Tangram"? Our targets are not actually limited to "mobile and embedded devices" since we have good support for macOS and Linux desktop platforms.

matkoniecz:
Changed.

Issue #2167, opened 2020-06-12
Incorrect icon marker channels on iOS


rwrx:
I have tried to add an icon to a marker in Tangram Swift Sample project - https://github.com/tangrams/tangram-swift-sample. Running it on iPhone 8, I got this result:

![screenshot](https://user-images.githubusercontent.com/2378743/84544052-cda37880-acfc-11ea-9eba-47eed9f3a0af.jpg)

The icon seems to have channels changed. I have found out that it is due to this for loop: https://github.com/tangrams/tangram-es/blob/514e9528b5936658527f111fada3273abcea779e/platforms/ios/framework/src/TGMarker.mm#L188

When I commented out this for loop, the icon seems to have correct channels.

This is the icon:

![location_icon](https://user-images.githubusercontent.com/2378743/84544661-009a3c00-acfe-11ea-9811-e04af189dccb.png)

and the Swift code I have added into Tangram Swift Sample into `ViewController.swift` file:

```
let newYork = CLLocationCoordinate2DMake(40.70532700869127, -74.00976419448854)
let marker = mapView.markerAdd()
marker.stylingString = "{ style: 'ux-location-gem-overlay', color: 'white', size: [35px, 35px], order: 10000, priority: 1, collide: false }"
marker.icon = UIImage(named: "location_icon")!
marker.point = newYork
```

matteblair:
That's very odd! This loop is for converting pixel formats in a way that was clearly intentional - I wonder if something changed in iOS? Can you tell me what version of iOS you are using?

rwrx:
I only have one device - iPhone 8, so unfortunately I cannot test it on other devices. It has iOS version 13.5.1. It should be the latest version.

matteblair:
I've reproduced the behavior you described. I'm still puzzled as to why the code worked this way in the first place! I'll see if I can dig up info in the repo history, but in the meantime I've drafted the fix here: https://github.com/tangrams/tangram-es/pull/2168

rwrx:
@matteblair thank you a lot for addressing this. So I can close this issue. Could I also ask you to have a look at https://github.com/tangrams/tangram-swift-sample/issues/3 when you will have some time?

PR #2168, opened 2020-06-15
Fix TGMarker bitmap channel swap

matteblair:
Addresses https://github.com/tangrams/tangram-es/issues/2167

For some reason, the code that processes bitmap data in `TGMarker` assumed that incoming bitmaps were in BGRA format, and so it swapped the channels to make it RGBA. This seems to be incorrect now, as it causes marker bitmaps to render with incorrect colors.

This PR removes the BGRA->RGBA conversion, while keeping the alpha un-pre-multiplication, which I believe is still necessary.

### DIFF
--- a/platforms/ios/framework/src/TGMarker.mm
+++ b/platforms/ios/framework/src/TGMarker.mm
@@ -170,7 +170,7 @@
 
     bitmap.resize(w * h);
 
-    CGColorSpaceRef colorSpace = CGImageGetColorSpace(cgImage);
+    CGColorSpaceRef colorSpace = CGColorSpaceCreateDeviceRGB();
     // iOS only supports contexts with pre-multiplied alpha, so we transform it below.
     CGContextRef cgContext = CGBitmapContextCreate(bitmap.data(), w, h, 8, w * 4,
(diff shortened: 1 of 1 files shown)


matteblair:
After taking another look at this, my best guess at why the previous code converted from BGRA pixels is that the CoreGraphics context used for the conversion was drawing pixels in the color space of the _image_, which may have been a BGR color space for the images originally tested. Instead, the context should always be drawing pixels in an RGB color space. I've now applied that change and tested that it produces expected results for both a JPG and a PNG with transparency.

matteblair:
I also added a correction to the math for un-premultiplying alpha that I spotted after checking the code that does this in stb_image.h (https://github.com/nothings/stb/blob/master/stb_image.h#L4910-L4925). Without this correction, colors with partial transparency will be off by one after going through the premultiply and un-premultiply operations.

Example:
- Original color has R = 100, A = 128
- Pre-multiply alpha, now R = 100 * 128 / 255 = 50
- Un-premultiply alpha, now R = 50 * 255 / 128 = 99
- Un-premultiply alpha, but with "1/2" added to correct for truncation, now R = (50 * 255 + 128 / 2) / 128 = 100

Issue #2165, opened 2020-05-26
MapController.lngLatToScreenPosition returns point 0,0 for certain parameters


westnordost:
**REPRODUCTION:**

Here is an example state for which it is reproducible.

Call `MapController.lngLatToScreenPosition` for

- `latitude = 37.80455166666666`
- `longitude = -122.30210333333335`

whilst the camera is at

- `latitude = 37.8034503901695`
- `longitude = -122.30177937658814`
- `rotation = 3.4500046`
- `tilt = 0.62831855`
- `zoom = 19.624928`

It incorrectly returns a point with

- `x = 0`
- `y = 0`

The used smartphone is a Sony Xperia Z1 Compact, in case screen size and pixel density play a role in this. If you want to reproduce it with a different device and you can't with these values, try zooming in further or moving the camera.

**EXPECTED BEHAVIOR:**

Here is an example state where it works. **Note, ALL the parameters are exactly the same as above, except that the zoom is a little lower.** So, the two examples are each at the edge of "result is correct" and "result is wrong". Hope that helps. 

Call `MapController.lngLatToScreenPosition` for

- `latitude = 37.80455166666666`
- `longitude = -122.30210333333335`

whilst the camera is at

- `latitude = 37.8034503901695`
- `longitude = -122.30177937658814`
- `rotation = 3.4500046`
- `tilt = 0.62831855`
- **`zoom = 19.341661`** <- this changed

It returns a point with

- **`y = 16659.263671875`** <- correct resultnow
- **`x = -1016.594482421875`**

**ENVIRONMENT:**

- Android 5.1, Sony Xperia Z1 Compact, tangram-es 0.12.0

**FURTHER INFORMATION:**

- It is not (only) related to the zoom level. In fact, if changing the rotation, the tilt or position offset beyond a certain value, the calculated point becomes correct again.
- **I could only ever reproduce it if the tilt is not 0**

westnordost:
(I added two more perhaps helpful observations to the further information section. Also if you would like to have something visual, follow the linked bug)

westnordost:
If you would like an example app for which you could play around with tilt, zoom, rotation etc while `MapController.lngLatToScreenPosition` is constantly called on each map movement with a clear visual indication if a calculation was wrong, you can take the current master of StreetComplete and do the following:
1. (Use your debug build of tangram-es instead of the maven dependency.)
[...]

PR #2179, opened 2020-07-13
Enable clipping to viewport when converting coordinates to screen position

matteblair:
This addresses some uncommon situations where converting a geographic coordinate to a screen position does not have a meaningful result, as described in https://github.com/tangrams/tangram-es/issues/2165

This change adds a method to convert coordinates to screen positions "clipped" to the viewport, internally resolving the problematic edge cases.

### DIFF
--- a/core/include/tangram/map.h
+++ b/core/include/tangram/map.h
@@ -308,9 +308,7 @@ public:
     // Given longitude and latitude coordinates, set the output coordinates to the
     // corresponding point in screen space (x right, y down); returns false if the
     // point is not visible on the screen, otherwise returns true
-    bool lngLatToScreenPosition(double _lng, double _lat, double* _x, double* _y);
-
-    bool lngLatToScreenPositionClipped(double lng, double lat, float* x, float* y);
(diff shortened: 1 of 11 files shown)


matteblair:
To summarize what this looks like for SDK users:

On Android, MapController has a new method that converts a LngLat into a screen position as a PointF and can optionally clip the result to the viewport. The return value is true when the original point is in the current viewport.
```java
public boolean lngLatToScreenPosition(@NonNull final LngLat lngLat, @NonNull final PointF screenPositionOut, boolean clipToViewport)
```

On iOS, TGMapView has one changed method and one new method (because out-parameters translate to Swift very poorly). First, the signature for `viewPositionFromCoordinate` now takes a boolean option for clipping to the viewport.
```objective-c
- (CGPoint)viewPositionFromCoordinate:(CLLocationCoordinate2D)coordinate clipToViewport:(BOOL)clip;
```
There's a separate method to check if a coordinate is in the current viewport.
```objective-c
-(BOOL)viewportContainsCoordinate:(CLLocationCoordinate2D)coordinate;
```

PR #2180, opened 2020-07-26
Faster macOS CI builds

matteblair:
Working on making this build fast enough to run on every push.


matteblair:
With ccache, the macOS build is now under 2 minutes! I think this is fast enough to run on every push now. I'm also going to try using ccache on other jobs too.

tallytalwar:
Bravo.. from Travis instances taking hours to even getting picked up to osx co builds under 2 mins. Journey!!

PR #2185, opened 2020-08-20
Add OPEN_FULLMUTEX flag when opening mbtiles

rwrx:
On Android without this flag showing maps from mbtiles works properly as somewhere I read that on Android SQLite is opened with thread-safety, but unfortunately on iOS without this flag showing maps from mbtiles crashes, so this adds thread-safety for iOS devices.


rwrx (commented):
Removed

matteblair:
Hey @rwrx, thanks for your patience with this PR! The change to update SQLiteCpp and apply the OPEN_FULLMUTEX flag looks good. I tried reverting the CMake changes because the iOS build was working on my machine without those flags. I think there may be a bug in CMake 3.17 related to iOS builds because it works fine using CMake 3.18.

rwrx:
@matteblair ok, I understand, thank you.

Issue #2177, opened 2020-07-09
Offset of marker behaves differently than offset of element in layer


westnordost:
I have been reluctant to report this because there is already such a stack of open bugs here plus I couldn't pin-point the problem until today. However, the following screenshots might give (you) more insight.

### Bug description

**The offsets given in the draw block in YAML seem to not be adjusted (correctly) to screen DPI, unlike the sizes of the icons and other sizes.**

The following screenshots are each made on Android each with screens set to different DPI. 
Depending on the DPI of the device, the outcome is different. In Android, there is a developer setting ("smallest width") in which one can simulate different DPIs on the same device:

![](https://user-images.githubusercontent.com/43007630/86927979-43500800-c134-11ea-826e-6d8294f90a3e.png)
**smallest width** : large (320) - normal (360) - small (400) - smaller (??) - smallest (514)

### Further information

There are three elements visible on the screen:
- `selected-quest-icons` is the pin on which a house can be seen
- `quest-dots` is the white dot that *should be* placed just so that it is obscured by the lower end of the pin
- `quest-selection` is a marker consisting of a blue circle and should be placed so that it surrounds the icon part of the pin

This is the relevant stylesheet:

```yaml
layers:
    streetcomplete_selected_quests:
        data: { source: streetcomplete_selected_quests }
        draw:
            selected-quest-icons:
                size: 66px
                sprite: function() { return feature.kind }
                sprite_default: quest
                collide: false
                offset: [-17px, -66px]
    streetcomplete_quests:
        dots_no_collide:
            filter: { $zoom: { min: 16 } }
            draw:
                quest-dots:
                    priority: 1
                    order: 1
                    size: 16px
                    collide: false
                    offset: [-2.5px, -22px]
```
and
```kotlin
val frame = ctx.resources.getBitmapDrawable(R.drawable.quest_selection_ring)
val w = frame.intrinsicWidth.toFloat().toDp(ctx)
val h = frame.intrinsicHeight.toFloat().toDp(ctx)

val marker = controller?.addMarker() ?: return null
marker.setStylingFromString("""{
    style: 'quest-selection',
    color: 'white',
    size: [${w}px, ${h}px],
    flat: false,
    collide: false,
    offset: ['1px', '-78px']
}""")
```

In case it matters, the styles for these elements are:

```yaml
styles:
    quest-selection:
        base: points
        blend: overlay
        blend_order: 3
    selected-quest-icons:
        base: points
        texture: quests
        blend: overlay
[...]

PR #2186, opened 2020-09-08
Fix DPI scaling on 'offset' parameter for points.

matteblair:
Resolves https://github.com/tangrams/tangram-es/issues/2177

The change that actually addresses this issue is https://github.com/tangrams/tangram-es/commit/6a6a8daf6d2fb2fd09c7bee2d4b29965556066c5, the rest is just cleanup and convenience functionality that happened while I was debugging.


tallytalwar (approved):
Thanks @matteblair . LGTM

westnordost:
Nice!

So what was the bug? Offset was assumed to be pixel scale ==1?

matteblair:
Yeah essentially. Labels drawn using the "Points" style were missing the step where pixel scale is applied to "offset". The "Text" style was applying pixel scale correctly, so the problem became very clear when multiple "Points" and "Text" meshes were drawn together.

Issue #2215, opened 2020-12-02
Memory allocation failure in MapController init


nikhiljha:
**TO REPRODUCE THE ISSUE, FOLLOW THESE STEPS:**

- Use an Android phone with GrapheneOS
- Run an app that uses the tangram library, 0.13.0

**RESULT:**

https://github.com/tangrams/tangram-es/blob/c4b024539f07dff1addc2e6424e7a535a1f0c191/platforms/android/tangram/src/main/java/com/mapzen/tangram/NativeMap.java#L10-L15

```
12-02 10:04:08.935 26927 26927 E AndroidRuntime: FATAL EXCEPTION: main
12-02 10:04:08.935 26927 26927 E AndroidRuntime: Process: de.westnordost.streetcomplete, PID: 26927
12-02 10:04:08.935 26927 26927 E AndroidRuntime: java.lang.RuntimeException: Unable to create a native Map object! There may be insufficient memory available.
12-02 10:04:08.935 26927 26927 E AndroidRuntime:        at com.mapzen.tangram.MapController.<init>(MapController.java:163)
12-02 10:04:08.935 26927 26927 E AndroidRuntime:        at com.mapzen.tangram.MapView.getMapInstance(MapView.java:276)
12-02 10:04:08.935 26927 26927 E AndroidRuntime:        at com.mapzen.tangram.MapView.initMapController(MapView.java:173)
12-02 10:04:08.935 26927 26927 E AndroidRuntime:        at com.mapzen.tangram.MapView$1.onLibraryReady(MapView.java:120)
12-02 10:04:08.935 26927 26927 E AndroidRuntime:        at com.mapzen.tangram.MapView$1InitTask.onPostExecute(MapView.java:236)
12-02 10:04:08.935 26927 26927 E AndroidRuntime:        at com.mapzen.tangram.MapView$1InitTask.onPostExecute(MapView.java:224)
12-02 10:04:08.935 26927 26927 E AndroidRuntime:        at android.os.AsyncTask.finish(AsyncTask.java:771)
12-02 10:04:08.935 26927 26927 E AndroidRuntime:        at android.os.AsyncTask.access$900(AsyncTask.java:199)
12-02 10:04:08.935 26927 26927 E AndroidRuntime:        at android.os.AsyncTask$InternalHandler.handleMessage(AsyncTask.java:788)
12-02 10:04:08.935 26927 26927 E AndroidRuntime:        at android.os.Handler.dispatchMessage(Handler.java:106)
[...]

PR #2217, opened 2020-12-10
Don't throw when native pointers have long values < 0

matteblair:
Android 11 and certain ROMs use "tagged pointers" that set the high bits on pointers, causing them to become negative as a long.

See https://github.com/tangrams/tangram-es/issues/2215

### DIFF
--- a/platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
+++ b/platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
@@ -670,7 +670,7 @@ public class MapController {
             return mapData;
         }
         final long pointer = nativeMap.addClientDataSource(name, generateCentroid);
-        if (pointer <= 0) {
+        if (pointer == 0) {
             throw new RuntimeException("Unable to create new data source");
         }
         mapData = new MapData(name, pointer, this);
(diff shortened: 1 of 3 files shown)


platforms/android/tangram/src/main/java/com/mapzen/tangram/MapController.java
```diff
@@ -670,7 +670,7 @@ public MapData addDataLayer(final String name, final boolean generateCentroid) {
             return mapData;
         }
         final long pointer = nativeMap.addClientDataSource(name, generateCentroid);
-        if (pointer <= 0) {
+        if (pointer == 0) {
```
bcamper, line 673: Not to be pedantic, but wouldn't it clarify intention here if it was just `pointer == NULL`? e.g. `pointer <= NULL` wouldn't have made sense to a casual reader.

matteblair: Yep, Java's `null` is not the same as `NULL` in C/C++ and it can't be compared with a `long`.

bcamper: sad

PR #2238, opened 2021-02-23
Fix pinch gesture handling and invalid scales

### LINKED ISSUE #2169: Rarely MapController.getCameraPosition().longitude returns NaN

Call `MapController.getCameraPosition().longitude`. Very rarely, it returns `NaN`.
Reproduction is unknown, I got a few crash reports because the app expects valid lat/lon pairs.

It happened once roughly when the map was displayed near latitude 48.864143, longitude 16.469961, quite an innocent position.

**ENVIRONMENT:**

at least Android 10, tangram-es 0.12.0

**OTHER:**

I am in contact with the person for which this happened several times. Should I ask him anything specific?

tallytalwar:
- Based on how scale for a handlePinchGesture is calculated, its
possible its negative (or zero). We need to guard against this!

Fixes #2169

### DIFF
--- a/core/src/util/inputHandler.cpp
+++ b/core/src/util/inputHandler.cpp
@@ -100,9 +100,15 @@ void InputHandler::handleFlingGesture(float _posX, float _posY, float _velocityX
 void InputHandler::handlePinchGesture(float _posX, float _posY, float _scale, float _velocity) {
     cancelFling();
 
+    // Early out for an invalid scale
+    if (_scale == 0.f) {
+        return;
+    }
     float z = m_view.getZoom();
     static float invLog2 = 1 / log(2);
-    m_view.zoom(log(_scale) * invLog2);
+
+    // Make sure log is valid with absolute value of scale
+    m_view.zoom(log(abs(_scale)) * invLog2);
 
     m_view.screenToGroundPlane(_posX, _posY);
     float s = pow(2, m_view.getZoom() - z) - 1;
--- a/core/src/util/inputHandler.cpp
+++ b/core/src/util/inputHandler.cpp
@@ -101,14 +101,15 @@ void InputHandler::handlePinchGesture(float _posX, float _posY, float _scale, fl
     cancelFling();
 
     // Early out for an invalid scale
-    if (_scale == 0.f) {
+    if (_scale <= 0.f) {
+        // 'scale' should be a ratio between two sizes with the same sign.
         return;
     }
+
     float z = m_view.getZoom();
     static float invLog2 = 1 / log(2);
 
-    // Make sure log is valid with absolute value of scale
-    m_view.zoom(log(abs(_scale)) * invLog2);
+    m_view.zoom(log(_scale) * invLog2);
 
     m_view.screenToGroundPlane(_posX, _posY);
     float s = pow(2, m_view.getZoom() - z) - 1;


tallytalwar:
This fixes the glfw App which was at least showing negative scale values. Though I am not sure how an android platform could return a negative (or zero) scale value (https://developer.android.com/reference/android/view/ScaleGestureDetector#getScaleFactor()).

I do not have a working android phone, if you could test this @matteblair?

core/src/util/inputHandler.cpp
```diff
@@ -100,9 +100,15 @@ void InputHandler::handleFlingGesture(float _posX, float _posY, float _velocityX
...
+    // Early out for an invalid scale
+    if (_scale == 0.f) {
+        return;
+    }
     float z = m_view.getZoom();
     static float invLog2 = 1 / log(2);
-    m_view.zoom(log(_scale) * invLog2);
+
+    // Make sure log is valid with absolute value of scale
+    m_view.zoom(log(abs(_scale)) * invLog2);
```
matteblair, line 111: If `_scale` is negative, zooming using the absolute value is probably not the expected result. Maybe instead we should exit early when `_scale <= 0`?

matteblair (approved):
Nice find - and thanks for the repro case in GlfwApp too.

I'll test this out on Android and report back!

tallytalwar:
Thanks Matt.

Also I see a few places were we can do better wrt avoiding NaN maths! I will try to address these soonish, but if you get to it before me, these are couple of interesting places I found where illegal math operations could happen because introduction of NaNs in the system.
- https://github.com/tangrams/tangram-es/blob/main/core/src/view/view.cpp#L448
- https://github.com/tangrams/tangram-es/blob/main/core/src/util/inputHandler.cpp#L31 (will happen everytime cancelFling gets called, but has no sideeffects because NaN comparison fails in subsequent lines).. however an early out should be preferred instead of introducing a NaN.

core/src/util/inputHandler.cpp
tallytalwar: Probably! But I did not see any adverse side effect of this in my very limiting testing. I have to think this through but I feel it might be fully accurate to simply ignore negative scale values.. Do you have a good reasoning behind ignoring negative scales, if so if you could take over and update with a code comment explaining the same? Thanks

matteblair:
I went ahead and made the change to ignore all `_scale` arguments that are `<= 0`. I explained briefly in a comment, but the reason is that the 'scale' here is supposed to be a ratio between two measurements of size. Since negative size isn't meaningful to this function, it implies an error (or some kind of geometric inversion that I don't even want to think about).

It's clear how the scrolling logic in `GlfwApp` would have produced non-positive scale values, but I'm still not sure how this happens on Android! I'll bet it's due to some vendor-specific code in gesture detectors.

tallytalwar:
Yeah I am also unsure about why android is behaving like this as mentioned in my previous comment. Lets take it from here and see if this resolves the errors @westnordost is seeing. (Thanks @westnordost for the clue on a great repro for this).

tallytalwar (commented):
LGTM. Merging

westnordost:
Nice! Do you plan to do a bugfix release for that, or will you bundle this fix together with other changes in a later version?

matteblair:
I'm planning to make a minor release with this fix before changing the publishing destinations, so that users still on jcenter can get this fix.

Issue #2234, opened 2021-02-10
Release artifacts need to move off of Bintray


matteblair:
JFrog is ending service for Bintray and JCenter, which is where we currently host our release artifacts for Android and iOS.

https://jfrog.com/blog/into-the-sunset-bintray-jcenter-gocenter-and-chartcenter

We need to move these artifacts onto a different service before May 1, 2021.

westnordost:
In Bird culture, this is considered a dick move.

Anyway, as I also have to migrate some libraries away from Dicktray, I read a bit into this, here is some useful info about it from [gradle.org](https://blog.gradle.org/jcenter-shutdown):

> Impact to packages publishing to Bintray
> 
> If you are publishing your builds to Bintray using the [com.jfrog.bintray](https://github.com/bintray/gradle-bintray-plugin) plugin, you’ll need to find another repository to host your package. You may be able to publish using the [built-in maven-publish](https://docs.gradle.org/current/userguide/publishing_setup.html#header).
> 
> If you’re migrating to Maven Central, the [io.github.gradle-nexus.publish-plugin](https://github.com/gradle-nexus/publish-plugin) plugin automates publishing to Nexus and Maven Central.

Also important for anyone **using** JCenter to date to get dependencies is this security consideration:

> As we have seen in recent years, dependency confusion and namespace collisions can be serious dangers for supply chain attacks. Each of these vulnerabilities relied on untrusted artifacts being used in place of trusted ones.
> 
> When you change the list of repositories you use to resolve dependencies, you may inadvertently expose yourself to these kinds of attacks.

Gradle offers a mechanism to defend against this, [described in their user guide](https://docs.gradle.org/current/userguide/dependency_verification.html#sub:enabling-verification). In a nutshell:

Execute `gradle --write-verification-metadata sha256 help` to bootstrap checksum verification of dependencies, then after that you can change the repository to fetch dependencies from and gradle will throw an error if the checksum of a dependency now located somewhere else is different.

PR #2239, opened 2021-03-01
Deploy Android artifacts to Maven Central via OSS Sonatype Nexus

matteblair:
This addresses the Android part of https://github.com/tangrams/tangram-es/issues/2234

We previously hosted Android artifacts with Sonatype and the configuration and credentials are still valid. This change is mostly a reversal of these two commits: b69d5956ec312cae1a3c22a906f5ac35aad912c4, 3fbfb97c9b8d74d5b8f16afd061bfef0368bfb9d.

The only non-trivial change was configuring signing, which is required for publishing to Maven Central. Since CircleCI does not allow you to provide files to a build as secrets, I decided to encode our GPG secret keyring file as base64, put the base64 text in an environment variable, and then `echo` and decode that variable to a file in the build workspace.

I've tested uploads of both snapshot and release artifacts using this setup locally and it all seems to work. All in all, not as bad as I had feared!


tallytalwar:
> I decided to encode our GPG secret keyring file as base64, put the base64 text in an environment variable, and then echo and decode that variable to a file in the build workspace.

Very smart :D

release-checklist.md
```diff
@@ -77,8 +79,8 @@ $ git push
 ### 6. Document release
 Document release notes at [https://github.com/tangrams/tangram-es/releases][5].
 
-[1]: https://circleci.com/gh/tangrams/tangram-es
-[2]: https://travis-ci.org/tangrams/tangram-es
-[3]: https://bintray.com/tangrams/maven/tangram
-[4]: https://bintray.com/bintray/jcenter
+[1]: https://app.circleci.com/pipelines/github/tangrams/tangram-es
+[2]: https://oss.sonatype.org/
+[3]: https://central.sonatype.org/pages/ossrh-guide.html
```
tallytalwar, line 84: If I remember this correctly (at least a couple of years since we did this), this has the manual step of "releasing" the artifact right?

tallytalwar (approved):
LGTM @matteblair .

Once we get to this, could you remove bintray environment variables from circle?

Issue #2235, opened 2021-02-11
Translucent background 


soheilarmin:
Currently, background opacity is ignored. Is there anyway we can have a background with alpha channel, or optionally entirely remove the background?

matteblair:
Hello! There currently is not a way to change the opacity of the map background, but it may be possible to add this as a feature. 

Can you tell me which platform(s) you are targeting and describe your use case in more detail?

soheilarmin:
Hi,
I'm using Tangram-es on Android. From my understanding, I need to change the color depth to 8,8,8,8. I'm not sure about all other changes that should take place. It would be nice if you can help. There are many many good features (like shaders) that Mapbox lacks of, so I'm deciding to use Tangram over Mapbox but this single feature about translucent background is really needed for my use case. 
I need to blend the map in several applications (like a media player). Those application have very simple backgrounds and map objects should be drawn over those background images. So, I will need to remove the current solid background color. As far as I know, I cannot apply a fragment shader on the background.

PR #2244, opened 2021-03-08
Enable translucent TextureView configuration

matteblair:
Resolves https://github.com/tangrams/tangram-es/issues/2235

The changes to the demo app will be reverted before merging. This change enables a map using a `TextureView` to use alpha blending with views behind it in the layout - like in this image:
Screenshot_20210308-014027

To achieve this, an application needs to:
1. Call `getMapAsync` with a `TextureViewHolderFactory` that has translucency enabled: 
```
view.getMapAsync(this, new TextureViewHolderFactory(true), httpHandler);
```
2. Render parts of the map scene with the desired opacity, such as the background:
```
scene:
  background:
    color: [0, 0, 0, 0]
```


tallytalwar:
Great use of TextureView handler. Will try to look at the code sometime next week, but if you want to merge before that, please go ahead.

Thanks Matt

Issue #2233, opened 2021-02-08
iOS scene files as bundles does not work properly


rwrx:
I have noticed that after this commit https://github.com/tangrams/tangram-es/commit/4114aeccbda6f9272d2f891ae04822a061203f57 when scene files are added into iOS project as bundles - zipped files (I am using them in this way so that they are included in the project) textures inside scene are not loaded or something wrong is happening. I have not digged into code change more, but it is broken. I have noticed that it is working when these scene files are added by internet url to zip it is working, but I need to use them as bundles which are contained inside application.

rwrx:
@matteblair maybe for you it is a quick fix, so I do not wanted to dig into it yet.

matteblair:
Hi, I've just tested out a scene file in a bundle on iOS and the textures seem to be loading correctly. Can you describe in more detail the problem you're seeing? Or, better yet, do you have an example of a scene bundle that isn't working for you?

rwrx:
Thank you a lot for looking into it. For example Walkabout style: https://cloud.vecturagames.com/index.php/s/ZmeesYiBiXPEo6E . But you need to uncomment sources, because they are commented out.

matteblair:
Thanks for the example - I've reproduced the issue and will report back when I find a fix.

tallytalwar:
@matteblair could you put some logs or your initial findings here, maybe I can recollect the logic of code around this and help direct indirectly. Thanks

PR #2245, opened 2021-03-12
Fix texture URL resolution in scene zip bundles

matteblair:
Resolves https://github.com/tangrams/tangram-es/issues/2233

When we resolve texture URLs during the scene import process, we need to handle the case where the URL of the scene refers to a zip file. Relative URLs in a scene bundle refer to entries within the zip archive, so they need to be resolved as "zip URLs" instead of being resolved  relative to the scene bundle itself. In the linked issue, the image URLs were resolving to:
```
file:///path/to/walkabout-style.zip/images/walkabout-green1.jpg
```
Instead they should resolve to:
```
zip://file%3A%2F%2F%2Fpath%2Fto%2Fwalkabout-style.zip/images/walkabout-green1.jpg
```


tallytalwar:
Looks good to me, thanks Matt.

tallytalwar (approved):

PR #2257, opened 2021-03-29
Add shader defines for all blend modes

### LINKED ISSUE #2256: Black background shown for all dashed line styles

Apply a dashed style to any line or polygon features.

**RESULT:**

image

**EXPECTED RESULT:**

The default dashed line background color should be transparent. This was the case until recently. I wasn't paying attention, so I'm not sure exactly when this started occurring, but I can state confidently that it used to work just fine.

**ENVIRONMENT:**

 - Tangram ES on iOS

As far as I know, our styles haven't changed. However, I can confirm that all of the dashed lines are now showing this black background by default.

matteblair:
Fixes a bug with dashed polyline rendering introduced by https://github.com/tangrams/tangram-es/pull/2137. On [this line in polyline.fs](https://github.com/tangrams/tangram-es/blob/main/core/shaders/polyline.fs#L77) we checked `#if defined(TANGRAM_BLEND_OPAQUE)`, but `TANGRAM_BLEND_OPAQUE` was never injected! This change injects shader defines for all blending modes, to more closely match Tangram JS shaders.

Resolves #2256

### DIFF
--- a/core/src/style/style.cpp
+++ b/core/src/style/style.cpp
@@ -41,11 +41,17 @@ void Style::build(const Scene& _scene) {
     constructVertexLayout();
     constructShaderProgram();
 
-    if (m_blend == Blending::inlay) {
-        m_shaderSource->addSourceBlock("defines", "#define TANGRAM_BLEND_INLAY\n", false);
-    } else if (m_blend == Blending::overlay) {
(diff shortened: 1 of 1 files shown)


tallytalwar:
Looks fine to me. Any idea when was the regression introduced?

matteblair:
I introduced this regression in #2137 where I was fixing an issue first found in Tangram JS. I used the same GLSL code because I assumed we were injecting the same defines.

PR #2279, opened 2021-07-18
Windows Port CI

matteblair:


mathisloge:
vcpkg cmake 3.21 bug is now resolved

matteblair:
The CircleCI configuration is now working the way I want it to, but I've discovered that CircleCI has very low usage limits for Windows on their free plan, even for OSS projects. We get 2500 free credits per week, the smallest Windows VM costs 40 credits per minute, and our median Windows build seems to take ~7 minutes, which works out to 8 builds per week. This seems like it won't be enough for periods of any significant activity, unfortunately. 

I'm going to look into using [Appveyor](https://www.appveyor.com/) for our Windows CI instead. From what I can tell, the usage limits on their free plan should be plenty for tangram-es.

mathisloge:
github actions might be a thing, too. The binary caching of vcpkg  (uploading nuget packages to the account) works fine with github action (also available for PRs)

matteblair:
Also a good option! I am fairly sure that Pipelines and Actions run on exactly the same VMs, so I'll check which one has more convenient configuration options.

PR #2281, opened 2021-07-31
Fixes needed for Raspberry PI build on latest Raspbian

ericnelsonaz:
These updates are needed in order to build on the latest version of Raspberry PI O/S.

### DIFF
--- a/platforms/rpi/config.cmake
+++ b/platforms/rpi/config.cmake
@@ -40,6 +40,7 @@ target_link_libraries(tangram
   /opt/vc/lib/libbrcmGLESv2.so
   /opt/vc/lib/libvchiq_arm.so
   /opt/vc/lib/libvcos.so
+  /usr/lib/arm-linux-gnueabihf/libatomic.so.1
 )
 
 target_compile_options(tangram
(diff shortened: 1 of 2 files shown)


platforms/rpi/config.cmake
```diff
@@ -40,6 +40,7 @@ target_link_libraries(tangram
   /opt/vc/lib/libbrcmGLESv2.so
   /opt/vc/lib/libvchiq_arm.so
   /opt/vc/lib/libvcos.so
+  /usr/lib/arm-linux-gnueabihf/libatomic.so.1
```
matteblair, line 43: Using just `atomic` here seems to work on my pi. If it works on yours too, can you replace the path here with `atomic`? I'd rather not hard-code libraries where it isn't necessary.

matteblair (changes requested):
Thanks for the URL fixes!

The `getKey()` changes didn't seem to make any difference on my Pi. My keyboard inputs are being ignored both with and without this change. But if this change helps for you then let's keep it!

I'd ask that you change the library configuration that I commented on, then this will be good to merge.

ETE-Design:
@matteblair Is it possible to use Tangram-es as an Navigation System with Route Directions on an Raspberry Pi 4?

Issue #2299, opened 2021-11-02
[Feature Request] MapData::setVisibility(Boolean)


westnordost:
# Use case

An application that uses layers (`MapController.addDataLayer`) to display different data or views on data will frequently want to switch what data is displayed.

As an example, the views I have in my app:

By default, I want to display "quest pins" on the map

- When viewing the edit history, recent edits should instead be displayed on the map
- When viewing the details of a "quest", stuff relevant to that quest should be displayed

Each when going back, the normal "quest pins" should again be displayed on the map.

# Current solution

Just switching the visibility of layers is currently not possible, so what is done instead is to completely clear a layer on "hiding" it. When showing it again, even though nothing changed, the data for the layer needs to be retrieved again.

A workaround for if this feature will not be implemented would be write a wrapper around tangram data layers that keep the map data in memory so it can be re-set when the layer should be shown again. But this will obviously be much slower than if there was some visibility-logic in tangram-es itself since the whole geojson passed to tangram-es needs to be parsed and transferred again on unhiding.

PR #2302, opened 2021-11-08
Add 'visible' toggle for Android MapData and iOS TGMapData

matteblair:
Allows apps to temporarily hide a client data source without completely removing it.

Resolves https://github.com/tangrams/tangram-es/issues/2299


westnordost:
🥳 

One thing: You exposed `isVisible` for iOS but not for Android.

Issue #2303, opened 2021-11-12
No system fonts fallback on Android 12


rwrx:
A user of our app reported that on Android 12 labels with Chinese characters are not shown on the map. I have checked it and unfortunately, it is the case that on Android 12 (Pixel 3a device) labels with Chinese characters are not shown. On Android 11 and older it is working properly. I have also tried to target Android 12 in our app and it did not helped. I have not digged into Tangram-ES what could be the case yet, I am just reporting it now. I am also attaching screenshots from our app (Walkabout style) and I have also noticed that it is not working in StreetComplete app too (@westnordost). Screenshots are showing Taiwan.

![gpxviewerpro](https://user-images.githubusercontent.com/2378743/141537791-6d002a01-0a57-4e40-9396-b92e3a5f2706.png)
![streetcomplete](https://user-images.githubusercontent.com/2378743/141537805-b327b9b2-fa1b-416c-9b83-3d12dc89cab7.png)

tallytalwar:
If I remember correctly we access system font files for part of our label rendering pipeline. Since it's working on android 11 but not on android 12, it would be good to verify if font file (fallback) exists and at the correct location on android 12. Maybe comparing font.xml on 11 and 12 should provide some hint.

rwrx:
@tallytalwar thank you a lot. I have checked fonts.xml on Android 11 (Samsung Galaxy Tab A) and on Android 12 (Google Pixel 3a) and there is a difference. It seems that parser cannot parse fonts.xml on Android 12 because it contains text and also inner tags: 

```
<familyset version="23">
    <!-- first font is default -->
    <family name="sans-serif">
    <font weight="100" style="normal">Roboto-Regular.ttf
      <axis tag="ital" stylevalue="0" />
      <axis tag="wdth" stylevalue="100" />
      <axis tag="wght" stylevalue="100" />
    </font>
    ...
```

this is from Android 11:
```
<familyset version="23">
    <!-- first font is default -->
    <family name="sans-serif">
      <font weight="100" style="normal">Roboto-Thin.ttf</font>
      ...
```

I am attaching these fonts.xml files:

Google Pixel 3a (Android 12) fonts.xml:
https://gist.github.com/rwrx/a1e16c0ce834b27d3855df40eab32f7d

Samsung Galaxy Tab A (Android 11) fonts.xml:
https://gist.github.com/rwrx/c6000e5187641fbb5d5fbf529e97190c

[...]

PR #2306, opened 2021-11-29
Android 12 system font support

matteblair:
Resolves https://github.com/tangrams/tangram-es/issues/2303

Despite the availability of a new `SystemFonts` API in Android API level 29 and above, it turned out to be easier and faster to modify the FontConfig XML parsing code to handle the format of `fonts.xml` in versions before and after Android 12. As mentioned in the issue though, we will still need to add support for font variation axes to render the bold or italic versions of many system fonts.

### DIFF
--- a/platforms/android/build.gradle
+++ b/platforms/android/build.gradle
@@ -5,7 +5,7 @@ allprojects {
       jcenter()
     }
     dependencies {
-      classpath 'com.android.tools.build:gradle:7.0.2'
+      classpath 'com.android.tools.build:gradle:7.0.3'
     }
   }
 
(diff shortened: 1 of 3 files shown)


westnordost:
With more and more people upgrading to Android 12, this is becoming more and more of a problem. My app has mostly users from the EU and US and I see now that 17% of users use Android 12, as opposed to 10% when the issue was first reported. As this is already fixed, are you planning to do a release soon?

Or were you waiting on feedback from the issue reporter or me whether the fix works correctly? I do have an Android 12 phone now, but I have no tangram-es development environment set up.

rwrx:
@westnordost I can confirm that it is foxed with this pull request.

matteblair:
I actually hadn't realized that this change had not been included in a release yet. I'll prioritize making a new release so that this becomes available to downstream users. Sorry to make you wait so long for this fix!

KS-Balaji:
I have pulled this source. Now the fonts rendering was slow in some android 12 devices. @matteblair any ideas?

Issue #2305, opened 2021-11-23
global constants from map style don't work for markers


westnordost:
Consider this style
```yaml
global:
  text_fill_color: '#0f0'
  text_stroke_color: '#f00'
  text_stroke: { color: global.text_stroke_color, width: 2.5px }
```

and this marker

```kotlin
val marker = mapController.addMarker()
marker.setStylingFromString(
"""{
  style: 'text',
  text_source: 'function() { return "test"; }',
  font: {
    fill: global.text_fill_color,
    size: '17px',
    stroke: global.text_stroke
  },
}
""")
```

This results in the following warnings
```
WARNING styleParam.cpp:381: Invalid color value: global.text_fill_color
WARNING styleParam.cpp:134: Unknown StyleParam font:stroke:global.text_stroke
```
and the text doesn't render (correctly).

Using styles previously defined in the map style however in the markers works without a problem.

PR #2309, opened 2021-12-19
Apply global variables within Marker styling strings

matteblair:
Resolves https://github.com/tangrams/tangram-es/issues/2305


westnordost:
🥳

PR #2312, opened 2021-12-24
Set up vcpkg binary caching via nuget on gh packages

matteblair:
Setup following instructions on https://github.com/microsoft/vcpkg/blob/master/docs/users/binarycaching.md

The binary caching feature in vcpkg seems to be as fast as the GitHub Actions caching, but with more accurate hashing and longer retention (because it is hosted on GitHub Packages). Cache retention should be effectively unlimited now, versus just 7 days with GitHub Actions.


tallytalwar:
Thanks Matt. There is a lot of changes to ci for tangram, is it possible for you to summarize some of this in the config files itself? - will help with contributions, etc.

PR #2313, opened 2022-01-02
Add iOS build to GH Actions

matteblair:
CircleCI limits our builds to 1 macOS job at a time, which can make release deployments slow. I'm trying out iOS builds in GitHub Actions instead.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -94,6 +94,7 @@ jobs:
             - ~/.ccache
       - store_artifacts:
           path: platforms/android/tangram/build/outputs/aar/tangram-release.aar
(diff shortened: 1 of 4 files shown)


matteblair:
The CI and release processes are getting a little complicated, but here's the basics:
- On pushes to all branches we run builds and tests in Debug mode for Linux, Windows, macOS, Android, and iOS.
- On pushes to `main` we also run builds in Release mode for the Android and iOS deployment artifacts.
- When making a new release, we take the Android and iOS deployment artifacts and attach them to the release on GitHub. 

If we move Android builds to GitHub Actions as well then I think we could automate a lot more of the release process. The `gh` CLI tool can create releases and upload files from a script.

PR #2344, opened 2023-10-12
increase max zoom from 20.5 to 25

### LINKED ISSUE #2264: FR: Allow for higher zoom values
Related, downstream issue: https://github.com/streetcomplete/StreetComplete/issues/2878

**Use case**: When displaying partially overlapping markers on a map, they are often hidden if their position is too close to each other. This often makes it impossible to click/read some markers, especially when the map is at max zoom already.

**Proposed solution**: Increase the maximum zoom level by 2-3 steps, e.g. to 22 or 23, so that applications can utilize higher zoom levels if they need to.

WebFreak001:
25 seems to be about 2-5m for one phone width, probably higher than what anyone will ever need, but the capability is there now if map devs actually want to do this.

Defaults are left unchanged, will only have an effect in applications if users call `MapController.setMaximumZoomLevel`

fix #2264

### DIFF
--- a/core/include/tangram/map.h
+++ b/core/include/tangram/map.h
@@ -261,8 +261,8 @@ public:
     // Get the minimum zoom level for the view.
     float getMinZoom() const;
 
-    // Set the maximum zoom level for the view; values greater than 20.5 will be
-    // clamped to 20.5; values less than the current minimum zoom level will set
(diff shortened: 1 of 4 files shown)


WebFreak001:
CI fail seems unrelated / flaky

